PR #85, opened 2020-04-30
fixes for tags mode

sverps:
## Status
**READY**

## Description
- Interface name was not properly pascal cased (the tag part was lowercase due to being concatenated before going into `pascal()` function)
- The tags mode didn't generate any output for operations that had no tag. This is changed by adding the `"default"` tag on such operations.


anymaniax:
Thanks!

PR #96, opened 2020-11-04
Merge branch 'master' of https://github.com/anymaniax/orval

ismail-codar:


src/core/getters/resReqTypes.ts
```diff
@@ -12,6 +12,7 @@ import { resolveObject } from '../resolvers/object';
 import { getRef } from './ref';
 
 const CONTENT_TYPES = [
+  '*/*',
```
anymaniax, line 15: why do you need this content type?

src/core/importers/openApi.ts
```diff
@@ -30,13 +30,17 @@ const importSpecs = (
...
-      swagger2openapi.convertObj(schema, {}, (err, { openapi }) => {
-        if (err) {
-          reject(err);
-        } else {
-          resolve(openapi);
-        }
-      });
+      swagger2openapi.convertObj(
+        schema,
+        { patch: true },
```
anymaniax: I saw in the doc that you can use that to fix minor errors. Do you have a use case?

src/core/getters/resReqTypes.ts
ismail-codar: > why do you need this content type?

It can be use in Spring Framework: https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/http/MediaType.java#L64
For example:
```java
@GetMapping(value = ["/my-path"], produces = [MediaType.ALL_VALUE])
```

src/core/importers/openApi.ts
ismail-codar: > I saw in the doc that you can use that to fix minor errors. Do you have a use case?

patch: true fixes my swagger v2 json problems without any side effect. So it may apply to other users.
Or we can add it as an option to the orval configuration.

anymaniax: yes maybe, it's better to add it as an option

anymaniax: What do you think about adding a 'fix' property who take a boolean in the input config and its default true

ismail-codar: > What do you think about adding a 'fix' property who take a boolean in the input config and its default true

I think that we must ensure that all swagger2openapi options can be passed with  `patch: true, warnOnly:true` by default
Like this: 
```js
swagger2openapi.convertObj(schema,  {patch: true, warnOnly:true, ...userSwagger2openapiOptions} 
```
Note: Default value of patch is true used in here
https://github.com/Mermade/oas-kit/blob/master/packages/swagger2openapi/oas-validate.js#L89

anymaniax: seems good to me who would you name the property in the orval config?

ismail-codar: I'm not good at naming. It would be ` swagger2openapi: {..some...options..}`

anymaniax: converterOptions: {..some...options..} would be could for you?

ismail-codar: > converterOptions: {..some...options..} would be could for you?

Yes good.

anymaniax: You feel comfortable to add it or I do it?

ismail-codar: We seem to agree on the job. I can do it.

anymaniax: ok top! thanks 🎉

src/bin/orval.ts
```diff
@@ -19,7 +19,11 @@ program
   .option('-p, --project <name>', 'focus a project of the config')
   .action((paths, cmd) => {
     if (isString(cmd.input) && isString(cmd.output)) {
-      generateSpec(process.cwd(), { input: cmd.input, output: cmd.output });
+      generateSpec(process.cwd(), {
+        input: cmd.input,
+        output: cmd.output,
+        converterOptions: undefined,
```
ismail-codar, line 25: undefined because converterOptions  from command line is not supported.

anymaniax: I would have put the converterOptions inside the input options

src/types/index.ts
```diff
@@ -120,4 +121,5 @@ export type ImportOpenApi = {
   input?: InputOptions;
   output?: OutputOptions;
   workspace: string;
+  converterOptions: any;
```
anymaniax: Could be cool if you can add a type for that in the swagger2openapi.d.ts file

anymaniax (changes requested):

src/types/index.ts
ismail-codar: swagger2openapi.d.ts options parameter is not typed. 
We can create typings like https://github.com/Mermade/oas-kit/tree/master/packages/swagger2openapi#usage But I don't think it's worth it. Because converterOptions used for only one time parameter passing. I think there is nothing code reusability this time or in the future.

anymaniax: Ok then maybe just add a comment with a link to the possible options

ismail-codar: I have added ConverterOptions type on latest commit

anymaniax: Awesome!

anymaniax:
Did you have time to do the last modification?

ismail-codar:
> Did you have time to do the last modification?

Sorry.

anymaniax:
implemented in the `5.0.0`

PR #97, opened 2020-11-08
Filter out imports that refer to the type defined in current file

antoni:
## Status

**READY**

## Description
Filter out imports that refer to the type defined in current file (OpenAPI recursive schema definitions)

### DIFF
--- a/src/core/generators/interface.ts
+++ b/src/core/generators/interface.ts
@@ -14,23 +14,29 @@ import { getScalar } from '../getters/scalar';
 export const generateInterface = (name: string, schema: SchemaObject) => {
   const { value, imports, schemas } = getScalar(schema, name);
   const isEmptyObject = value === '{}';
+  const definitionName = pascal(name);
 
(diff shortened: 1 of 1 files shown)


src/core/generators/interface.ts
```diff
@@ -25,12 +25,17 @@ export const generateInterface = (name: string, schema: SchemaObject) => {
     model += `export type ${pascal(name)} = ${value};\n`;
   }
 
+  // Filter out imports that refer to the type defined in current file (OpenAPI recursive schema definitions)
+  const externalModulesImportsOnly = imports.filter(
+    (importName) => importName !== name,
```
anymaniax: name should be pascal(name) and maybe we should create a property.  Like this 
`const definitionName = pascal(name)`

antoni: Changed

anymaniax (approved):
Top. Thanks 🎉

PR #109, opened 2021-01-15
add content-type text/yaml as supported content-type to import

wouterkroes:
## Status

**READY**

## Description

add content-type text/yaml as supported content-type to import

## Related PRs
https://github.com/anymaniax/orval/discussions/108

branch | PR
------ | ------
other_pr_production | [link]()
other_pr_master | [link]()

## Todos
- [x] Tests
- [x] Documentation

## Notes

`headers['content-type']` comparison for `text/yaml`, `text/plain` and `application/json` now all in the same way (with `includes())`. Because for instance for `application/json` the actual content-type could be `application/json; charset=utf-8`.

### DIFF
--- a/src/core/importers/specs.ts
+++ b/src/core/importers/specs.ts
@@ -7,6 +7,7 @@ import { getGithubOpenApi } from '../../utils/github';
 import { isObject, isString } from '../../utils/is';
 import { dynamicReader } from '../../utils/reader';
 import { importOpenApi } from './openApi';
+import { ImportOpenApi } from '../../types';
 
 export const importSpecs = async (
   workspace: string,
@@ -24,7 +25,7 @@ export const importSpecs = async (
     throw new Error('You need to provide an output');
   }
 
-  const format = getExtension(path);
+  let format: ImportOpenApi['format'] = getExtension(path);
 
   let data: string | object;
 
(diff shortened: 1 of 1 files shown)


src/core/importers/specs.ts
```diff
@@ -34,12 +35,19 @@ export const importSpecs = async (
...
+          'application/json',
+        );
 
-        if (
-          headers['content-type'] === 'application/json' ||
-          headers['content-type'].includes('text/plain')
-        ) {
+        if (isContentTypeJson || isContentTypeYaml || isContentTypeTextPlain) {
           data = specification;
+          if (isContentTypeTextPlain && typeof data === 'string') {
```
anymaniax, line 54: you need to do that for isContentTypeJson and isContentTypeYaml too. Just assign the format directly (format = 'json' for isContentTypeJson).

anymaniax (changes requested):

wouterkroes:
@anymaniax any thoughts on my comment regarding your requested code change?

anymaniax:
@wouterkroes which comment? There is only my request to assign the format linked to the content-type right?

wouterkroes:
Ah yes, and I replied to your code change request:
image

anymaniax:
You comment is in pending that why I don’t see it I think. I ask you to do that for the case where in the path you only have /specification without an extension. In that case you still correctly set the good format and avoid a possible error.

wouterkroes:
Updated! Hopefully, this is the desired implementation for the case you mentioned

anymaniax (approved):
Thanks for this new feature 🎉🚀

PR #118, opened 2021-05-10
First commit

Eddie-CooRo:
## Status

READY

## Description
Right now if you try to put models into some deep folder, while the parent directories does not exist, you'll see an error. this MR solves that by passing recursive: true to all of mkdir requests.


anymaniax:
Can you try with the version 5. Normally fixed in the next version. You can install it with orval@next

PR #119, opened 2021-05-10
feat: pass recursive: true to all mkdir calls

Eddie-CooRo:
## Status

READY

## Description
Right now if you try to put models into some deep folder, while the parent directories does not exist, you'll see an error. this MR solves that by passing
```ts
mkdirSync(path, { recursive: true });
```
to all of mkdir calls to make sure it will create parent paths as well if they don't exist.


anymaniax:
Can you try with the version 5 (orval@next) should be fixed in the next release

Eddie-CooRo:
BTW, where should I branch out if I want to push some (other) quick fixes directly to v5?

PR #121, opened 2021-05-11
fix: implement & use relativeSafe to keep leading ./ in import paths

### LINKED ISSUE #120: [v5] schema import doesn't work correctly when schemas is at the same level as the endpoints file
1. build a project with the following config:
```js
module.exports = {
  petstore: {
    output: {
      mode: 'split',
      target: './api/petstore/index.ts',
      schemas: './api/petstore/models',
	  client: 'react-query'
    },
    input: {
      target: './petstore.yaml',
    },
  }
}
```

## What happens?
in the `api/petstore/index.ts` the models are getting imported from `models`
image

The import path should be `./models`
image

**Package Version:** 5.1.2

Eddie-CooRo:
## Status

**READY**

## Description
Resolves #120 By introducing a new function named `relativeSafe` and resolve the import path using it. `relativeSafe` is the `path.relative` doppelganger that respects the meaningful leading "./"

### DIFF
--- a/src/core/generators/mutator.ts
+++ b/src/core/generators/mutator.ts
@@ -1,7 +1,7 @@
-import { relative } from 'upath';
 import { Mutator } from '../../types';
 import { getFileInfo } from '../../utils/file';
 import { isString } from '../../utils/is';
+import { relativeSafe } from '../../utils/path';
 
 const getImport = (output: string, mutator: Mutator) => {
   const outputFileInfo = getFileInfo(output);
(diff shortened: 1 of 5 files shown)


Eddie-CooRo:
Don't merge just yet. investigating something

anymaniax:
Thanks can you do the same fix for the other mode please? (tags-split, single, tags)

anymaniax:
I will create a fix release `5.1.3`

PR #133, opened 2021-05-22
Fixes relative paths bug in singleMode

### LINKED ISSUE #132: Bug: incorrect import path generation
In `writers/SingleMode.ts` there's a part producing incorrect relative schemas path 
Code here https://github.com/anymaniax/orval/blob/master/src/core/writers/singleMode.ts#L43

```typescript
export const writeSingleMode = (...) {

// (1)
 const { path, dirname } = getFileInfo(
     join(workspace, targetedPath),
    camel(info.title),
  );

// (2)
const schemasPath = relative(dirname || '', output.schemas); // ←—— This line
```

1. setup output with relative parts:
```js
mode    : 'single',
target  : `../../openapi/orval-build/endpoints/openapi.ts`,
schemas : `../../openapi/orval-build/model`,
```
2. run build

## What happens?
`endpoints/openapi.ts` output file will get incorrect import path:
```
from '..\..\..\..\..\openapi\orval-build\model'
```

but expected to get `from '../model';`

## Any other comments?
So we are getting 2 problems: 
1) inverted slashes (OS separators could be straight or back slashes, but imports must not be backslash).
2) incorrect relative part.

# The solution.
1) Slashes direction could be easily solved by adding `.replace(/\\/g, '/');`
2) When using `relative(dirname || '', output.schemas);`, `dirname` is already absolute path, processed with `path.join` (`(1)`). But `output.schemas` is still dirty input from config file `(2)`. So the result is incorrect.
Fix could look this way:
```js
relative(dirname, join(workspace, output.schemas)).replace(/\\/g, '/');
```

---

### What versions are you using?
**Operating System:** Win10 
**Node** 14

Boorj:
## Status
**READY**

## Description
Fixes #132, makes correct path generation in singleMode imports.

### DIFF
--- a/src/core/writers/singleMode.ts
+++ b/src/core/writers/singleMode.ts
@@ -40,7 +40,7 @@ export const writeSingleMode = ({
   let data = getFilesHeader(info);
 
   if (isObject(output) && output.schemas) {
-    const schemasPath = relative(dirname || '', output.schemas);
+    const schemasPath = relative(dirname || '', join(workspace, output.schemas)).replace(/\\/g, '/');
 
     data += generateClientImports(output.client, implementation, [
       { exports: imports, dependency: schemasPath },
--- a/src/core/writers/singleMode.ts
+++ b/src/core/writers/singleMode.ts
@@ -40,7 +40,10 @@ export const writeSingleMode = ({
   let data = getFilesHeader(info);
 
   if (isObject(output) && output.schemas) {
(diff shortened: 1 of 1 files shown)


anymaniax:
Hello, can you try with the orval@next should be fixed it's the rc

anymaniax:
Fixed with the new release `5.3.2`

PR #134, opened 2021-05-23
Fix of undefined case of definition in react-query.ts

Boorj:
## Status
**READY**

## Description
Fixes little bug here: https://github.com/anymaniax/orval/issues/132#issuecomment-846464157

> this line produces incorrect code if definition is undefined
https://github.com/anymaniax/orval/blob/master/src/core/generators/react-query.ts#L82

### DIFF
--- a/src/core/generators/react-query.ts
+++ b/src/core/generators/react-query.ts
@@ -27,9 +27,7 @@ const REACT_QUERY_DEPENDENCIES: GeneratorDependency[] = [
       { name: 'useInfiniteQuery', values: true },
       { name: 'useMutation', values: true },
       { name: 'UseQueryOptions' },
-      {
-        name: 'UseInfiniteQueryOptions',
-      },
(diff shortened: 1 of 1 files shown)


src/core/generators/react-query.ts
```diff
@@ -69,7 +67,9 @@ const generateAxiosFunction = (
           )?.slice(1, -1)} ...options}`
         : '// eslint-disable-next-line\n// @ts-ignore\n options'
       : '';
-
+   
+    const mutatorTyping = `<Data extends unknown ? ${ response.definition || 'Data' } : Data>`;
```
anymaniax: In that case, I prefer the following solution which remove the conditional type and replace it only with the generic

```suggestion
    const mutatorTyping = response.definition ? `<Data extends unknown ? ${ response.definition} : Data>` : '<Data>';
```

anymaniax (changes requested):

anymaniax (approved):

PR #141, opened 2021-05-23
Add custom client generators

### LINKED ISSUE #135: Customize client generators
## Generators customization

I want to change the behavior of generator. For example, add some JSDoc generation. It will be huge and quite easy boost for whole library, if we do next way.

### 1 Create custom client object
```ts
  [OutputClient.REACT_QUERY]: {
    client: myCustomGeneratorHere, // <---- HERE
    msw: generateMSW,
    header: generateReactQueryHeader,
    dependencies: getReactQueryDependencies,
    footer: generateReactQueryFooter,
    title: generateReactQueryTitle,
  }, //...
```
### 2 Include that in project
Could be done in 2 ways
#### 2A 
tell program, that we want to use `customClientGeneratorFunc` here:
https://github.com/anymaniax/orval/blob/master/src/core/generators/client.ts#L177
```ts
const generator = GENERATOR_CLIENT[outputClient];
const client = generator.client(verbOption, options);
```
Custom client could be passed 
- as `override.client` property
- as `output.client` as function:
```js
  const generator = typeof outputClient === 'function' ? outputClient : GENERATOR_CLIENT[outputClient];
  // this should be done in all spots where GENERATOR_CLIENT is invoked.
```

#### 2B
Make `GENERATOR_CLIENT` a singleton, which could be mutated.

```js
// generators/client.ts
const GENERATOR_CLIENT = {...};

export function addGeneratorClient(key, obj) {
    GENERATOR_CLIENT[key] = obj;
}

// in orval.config.js 
import { addGeneratorClient } from 'orval';
addGeneratorClient('custom-query', {
  client : foo(){},
 // ...
});

export default {
   output: { client: 'custom-query' }
}
```

Of course that could be realized with class and private properties, anyway - idea was modifying global object with own generators.

## steps are
1. make this tiny code change
2. adopt change typings for new changes
3. update documentation

---

## Profit
Users can change main functions with their own just in few operations. It would be easy to change just one generator (i. e. `client`) and the rest ones (`msw`, `title`)  could be imported as is.

Boorj:
## Status
**READY**

## Description
adds ability to mutate default client generators list and closes #135

### DIFF
--- a/src/core/generators/client.ts
+++ b/src/core/generators/client.ts
@@ -6,6 +6,7 @@ import {
   GeneratorOptions,
   GeneratorVerbOptions,
   GeneratorVerbsOptions,
+  ClientGeneratorsBuilder,
 } from '../../types/generator';
 import { asyncReduce } from '../../utils/async-reduce';
 import { pascal } from '../../utils/case';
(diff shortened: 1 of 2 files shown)


src/core/generators/client.ts
```diff
@@ -90,6 +91,50 @@ const GENERATOR_CLIENT = {
   },
 };
 
+export const addGeneratorClient = (
```
anymaniax, line 94: I am not a big fan to expose a function. Orval is primarily a tool and cli. Could we maybe investigate a solution passing by the orval config? Like if you provide a function to the client property and we pass in arg the `GENERATOR_CLIENT`?

Boorj: I agree. It's a kind of hot fix, which empowers the abilities. It doesn't ruin anything, it's quite safe to use in API (not in CLI). And it can help for my usage or whoever wants to customize output. It even does have gitignore, which is a dirty hack, relying on the fact, that it won't be called by people who don't know about this little hacky feature.

The neat solution with clientGenerator could look like a path of a file with `module.exports = {client :..., msw :...}`, injected to overrides section. And that could be used as a CLI solution.

I understand all disadvantages and problems of maintaining FOSS (that's why I do contribute sometimes). 

Anyway, this little hack doesn't break anything and adds nice ability. And could be assisted with an `@experimental` tag or whatever. Like "don't rely on this, interface can be changed". But it could help people, i suppose. Probably to develop their own builder and probably, contribute in easier way.

anymaniax: I can easily do the one in the orval config too if you want. I prefer to do that than add hidden hack

anymaniax: I opened the #142 if you want to check

PR #142, opened 2021-05-23
feat(generator): add ability to provide your own client

anymaniax:
## Status
**READY**

## Description
add the ability to provide your own client

### DIFF
--- a/src/core/generators/client.ts
+++ b/src/core/generators/client.ts
@@ -38,15 +38,15 @@ import {
 
 const DEFAULT_CLIENT = OutputClient.AXIOS;
 
-export const GENERATOR_CLIENT = {
-  [OutputClient.AXIOS]: {
+export const GENERATOR_CLIENT: GeneratorClients = {
+  axios: {
     client: generateAxios,
     header: generateAxiosHeader,
     dependencies: getAxiosDependencies,
     footer: generateAxiosFooter,
     title: generateAxiosTitle,
   },
-  [OutputClient.AXIOS_FUNCTIONS]: {
+  'axios-functions': {
     client: (verbOptions: GeneratorVerbOptions, options: GeneratorOptions) => {
       const { implementation, imports } = generateAxios(verbOptions, options);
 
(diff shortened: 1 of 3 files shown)


anymaniax:
@Boorj what do you think about this?

Boorj:
Haha) Very quick) I'm implementing this to new project, so need to test.
✔️  Reviewed code. I see that you took my part of typings. But i need to warn you: i haven't checked over the full code and tested all invocations of those types at all clients (angular, svelte, etc). So probably in some calls arguments could differ.
If it works - cool. But my opinion, later that `CLIENT GENERATORS` should be typed with this line:
https://github.com/anymaniax/orval/blob/51544ed094b9303e6aab355f6f8a08cc66984243/src/types/generator.ts#L148

```ts 
export const GENERATOR_CLIENT = { //...

export const GENERATOR_CLIENT :GeneratorClients  = { //...
```

anymaniax:
If you are ok with it. I can merge it and create a new version for it. I check the type that you did and they seem pretty good. I also updated the merge request to avoid the cast and type correctly the GENERATOR_CLIENT

Boorj:
Anyway, it's very cool fix, since it's solves the biggest problem. I'll try to implement it asap.

Idea for enhancement - configurable writers. For example, to change structure of files, or get custom naming.

**Optional notes about code base**
Btw,  this name 'client generator' is a bit confusing. As a fact, GeneratorClients is a list of objects for generating strings, functions and objects, and not only clients.

Also as a reviewer i prefer not to use those tiny dependencies like 'isFunction', because outside of context viewer college need to browse to the definition to get sure, that it's `typeof foo === 'function'` and no other conditions. So KISS code in such cases i prefer more, but it's up to personal preferneces, of course.

Also, there could be a method like "normalizeAndValidateConfig" in the beginning, so it ensures {input, output} is valid and in normalized form. 
Like `{input : './path'}` to `{input: {target: 'c:/apps/some/path'}}` . So no need to spoil code with joins, resolves and everything could be strictly typed.

Boorj:
> If you are ok with it. I can merge it and create a new version for it. I check the type that you did and they seem pretty good. I also updated the merge request to avoid the cast and type correctly the GENERATOR_CLIENT

Thank you, Victor, everything is great. Looking forward to update to new version and check.

anymaniax:
@Boorj could be good to normalize and validate the config upfront I will create an issue for that.

PR #146, opened 2021-05-23
fixing slashes bug in specs.ts

Boorj:
## Status
**READY**

## Description
fixes imports bug, #132
https://github.com/anymaniax/orval/issues/132#issuecomment-846616351

### DIFF
--- a/src/core/importers/specs.ts
+++ b/src/core/importers/specs.ts
@@ -36,6 +36,13 @@ export const importSpecs = async (
     : parserDefaultOptions;
   const data = (await SwaggerParser.resolve(path, parserOptions)).values();
 
+  // normalizing slashes after SwaggerParser
+  Object.keys(c).forEach(key => {
+    const value = c[key];
+    delete c[key];
+    c[resolve(key)] = value;
+  });
+  
   return importOpenApi({
     data,
     ...(isObject(input) && {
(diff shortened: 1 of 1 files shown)


src/core/importers/specs.ts
```diff
@@ -36,6 +36,13 @@ export const importSpecs = async (
     : parserDefaultOptions;
   const data = (await SwaggerParser.resolve(path, parserOptions)).values();
 
+  // normalizing slashes after SwaggerParser
+  Object.keys(data).forEach(key => {
+    const value = data[key];
+    delete data[key];
+    data[resolve(key)] = value;
+  });
+
```
anymaniax:
```suggestion
const resolveSpecs = async (
  path: string,
  parserOptions: SwaggerParser.Options,
) => {
  const data = (await SwaggerParser.resolve(path, parserOptions)).values();

  // normalizing slashes after SwaggerParser
  return Object.fromEntries(
    Object.entries(data).map(([key, value]) => [resolve(key), value]),
  );
};

const data = await resolveSpecs(path, parserOptions);
```
If you want to use the Object.fromEntries you need to update the tsconfig lib also change `es2015` to `es2019`

I would prefer something like this. What do you think?

Boorj: Nice solution! 
I do still have some issues with codestyle, when 3+ actions are inlined and you need to parse long lines by eyes, instead of quick - "1-2 actions per line" falling down. Anyway, that sticks to the whole project codestyle. Also using of ->entries-> is consistent. So yes, sure!

anymaniax (approved):

PR #151, opened 2021-05-24
fix: remove P extends unknown check from SecondParameter

### LINKED ISSUE #150: [v5] react-query mutator config type is not working
1. Pass a custom mutator and build the project
2. Try to pass `request` config to any `useQuery` or `useMutation` calls.

## What happens?
The request config type is `Record<string, any>`

It should be the type that the fetcher accepts as its second parameter.

[Code sandbox](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzEAXMALEMBc8AUYA9gHaICWA5rlMQJ4A08IAHijFAMInlXwDObMsQoBKanQDcAWABQslLQAOCAMrgSwAAqwoAW1RYAPABUmrEMWB98AOjuwKfcbRHwAvAD54NWl7fxTFhQLK3xZeHgiUkpnenDvGEdcIWQYeE04mVdPbzp4gH50s2DLawBXYgBrYkIAd2IC+AAldRhgQwEYIQpGHw943E0B+GIQADcsaTlphWV4AEF3eDUorR19YJhDWZBCRCRUDCwPKYigA)

## Any other comments?
[The axios generator](https://github.com/anymaniax/orval/blob/4acd3bc2de568766b1e945e38cc31ca1c28d1558/src/core/generators/axios.ts) doesn't have the `extends unknown` part. As everything in typescript extends unknown, [this part](https://github.com/anymaniax/orval/blob/4acd3bc2de568766b1e945e38cc31ca1c28d1558/src/core/generators/react-query.ts#L395) seems to be the cause of the issue in the `react-query` generator. AFAIK, removing it should be safe.

**Package Version:** 5.3.2

Eddie-CooRo:
## Status

**READY**

## Description
Fixes #150

### DIFF
--- a/src/core/generators/query.ts
+++ b/src/core/generators/query.ts
@@ -417,9 +417,7 @@ ${
   config: any,
   args: infer P,
 ) => any
-  ? P extends unknown
-  ? Record<string, any>
-  : P
+  ? P
   : never;\n\n`
     : ''
 }`;


src/core/generators/query.ts
```diff
@@ -417,9 +417,7 @@ ${
   config: any,
   args: infer P,
 ) => any
-  ? P extends unknown
-  ? Record<string, any>
-  : P
+  ? P
```
anymaniax, line 420: What is the problem with the current type?

eddiecooro: https://github.com/anymaniax/orval/issues/150

anymaniax (approved):

PR #186, opened 2021-07-22
fix: quote object mock key if it is an identifier name

alanpoulain:
## Status
**READY**

## Description
If the OpenAPI specification uses properties with special characters like `@`, the generated mocks are invalid JS since the object key is not a valid [identifier name](http://es5.github.io/x7.html#x7.6).
In this case, the key needs to be quoted.

## Related PRs
N/A

## Todos
- [x] Tests
- [ ] Documentation


anymaniax (approved):

anymaniax:
Thanks @alanpoulain  🎉

PR #197, opened 2021-08-20
feat(client): update HttpClientOptions to be compatible with angular v12

jogelin:
## Status
**READY**

## Description
Since Angular 12, it is possible to provide context to the interceptors: https://netbasal.com/new-in-angular-v12-passing-context-to-http-interceptors-308a1ca2f3dd

But the current type `HttpClientOptions` is not compatible.


src/core/generators/angular.ts
```diff
@@ -51,14 +52,15 @@ ${
...
-  observe?: any;
+  context?: HttpContext;
+  observe?: 'body';
   params?: HttpParams | {
-      [param: string]: string | string[];
+    [param: string]: string | number | boolean | ReadonlyArray<string | number | boolean>;
   };
   reportProgress?: boolean;
-  responseType?: any;
+  responseType?: 'json';
```
anymaniax: cannot only be 'json' or will limit the user

src/core/generators/angular.ts
```diff
@@ -51,14 +52,15 @@ ${
   isRequestOptions && !isGlobalMutator
     ? `type HttpClientOptions = {
   headers?: HttpHeaders | {
-      [header: string]: string | string[];
+    [header: string]: string | string[];
   };
-  observe?: any;
+  context?: HttpContext;
+  observe?: 'body';
```
anymaniax: cannot only be body or will limit the user

anymaniax (changes requested):

src/core/generators/angular.ts
jogelin: Correct! Sorry, I copy/paste one of the overloads...

I updated it with the real combination of overloads from: https://github.com/angular/angular/blob/c389052bab0db14d4451bb93cf356809a3ad97e3/packages/common/http/src/client.ts#L2552

src/core/generators/angular.ts
```diff
@@ -53,14 +54,15 @@ ${
   headers?: HttpHeaders | {
       [header: string]: string | string[];
   };
-  observe?: any;
+  context?: HttpContext;
+  observe?: 'body' | 'events' | 'response';
```
anymaniax: Will also cause typescript error. For the get you cannot have responseType 'json' and observe 'events'. It's why that was any because I didn't found a good solution for that.

src/core/generators/angular.ts
```diff
@@ -53,14 +54,15 @@ ${
...
   params?: HttpParams | {
-      [param: string]: string | string[];
+    [param: string]: string | number | boolean | ReadonlyArray<string | number | boolean>;
   };
   reportProgress?: boolean;
-  responseType?: any;
+  responseType?: 'arraybuffer' | 'blob' | 'json' | 'text';
```
anymaniax: same as above

anymaniax (changes requested):

src/core/generators/angular.ts
jogelin: ok reverted ;)

jogelin: Maybe the best approach would be to check the verb used by each path and generating the method name, use directly the type `Parameters<HttpClient['post']>`:
```javascript
  retrieveBrand<TData = BrandDto>(retrieveBrandBody: RetrieveBrandBody, options?: Parameters<HttpClient['post']>[2]): Observable<TData> {
    return this.http.post<TData>(`/api/Brand/RetrieveBrand`, retrieveBrandBody, options);
  }
```

src/core/generators/angular.ts
```diff
@@ -55,12 +56,12 @@ ${
   };
   observe?: any;
```
anymaniax, line 58: you also removed the `context?: HttpContext;` by mistake I think?

anymaniax (changes requested):

src/core/generators/angular.ts
jogelin: hard day... :S

anymaniax (approved):

anymaniax:
Thanks 🎉

PR #213, opened 2021-09-28
fix: manage enum plus number

alanpoulain:
## Status

**READY**

## Description

Minus number was managed for enum but not plus numbers (for instance for calling codes).

## Todos
- [x] Tests
- [ ] Documentation


anymaniax:
thanks 🎉

PR #214, opened 2021-09-28
fix: deduplicate enum values

alanpoulain:
## Status

**READY**

## Description

Deduplicate enum values by using a `Set`.

I'm not sure if this PR is OK since the problem is also because the OpenAPI schema is not "right".

## Todos
- [ ] Tests
- [ ] Documentation


anymaniax (commented):
True it's a problem with the specification but I think that it's ok if we can fix it easily. So thanks 🎉

anymaniax (approved):

PR #215, opened 2021-09-28
fix: Add types for pageParam when useInfiniteQueryParam is specified

neemski:
## Status

**READY**

## Description
When `useInfiniteQueryParam` is specified in the Orval config, `pageParam` has an implicit `any` type:

Screen Shot 2021-09-28 at 3 49 48 PM
Screen Shot 2021-09-28 at 3 50 01 PM

I addressed this by adding an inline type to the query generator: `{ pageParam?: string }`

**Note:** I noticed the sample config for react-query has `useInfinite` and `useInfiniteQueryParam` defined: https://github.com/anymaniax/orval/blob/master/samples/react-app-with-react-query/orval.config.ts#L43. However, the generated endpoint file doesn't have any `useInfinite` hooks: https://github.com/anymaniax/orval/blob/master/samples/react-app-with-react-query/src/api/endpoints/petstoreFromFileSpecWithTransformer.ts. Should we address that first? Does the `petstore.yaml` file need to be updated?

## Todos
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/core/generators/query.ts
+++ b/src/core/generators/query.ts
@@ -344,7 +344,7 @@ export const ${camel(
   const queryKey = queryOptions?.queryKey ?? ${queryKeyFnName}(${properties});
   const queryFn = (${
     queryParam && props.some(({ type }) => type === 'queryParam')
-      ? `{ pageParam }`
+      ? `{ pageParam }: { pageParam?: string }`
       : ''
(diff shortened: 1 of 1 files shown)


anymaniax:
Hello @neemski, I did a fix this morning that will solve the problem. Thanks for reporting the bug!

neemski:
Awesome, thanks @anymaniax!

Issue #247, opened 2021-11-04
TypeError: Cannot read properties of undefined (reading '$ref')


jitbasemartin:
## What are the steps to reproduce this issue?

1. Create following specs file `openapi.yaml`

```yaml
openapi: 3.0.0
info:
  title: test
  version: 0.0.1
paths:
  /test:
    get:
      tags:
        - TestController
      responses:
        '200':
          description: Test
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/test'
      parameters:
        - name: where
          in: query
          content:
            application/json:
              schema:
                type: object
                title: test.WhereFilter
                additionalProperties: true
      operationId: test
servers:
  - url: http://localhost:3000/api
components:
  schemas:
    test:
      type: object
      title: test
      properties:
        test:
          type: number
```

2. Run Orval v6.2.4

```sh
npx orval --input ./openapi.yaml --output ./openapi.ts 
```

## What happens?

Got following error

```txt
🍻 Start orval v6.2.4 - A swagger client generator for typescript
🛑  TypeError: Cannot read properties of undefined (reading '$ref')
```

## Comments

The problem is caused by `parameters`, I think Orval expect `$ref` after `content > application/json > schema`

Thanks

mwlazlo:
I also ran into this bug this morning +1

tynanspense:
This is also a problem I'm facing +1

anymaniax:
Should be fixed with `6.4.0` can you test when you have time, please?

anymaniax:
Closing this issue since it's normally fixed. Don't hesitate if you have any problem

PR #250, opened 2021-11-08
fix(query): check content for query schema

mwlazlo:
## Status
**READY**

## Description
Fix issue #247 

## Steps to Test or Reproduce
check issue #247

### DIFF
--- a/src/core/getters/queryParams.ts
+++ b/src/core/getters/queryParams.ts
@@ -1,4 +1,4 @@
-import { SchemaObject } from 'openapi3-ts';
+import { ContentObject, SchemaObject } from 'openapi3-ts';
 import { ContextSpecs } from '../../types';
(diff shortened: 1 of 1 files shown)


anymaniax:
Hello @mwlazlo, thanks for the pull request. I didn't know that a parameter could have a content there are some cases where it's needed instead of a schema directly? Also would be interesting to not only handle `application/json` in that case no?

mwlazlo:
Hi @anymaniax! Yes, we didn't realise it either until the generator tool we use for backend code failed to emit a parser for an object unless we specified the mime type.

Currently, Orval will automatically emit application/json if the schema is an object, so this fix is equivalent while supporting the spec (admittedly only for this specific use case).

Besides, I have no idea how or why you would render an object as, say, text/plain :-)

anymaniax (approved):
Thanks 🎉

PR #255, opened 2021-11-10
fix: add support for delete body

RichardLindhout:
Fixes https://github.com/anymaniax/orval/issues/254

### DIFF
--- a/src/constants/index.ts
+++ b/src/constants/index.ts
@@ -15,6 +15,6 @@ export const generalJSTypesWithArray = generalJSTypes.reduce<string[]>(
   [],
 );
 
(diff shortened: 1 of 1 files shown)


anymaniax:
Thanks for the merge request but it will cause a problem with the Axios generation that don't accept a body for a delete method

anymaniax:
Will also need to update the client generators to handle that case and pass it in the axios config with the property data

anymaniax:
I just did the implementation and it will be in the next release so I close this pull request. Thanks

RichardLindhout:
Thank you!

PR #260, opened 2021-11-18
fix(generator): apply nullish? operator in schema.ts

Talor-A:
## Status

**READY**

## Description
fixes https://github.com/anymaniax/orval/issues/259 . I don't know if this is a "proper" fix but it solves the example reproduction case in the issue body. 

## Todos
- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce
see issue for details. the files specified will fail yo

### DIFF
--- a/src/core/getters/schema.ts
+++ b/src/core/getters/schema.ts
@@ -8,13 +8,13 @@ export const getSchema = (
   specKey?: string,
 ) => {
   const schemas = Object.entries(
-    context.specs[specKey || context.specKey].components?.schemas || [],
+    context.specs[specKey || context.specKey]?.components?.schemas || [],
   ).reduce((acc, [name, type]) => ({ ...acc, [name]: type }), {}) as {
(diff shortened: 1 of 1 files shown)


Talor-A:
actually, this fixes the example case but in my spec (it's proprietary) I get only interfaces generated, and no hooks/methods/mocks even though the generation succeeds.

Talor-A:
converting to draft, need a better solution.

Talor-A:
reopening as I experienced 2 related bugs, this fixes the first.

anymaniax:
Hey @Talor-A, mmh weird this case shouldn't happen I need to investigate that more. Will have time this weekend normally

Talor-A:
ok no problem. I've pulled this commit into my fork as it fixes it for my situation, but I don't doubt there's a better solution. thanks for the response!

anymaniax:
thanks fixed with an other pull request

PR #262, opened 2021-11-19
feat(generator): cast return type of queries

Talor-A:
## Status

**READY**

## Description
Typescript currently doesn't know how to make sense of the type that results from:
```tsx
{
  ...query,
  queryKey
}
```

intellisense in many cases thinks that it's not assignable to `UseQueryResult`. an explicit cast fixes this.

## Todos
- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce
before: the return type is a union of tons of different objects.

Screen Shot 2021-11-18 at 5 27 34 PM

after: the return type is `useQueryResult` | `useInfiniteQueryResult`
Screen Shot 2021-11-18 at 5 28 14 PM

### DIFF
--- a/samples/react-query/basic/src/api/endpoints/petstoreFromFileSpecWithTransformer.ts
+++ b/samples/react-query/basic/src/api/endpoints/petstoreFromFileSpecWithTransformer.ts
@@ -9,6 +9,9 @@ import {
   useInfiniteQuery,
   useMutation,
   UseQueryOptions,
+  UseQueryResult,
+  UseInfiniteQueryResult,
+  QueryKey,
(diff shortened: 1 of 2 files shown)


anymaniax:
Hey @Talor-A, thanks for the pr. Instead of the cast, it would be better to add a return type no?

Talor-A:
Yeah that makes sense! I normally wouldn’t want to cast like this in my code (because sometimes it can lead to bugs) but I figured since this is all generated it’s ok. But I agree, no need to use the cast when a return type annotation will do. I’ll adjust it.

anymaniax:
Hey, thanks! I did it in another pull request

PR #277, opened 2021-12-07
feat(generators): add provde in option angular client

anymaniax:


kfrederix:
Hello, super happy with these changes, thank you for this!
Just 1 question:
I don't really see how the removal of `generateQueryFormDataFunction` is related to this. Was it just a clean-up on the side of something which was not actually used?

anymaniax:
@kfrederix it's unused code that I cleanup

PR #278, opened 2021-12-07
fix(core): path parameters default values

DavideWalder:
## Status

**READY**

## Description
Default string values were rendered without quotes.
Before:
```typescript
export const listPetsByCountry = <TData = AxiosResponse<Pets>>(
  params?: ListPetsByCountryParams,
  country = UY, // <---
  options?: AxiosRequestConfig,
): Promise<TData> => {
  return axios.get(`/pets-by-country/${country}`, {
    params,
    ...options,
  });
};
```
After:
```typescript
export const listPetsByCountry = <TData = AxiosResponse<Pets>>(
  params?: ListPetsByCountryParams,
  country = 'UY', // <---
  options?: AxiosRequestConfig,
): Promise<TData> => {
  return axios.get(`/pets-by-country/${country}`, {
    params,
    ...options,
  });
};
```

I implemented it in a separate function so that it can be expanded if other types need special formatting.
The tests cover strings (as enums) and integers and use a separate definition.
## Todos
- [x] Tests
- [ ] Documentation

## Steps to Test or Reproduce
1.
```bash
> git fetch --all
> git checkout endpoint-parameters
> yarn; yarn build
> cd tests
> yarn; yarn generate
```
2. Check `tests/generated/default/endpointParameters/endpoints.ts`

### DIFF
--- a/src/core/getters/params.ts
+++ b/src/core/getters/params.ts
@@ -1,6 +1,6 @@
 import { ContextSpecs } from '../../types';
 import { GetterParameters, GetterParams } from '../../types/getters';
-import { sanitize } from '../../utils/string';
+import { sanitize, stringify } from '../../utils/string';
 import { resolveValue } from '../resolvers/value';
 
 /**
@@ -23,20 +23,6 @@ export const getParamsInPath = (path: string) => {
   }
 
   return output;
-};
-
-/**
- * objToString handles the rendering for specific types
- * 
- * @param obj
- */
-const objToString = (obj: any): string => {
-  switch (typeof obj) {
-    case 'string':
-      return `'${obj}'`;
-    default:
-      return `${obj}`;
-  }
 };
 
 export const getParams = ({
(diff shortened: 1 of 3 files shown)


src/core/getters/params.ts
```diff
@@ -89,7 +103,7 @@ export const getParams = ({
       }${
         !resolvedValue.originalSchema!.default
           ? `: ${resolvedValue.value}`
-          : `= ${resolvedValue.originalSchema!.default}`
+          : `= ${objToString(resolvedValue.originalSchema!.default)}`
```
anymaniax: I already have a util function for that you can find it [here](https://github.com/anymaniax/orval/blob/master/src/utils/string.ts#L11)

anymaniax (changes requested):

src/core/getters/params.ts
DavideWalder: Thanks, exchanged it

anymaniax (approved):
thanks 🎉

PR #284, opened 2021-12-16
fix(types): remove generic from axios query function generator

### LINKED ISSUE #283: Get rid of TData generic in query function declaration
Right now every query function has a possibility to have on overwritten return type by using a `TData` generic. There is the [responsible code](https://github.com/anymaniax/orval/blob/bb205cb7505fdcda5ccc206ad508842b73722c00/src/core/generators/axios.ts#L86-L97). This generic looks strange because query functions will always return the same data. Also there is another problem with this approach: typescript unable to calculate queries return type dynamically because there is no guarantee that they wasn't overwritten.

I think it should look like this:
```javascript
return `const ${operationName} = (\n ${toObjectString(props, 'implementation')}\n ${
      isRequestOptions && isMutatorHasSecondArg
        ? `options?: SecondParameter<typeof ${mutator.name}>`
        : ''
    }) => {${bodyForm}
      return ${mutator.name}<${response.definition.success || 'unknown'}>(
      ${mutatorConfig},
      ${requestOptions});
    }
  `;
```

yekver:
Fix #283

### DIFF
--- a/src/constants/index.ts
+++ b/src/constants/index.ts
@@ -15,6 +15,12 @@ export const generalJSTypesWithArray = generalJSTypes.reduce<string[]>(
   [],
 );
 
-export const VERBS_WITH_BODY = [Verbs.POST, Verbs.PUT, Verbs.PATCH, Verbs.DELETE];
+export const VERBS_WITH_BODY = [
+  Verbs.POST,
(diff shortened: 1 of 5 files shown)


anymaniax (approved):
Thanks !

PR #287, opened 2021-12-17
fix(msw): declaration 'req' is exempted from the unused parameter che…

wouterkroes:
## Status
READY

## Description
Declaration 'req' is [exempted](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#flag-unused-declarations-with---nounusedparameters-and---nounusedlocals) from the unused parameter checking in handler.


anymaniax (approved):
Thanks!

PR #292, opened 2021-12-18
fix(code): reducers speed & memory optimisation

yekver:
Global refactoring to avoid coping all data on every reducer iteration.

This consumes lots of memory and work as `O(n^2)`. Current optimization increases reducers speed to `O(n)`

Also some TS types were fixed a bit.


anymaniax:
Thanks a lot for the contribution @yekver. I am a big fan of functional programming but I have to admit that this way it will be way more performant. And I have also the impression that it's more readable like this.

anymaniax (approved):
Thanks! Really clean pull request

PR #297, opened 2022-01-14
Add eslint + tslint ignore to generated files

wouterkroes:
## Status
READY

## Description
Add eslint + tslint ignore to generated files

**! I tried running my change locally with `npm run dev` but I get an error**:

```sh
orval git:(master) npm run dev                                                                      

> orval@6.5.1 dev
> tsup ./src/bin/orval.ts ./src/index.ts --clean --watch src --onSuccess 'yarn generate-api'

CLI Building entry: ./src/bin/orval.ts, ./src/index.ts
CLI Using tsconfig: tsconfig.json
CLI tsup v5.10.0
CLI Running in watch mode
CLI Target: node12
CLI Cleaning output folder
CJS Build start
CJS Build success in 96ms
CLI Watching for changes in "src"
CLI Ignoring changes in "**/{.git,node_modules}/**" | "dist"
yarn run v1.22.0
$ node ./dist/bin/orval.js --config ./samples/react-query/basic/orval.config.ts --watch
🍻 Start orval v6.5.1 - A swagger client generator for typescript
node:internal/process/promises:246
          triggerUncaughtException(err, true /* fromPromise */);
          ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "failed to load from /Users/wouter.kroes/Projects/orval/samples/react-query/basic/orval.config.ts => Error: Cannot find module 'orval'
Require stack:
- /Users/wouter/Projects/orval/samples/react-query/basic/orval.config.ts
- /Users/wouter/Projects/orval/dist/bin/orval.js".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
node.js v16.13.2

## Related PRs
n/a

branch | PR
------ | ------
n/a

## Todos
- [x] Tests
- [x] Documentation


anymaniax:
You normally doesn't need this. You have some problems in the code generated? If you really need to do that you can also do it with [this](https://orval.dev/reference/configuration/output#header)

wouterkroes:
In my project ESLint extends `airbnb-base`, which has the rule https://github.com/airbnb/javascript#modules--prefer-default-export.  This conflicts with the generated `index.ts`, that exports a named const/function.

At the moment I have added the Orval generated folder to the `.eslintignore`.
I did not think of doing it your way, with the output#header.
Thanks for the suggestion!

But It still makes me wonder; the Orval files are generated, why not add these ignore lines?
A MSW does this for instance https://github.com/mswjs/msw/blob/main/src/mockServiceWorker.js

Would you be so kind and explain why you rather not add these ignore lines of code?

anymaniax:
Hey @wouterkroes, you are right I don't really have any issue with adding those two lines. I just really try to keep the generated code similar to what I would write If I needed to do it myself. Like that, if tomorrow you have an issue or need to stop to generate the code that's not a problem.

wouterkroes:
@anymaniax thanks for the reply.

And I totally get the hesitation about addings the ignore lines. 
Both choices are valid choices to make 😉 

Regardless of the choice, thanks for this great generator!

anymaniax:
can be added with the orval config and I prefer that it's stay like that

PR #298, opened 2022-01-20
fix(getters): enum can contain dashes

thhermansen:
## Status

**READY**

## Description

This PR fixes generated TS code when an enum contains a dash. Example of usage will be query param `sort`: 

```yaml
paths:
  /pets:
    get:
      summary: List all pets
      operationId: listPets
      tags:
        - pets
      parameters:
        - name: limit
          in: query
          required: false
          schema:
            type: string
        - name: sort
          in: query
          description: |
            Which property to sort by?
            Example: name sorts ASC while -name sorts DESC.
          required: false
          schema:
            type: string
            enum:
              - name
              - -name
              - email
              - -email
```

### Generated code before

```ts
export const ListPetsSort = {
  name: 'name' as ListPetsSort,
  name: '-name' as ListPetsSort, // Note the duplicated key
  email: 'email' as ListPetsSort,
  email: '-email' as ListPetsSort, // Note the duplicated key
};
```

### Generated code after

```ts
export const ListPetsSort = {
  name: 'name' as ListPetsSort,
  '-name': '-name' as ListPetsSort,
  email: 'email' as ListPetsSort,
  '-email': '-email' as ListPetsSort,
};
```

Which gets us with the desired result:

image

## Todos
- [x] Tests
- [x] Documentation

### DIFF
--- a/src/core/getters/enum.ts
+++ b/src/core/getters/enum.ts
@@ -26,7 +26,7 @@ export const getEnumImplementation = (
     const key =
       isNumber || isTypeNumber
         ? toNumberKey(isTypeNumber ? val.toString() : val.slice(1, -1))
-        : sanitize(val, { underscore: '_', whitespace: '_' });
+        : sanitize(val, { underscore: '_', whitespace: '_', dash: '-' });
 
     return (
       acc +
--- a/src/utils/string.ts
+++ b/src/utils/string.ts
@@ -51,10 +51,11 @@ export const sanitize = (
     whitespace?: string | true;
     underscore?: string | true;
     dot?: string | true;
+    dash?: string | true;
   },
 ) => {
-  const { whitespace = '', underscore = '', dot = '' } = options || {};
-  let newValue = value.replace(/[^\w\s.]/g, '');
+  const { whitespace = '', underscore = '', dot = '', dash = '' } = options || {};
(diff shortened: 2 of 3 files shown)


anymaniax (approved):
Thanks!

thhermansen:
I think it is appropriate for me to thank you too :-)

Our team tried https://openapi-generator.tech/ before we found Orval and this seems so generate simpler code and also could generate code for useSWR out of the box so it seems good. Hopefully it works with all our API specifications without too much trouble.

### A bit off topic

We had a couple of other issues with that our OpenAPI spec, generated from [fastify-swagger](https://github.com/fastify/fastify-swagger). Specifically:

```yml
completedAt:
  anyOf:
    - format: date-time
      type: string
      nullable: true
    - type: 'null'
```

where it didn't like the `- type: 'null'` (I *think* this is an OpenAPI 3.1 feature) and:

```yml
status:
  anyOf:
    - type: string
      enum:
        - PENDING
    - type: string
      enum:
        - SUCCESS
    - type: string
      enum:
        - FAILED
```

I made both of these cases work by a little input transformer magic, and I presume at least the first one will work with support of OpenAPI 3.1 (?).

anymaniax:
Thanks for the feedback! You can create an issue for both of them I will check when I have a bit of time.

PR #302, opened 2022-01-20
feat(axios): axios factory function accepts an axios instance

thhermansen:
## Status

**READY**

## Description

This provides a dynamic alternative to
the configuration option `output.override.mutator`
where we can provide a axios instance to
make http requests through.

Old code generated:

```ts
  export const getSwaggerPetstore = () => {
/**
 * @summary List all pets
 */
const listPets = <TData = AxiosResponse<Pets>>(
    params?: ListPetsParams,
    version= 1, options?: AxiosRequestConfig
 ): Promise<TData> => {
    return axios.get(
      `/v${version}/pets`,{
        params,
    ...options}
    );
  }
```

New code generates:

```ts
  export const getSwaggerPetstore = ({axios: localAxios = axios}: {axios?: AxiosInstance} = {}) => {
/**
 * @summary List all pets
 */
const listPets = <TData = AxiosResponse<Pets>>(
    params?: ListPetsParams,
    version= 1, options?: AxiosRequestConfig
 ): Promise<TData> => {
    return localAxios.get(
      `/v${version}/pets`,{
        params,
    ...options}
    );
  }
```

`localAxios` is set to global axios, so I believe the old behaviour is kept.

## Why?

Be able to do something like this:

```ts
import {getSwaggerPetstore} from 'generated'

export function prepareApi(someArgDictatingAuthStrategyAndOtherAxiosConfigs: any) {
  const axios = Axios.create()

  // Configure axois instance according to someArgDictatingAuthStrategyAndOtherAxiosConfigs

  return getSwaggerPetstore({axios})
}
```

In other words, use a specific Axios instance each time `prepareApi` is called.

## Question

Maybe I have missed something and there is a better way of archiving this?

## Todos
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/core/generators/axios.ts
+++ b/src/core/generators/axios.ts
@@ -26,6 +26,7 @@ const AXIOS_DEPENDENCIES: GeneratorDependency[] = [
       },
       { name: 'AxiosRequestConfig' },
       { name: 'AxiosResponse' },
+      { name: 'AxiosInstance' },
     ],
     dependency: 'axios',
   },
@@ -46,7 +47,7 @@ const generateAxiosImplementation = (
     formData,
     formUrlEncoded,
   }: GeneratorVerbOptions,
-  { route, context }: GeneratorOptions,
+  { route, context, noFunction }: GeneratorOptions & {noFunction?: boolean},
 ) => {
   const isRequestOptions = override?.requestOptions !== false;
   const isFormData = override?.formData !== false;
@@ -114,7 +115,7 @@ const generateAxiosImplementation = (
   }>>(\n    ${toObjectString(props, 'implementation')} ${
     isRequestOptions ? `options?: AxiosRequestConfig\n` : ''
   } ): Promise<TData> => {${bodyForm}
-    return axios${
+    return ${noFunction === true ? 'axios' : 'localAxios'}${
       !isSyntheticDefaultImportsAllowed ? '.default' : ''
     }.${verb}(${options});
   }
@@ -147,14 +148,14 @@ export const generateAxiosHeader = ({
   : never;\n\n`
     : ''
 }
-  ${!noFunction ? `export const ${title} = () => {\n` : ''}`;
+  ${!noFunction ? `export const ${title} = ({axios: localAxios = axios}: {axios?: AxiosInstance} = {}) => {\n` : ''}`;
 
 export const generateAxiosFooter = (operationNames: string[] = []) =>
   `return {${operationNames.join(',')}}};\n`;
 
 export const generateAxios = (
   verbOptions: GeneratorVerbOptions,
-  options: GeneratorOptions,
+  options: GeneratorOptions & {noFunction?: boolean},
 ): GeneratorClient => {
   const imports = generateVerbImports(verbOptions);
   const implementation = generateAxiosImplementation(verbOptions, options);
--- a/src/core/generators/client.ts
+++ b/src/core/generators/client.ts
@@ -56,7 +56,7 @@ export const GENERATOR_CLIENT: GeneratorClients = {
   },
   'axios-functions': {
     client: (verbOptions: GeneratorVerbOptions, options: GeneratorOptions) => {
-      const { implementation, imports } = generateAxios(verbOptions, options);
+      const { implementation, imports } = generateAxios(verbOptions, {...options, noFunction: true });
 
       return {
         implementation: 'export ' + implementation,


anymaniax:
Hey @thhermansen, Thanks for the pull request. Do you have a blocking point with the mutator that you need to pass the axios instance like this?

thhermansen:
Hi,

We want bind a custom Axios instance when we call the function to retrieve the object of api functions (example: `getSwaggerPetstore()`). The Axios instance will be different each time we retrieve the set of api functions.

I guess one solution with the current version could be:

1. Set up a custom client with mutator which looks something like this:

```ts 
import Axios, {AxiosInstance, AxiosRequestConfig} from 'axios'

export const customInstance = async <TReturn>(
  config: AxiosRequestConfig,
  {axios, ...axiosRequestConfig}: AxiosRequestConfig & {axios: AxiosInstance}
): Promise<TReturn> => {
  try {
    const {data} = await axios({...config, ...axiosRequestConfig})
    return data
  } catch (error) {
    if (Axios.isAxiosError(error)) throw new DriftiClientError(error)
    throw error
  }
}
```

2. Generated api function will look like this:

```ts
const getFile = (id: string, options?: SecondParameter<typeof customInstance>) => {
  return customInstance<GetFile200>( {url: `/file/${id}/info`, method: 'get' }, options);
}
```

2. Make a factory function which takes an axios instance and calls the generated function to retrieve the object of API functions. We must now create new functions for each and return a object with "wrapped" functions so that for instance when the user calls `getFile('1')` the generated `getFile` function is called with: `getFile('1', {axios: axiosInstancePassedInToFactoryFunction})`.

This seems a bit complex rater than just leveraging Javascript's function scope binding. I also see a challenge to this approach if if the api call functions signature is like this, and it is not called by an id:

```ts
const getAllOrById = (id?: string, options?: SecondParameter<typeof customInstance>) => {
  return customInstance( {url: `/some`, ... }, options);
}
```

In order to figure out which argument position to inject axios to in this case would require .. hmm .. I'm not sure actually. One suggestion was to `toString()` the function and "parse" it to reflect on it's number of arguments.

### tl;dr

I feel like this is harder than it needs to be to overcome with the mutator approach I have looked at so far. Maybe there is a simpler solution than I have been able to come up with using it? If there isn't would be happy for this PR being considered merged.

src/core/generators/axios.ts
```diff
@@ -147,14 +148,14 @@ export const generateAxiosHeader = ({
   : never;\n\n`
     : ''
 }
-  ${!noFunction ? `export const ${title} = () => {\n` : ''}`;
+  ${!noFunction ? `export const ${title} = ({axios: localAxios = axios}: {axios?: AxiosInstance} = {}) => {\n` : ''}`;
```
anymaniax, line 151: I think that we also need to check if there is a global mutator here. Because if that's the case we don't use the axios instance

thhermansen:
Let me try a bit more with the current version of Orval to see if I can work around it as I tried to describe above. I'll write an update here when I have tried a bit more.

thhermansen:
I think we managed to get it working the way we wanted. I have provided the solution for anyone who ends up in this PR having a similar need. 

We used the following output mutator:

```ts
import Axios, {AxiosInstance, AxiosRequestConfig} from 'axios'

export const customInstance = async <TReturn>(
  config: AxiosRequestConfig,
  options?: AxiosRequestConfig & {axios?: AxiosInstance}
): Promise<TReturn> => {
  const {axios, ...rest} = options ?? {}
  if (!axios) throw new Error('Missing Axios')

  const {data} = await axios({...config, ...rest})
  return data
}
```

And then defined a factory function like this which takes a set of Orval's api functions:

```ts
type OrvalApiFactoryFunction = () => Record<string, (...args: any[]) => any>
type OrvalApiFactoryMap = Record<string, OrvalApiFactoryFunction>

function injectAxiosIntoLastArgument(
  fn: (...args: any[]) => any,
  {axios, apiName, apiFunctionName}: {axios: AxiosInstance; apiName: string; apiFunctionName: string}
) {
  return (...args: any[]) => {
    const lastArgIndex = fn.length - 1
    const lastArg = args[lastArgIndex]

    if (lastArg !== undefined && !(lastArg !== null && typeof lastArg === 'object')) {
      throw new Error(
        `Expected last argument in ${apiName}.${apiFunctionName} to be undefined or an object. Cannot inject Axios instance as value was: ${typeof lastArg}`
      )
    }

    args[lastArgIndex] = {...lastArg, axios}

    return fn(...args)
  }
}

export function apiClient<
  TApiFactoryMap extends OrvalApiFactoryMap,
  K extends keyof TApiFactoryMap = keyof TApiFactoryMap,
  TReturnApis = {[k in K]: ReturnType<TApiFactoryMap[k]>}
>(
  options: CreateApiClient,
  apisToWrap: TApiFactoryMap = {} as TApiFactoryMap
): TReturnApis & {
  meta: {
    /** The axios instance used by APIs to make requests */
    axios: AxiosInstance
  }
} {
  const config = options.config ?? new CoreConfig(process.env)

  const axios = options.axios ?? Axios.create()
  axios.defaults.baseURL = config.baseUrl
  axios.interceptors.request.use(makeAuthInterceptor({...options, config}))

  const apis: TReturnApis = Object.keys(apisToWrap).reduce((apisOut, apiName) => {
    const apiFunctions = apisToWrap[apiName]()

    apisOut[apiName] = Object.keys(apiFunctions).reduce((functionsOut, apiFunctionName) => {
      const apiFn = apiFunctions[apiFunctionName]
      functionsOut[apiFunctionName] = injectAxiosIntoLastArgument(apiFn, {axios, apiName, apiFunctionName})

      return functionsOut
    }, {} as any)

    return apisOut
  }, {} as any)

  return {...apis, meta: {axios}}
}
```

So now we can take for example `getSwaggerPetstore` and do this:

```ts
const client = apiClient(
  {some: 'options', to: 'configure', axios: 'auth interceptor'}, 
  {pets: getSwaggerPetstore}
)

const data = await client.pets.listPets() // client and functions is typed. Axios is passed when called
```

anymaniax:
Ok cool that you found a work around. I will still check to add this feature when I have a bit of time.

PR #307, opened 2022-02-03
Fix #106: Deserialize to JS Date

### LINKED ISSUE #106: Deserializing OpenAPI date and datetimes as JavaScript Date objects
I would like to add an option to the library, so that the user can deserialize date and datetime objects ([both have type `string` and format `date` and `date-time`, respectively](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#data-types)) as JavaScript `Date` object.

I have already tested it and this change in `scalar.ts` seems to actually be the only change needed (to be added inside the `string` case statement):

```
      if(item.format === 'date' || item.format === 'date-time') {
        return {
          value: 'Date' + nullable,
          isEnum: false,
          type: 'string',
          imports: [],
          schemas: [],
        };
      }
```

This would only support ISO 8601 dates and date times (default in the OpenAPI specification) but that is what most of the people use in the REST APIs these days.

However, there will also be change needed in the Axios instance, so that dates are properly deserialized as `Date` objects (currently these are `string`s).

Would you be open for such change? Do you think we should make it optional, via dedicated field in the configuration object?

melloware:
## Status

**READY**

## Description
Converts OpenAPI Dates to Javascript Date

## Related PRs
NONE

## Todos
- [ ] Tests
- [ ] Documentation


src/core/getters/scalar.ts
```diff
@@ -87,6 +87,10 @@ export const getScalar = async ({
         value = 'Blob';
       }
 
+      if (item.format === 'date' || item.format === 'date-time') {
```
anymaniax: Hey, thanks for the pull request. It's not only that. Axios should also parse the string into a js Date then

anymaniax: It should also be an option in the config because it could be a breaking change for some codebase

melloware: I will work on that

melloware:
Ahh yes I see in my OpenAPi TypeScript Fetch code they do this with the model object.

```typescript
export interface TrainingRecord {
    /**
     * 
     * @type {string}
     * @memberof TrainingRecord
     */
    mudid?: string;
    /**
     * 
     * @type {Date}
     * @memberof TrainingRecord
     */
    timestamp?: Date;
    /**
     * 
     * @type {Array<string>}
     * @memberof TrainingRecord
     */
    courses?: Array<string>;
}

export function TrainingRecordFromJSON(json: any): TrainingRecord {
    return TrainingRecordFromJSONTyped(json, false);
}

export function TrainingRecordFromJSONTyped(json: any, ignoreDiscriminator: boolean): TrainingRecord {
    if ((json === undefined) || (json === null)) {
        return json;
    }
    return {
        
        'mudid': !exists(json, 'mudid') ? undefined : json['mudid'],
        'timestamp': !exists(json, 'timestamp') ? undefined : (new Date(json['timestamp'])),
        'courses': !exists(json, 'courses') ? undefined : json['courses'],
    };
}

export function TrainingRecordToJSON(value?: TrainingRecord | null): any {
    if (value === undefined) {
        return undefined;
    }
    if (value === null) {
        return null;
    }
    return {
        
        'mudid': value.mudid,
        'timestamp': value.timestamp === undefined ? undefined : (value.timestamp.toISOString()),
        'courses': value.courses,
    };
}
```

melloware:
OK so this guys approach is best using Axios interceptor.

https://stackoverflow.com/a/66238542/502366

Going to update this PR.

melloware:
OK I think I did this right I did the following:

1. Add a new `override` option for `useDates`
2. If `useDates=true` we convert the TypeScript object to Date
3. Updated the docs to show how to use this configuration with Axios Intercept to convert the dates

src/types/index.ts
```diff
@@ -234,6 +235,7 @@ export type OverrideOutput = {
     verb: Verbs,
   ) => string;
   requestOptions?: Record<string, any> | boolean;
+  useDates: boolean;
```
anymaniax: it's not required here

docs/src/pages/reference/configuration/output.md
```diff
@@ -822,3 +822,57 @@ export const customFormUrlEncodedFn = <Body>(body: Body): URLSearchParams => {
...
+  return value && typeof value === "string" && isoDateFormat.test(value);
+}
+
+export function handleDates(body: any) {
+  if (body === null || body === undefined || typeof body !== "object")
+    return body;
+
+  for (const key of Object.keys(body)) {
+    const value = body[key];
+    if (isIsoDateString(value)) body[key] = parseISO(value);
```
anymaniax: from where come `parseISO`?

anymaniax:
Thanks a lot for the work!

melloware:
Removed `parseIso` which is from `date-fns` library for just `new Date()` which assumes its an ISO string.

src/types/index.ts
melloware: Removed

docs/src/pages/reference/configuration/output.md
melloware: Switched to `new Date()`

melloware:
I had to put useDates back as your build fails with `Error: src/utils/options.ts(165,43): error TS2339: Property 'useDates' does not exist on type 'OverrideOutput'.` without it.

anymaniax:
Hey @melloware, the useDates should be there but optional so just like this useDates?:boolean;

melloware:
Added text and even commented out examples for Luxon, MomentJS, etc.  And changed `useDates?` to optional.

anymaniax (approved):

anymaniax:
Thanks !!! 🎉

Thullner:
@melloware Where can i find this? Because i would like to add conversion to Day.js.

`Added text and even commented out examples for Luxon, MomentJS, etc. And changed useDates? to optional.
`

melloware:
@Thullner 

#### useDates

Type: `Boolean`

Valid values: true or false. Defaults to false.

Use this property to convert OpenAPI `date` or `datetime` to JavaScript `Date` objects instead of `string`.


Example:

```js
module.exports = {
  petstore: {
    output: {
      override: {
        useDates: true,
      },
    },
  },
};
```

**Note:** You must provide and Axios converter to convert these to dates as this just makes the TypeScript definition a Date.
You can choose to use any Date library you want like Moment, Luxon, or native JS Dates.

```ts
// type signature
const client = axios.create({
    baseURL: ''
});

client.interceptors.response.use(originalResponse => {
    handleDates(originalResponse.data);
    return originalResponse;
});

export default client;

const isoDateFormat = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d*)?(?:[-+]\d{2}:?\d{2}|Z)?$/;

function isIsoDateString(value: any): boolean {
    return value && typeof value === "string" && isoDateFormat.test(value);
}

export function handleDates(body: any) {
    if (body === null || body === undefined || typeof body !== "object")
        return body;

    for (const key of Object.keys(body)) {
        const value = body[key];
        if (isIsoDateString(value)) {
            body[key] = new Date(value); // default JS conversion
            // body[key] = parseISO(value); // date-fns conversion
            // body[key] = luxon.DateTime.fromISO(value); // Luxon conversion
            // body[key] = moment(value).toDate(); // Moment.js conversion
        } else if (typeof value === "object") {
            handleDates(value);
        }
    }
}
```

PR #311, opened 2022-02-08
fix(config): allow to pass an operationId modifier in config

CPatchane:
## Status

**READY**

## Description
Like we allow to set `operationName`, we should be able to customize the way that `operationId` are generated from the endpoint path.

I use the v6.4.3 codebase since te 6.5 doesn't work on my side due to [that issue](https://github.com/anymaniax/orval/issues/301)

## Todos
- [ ] Tests: I don't see any tests about the config, is there any?
- [x] Documentation
- [ ] Changelog Entry (unreleased) Where is it?

## Steps to Test or Reproduce
```bash
> git pull --prune
> git checkout <branch>
> grunt jasmine
```
1.

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -732,6 +732,42 @@ Type: `Object`.
 
 Exactly the same as the `override.operations` but this time you can do it by <a href="https://swagger.io/docs/specification/grouping-operations-with-tags/" target="_blank">tags</a>
 
+#### operationId
+
+Type: `Function`.
+
+Function to override the generate operation id. The id is then used to generate all types and functions names.
+You can see <a href="https://github.com/anymaniax/orval/blob/next/src/core/getters/operation.ts" target="_blank">here</a> how we compute that id by default (`getOperationId`).
+
+**Important notice:** The id returned here should always be unique by route and verb otherwise you'll get naming conflicts when generating types/functions.
+
+
+```ts
+// type signature
+(operation: OperationObject, route: string, verb: Verbs) => string;
+// Example
+import { capitalize, camelCase } from 'lodash';
+const sanitizeId = (str) =>
+  str.replace(/[^\w{}]/g, '').replace(/{(\w+)}/g, 'By-$1');
+
+module.exports = {
+  petstore: {
+    output: {
+      override: {
+        // From: /api/endpoint/objects/{id}/index
+        // To: <verb>ApiEndpointObjectsByIdIndex
+        opearationId: (operation, route, verb) => {
+          return capitalize(
+            camelCase(
+              [verb, ...route.split('/').map((p) => sanitizeId(p))].join('-'),
+            ),
+          );
+        },
+      },
+    },
+  },
+};
+```
 #### operationName
 
 Type: `Function`.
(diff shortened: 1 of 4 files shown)


docs/src/pages/reference/configuration/output.md
```diff
@@ -726,6 +726,20 @@ Type: `Object`.
...
+
+```ts
+// type signature
+(operation: OperationObject, route: string, verb: Verbs) => string;
+```
+
+Function to override the generate operation id. The id is then used to generate all types and functions names.
+You can see <a href="https://github.com/anymaniax/orval/blob/next/src/core/getters/operation.ts" target="_blank">here</a> how we compute that id by default (`getOperationId`).
+
+__Important notice:__ The id returned here should always be unique by route and verb otherwise you'll get naming conflicts when generating types/functions.
```
melloware: Can you add a configuration example showing how you would use this?    I think that would help users understand for example what you would do.

src/types/index.ts
```diff
@@ -81,6 +81,11 @@ export type NormalizedOverrideOutput = {
     route: string,
     verb: Verbs,
   ) => string;
+  operationId?: (
+    operation: OperationObject,
+    route: string,
+    verb: Verbs,
+  ) => string;
```
melloware, line 89: Will this help with my issue also: https://github.com/anymaniax/orval/issues/305  ???

melloware (changes requested):

docs/src/pages/reference/configuration/output.md
CPatchane: Sure, I added with an actual example 👍

src/types/index.ts
CPatchane: It could help but I don't think it's the right fix.
In your issue something is wrong, you shouldn't get a function with only `delete`, it should be `deleteAuthorizationAuthorizationId` if I understand correctly 🤔

anymaniax:
Hey @CPatchane, thanks for this pull request. I am not sure to see what this will add that the `operationName` config doesn't have? Since at the end the `operationName` is used to generate everything

CPatchane:
> Hey @CPatchane, thanks for this pull request. I am not sure to see what this will add that the operationName config doesn't have? Since at the end the operationName is used to generate everything

I tried the `operationName` before doing this but `operationName` only changes the functions (hooks, queries, ...) but not the models (types) while the `operationId` is more global to all naming generation. 
This might deprecate the `operationName` if we say that we shouldn't be able to change the functions naming without changing the models one.

anymaniax:
I thought it was already the case for the `operationName`. It's a bug but could also be a breaking change for some ppl

CPatchane:
For example, to generate the responses types, we use only the `operationId`: https://github.com/anymaniax/orval/blob/master/src/core/getters/response.ts#L29

anymaniax:
Yes, I did some tests and we need to change this [response](https://github.com/anymaniax/orval/blob/master/src/core/generators/verbsOptions.ts#L82) [body](https://github.com/anymaniax/orval/blob/master/src/core/generators/verbsOptions.ts#L84) and pass the `operationName` there

CPatchane:
> Yes, I did some tests and we need to change this [[response]](https://github.com/anymaniax/orval/blob/master/src/core/generators/verbsOptions.ts#L82) [[body]](https://github.com/anymaniax/orval/blob/master/src/core/generators/verbsOptions.ts#L84) and pass the operationName there

Cool then. Are you on it or do you want me to do a new PR?

anymaniax:
You can do a new pull request that fixes it. If you are ok with it

PR #313, opened 2022-02-09
fix(config): fix operationName bug not changing models naming

CPatchane:
## Status

**READY**

## Description
Close  #311 since that  was actually a bug with `operationName` config.


src/core/getters/body.ts
```diff
@@ -1,19 +1,18 @@
 import { ReferenceObject, RequestBodyObject } from 'openapi3-ts';
 import { generalJSTypesWithArray } from '../../constants';
 import { ContextSpecs } from '../../types';
-import { GeneratorImport, GeneratorSchema } from '../../types/generator';
```
CPatchane, line 4: Unused imports

anymaniax (approved):

PR #314, opened 2022-02-09
fix(server): fix operation names using ES5 keywords

melloware:
## Status
**READY**

## Description
Fixes any ES5 keyword that may be an operationName like `delete` becomes `_delete`

### DIFF
--- a/src/core/generators/verbsOptions.ts
+++ b/src/core/generators/verbsOptions.ts
@@ -5,6 +5,7 @@ import {
   PathItemObject,
   ReferenceObject,
 } from 'openapi3-ts';
+import { keyword } from 'esutils';
 import {
   ContextSpecs,
   NormalizedOperationOptions,
@@ -75,9 +76,12 @@ const generateVerbOptions = async ({
 
   const overrideOperationName =
     overrideOperation?.operationName || output.override?.operationName;
-  const operationName = overrideOperationName
+  const overriddenOperationName = overrideOperationName
     ? overrideOperationName(operation, route, verb)
     : camel(operationId);
+  const operationName = keyword.isKeyWordES5(overriddenOperationName)
+    ? `_${overriddenOperationName}`
+    : overriddenOperationName;
 
   const response = await getResponse(responses, operationName, context);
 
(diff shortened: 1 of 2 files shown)


src/core/generators/verbsOptions.ts
```diff
@@ -5,6 +5,7 @@ import {
   PathItemObject,
   ReferenceObject,
 } from 'openapi3-ts';
+import { keyword } from 'esutils';
```
anymaniax, line 8: you also need to update the esutils.d.ts to add the type of the function

melloware: Fixing.

melloware:
OK I don't understand the build failure?

src/core/generators/verbsOptions.ts
```diff
@@ -75,9 +76,12 @@ const generateVerbOptions = async ({
 
   const overrideOperationName =
     overrideOperation?.operationName || output.override?.operationName;
-  const operationName = overrideOperationName
+  const overriddenOperationName = overrideOperationName
     ? overrideOperationName(operation, route, verb)
     : camel(operationId);
+  const operationName = keyword.isKeyWordES5(overriddenOperationName, true)
```
anymaniax: it's `isKeywordES5 `

melloware: DOH

anymaniax (approved):
Thanks!

Pierstoval:
Hey folks, sorry to stumble a bit late on this, but I just generated an Axios-based client with Orval from an OpenAPI schema, and here I have parameters that are named `import` and it generates invalid Typescript as for the "delete" keyword before:

```typescript
/**
 * Edit an import by it's id
 * @summary Edit an import by it's id
 */
const apiimportsStyxIdPatch = <TData = AxiosResponse<Import>>(
    id: string,
    import: Import, options?: AxiosRequestConfig
 ): Promise<TData> => {
    return axios.patch(
      `/styx/imports/${id}`,
      import,options
    );
  }
```

This code is invalid because `import` is a reserved keyword.

How can I make sure it creates an `_import` variable?

melloware:
Looks like ES does not consider import a reserved word: https://github.com/estools/esutils/blob/master/lib/keyword.js

PR #318, opened 2022-02-11
feat(query): generate query return type along with hooks

CPatchane:
## Status

**READY**

## Description
When dealing with hooks generated using `react-query`, it's currently difficult to get the actual return types from the queries (not those from the schemas).
In a case where we process the data from the API (changing the case for example) using a custom client, we can't directly rely on the schemas types that have been generated, we have to get the response type from the queries generated instead.
This PR exported those new types for each functions in order to use them in components that used data from those, looking like:
```typescript
export type GetDemoQueryResult = NonNullable<AsyncReturnType<typeof getDemo>>
export type GetDemoQueryError = ErrorType<GetDemo400>

// Then we can use like
type MyComponentPropsType = {
  foo: GetDemoQueryResult["foo"]
  bar: GetDemoQueryResult["bar"]
}
```

Note: We have to use `NonNullable` for some cases when the client can return `undefined` (201 requests) and avoid type with `| undefined`. In that case, the final type will be `never`.

Wdyt?
I guess it could be considered as a breaking change, should we maybe put this behind a config property and make it optional?

### DIFF
--- a/src/core/generators/query.ts
+++ b/src/core/generators/query.ts
@@ -552,7 +552,16 @@ const generateQueryHook = (
       : response.definition.errors || 'unknown';
   }
 
+  const dataType = mutator?.isHook
+    ? `ReturnType<typeof use${pascal(operationName)}Hook>`
+    : `typeof ${operationName}`;
+
   return `
+    export type ${pascal(
+      operationName,
+    )}QueryData = NonNullable<AsyncReturnType<${dataType}>>
+    export type ${pascal(operationName)}QueryError = ${errorType}
+
     export const ${camel(`use-${operationName}`)} = <TError = ${errorType},
     ${!definitions ? `TVariables = void,` : ''}
     TContext = unknown>(${generateQueryArguments({
@@ -580,13 +589,9 @@ const generateQueryHook = (
       }
 
 
-      const mutationFn: MutationFunction<AsyncReturnType<${
-        mutator?.isHook
-          ? `ReturnType<typeof use${pascal(operationName)}Hook>`
-          : `typeof ${operationName}`
-      }>, ${definitions ? `{${definitions}}` : 'TVariables'}> = (${
-    properties ? 'props' : ''
-  }) => {
+      const mutationFn: MutationFunction<AsyncReturnType<${dataType}>, ${
+    definitions ? `{${definitions}}` : 'TVariables'
+  }> = (${properties ? 'props' : ''}) => {
           ${properties ? `const {${properties}} = props || {}` : ''};
 
           return  ${operationName}(${properties}${properties ? ',' : ''}${
--- a/src/core/generators/query.ts
+++ b/src/core/generators/query.ts
@@ -381,7 +381,7 @@ const generateQueryImplementation = ({
     : `typeof ${operationName}`;
 
   return `
-export type ${pascal(name)}QueryData = NonNullable<AsyncReturnType<${dataType}>>
+export type ${pascal(name)}QueryResult = NonNullable<AsyncReturnType<${dataType}>>
 export type ${pascal(name)}QueryError = ${errorType}
 
 export const ${camel(
@@ -559,7 +559,7 @@ const generateQueryHook = (
   return `
     export type ${pascal(
       operationName,
-    )}MutationData = NonNullable<AsyncReturnType<${dataType}>>
+    )}MutationResult = NonNullable<AsyncReturnType<${dataType}>>
     export type ${pascal(operationName)}MutationError = ${errorType}
 
     export const ${camel(`use-${operationName}`)} = <TError = ${errorType},
--- a/src/core/generators/query.ts
+++ b/src/core/generators/query.ts
@@ -376,18 +376,25 @@ const generateQueryImplementation = ({
       : response.definition.errors || 'unknown';
   }
 
+  const dataType = mutator?.isHook
+    ? `ReturnType<typeof use${pascal(operationName)}Hook>`
+    : `typeof ${operationName}`;
+
   return `
-export const ${camel(`use-${name}`)} = <TData = AsyncReturnType<${
-    mutator?.isHook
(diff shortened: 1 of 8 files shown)


src/core/generators/query.ts
```diff
@@ -562,6 +562,8 @@ const generateQueryHook = (
   }
 
   return `
+    export type ${pascal(operationName)}QueryError = ${errorType}
+
```
CPatchane, line 566: Is it expected that we can't return any data from a mutation?

src/core/generators/query.ts
```diff
@@ -555,6 +562,8 @@ const generateQueryHook = (
...
     export const ${camel(`use-${operationName}`)} = <TError = ${errorType},
     ${!definitions ? `TVariables = void,` : ''}
     TContext = unknown>(${generateQueryArguments({
```
CPatchane, line 569: Is it expected that we can't return any data from a mutation?

melloware:
Just to clear up my confusion can you post what it generates NOW and what your proposed patch will generate?  Or is what you are saying you are generating these "additional" values?

CPatchane:
Sorry for the confusion, indeed this PR generates new types

anymaniax:
Hello @CPatchane, it's a good idea! There is no breaking change from what I see no? You just add new types?

melloware (dismissed):

CPatchane:
> Hello @CPatchane, it's a good idea! There is no breaking change from what I see no? You just add new types?

Indeed, this is only new types for the queries part only.
I was thinking maybe that new types can generate more things some people might not need, but I am clearly ok to add this feature as it and maybe put behind a config according to future feedbacks.

Btw, I just pushed a new change to use `MutationData` and `MutationError` naming suffix for mutations for better consistency.

melloware:
I personally love this @CPatchane .  Since its add new code its really just adding new features and flexibility for users without breaking backwards compatibility I am all for it!

I just started using React Query and I can't believe I did not know about it before but its an absolutely amazing library and this Orval generator makes my life so much easier since I already have OpenAPI generated from my Spring Boot apps!

anymaniax (approved):

anymaniax:
It's only done for react-query here but we should do it for all the clients

CPatchane:
@anymaniax Is it normal we don't generate hooks for `swr` mutations? When I generate on my side I only have hook for queries, for mutations I only have functions 🤔

anymaniax:
Swr doesn't have mutators like react query

CPatchane:
@anymaniax Let me now, it should be good for the other clients :) (swr, angular and axios)

src/types/generator.ts
```diff
@@ -147,7 +147,11 @@ export type ClientHeaderBuilder = (params: {
   provideIn: boolean | 'root' | 'any';
 }) => string;
 
-export type ClientFooterBuilder = (operationIds?: string[]) => string;
+export type ClientFooterBuilder = (options: {
```
anymaniax: it's a breaking change 🙁

anymaniax: what we could do is instead of generating them directly in this file. We add them with the other models? WOuld be better and more consistent no?

CPatchane: We could do that, but how would you get the operationName from there?
I still think the best place is here since it's very related to the functions/hooks and less to models/schemas

CPatchane: this client footer builder is overridable here?

anymaniax: yes

anymaniax: You could generate the new models here but you can return an array of `GeneratorSchema` by the client. Like that they can be added to the other models/schemas

CPatchane: Ok, sorry but I might need more explanations here. I don't understand why this is breaking, for me the footer builder is not exposed to the configuration itself, it seems to be used only internally. What did I missing?

Do you mean I should have a new prop `schemas` [here](https://github.com/CPatchane/orval/blob/de2e0e622cd24340bdbb4810180dd465256c9dfe/src/core/generators/client.ts#L258) to then be used in api.ts?

anymaniax: Look the type [here](https://github.com/anymaniax/orval/blob/master/src/types/index.ts#L42). You will see that the function type is exposed

anymaniax: To avoid the breaking change you could keep the two `signature` and check when it's an array to handle it like before and add a warning no?

CPatchane: It make things much more complicated to handle, using TS function overload will make types breaking since TS will expect to handle the two types of arguments in the custom footer function and in our internal functions too.
Even if it's a breaking change (I don't think it will impact that much people tbh), I still think this change is necessary since it make things more consistent between the footer and the header (both functions is using an options object as parameter now) and more maintainable if we want to add more options in the future.
I am afraid I can't add something simple/not ugly to handle that case without that :/ (if you still want to do it I can push a commit that should work)

CPatchane: @anymaniax Any updates on this 🙏 ?

anymaniax: Sorry, I am a bit busy this week and next week. I will take a look as soon as possible

anymaniax: I quick solution, for now, would be to have two arguments no? keep the operationIds in first and have the options in second. With a warning that it's deprecated for the next version?

CPatchane: Indeed that could be a quick solution since the second argument would be optional then

CPatchane: I'll rework using that quick fix then 👍

CPatchane: It makes thing more difficult to maintain actually, you'll have to call the function each time wtih `undefined` as first argument in the "new code".
What do you think of the last commit instead? If this is a custom client, I use a string arrays and if it works the developer will get a warn about that deprecation. Using that, it will be easier for you to remove that part when you'll release a new major version and you can use the new signature without any issues.

anymaniax (approved):

anymaniax:
Really great job

PR #319, opened 2022-02-13
feat(query): allow body type customisation through mutator for mutations

CPatchane:
## Status

**HOLD**

## Related PRs
~~Wait merging PR #318 before merging this one.~~ ✅ 

## Description
Look only at the last commit, the previous ones are from #318.
With those two PRs combined, we're now able to customise via the mutator the return type (see PR #318) (in my example, a server returning snake cased data that we want to camel case in our TS code) but also the body type (this PR) expecting by the operation function (since that the server is expecting snake cased keys, the mutator can handle that case change systematically) through a dedicated and optional type named `BodyType`.

## How it works

In the mutator, we can export a type named `BodyType` like following (similar to what we can do for `ErrorType`):
```ts
// ./custom-mutator.ts
export type BodyType<BodyData> = CamelCase<BodyData>
```

Then, it will generated for the mutation:
```ts
export const postMyDemo = (id: string, postMyDemoBody: BodyType<PostMyDemoBody>) => {
  return apiClient<void>({
    url: `/my/demo`,
    method: "post",
    data: postMyDemoBody,
  })
}

export type PostMyDemoMutationResult = NonNullable<AsyncReturnType<typeof postMyDemo>>
export type PostMyDemoMutationBody = BodyType<PostMyDemoBody>
export type PostMyDemoMutationError = unknown

export const usePostMyDemo = <TError = unknown,  TContext = unknown>(
options?: {
  mutation?: UseMutationOptions<
    AsyncReturnType<typeof postMyDemo>,
    TError,
    {
      id: string
      data: BodyType<PostMyDemoBody>
    },
    TContext
  >
}) => {
  const { mutation: mutationOptions } = options || {}

  const mutationFn: MutationFunction<
    AsyncReturnType<typeof postMyDemo>,
    {
      id: string
      data: BodyType<PostMyDemoBody>
    }
  > = (props) => {
    const { id, data } = props || {}

    return postMyDemo(id, data)
  }

  return useMutation<
    AsyncReturnType<typeof postMyDemo>,
    TError,
    {
      id: string
      data: BodyType<PostMyDemoBody>
    },
    TContext
  >(mutationFn, mutationOptions)
}
```

Wdyt?
It might not be related to this PR but I personally think we might need more tests (with a snapshotting system?), at least to be able to see when a fix/feature is added how this is impacting the different scenarios because I think that TS won't be able catch everything imho.


melloware:
Again since this is adding new code I think this is another good improvement.

melloware (dismissed):

anymaniax:
What is the status of this one? Since the other one is merged

anymaniax:
ok thanks for the great work on this again

CPatchane:
@anymaniax I had to update the samples due to the last PR too.
I did the changes to handle BodyType, but it will work only for queries with custom clients. I don't think we need it for the other clients but I might be wrong. Wdyt?

src/core/generators/imports.ts
```diff
@@ -63,12 +64,14 @@ export const generateMutatorImports = (
       const importDefault = mutator.default
         ? `${mutator.name}${
             mutator.hasErrorType
-              ? `, { ErrorType as ${mutator.errorTypeName} }`
+              ? `, { ErrorType as ${mutator.errorTypeName}${
```
anymaniax: you don't handle the case where you only have the bodyTypName no?

CPatchane:
I updated the samples and add a new generated test named "custom-client" allowing to test the `BodyType` usage. Let me know if it's good :)

src/core/generators/imports.ts
```diff
@@ -63,12 +64,18 @@ export const generateMutatorImports = (
       const importDefault = mutator.default
         ? `${mutator.name}${
             mutator.hasErrorType
```
anymaniax: what do you think of doing something like this mutator.hasErrorType || mutator.bodyTypeName and then handling rest inside it?

CPatchane: Indeed it will be clearer 👍 
I did the change, let me know :)

anymaniax (approved):
Thanks 🎉! Will do a release soon

PR #321, opened 2022-02-14
fix(schema): generate 'any' instead of '{}'

melloware:
## Status
**READY**

Fix #321 

## Description
Instead of generating an empty type like...

```ts
export type Health200 = {};
```

it generates it as an ANY

```ts
export type Health200 = any;
```

@CPatchane @anymaniax Can you review this it looks to me like the first `if` and the last `else` were doing the exact same thing so I simplified the logic "I think".

### DIFF
--- a/src/core/generators/schemaDefinition.ts
+++ b/src/core/generators/schemaDefinition.ts
@@ -71,7 +71,8 @@ export const generateSchemasDefinition = async (
           );
 
           if (!imp) {
-            output += `export type ${schemaName} = ${resolvedValue.value};\n`;
+            const outputValue = resolvedValue.value === '{}' ? 'any' : resolvedValue.value;
+            output += `export type ${schemaName} = ${outputValue};\n`;
           } else {
             const alias = imp?.specKey
               ? `${pascal(getSpecName(imp.specKey, context.specKey))}${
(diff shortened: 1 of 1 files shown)


melloware:
Ahh nevermind I see the logic now that it broke a unit test the else's were not in the same if.

melloware:
Fixed https://github.com/anymaniax/orval/issues/308

src/core/generators/schemaDefinition.ts
```diff
@@ -71,22 +71,25 @@ export const generateSchemasDefinition = async (
...
+            output += `export type ${schemaName} = ${outputValue};\n`;
           } else {
-            const alias = imp?.specKey
+            let alias = imp?.specKey
               ? `${pascal(getSpecName(imp.specKey, context.specKey))}${
                   resolvedValue.value
                 }`
               : `${resolvedValue.value}Bis`;
 
+            alias = alias === '{}' ? 'any' : alias;
```
anymaniax: will never be `{}` in that case

CPatchane: I don't think you need to change the alias here, just the output value should be changed or am I wrong?

anymaniax: In that case, that will be always a schema so it's always a name

melloware:
OK updated and remove the `alias` part.

PR #322, opened 2022-02-15
feat(params): allow [_-] in url params

CPatchane:
## Status

**READY**

## Description
Allow to use `_` and `-` in path params name.

```yaml
# openapi.yml
paths:
  /api/demo/{id}/index
    ...
  /api/demo/users/{user_id}/index
    ...
```

Before:
```ts
export const getApiDemoIdIndex = (id: string) => {
  return client<GetApiDemoIdIndex200>({
    url: `/api/demo/${id}/index`,
    method: "get",
  })
}

// Won't work at all, but the function naming is also breaking:
export const getApiDemoUseridIndex = ...
```

After:
```ts
export const getApiDemoIdIndex = (id: string) => {
  return client<GetApiDemoIdIndex200>({
    url: `/api/demo/${id}/index`,
    method: "get",
  })
}

export const getApiDemoUsersUserIdIndex = (userId: string) => {
  return client<GetApiDemoUsersUserIdIndex200>({
    url: `/api/demo/users/${userId}/index`,
    method: "get",
  })
}
```

### DIFF
--- a/src/core/getters/route.ts
+++ b/src/core/getters/route.ts
@@ -1,17 +1,19 @@
+import { camel } from '../../utils/case';
 import { sanitize } from '../../utils/string';
 
+const hasParam = (path: string) => /[^{]*{[\w_-]*}.*/.test(path)
+
 const getRoutePath = (path: string) => {
-  return path.split('').reduce((acc, letter) => {
-    if (letter === '{') {
-      return acc + '${';
-    }
-
-    if (letter === '}') {
-      return acc + '}';
-    }
-
-    return acc + sanitize(letter, { dot: true });
-  }, '');
+  const matches = path.match(/([^{]*){?([\w_-]*)}?(.*)/)
+  if (!matches?.length) return path // impossible due to regexp grouping here, but for TS
+  const prev = matches[1]
+  const param = camel(sanitize(matches[2], { underscore: true, dash: true, dot: true }))
+  const next: string = hasParam(matches[3]) ? getRoutePath(matches[3]) : matches[3]
+  if (hasParam(path)) {
+    return `${prev}\${${param}}${next}`
+  } else {
+    return `${prev}${param}${next}`
+  }
 };
 
 export const getRoute = (route: string) => {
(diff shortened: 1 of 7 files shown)


src/core/getters/route.ts
```diff
@@ -1,17 +1,12 @@
...
-  return path.split('').reduce((acc, letter) => {
-    if (letter === '{') {
-      return acc + '${';
-    }
-
-    if (letter === '}') {
-      return acc + '}';
-    }
-
-    return acc + sanitize(letter, { dot: true });
```
CPatchane, line 13: We were allowing dot here but it can't work since in `params.ts` we wasn't allowing it 🤷

anymaniax: dot are not for the params

anymaniax: you don't handle the case where you have something like this /{petId}.js

src/utils/string.ts
```diff
@@ -55,7 +55,7 @@ export const sanitize = (
   },
 ) => {
   const { whitespace = '', underscore = '', dot = '', dash = '' } = options || {};
-  let newValue = value.replace(/[^\w\s.-]/g, '');
```
anymaniax: don't need to add the _ was already working like that normally

CPatchane: Indeed, TIL `underscore` is included in `\w`, my bad

src/core/getters/route.ts
CPatchane: Indeed, I rewrote that part and I added jest to be able to have unit tests for this, much simpler to test and make it robust. Wdyt?

jest.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
anymaniax: really great idea. Would you be ok to use https://vitest.dev/ instead?

CPatchane: Didn't know vitest but let's do it 👍

CPatchane: Much simpler to config, it's done :)

anymaniax:
Would be great to also run the test on pre-commit and in the ci

CPatchane:
> Would be great to also run the test on pre-commit and in the ci

Indeed, done :)

anymaniax (approved):
Really nice job! 🎉

PR #325, opened 2022-02-15
fix(mutator-parsing): stop loading the mutator and instead parse the …

anymaniax:
…code with acorn

## Status

**READY**

## Description
Stop trying to load the mutator file and instead parse it with acorn

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
   },
   "dependencies": {
     "@apidevtools/swagger-parser": "^10.0.3",
+    "acorn": "^8.7.0",
     "cac": "^6.7.12",
     "chalk": "^4.1.2",
     "chokidar": "^3.5.3",
--- a/src/core/generators/mutator.ts
+++ b/src/core/generators/mutator.ts
@@ -1,6 +1,7 @@
+import { Parser } from 'acorn';
 import chalk from 'chalk';
 import { readFile } from 'fs-extra';
-import { GeneratorMutator } from '../..';
+import { GeneratorMutator, GeneratorMutatorParsingInfo } from '../..';
 import { NormalizedMutator, Tsconfig } from '../../types';
 import { pascal } from '../../utils/case';
 import { getFileInfo, loadFile } from '../../utils/file';
@@ -47,26 +48,22 @@ export const generateMutator = async ({
     ? 'ErrorType'
     : `${pascal(name)}ErrorType`;
 
-  const { file, cached } = await loadFile<Record<string, Function>>(
-    importPath,
-    {
-      isDefault: false,
-      root: workspace,
-      alias: mutator.alias,
-      tsconfig,
-    },
-  );
+  const { file, cached } = await loadFile<string>(importPath, {
+    isDefault: false,
+    root: workspace,
+    alias: mutator.alias,
+    tsconfig,
+    load: false,
+  });
 
   if (file) {
-    const mutatorFn =
-      file[mutator.default || !mutator.name ? 'default' : mutator.name];
+    const mutatorInfoName = isDefault ? 'default' : mutator.name!;
+    const mutatorInfo = parseFile(file, mutatorInfoName);
 
-    if (!mutatorFn) {
+    if (!mutatorInfo) {
       createLogger().error(
         chalk.red(
-          `Your mutator file doesn't have the ${
-            mutator.default ? 'default' : mutator.name
-          } exported function`,
+          `Your mutator file doesn't have the ${mutatorInfoName} exported function`,
         ),
       );
       process.exit(1);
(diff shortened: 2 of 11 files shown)


melloware (dismissed):

src/core/generators/query.ts
```diff
@@ -162,7 +162,7 @@ const generateQueryRequestFunction = (
       isFormUrlEncoded,
     });
 
-    const isMutatorHasSecondArg = mutator.mutatorFn.length > 1;
+    const isMutatorHasSecondArg = mutator.numberOfParams > 1;
```
yekver: How about extracting `isMutatorHasThirdArg`, `isMutatorHasSecondArg` and `isMutatorHook` in a separate helper that will accept `mutator` as a parameter to avoid duplication?

src/core/generators/mutator.ts
```diff
@@ -97,9 +98,114 @@ export const generateMutator = async ({
...
+  );
+
+  return parseFunction(ast, property.value.body.name);
+};
+
+const parseFunction = (
+  ast: any,
+  name: string,
+): { numberOfParams: number } | undefined => {
+  const node = ast?.body?.find((node: any) => {
```
yekver: it's better to choose a different name for the inner `node` variable to avoid naming clash

package.json
```diff
@@ -92,6 +92,7 @@
   },
   "dependencies": {
     "@apidevtools/swagger-parser": "^10.0.3",
+    "acorn": "^8.7.0",
```
yekver, line 95: How about adding a [typings](https://www.npmjs.com/package/@types/acorn) to this package?

src/core/generators/mutator.ts
```diff
@@ -97,9 +98,114 @@ export const generateMutator = async ({
...
       hasErrorType,
       errorTypeName,
     };
   }
 };
+
+const parseFile = (
+  file: string,
+  name: string,
+): { numberOfParams: number } | undefined => {
```
yekver: Seems like the return type of `parseFile` and `parseFunction` should be shared (not duplicated)

src/core/generators/mutator.ts
```diff
@@ -47,21 +48,21 @@ export const generateMutator = async ({
...
+    tsconfig,
+    load: false,
+  });
 
   if (file) {
-    const mutatorFn =
-      file[mutator.default || !mutator.name ? 'default' : mutator.name];
+    const mutatorInfo = parseFile(
+      file,
+      !mutator.default ? mutator.name! : 'default',
```
yekver: How can you be sure that `mutator.name` is defined? During the [declaration](https://github.com/anymaniax/orval/pull/325/files#diff-3bf59b28833123acbdd18b785a76b03fe4e417ad1426e394d89cdd93d9dad23dR38) of `importName` variable we check this case.

yekver: also it's better to use `isDefault` and invert the ternary condition to avoid using the `!` prefix

anymaniax: Normally, if it's not default we should always have a name but true we should check that in the normalization phase

anymaniax: Was already done in fact

package.json
anymaniax: the base types are not good and they are not simple to manage

src/core/generators/mutator.ts
anymaniax: done

src/core/generators/mutator.ts
anymaniax: done

src/core/generators/query.ts
anymaniax: done

yekver (approved):
Good job!

yekver (approved):

src/core/generators/mutator.ts
```diff
@@ -47,21 +48,18 @@ export const generateMutator = async ({
...
+    root: workspace,
+    alias: mutator.alias,
+    tsconfig,
+    load: false,
+  });
 
   if (file) {
-    const mutatorFn =
-      file[mutator.default || !mutator.name ? 'default' : mutator.name];
+    const mutatorInfo = parseFile(file, isDefault ? 'default' : mutator.name!);
```
yekver: Looks like the condition `isDefault ? 'default' : mutator.name!` can be extracted into a separate variable and reused [at line 66](https://github.com/anymaniax/orval/pull/325/files#diff-3bf59b28833123acbdd18b785a76b03fe4e417ad1426e394d89cdd93d9dad23dR66)

anymaniax: done

yekver (approved):

PR #356, opened 2022-04-07
fix(generator): add prefix to es5keyword parameters

### LINKED ISSUE #350: Bug: parameter names identical to js keywords are not prefixed
```
/method/{default}:
    get:
      parameters:
        - name: default
```

Ends up with `Declaration or statement expected.ts(1128)` in generated file. Because parameter name `default` is not prefixed with `_`

yekver:
## Status

**READY**

## Description
Fix #350


src/core/getters/params.ts
```diff
@@ -58,7 +58,7 @@ export const getParams = ({
         schema,
       } = pathParam.parameter;
 
-      const name = sanitize(camel(nameWithoutSanitize));
+      const name = sanitize(camel(nameWithoutSanitize), { es5keyword:true });
```
anymaniax, line 61: will remove underscore and dash on the string if you use the sanitize function like this

PR #358, opened 2022-04-08
docs(configuration): update

kfrederix:
Update documentation related to Angular-specific configuration option: `provideIn`.

## Description
I noticed that the documentation related to `provideInRoot` was a bit outdated, here's my first attempt to fix that.

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -569,7 +569,7 @@ module.exports = {
       ...
       override: {
         angular: {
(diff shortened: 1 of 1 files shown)


kfrederix:
@anymaniax I am not sure if we should document the `Boolean` type for this option or not. What do you think? The behaviour of this option is a bit complicated due to backwards compatibility with the now deprecated option `provideInRoot`.

anymaniax:
It's perfect like that. Thanks

anymaniax (approved):

PR #360, opened 2022-04-15
Replace AsyncReturnType with Awaited<ReturnType<>>

krishnaglick:
## Status

**READY**

## Description
If a custom mutator function returns a non-promise, `AsyncReturnType` breaks.
This removes `AsyncReturnType` and just uses `Awaited<ReturnType<>>` in its place which:
- Doesn't break with sync functions
- Relies on internal Typescript functionality, and thus should be more stable

Additionally the footer generation function no longer needs `hasMutator` so it has been removed.

This was built off of v6.7.1 so I could test locally as `master` had issues with generating my output.

## Todos
- [X] Tests
- [X] Documentation

### DIFF
--- a/src/core/generators/axios.ts
+++ b/src/core/generators/axios.ts
@@ -92,11 +92,11 @@ const generateAxiosImplementation = (
       (title?: string) =>
         `export type ${pascal(
           operationName,
-        )}Result = NonNullable<AsyncReturnType<${
+        )}Result = NonNullable<Awaited<ReturnType<${
           title
             ? `ReturnType<typeof ${title}>['${operationName}']`
             : `typeof ${operationName}`
-        }>>`,
+        }>>>`,
     );
 
     return `const ${operationName} = (\n    ${toObjectString(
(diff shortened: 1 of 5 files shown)


melloware:
cc @CPatchane

anymaniax:
Really nice one. Which can of problem did you had with the master branch?

krishnaglick:
> Really nice one. Which can of problem did you had with the master branch?

On `master` the generated props for one function with a specific keyword was incorrectly generated. I could get you the details Monday!

krishnaglick:
Did you want me to get you that error? The project I work on is open source so you could replicate it locally yourself @anymaniax

CPatchane (approved):
Always better to rely on a dedicated TS utility instead of a custom one 👍

anymaniax:
Can you also replace it in the other clients?

krishnaglick:
> Can you also replace it in the other clients?

Not to be rude but that greatly expands the surface area of this change, and in a manner I don't have setup to test locally. I'm not confident that it won't break something.
Additionally I've had to change my implementation and no longer require this change, but feel it's still advantageous to implement.

yekver (approved):

melloware (approved):

anymaniax (approved):

krishnaglick:
Since this has received 4 approvals what needs to be done to merge? I see the vercel deploy requires some sort of authorization making this  red.

anymaniax:
Will merge it don't worry. Just didn't had time yet

PR #366, opened 2022-04-25
Simplify generated code for MSW

arlyon:
## Status

**READY**

## Description

This PR introduces two commits:

- replace the now-deprecated `faker.helpers.randomize` (which logs an error when used) with `faker.random.arrayElement`
- simplify the logic for lists of enums using `faker.random.arrayElements` (see here: https://fakerjs.dev/api/random.html#arrayElements)

## Steps to Test or Reproduce

These changes do not introduce any new logic so existing tests should be sufficient.

Note: I deliberately did not update the samples and readme to use faker.random.arrayElement but can do so as well.

### DIFF
--- a/src/core/generators/msw.ts
+++ b/src/core/generators/msw.ts
@@ -89,7 +89,7 @@ export const generateMSW = async (
   if (mockData) {
     value = mockData;
   } else if (definitions.length > 1) {
-    value = `faker.helpers.randomize(${definition})`;
+    value = `faker.random.arrayElement(${definition})`;
   } else if (definitions[0]) {
     value = definitions[0];
   }
--- a/src/core/getters/combine.mock.ts
+++ b/src/core/getters/combine.mock.ts
@@ -55,9 +55,9 @@ export const combineSchemasMock = async ({
       if (!index && !combine) {
         if (resolvedValue.enums || isOneOf) {
           if (arr.length === 1) {
-            return `faker.helpers.randomize([${resolvedValue.value}])`;
+            return `faker.random.arrayElement([${resolvedValue.value}])`;
           }
-          return `faker.helpers.randomize([${resolvedValue.value},`;
+          return `faker.random.arrayElement([${resolvedValue.value},`;
         }
         if (arr.length === 1) {
           return `{${resolvedValue.value}}`;
--- a/src/core/getters/scalar.mock.ts
+++ b/src/core/getters/scalar.mock.ts
@@ -125,13 +125,7 @@ export const getMockScalar = async ({
         );
         const enumValue = enumImp?.name || name;
         return {
-          value: `[...Array(faker.datatype.number({min:1, max: ${enums.length}}))].reduce(({values, enums}) => {
-            const newValue = enums[faker.datatype.number({min:0, max: enums.length - 1})];
-            return {
-              values: [...values, newValue],
-              enums: enums.filter((v: ${enumValue}) => newValue !== v)
-            }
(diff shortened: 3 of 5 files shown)


src/core/getters/combine.mock.ts
```diff
@@ -55,9 +55,9 @@ export const combineSchemasMock = async ({
       if (!index && !combine) {
         if (resolvedValue.enums || isOneOf) {
           if (arr.length === 1) {
-            return `faker.helpers.randomize([${resolvedValue.value}])`;
+            return `faker.random.arrayElement([${resolvedValue.value}])`;
```
yekver, line 58: I didn't get, what is the point to get random element from array with length === 1 ?

arlyon: That's a good question. I didn't write the original code so for the sake of keeping the PR as small as possible I decided not to touch the logic in case I misunderstood its purpose, and simply replace the deprecated function with its recommended replacement.

I could remove the call to faker altogether instead if that is preferred.

anymaniax: it's a list of enums, there is maybe only one enum in the list but an enum can have multiple values

src/core/generators/msw.ts
```diff
@@ -89,7 +89,7 @@ export const generateMSW = async (
   if (mockData) {
     value = mockData;
   } else if (definitions.length > 1) {
-    value = `faker.helpers.randomize(${definition})`;
+    value = `faker.random.arrayElement(${definition})`;
```
anymaniax, line 92: what is the difference between arrayElement and randomize?

anymaniax: ok I see it's the new api

src/core/getters/scalar.mock.ts
```diff
@@ -125,13 +125,7 @@ export const getMockScalar = async ({
...
         const enumValue = enumImp?.name || name;
         return {
-          value: `[...Array(faker.datatype.number({min:1, max: ${enums.length}}))].reduce(({values, enums}) => {
-            const newValue = enums[faker.datatype.number({min:0, max: enums.length - 1})];
-            return {
-              values: [...values, newValue],
-              enums: enums.filter((v: ${enumValue}) => newValue !== v)
-            }
-          },{ values: [], enums: Object.values(${enumValue})}).values`,
+          value: `faker.random.arrayElements(Object.values(${enumValue}), faker.datatype.number({min:1, max: ${enums.length}}))`,
```
anymaniax: I am not sure we need the second argument with the new API. I think we can remove it no?

anymaniax:
nice one thanks

src/core/getters/scalar.mock.ts
arlyon: Yes, actually. I had it in my mind the second arg was required. I just force-pushed now :)

src/core/getters/combine.mock.ts
anymaniax: @yekver is it ok for you?

yekver: > @yekver is it ok for you?

According to the scope of current PR and it's purpose - yes, it's ok for me.

P.S.
I was confused by `arr.length === 1` and didn't realize that `resolvedValue.value` could be an `array`.
I think this part might be refactored a bit in future.

yekver (approved):

rsanchez:
Looking forward to this getting merged. The code currently generated with for an enum fails TS compilation:

Screen Shot 2022-04-29 at 5 27 06 PM

PR #380, opened 2022-05-07
docs: update custom-axios.md

### LINKED ISSUE #379: Pass an additional option object in fetcher function when using custom axios instance
1. The fetchcer function in generated code is correct with default config

orval.config.ts:

```ts

export default defineConfig({
  test: {
    output: {
      mode: 'split',
      target: './swag.ts',
      client: 'react-query',
    },
    input: {
      target: './swagger.json',
    }
  }
})
```

result:

```ts
export const auth = (
     options?: AxiosRequestConfig // here we have an options here
 ): Promise<AxiosResponse<X>> => {
    return axios.get(
      `/api/v1/auth`,options
    );
  }
```
2. But with custom axios instance config:

```diff
export default defineConfig({
  test: {
    output: {
      mode: 'split',
      target: './swag.ts',
      client: 'react-query',
+      override: {
+        mutator: {
+          path: './custom-instance.ts',
+         name: 'customInstance',
+       },
      },
    },
    input: {
      target: './swagger.json',
    }
  }
})
```

The options param no longer remains..

```ts
export const auth = ( // here no options can be passed in
     
 ) => {
      return customInstance<X>(
      {url: `/api/v1/auth`, method: 'get'
    },
      );
    }
```

I want to pass some different option to some request, for example a longer timeout for certain requests. But with custom axios instance it behaves different than default one, it'll need to make custom axios instance also have fetcher function accept an options object as param.

**Package Version:**  v6.7.1

zoubingwu:
Specify the usage of adding second `options` argument in custom axios instance. See https://github.com/anymaniax/orval/issues/282 and https://github.com/anymaniax/orval/issues/379

## Status

**READY**

## Description
Update docs

fix #379

### DIFF
--- a/docs/src/pages/guides/custom-axios.md
+++ b/docs/src/pages/guides/custom-axios.md
@@ -26,9 +26,10 @@ import Axios, { AxiosRequestConfig } from 'axios';
 
 export const AXIOS_INSTANCE = Axios.create({ baseURL: '<BACKEND URL>' }); // use your own URL here or environment variable
 
-export const customInstance = <T>(config: AxiosRequestConfig): Promise<T> => {
+// add a second `options` argument here if you want to pass extra options to each generated query 
+export const customInstance = <T>(config: AxiosRequestConfig, options?: AxiosRequestConfig): Promise<T> => {
   const source = Axios.CancelToken.source();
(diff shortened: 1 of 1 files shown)


melloware (approved):

anymaniax (approved):
Thanks ❤️

Issue #136, opened 2021-05-23
Run prettier on generated code.


Boorj:
Prettier is already imported. I see no disadvantages to use it for generated code. That will correct all ugly linebreaks produced

1. add `src/utils/formatter.ts`
```ts
import Prettier from 'prettier';
export function prettifyTypescript(data){
    return Prettier.format(data, {
        parser: 'typescript',
        "useTabs": false,
        "tabWidth": 4,
        "singleQuote": true,
        "semi": true,
        "printWidth": 100,
        "trailingComma": "all",
        "proseWrap": "never"
    });
}
```

2. apply `prettifyTypescript` in all writers before writing TS-code.

anymaniax:
Prettier is use as dev dependency at the moment. That's a new feature that can be added but should be config in the orval config file. Also could be good to try to find a prettier config in the project and not only apply a custom config.

Boorj:
Have some concerns about using hardcoded config instead of flexible user config. But as a default option - definitly. 

p.s. @anymaniax, btw have you checked https://openapi.tools ? Orval deserves to be in the list.

anymaniax:
@Boorj I didn't know about https://openapi.tools could be interesting to add Orval there

anymaniax:
@Boorj We should only use the hardcoded config when we cannot find the user config in its own project.

melloware:
I am having trouble with this feature.  I have Orval and Prettier in my `devDependencies`

```json
        "orval": "^6.5.3",
        "prettier": "^2.5.1",
```

When I run `npm run codegen` i get this output..

⚠️  authorization - Prettier not found
🎉 authorization - Your OpenAPI spec has been converted into ready to use orval!

anymaniax:
Probably because with [this](https://github.com/anymaniax/orval/blob/master/src/core/writers/specs.ts#L139) code it's only using the global installed prettier

melloware:
hmmm i dont have prettier installed globally.

anymaniax:
yep, it's a bug need to find a better solution for this

c0m1t:
You can probably run prettier on `pre-commit` hook. Also, I made a pull request #459 which you can run a script (prettier, lint, etc...) when client is generated by `orval`.

melloware:
This is easy too..

```JS

"scripts": {
		"format": "prettier --write \"{src,__tests__}/**/*.{ts,tsx}\"",
		"codegen": "orval && npm run format",
[...]

Issue #234, opened 2021-10-15
Prettier doesn't run without schemas


ptrckhjnl:
## What are the steps to reproduce this issue?
1. Create a config that does not have `{ output: { schemas: ... } }` (but it has `target: ...` and `prettier: true`)
2. Run `orval`

## What happens?
The generated code is not formatted.

## What were you expecting to happen?
The code to be formatted.

## Any other comments?
It's probably because of this line: https://github.com/anymaniax/orval/blob/master/src/core/writers/specs.ts#L140

Here it should probably also check if the config contains a `target`.

But according to the docs (https://orval.dev/reference/configuration/output#schemas) if no `schemas` is given, `target` is used as the default value.

anymaniax:
Everything seems to work correctly on my side. Do you have prettier in your dev dependencies?

melloware:
This is easy enough to do. 

```JS

"scripts": {
		"format": "prettier --write \"{src,__tests__}/**/*.{ts,tsx}\"",
		"codegen": "orval && npm run format",
		"start": "react-app-rewired start"

```

PR #390, opened 2022-05-15
Reworked prettier setup

reichhartd:
This PR is best reviewed by commits. The first commit is the revised setup. In the second, only Prettier was run over all files.

## Status

**READY**

## Description

Revision of the Prettier setup:
- Package `pretty-quick` added, which runs over the staged files before the commit.
- Prettier command was adapted to run over all files that Prettier supports. Markdown is now one of them.
- Best practice: Fixed version of Prettier, because even a different patch version formats the code differently ([Prettier install](https://prettier.io/docs/en/install.html)).

Apart from that, Prettier was also run over all files.


melloware:
Does this fix #136 and #234 ???

reichhartd:
No, neither of them.

anymaniax (approved):

PR #391, opened 2022-05-15
`orval.config.ts` in docs

reichhartd:
Screenshot 2022-05-15 at 18 54 44

## Status

**READY**

## Description

Currently, there isn't a section in the documentation that explains, that it is also possible to save your config file as a `mjs` or `ts`. This has been expanded in the documentation.

A tabs switch would look best, similar to [Docusaurus](https://docusaurus.io/docs/markdown-features/tabs). However, I haven't found a good package for MDX right away, and writing a component myself is more complicated than I thought at first. It is important that the children, which are Markdown elements, are recognised as such and rendered correctly.
Maybe in a future PR.

## Todos
- [ ] Tests
- [X] Documentation


anymaniax (approved):
Thanks

PR #415, opened 2022-06-02
feat(generator): add tRPC generator

DaleWebb:
## Status

**READY**

## Description
This PR introduces a generator for [tRPC](https://trpc.io/). tRPC is a library for creating type-safe APIs.

The generator will create a router for each Open API input, with routes for each `GET`, `POST` etc.

This includes changes to existing functions. They are to un-`promise` their response - the `Promise` didn't appear to be required:
- `src/core/getters/object.ts`
- `src/core/getters/ref.ts`
- `src/core/resolvers/ref.ts`

## Todos

- [x] Tests
- [x] Documentation


src/core/getters/ref.ts
```diff
@@ -30,15 +30,15 @@ const regex = new RegExp('~1', 'g');
  *
  * @param $ref
  */
-export const getRefInfo = async (
+export const getRefInfo = (
   $ref: ReferenceObject['$ref'],
   context: ContextSpecs,
-): Promise<{
```
DaleWebb, line 36: There doesn't appear to be a reason why this should be a promise.

src/core/resolvers/ref.ts
```diff
@@ -15,23 +15,17 @@ type ComponentObject =
...
-export const resolveRef = async <
-  Schema extends ComponentObject = ComponentObject,
->(
+export const resolveRef = <Schema extends ComponentObject = ComponentObject>(
   schema: ComponentObject,
   context: ContextSpecs,
   imports: GeneratorImport[] = [],
-): Promise<{
+): {
   schema: Schema;
```
DaleWebb, line 23: There doesn't appear to be a reason why this should be a promise.

anymaniax:
Hello @DaleWebb, thanks for the pull request. This new generator only use trpc as gateway to use another api?

DaleWebb:
> Hello @DaleWebb, thanks for the pull request. This new generator only use trpc as gateway to use another api?

Hey @anymaniax - thanks for responding. Effectively, this implementation of tRPC will provide a typesafe API to endpoints in an OpenAPI spec

anymaniax:
can be introduce with v7 #434 and with your own template then maybe

anymaniax:
I will do some tests this week to see if we can merge it could be a good template. I add another idea about trpc since moment where instead of the mutator or the http call function we could add a sort of handler to have some business logic directly there. We currently could do it with the mutator but would be annoying to do.

anymaniax:
Hello @DaleWebb, you can now check the new architecture and create your own client with the extracted core. The current version is `@orval/core@6.11.0-alpha.2`. If you do so I would be happy to add a section about your client in orval doc


You can now do something like this.

```ts
import query from '@orval/query';

export default defineConfig({
  petstore: {
    output: {
       client: query() // query({type: 'svelte-query'}) for another query client,
     }
  }
});
```

so you could do exactly the same with your own client

melloware:
No response from OP and PR is 2 years old

PR #475, opened 2022-07-10
added min and max so user can specifiy from swagger doc

kh45:
Problem: When a user specifies a minimum and/or maximum value for a property of type `number` or `integer`, these values are not passed to the faker function that generates the number in the mock api files, and thus the mock returns values that are not wanted.

This change addresses that issue.

Example:

This specification in the swagger doc:
Screen Shot 2022-07-10 at 11 41 48 AM

Will lead to this in the mock api file:
Screen Shot 2022-07-10 at 12 20 38 PM

If no minimum or maximum is provided, those values will be passed as `undefined` in the faker function
Screen Shot 2022-07-10 at 12 23 19 PM

[which will lead to the same default behavior before this change, of returning a random number between 0 and 99999.](https://fakerjs.dev/api/datatype.html#number)

### DIFF
--- a/src/core/getters/scalar.mock.ts
+++ b/src/core/getters/scalar.mock.ts
@@ -81,7 +81,7 @@ export const getMockScalar = async ({
     case 'number':
     case 'integer': {
       return {
-        value: getNullable('faker.datatype.number()', item.nullable),
+        value: getNullable(`faker.datatype.number({min: ${item.minimum}, max: ${item.maximum}})`, item.nullable),
         imports: [],
         name: item.name,
       };


src/core/getters/scalar.mock.ts
```diff
@@ -81,7 +81,7 @@ export const getMockScalar = async ({
     case 'number':
     case 'integer': {
       return {
-        value: getNullable('faker.datatype.number()', item.nullable),
+        value: getNullable(`faker.datatype.number({min: ${item.minimum}, max: ${item.maximum}})`, item.nullable),
```
anymaniax, line 84: Thanks for the pull request. You also need to handle the case when it's not defined in the specification

kh45: Of course! It's my first attempt at contributing to anything open source so i'm excited. For the case where it is not defined in the specification, I mentioned in the top comment it will simply pass `undefined` to the faker function which will lead to the same default behavior that was already happening (it will use the default values for min and max which are 0 and 99999). Is this sufficient or should I handle it some other way?

anymaniax:
thanks 🎉

PR #481, opened 2022-07-11
[swr] Optional swrKey and conditional fetching for generated SWR requests

RobbyUitbeijerse:
Hi there!

the `swrKey` option currently is currently defined required if you pass the `swr` object, but we would like the flexibility of passing down options to the useSWR instance without necessarily passing a custom key. 

Making it optional seems completely valid here, as already seems to be handled properly (output example from docs):

```
   const swrKey =
     swrOptions?.swrKey ?? (() => (isEnable ? getShowPetByIdKey(petId) : null));
```

**[edit]**
I realised it also isn't as straightforward to use the conditional fetching from SWR by passing `null` as the key, as it will just use the generated key as a fallback. In some cases, it might actually be useful to pause requests from SWR since we for example want to await a certain state update, while the state update itself might not be a parameter necessary for the request itself. From the swr docs;

```
// conditionally fetch
const { data } = useSWR(shouldFetch ? '/api/data' : null, fetcher)
```

In order to facilitate this I also added an `shouldFetch` option to the `swr` configuration object, but there might be other simple solutions I'm not seeing here.

The new output of the petstore would simply always contain the `isEnable`, regardless whenever there would be parameters or not:

```
  const isEnable = shouldFetch !== false
  const swrKey = swrOptions?.swrKey ?? (() => isEnable ? getShowPetByIdKey(petId,version) : null);
 ```

Lmk!

### DIFF
--- a/docs/src/pages/guides/swr.md
+++ b/docs/src/pages/guides/swr.md
@@ -45,16 +45,17 @@ export const useShowPetById = <TError = Error>(
   petId: string,
   options?: {
     swr?: SWRConfiguration<AsyncReturnType<typeof showPetById>, TError> & {
-      swrKey: Key;
+      swrKey?: Key;
+      enabled?: boolean;
     };
     axios?: AxiosRequestConfig;
   },
 ) => {
   const { swr: swrOptions, axios: axiosOptions } = options ?? {};
 
-  const isEnable = !!petId;
+  const isEnabled = swrOptions?.enabled !== false && !!petId;
   const swrKey =
-    swrOptions?.swrKey ?? (() => (isEnable ? getShowPetByIdKey(petId) : null));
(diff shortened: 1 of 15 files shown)


anymaniax:
Hello @RobbyUitbeijerse, thanks for the merge request what do you think about naming it `enabled` to align with react query?

RobbyUitbeijerse:
@anymaniax makes sense, I renamed `shouldFetch` to `enabled`, renamed the variable `isEnable` to `isEnabled`. Also updated the swr guide, that should cover things.

anymaniax (approved):

RobbyUitbeijerse:
thanks for the merge @anymaniax! I'm currently running a patched version of Orval based on the pr, any plans of releasing this soonish? 😁

PR #489, opened 2022-07-17
fix: correct keys enum

ratmirslv:
## Status

**READY**

## Description

Hello! When i generate types for my json schema - i seeing incorrect types for enum:
```
export const PetColors = {
  '': 'Желтый',
  '': '青い',
  '': 'ლურჯი',
} as const;
```
Link codesandbox: [https://codesandbox.io/s/typescript-node-forked-hhgs70?file=/src/api/hooks.schemas.ts:264-291](url)
Your regex expression deleting keys when i use in my scheme - non English letters.

Maybe use transliteration?


anymaniax:
Hello @ratmirslv, seems cool it's not ready yet?

ratmirslv:
Hello, @anymaniax, Everything is ready, but I'm waiting for your comments.

anymaniax:
Seems good to me

ratmirslv:
Hello, @anymaniax . It's better to use keys as values:
1. Because there is a problem when values - Unicode.
2. The user will explicitly understand by the key - what value he will receive.

src/core/getters/enum.ts
```diff
@@ -20,16 +20,7 @@ export const getEnumImplementation = (value: string, type: string) => {
...
-    const key =
-      isNumber || isTypeNumber
-        ? toNumberKey(isTypeNumber ? val.toString() : val.slice(1, -1))
-        : sanitize(val, { underscore: '_', whitespace: '_', dash: '-' });
-
-    return (
-      acc +
-      `  ${keyword.isIdentifierNameES5(key) ? key : `'${key}'`}: ${val},\n`
-    );
+    return acc + ` ${isTypeNumber ? toNumberKey(val) : val}: ${val},\n`;
```
anymaniax: you cannot simplify it like that. this implementation was there for some reason

anymaniax: and it's also a breaking change

anymaniax: In you case I think it's just that we need to tweak the sanitize to accept those characters

anymaniax: something like this should do the job

Screen Shot 2022-07-19 at 12 51 39

src/core/getters/enum.ts
```diff
@@ -20,11 +20,32 @@ export const getEnumImplementation = (value: string, type: string) => {
...
+      /*
+        Potentially sanitize can strip to much from a value
+        For example:
+        sanitize("привет, мир!") -> "_"
+        sanitize("❤") -> ""
+        In that case it's more reasonable to return a full unchanged value
+        which will be later surrounded with quotes
+       */
+      if (key.length < valueWithoutQuotes.length) {
+        key = valueWithoutQuotes;
```
anymaniax: You could have case where you don't replace the space by _ anymore with this which was a convention added from the beginning

ratmirslv: Sorry, I'm not sure what you mean

Unicode symbols in enum values are trimmed in `sanitize` function with the regex below
https://github.com/anymaniax/orval/blob/3f7a468afa0eb7db60000a2ce251eda996366905/src/utils/string.ts#L66

As a result some keys are reduced to empty strings. For example
```ts
sanitize('привет') // ''
```

Maybe you could suggest a better idea how to preserve original unicode values?

anymaniax: With your implementation, if you have something like this `привет  привет` the space is not replaced by an underscore and that's the issue because by convention since the beginning orval does that and so it's a breaking change

ratmirslv: Thanks, for the explanation. I've fixed it 🙏

anymaniax:
should be fixed with 6.9.2

Issue #352, opened 2022-04-04
Required field ignored in case of allOf


yekver:
I have a `MyType` with a `date` field marked as a required. Unfortunately this requirement is mistakenly ignored by orval.

Example:
```yaml
MyAnotherType:
    type: object
    properties:
      myProp:
        type: string
MyType:
  required:
    - date # ignored
  allOf:
    - $ref: '#/components/schemas/MyAnotherType'
    - type: object
      properties:
        date:
          type: string
```

Produces:
```typescript
export interface MyAnotherType {
  myProp?: string;
}

export type MyTypeAllOf = {
  date?: string;
};

// Here MyType['date'] will be an optional field
export type MyType = MyAnotherType & MyTypeAllOf;
```

anymaniax:
the required work at the schema level right now in the allOf. Don't know if it should also work there but probably

yekver:
> the required work at the schema level right now in the allOf. Don't know if it should also work there but probably

I think that it should because yaml compiles without any errors.

soartec-lab:
maybe this issue has already been fixed by several another improvements

PR #503, opened 2022-07-22
Feature/352 support all of

yokoyama-takuya:
## Status

**READY**

## Description

fix allOf type definition 
e.g. https://github.com/anymaniax/orval/issues/352


yokoyama-takuya:
@anymaniax 
Thank you for the awesome library!
I used it for work and had the same problem, so I tried to fix it, but what about?
I'd be happy if you merge it.

src/core/resolvers/mergeAllof.ts
```diff
@@ -0,0 +1,33 @@
...
+import { OpenAPISchema } from 'redoc/typings/types';
+
+export const mergeAllOf = (
+  parser: OpenAPIParser,
+  inputSchema: OpenAPISchema,
+) => {
+  let schema = { ...inputSchema };
+  // test case [TopLevelAllOf]
+  if (schema?.allOf) {
+    schema = { ...schema, ...parser.mergeAllOf(schema) };
```
koenvangeert: Hey @yokoyama-takuya. We use allOf in our schemas in combination with oneOf that reference each other.
I tried your pr and I had to make a small change to make this use case work. Not sure if the same is also needed for nestedAllOf

Before:
```typescript
 schema = { ...schema, ...parser.mergeAllOf(schema) };
```

after
```typescript
    const oneOf = schema.oneOf;
    schema = {
      ...schema,
      ...parser.mergeAllOf(schema),
    };
    schema.oneOf = oneOf;
```
Short example:
```
Pet:
  oneOf:
   - Cat
   - Dog
  Properties:
   - id
   - name
Cat: 
  allOf:
   - Pet
Dog: 
  allOf:
   - Pet
```

yokoyama-takuya: Hi! @koenvangeert  Thanks to try this!
If it's okay until next week , I'll make an additional commit to this PR.

anymaniax:
Thanks for the pull request. I still need to test and check myself. What I am not a fan of is that with that we will have two different parser it should be possible with `@apidevtools/swagger-parser` no?

yokoyama-takuya:
@anymaniax  Thank you for your reply!
It might be possible, if you want.
My motivation I simply wanted to match the content and type displayed in Redoc. But, Test data's is Swagger UI same.
(i.e.... allOf has different priority when the same key exists. This fix conforms to Swagger UI behavior)

I wasn't sure if @apidevtools/swagger-parser could do the same thing

anymaniax:
Also by doing that it will cause a breaking change, I think because some of the models that were generated won't be anymore. Your issue is the same as #352? if it's the case I have maybe a solution for it

yokoyama-takuya:
@anymaniax Yes! It is the same as https://github.com/anymaniax/orval/issues/352. But, There were other cases that I expected. All patterns are left in the test code of this PR.

anymaniax:
I did some tests and with your implementation it will case some breaking changes. Also I saw some case where it doesn't seems to work correctly with multi files specifications. I thought about adding an utility type. It would be added for top level required.

Something like this

type WithRequired<T, K extends keyof T> = T & { [P in K]-?: T[P] }

yokoyama-takuya:
@anymaniax We solved by using `@apidevtools/swagger-cli` before orval. It is merge tool to one source file.
In case of have same property key, `Intersection Types` results type to `never `.  So before generate type, I think it is better merge.

Do you have an example this breaking changes?

anymaniax:
@yokoyama-takuya for the breaking change some models aren't generated anymore or are generated differently. Also for the case that is not working with the implementation. It's when you have the ref pointing to another specification file in an allOf.

yokoyama-takuya:
@anymaniax 

> some models aren't generated 

I want that openapi spec.

> Another big issue is that you also need to install as dependencies a lot of stuff that are not useful for the project.

Yeah, I think so too.

anymaniax:
@yokoyama-takuya 

You can check [here](https://github.com/anymaniax/orval/blob/master/tests/specifications/multi-files/api.yaml) and replace a ref in Pet.yaml or Person.yaml by a allOf with the same ref for example

PR #530, opened 2022-08-11
feat: sort schema barrel file

### LINKED ISSUE #483: Sort Schema barrel file
First of all thanks a bunch for this library. Really awesome and amazing work, it's my go-to for projects now.

1. Clone orval@6.8.1
2. Open `samples/react-query/basic`
3. run `npm install`
4. run `orval --config ./orval.config.ts`
5. Look at the output of `samples/react-query/basic/src/api/model/index.ts`
5. Delete the file (or a single line)
6. re-run `orval --config ./orval.config.ts`
7. Note how the file contents is not stable, the contents can be in any order

## What happens?
The contents of the schema barrel files (model/index.ts in this example) are not stable - they do not produce the same results if you clear the output and re-run.

I expect the outputs of all files to be stable. If I clear the output and run orval I should get the same result as long as the input did not change.

## Any other comments?
Root cause is due to the append logic within `writeSchemas` ([see source](https://github.com/anymaniax/orval/blob/9440969d21e1351ce9ff3706f89587f0df767a8d/src/core/writers/schemas.ts#L73))

Wondering your thoughts on the following proposed solution:
Refactor `writeSchemas` to do the following:
1. Build the entire contents of the schemas list in memory (no IO required). The lines would be sorted by name.
2. Do a single IO write to to overwrite the entire contents of the file (instead of a loop of promises)

This should also result in better performance.

The current solution (promise.all) would be difficult to solve because the promises can resolve in any order.

**Operating System:** Windows 10
**Package Version:** 6.8.1

SamKirkland:
Fix #483 sort schema barrel file

## Note
When testing this change I found the output of many example folders to be out of date. I didn't want to pollute this PR so I left those files out of this PR. I did leave the `samples/react-query/basic/src/api/model/index.ts` file in this PR as an example of what this change accomplishes.

## Status
**READY**

## Description
The contents of the schema barrel files (model/index.ts) should be stable between builds - the output between re-runs should be identical. This has several benefits, including less noisy git diffs.

This change also reduces IO trashing because it removes 2 IO operations that were previously within a loop (read file, append file)

## Related PRs
Non

## Todos

- [x] Tests
- [ ] Documentation - N/A

## Steps to Test or Reproduce

1. Clone orval@6.8.1
2. Open `samples/react-query/basic`
3. run `npm install`
4. run `orval --config ./orval.config.ts`
5. Look at the output of `samples/react-query/basic/src/api/model/index.ts`
5. Delete the file (or a single line)
6. re-run `orval --config ./orval.config.ts`
7. Note how the file contents is not stable, the contents can be in any order


src/core/writers/schemas.ts
```diff
@@ -85,18 +74,23 @@ export const writeSchemas = async ({
...
-      writeSchema({
-        path: schemaPath,
-        schema,
-        target,
-        isRootKey,
-        specsName,
-        header,
-      }),
-    ),
+  schemas.map((schema) =>
```
anymaniax: should stay a promise all to be sure to wait that everything has been written

src/core/writers/schemas.ts
```diff
@@ -85,18 +74,23 @@ export const writeSchemas = async ({
...
+      schema,
+      target,
+      isRootKey,
+      specsName,
+      header,
+    }),
   );
+
+  try {
+    await writeFile(schemaFilePath, generateModelImports({ schemas }));
```
anymaniax, line 94: The `appendFile` was there to not override the file completely. For example, if you already have some export in that file that you don't want orval to override each time. For example, having other models, not in the specification.

SamKirkland: Ah I understand now.

Is that a requirement or something you are willing to change?
If users want to add custom code they should probably be re-exporting in another barrel file.

anymaniax: I am not sure anyone use that feature but I still prefer to maintain it because otherwise it's a breaking change

SamKirkland: Because this is a undocumented feature I normally use a minor release digit bump. But I understand if you want to use strict semantic versioning.

I can change the PR target to your next version but I don't see a `v7` branch and `next` is actually behind `master`. How would you like to proceed?

SamKirkland: > "Be carefull clean all output target and schemas folder."
----- https://orval.dev/reference/configuration/output#clean

The documentation on `clean` sort of mentions this feature... But I wasn't sure what that meant when initially integrating orval 😄

anymaniax: We can merge it like that it is something that nobody knows and uses anyway

src/core/writers/schemas.ts
SamKirkland: Added!

SamKirkland:
Changes made, requesting re-review

anymaniax (approved):

anymaniax:
@SamKirkland Thanks

PR #545, opened 2022-08-21
fix: prevent numeric prefixes in response types

knyzorg:
## Status

**READY**

## Description

This is a limited fix for an issue I was dealing with. I have a swagger file which returns response types that are prefixed with the response code: e.g. `404Response`.

There might be a better way to do this, but it fixes it for me.


anymaniax:
Hey @knyzorg, thanks for the pull request. I will close this one because a fix has been done and will be in the next release

PR #563, opened 2022-08-31
fix(code): generated code for arrays with nullable items

codebutler:
## Status

**READY**

## Description

The code generator for arrays outputs ambiguous types, and breaks for nullable array items. 
Currently, the generated code looks like: `string | null[]`
This is interpreted as "string or an array of null", which is incorrect.
With this fix, the generated code is: `(string | null)[]`

This change wraps _all_ array types in parenthesis (`()`). This is not necessarily needed for a single type, but it causes no harm, avoids any ambiguity, and avoids the need for more complex changes to the code generator.

I also added a regression test that fails to compile without this fix. I didn't see an existing pattern for regression tests so I came up with something. Let me know if you'd prefer a different approach.

## Related PRs

n/a

## Todos

- [X] Tests
- [ ] Documentation

## Steps to Test or Reproduce

> **Note** See tests for complete example.

1. Create a schema like:
    ```yaml
     nullable_items:
       type: array
       items:
         type: string
         nullable: true
    ```
2. Generate code, attempt to assign a value to the type, see error:
    ```
    TS2322: Type '[string]' is not assignable to type 'string | null[]'.   Type '[string]' is not assignable to type 'null[]'.     Type 'string' is not assignable to type 'null'.
    ```

### DIFF
--- a/src/core/getters/array.ts
+++ b/src/core/getters/array.ts
@@ -24,7 +24,7 @@ export const getArray = ({
       context,
     });
     return {
-      value: `${resolvedObject.value}[]`,
+      value: `(${resolvedObject.value})[]`,
       imports: resolvedObject.imports,
       schemas: resolvedObject.schemas,
       isEnum: false,
--- a/tests/configs/default.config.ts
+++ b/tests/configs/default.config.ts
@@ -42,4 +42,11 @@ export default defineConfig({
       schemas: '../generated/default/translation/model',
     },
   },
+  regressions: {
+    input: '../specifications/regressions.yaml',
+    output: {
+      target: '../generated/default/regressions/endpoints.ts',
+      schemas: '../generated/default/regressions/model',
+    },
+  },
 });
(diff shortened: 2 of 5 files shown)


hallaca (approved):

melloware (approved):

src/core/getters/array.ts
```diff
@@ -24,7 +24,7 @@ export const getArray = ({
       context,
     });
     return {
-      value: `${resolvedObject.value}[]`,
+      value: `(${resolvedObject.value})[]`,
```
anymaniax, line 27: We could maybe add a check if the string contain a |. Like that we can just add the parentheses when needed.

codebutler: imo its not worth adding more complexity here. can use the `--prettier` if you really care about clean generated code

codebutler:
can someone please authorize this?

anymaniax:
I will just let me the time to review your comment. And also It's complicated for me lately. That's why it take a bit of time to be merged and review too

anymaniax (approved):
Thanks for the fix!

melloware:
thanks @anymaniax any chance we can get a 6.9.7 release?

Issue #564, opened 2022-09-02
Config Option (useDeprecatedOperations): Do not generate deprecated APIs


melloware:
You need an OpenAPI file with an API call marked `deprecated: true`

Here is an example operation with ` "deprecated": true,` I would love an Orval config option `useDeprecatedOperations: false` to basically not generate any React Query code for any deprecated APIs.  It will reduce my TypeScript code and prevent someone from calling deprecated APIs.

```json
 "/submit-re": {
      "post": {
        "tags": [
          "Manifest"
        ],
        "summary": "Deprecated please use /v2/submit-toc-file",
        "operationId": "createPackageFromFiles",
        "parameters": [
          {
            "name": "Authorization",
            "in": "header",
            "required": false,
            "schema": {
              "type": "string"
            }
          },
          {
            "name": "refresh-data",
            "in": "query",
            "required": false,
            "schema": {
              "type": "boolean",
              "default": false
            }
          }
        ],
        "requestBody": {
          "content": {
            "multipart/form-data": {
              "schema": {
                "required": [
                  "file"
                ],
                "type": "object",
                "properties": {
                  "file": {
                    "type": "string",
                    "format": "binary"
                  }
                }
              }
            }
          }
        },
        "responses": {
          "200": {
            "description": "OK",
            "content": {
              "application/json": {
                "schema": {
                  "type": "object",
                  "additionalProperties": {
                    "type": "object"
                  }
                }
              }
            }
          }
        },
        "deprecated": true,
      }

melloware:
works like a charm in production removing my deprecated operations!

PR #565, opened 2022-09-03
feat: useDeprecatedOperations to exclude deprecated (#564)

melloware:
## Status

***READY**

## Description

Adds new `useDeprecatedOperations` option to allow excluding deprecated operations from even being generated in the code!

Adding this to config will stop them from being generated the value is `true` by default to not change any existing behavior.

```js
module.exports = {
  petstore: {
    output: {
      override: {
        useDeprecatedOperations: false,
      },
    },
  },
};
```

### DIFF
--- a/src/core/generators/query.ts
+++ b/src/core/generators/query.ts
@@ -886,6 +886,15 @@ export const generateQuery: ClientBuilder = (
   outputClient,
 ) => {
   const imports = generateVerbImports(verbOptions);
+
+  // GitHub #564 check if we want to exclude deprecated operations
+  if (verbOptions.exclude === true) {
+    return {
+      implementation: ``,
+      imports,
+    };
+  }
+
   const functionImplementation = generateQueryRequestFunction(
     verbOptions,
     options,
--- a/src/core/generators/query.ts
+++ b/src/core/generators/query.ts
@@ -8,7 +8,6 @@ import {
   ClientHeaderBuilder,
   ClientTitleBuilder,
   GeneratorDependency,
-  GeneratorImport,
   GeneratorMutator,
   GeneratorOptions,
   GeneratorVerbOptions,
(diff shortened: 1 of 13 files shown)


src/core/generators/verbsOptions.ts
```diff
@@ -206,7 +206,12 @@ export const generateVerbsOptions = ({
           context,
         });
 
-        acc.push(verbOptions);
+        // GitHub #564 check if we want to exclude deprecated operations
+        const includeOperation = !operation.deprecated || output.override.useDeprecatedOperations === true;
+
+        if (includeOperation) {
+          acc.push(verbOptions);
+        }
```
melloware: I apologize if this is not the correct place for this check and exclusion but it felt like the only place that made sense but feel free to make a better suggestion!

melloware: I have more work to do here. Wasn't as simple as I hoped. Was hoping to just pull the whole operation out of the list. Before it's processed.

melloware:
OK I was able to get it all working.

src/core/generators/query.ts
```diff
@@ -885,7 +886,20 @@ export const generateQuery: ClientBuilder = (
   options,
   outputClient,
 ) => {
-  const imports = generateVerbImports(verbOptions);
+  let imports = [] as GeneratorImport[];
+
```
anymaniax: By adding the logic here you only doing it for the query templates. I think it would be better to do it as soon as possible. Also in the next big release this will probably be removed because I want to add a better way to generate the endpoints more customizable

melloware:
OK I added to angular and swr generators too.  Until the V7 overhaul I would still love to see this feature in 6.9.7 only because I have an API with a lot of deprecated methods and its generating so much TypeScript code I don't want the UI developers to see or use.

anymaniax:
Would be better to do it [here](https://github.com/anymaniax/orval/blob/master/src/core/generators/api.ts#L19) no?

melloware:
You are right!  PR updated that was a MUCH better place for it. Way cleaner and really what I wanted to do in the first place.

anymaniax (approved):
Seems good thanks!

PR #581, opened 2022-09-12
fix(writers): fix handling of duplicate kebab'd tag names

codebutler:
## Status

**READY**

## Description

The splitTagsMode writer uses kebab(tag) to generate output filenames. This is a problem if the
input schema has multiple tags that "kebab" to the same value.
For example: "dup_tag", "dup tag", "dupe-tag", "DUP TAG"

The current behavior is that files are overwritten non-deterministically.

This change fixes the problem by kebab-ing tag names earlier in the process so they can be merged
together into a single file.

## Related PRs

NA

## Todos

- [X] Tests
- [ ] Documentation

## Steps to Test or Reproduce

See regression tests.

### DIFF
--- a/src/core/writers/splitTagsMode.ts
+++ b/src/core/writers/splitTagsMode.ts
@@ -2,7 +2,7 @@ import { outputFile } from 'fs-extra';
 import { join } from 'upath';
 import { OutputClient } from '../../types';
 import { WriteModeProps } from '../../types/writers';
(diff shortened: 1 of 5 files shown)


melloware (approved):

anymaniax (approved):

codebutler:
thanks for getting my fixes into the latest release!

anymaniax:
@codebutler  Thanks for the fix 😊

PR #593, opened 2022-09-22
[Documentation] Update output.override.header example

SamKirkland:
## Status
**READY**

## Description
The example for output.override.header does not build/work.
The example uses `title` `description` and `version` directly. But the variables are not in scope. They should either be destructed or use `info.xxx`


anymaniax (approved):
Thanks!

PR #680, opened 2022-11-26
docs: fix position of mock options

Kanary159357:
## Status

**READY**

### Set node engine limits to 16
There is error that starting docs is not working when node version is over 16. 
https://github.com/webpack/webpack/issues/15900

It could be resolved by updating librarys in docs folder, but it would be too large and may not be the way all contributor wants.  So just update package.json file.

### Integrate mocks options

Mock options has same depth as mock itself so it is hard to recognize that it is property of mock without looking at example. So fixed it according to https://github.com/anymaniax/orval/blob/68f84a8b49a78e19b45c4623b2f659b2f2d0481d/packages/core/src/types.ts#L59

AS_IS
image

TO_BE
image

## Related PRs

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

check deployed docs.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -79,6 +79,6 @@
     "webp-loader": "^0.6.0"
   },
   "engines": {
-    "node": ">=12.x"
+    "node": ">=12.x <17"
   }
 }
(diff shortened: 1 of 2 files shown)


anymaniax (approved):
Thanks!

PR #683, opened 2022-11-27
fix: required query parameters with content

einartryggvi:
## Status

**READY**

## Description

Support required query parameters with content instead of schema along with tests that verify that code generation for required/non-required and with/without default generates types as optional correctly.

Currently client generation fails for required query parameters with content, but works for required query parameters with schema.

The error is `TypeError: Cannot read properties of undefined (reading 'default')` since the schema variable is undefined when you specify content.

    # Works
    queryParamWithContentNotRequired:
      name: queryParamWithContentNotRequired
      in: query
      required: false
      content:
        application/json:
          schema:
            type: object
            properties:
              prop1:
                type: string
              prop2:
                type: string

    # Works
    queryParamWithSchemaNotRequired:
      name: queryParamWithSchemaNotRequired
      in: query
      required: false
      schema:
        type: object
        properties:
          prop1:
            type: string
          prop2:
            type: string

    # Works
    queryParamWithSchemaRequired:
      name: queryParamWithSchemaRequired
      in: query
      required: true
      schema:
        type: object
        properties:
          prop1:
            type: string
          prop2:
            type: string

    # Doesn't work
    queryParamWithContentRequired:
      name: queryParamWithContentRequired
      in: query
      required: true
      content:
        application/json:
          schema:
            type: object
            properties:
              prop1:
                type: string
              prop2:
                type: string

### DIFF
--- a/packages/core/src/getters/query-params.test.ts
+++ b/packages/core/src/getters/query-params.test.ts
@@ -1,0 +1,123 @@
+import { ParameterObject } from 'openapi3-ts';
+import { describe, expect, it } from 'vitest';
+import { ContextSpecs } from '../types';
+import { getQueryParams } from './query-params';
+
+// Mock context for getQueryParams
+const context: ContextSpecs = {
+  specs: {},
+  // @ts-ignore
+  override: {
+    useDates: true,
+  },
+};
+
+const queryParams: {
+  parameter: ParameterObject;
+  optional: boolean;
+}[] = [
+  {
+    parameter: {
+      name: 'queryParamWithSchemaNotRequired',
+      in: 'query',
+      schema: {
+        type: 'string',
+      },
+      required: false,
+    },
+    optional: true,
+  },
+  {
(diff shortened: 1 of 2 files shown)


einartryggvi:
@anymaniax Any chance this could be merged soon?

anymaniax:
Hello @einartryggvi, I maybe miss something but I don't see any fix in your code.

einartryggvi:
@anymaniax true, it's not obvious from the diff! I could have been clearer in the description of where the error occurs in the code. Sorry about that!

So the schema variable is being used elsewhere in the function, and since that isn't fallbacking on `content['application/json']` it crashes.

There are a few lines that attempt to access `schema.default`, mainly this one: https://github.com/anymaniax/orval/blob/master/packages/core/src/getters/query-params.ts#L80

anymaniax:
ok gotcha thanks!

anymaniax (approved):

PR #731, opened 2023-01-12
chore: remove deprecated 'url' dependency

MatthijsKok:
## Status

**READY**

## Description

Downstream users of `orval` have been getting a deprecation warning.
The 'url' dependency relied on 'querystring', which generated the warning.
I've found only one usage of 'url', and this PR replaces that usage.
Deprecation and migration link: https://nodejs.org/api/url.html#urlresolvefrom-to

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout <branch>
> grunt jasmine
```

1.


packages/core/src/getters/ref.ts
```diff
@@ -59,8 +58,18 @@ export const getRefInfo = (
     };
   }
 
+  const resolve = (from: string, to: string): string => {
```
anymaniax: What do you think about moving it outside of the getRefInfo function? And rename it `resolveUrl`?

anymaniax (approved):

Issue #338, opened 2022-02-25
Add additional support for more JSON Schema keywords (minLength, maxLength, etc.)


achendrick:
## What are the steps to reproduce this issue?
1. I ran orval against an OpenApi 3.0 spec with some of the following rules and saw no reference to them:
```jsonc
AnObjectDto: {
  type: "object",
  properties: {
    key: {
      maxLength: 30,
      minLength: 0,
      pattern: "\A(\w|-)+\Z",
      type: "string"
    },
    name: {
      maxLength: 100,
      minLength: 0,
      pattern: "\A(?!\s*\Z).+\Z",
      type: "string"
    },
    // ...
}
```

## What happens?
I see no references to `maxLength`, `minLength` or `pattern` and I'd suspect other [keywords](https://swagger.io/docs/specification/data-models/keywords/) are missing as well.
I _do_ see references to enums, which is great!

## What were you expecting to happen?
I'd expect either an object with references to these rules to be created or even individual constants created similar to how `enum` is treated.

Ideally:
```typescript
export const AnObjectDtoKey = {
  maxLength: 30,
  minLength: 0,
  pattern: "\A(\w|-)+\Z"
};
```
Or:
```typescript
export const AnObjectDtoKeyMaxLength = 30;
export const AnObjectDtoKeyMinLength = 0;
export const AnObjectDtoKeyPattern = "\A(\w|-)+\Z";
```

## Any logs, error output, etc?
None seen

## Any other comments?
Love the library and I'm really hoping I can leverage it to help with data validation.
I primarily use [Yup](https://github.com/jquense/yup) for client validation and I could take one of these generated values and use it along with it.

## What versions are you using?
**Operating System:** macOS 12.2.1 (Monterey)
**Package Version:** `6.5.3`
**Browser Version:** Brave [Version 1.35.103 Chromium: 98.0.4758.102 (Official Build) (x86_64)](https://brave.com/latest/)

soartec-lab:
Currently we support `zod` and you can use these in your schema validation.

PR #777, opened 2023-02-26
feat: #338 add export of additional keywords

aodinok:
## Status

**READY**

## Description

Adds additional keywords export in generated models e.g. **minimum** / **maximum**, **minLength** / **maxLength**.

These could be then used to for validation logic. One could argue that validation should not be part of generated code since it is not exactly related to data fetching / updating logic and it may be better to separate client side validation from data layer, or even use something like https://github.com/igtm/openapi-yup-generator.

But I see it as a very nice improvement in development experience and also do not see big harm in exporting additional keywords as conts. We are not adding restrictions on top of those constants, like wrapping them in Yup schemas or other validation lib, leaving it completely up to client to decide how to use const, or ignore it at all.

I was also thinking that maybe it would be nice to add setting to control const generation? 

Example:

```yaml
    Labradoodle:
      type: object
      required: ['cuteness']
      properties:
        cuteness:
          type: integer
          minimum: 1
          maximum: 128
```

Generated model:

```ts
export interface Labradoodle {
  cuteness: number;
  breed: LabradoodleBreed;
}
export const LabradoodleCutenessMinimum = 1;
export const LabradoodleCutenessMaximum = 128;
```

Client side validation:

```ts
const schema = yup.object().shape({
  cuteness: yup
    .number()
    .required()
    .min(LabradoodleCutenessMinimum)
    .max(LabradoodleCutenessMaximum)
})
```

After this BE can control the client side validation and there is no need to keep duplicated and manually hardcoded value anymore.

## Todos

- [x] Tests - not sure I understood completely how they work, but I added some keywords to basic yaml and saw it in generated
- [ ] Documentation - want to see if that even makes sense @anymaniax  before adding this

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout <branch>
> grunt jasmine
```

1.

### DIFF
--- a/packages/core/src/generators/consts.ts
+++ b/packages/core/src/generators/consts.ts
@@ -1,0 +1,68 @@
+import { isSchemaObject, SchemaObject } from 'openapi3-ts';
+import { resolveRef } from '../resolvers';
+import { ContextSpecs } from '../types';
+import { isReference, pascal } from '../utils';
+
+/**
+ * Generate the consts based on passed schema
+ * Generated consts will be exported along with model to be used for validation purposes
+ *
+ * @param name Model name
+ * @param schema Model schema
+ */
+export const generateConsts = ({
+  name,
+  schema,
+  context,
+  initialOutput,
+}: {
+  name: string;
+  schema: SchemaObject;
+  context: ContextSpecs;
+  initialOutput?: string;
+}): string => {
+  let output = initialOutput ?? '';
+
+  if (isSchemaObject(schema) && schema.properties) {
+    output += [
+      ...new Set(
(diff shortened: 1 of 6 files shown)


aodinok:
@anymaniax Can you please check this when you have time 🙏

anymaniax:
Hello @aodinok, thanks for the pull request. I wouldn't add this by default. It's really specific. What do you think about creating a new client for that? Maybe a yup or Zod client?

anymaniax:
#785

aodinok:
Closing it favor of https://github.com/anymaniax/orval/pull/785

PR #785, opened 2023-03-02
feat(clients): add zod

anymaniax:
Due to the demand, I add the new client zod 🎉.

Thanks to @stijnvanhulle, who previously did the implementation. I hope it's ok for you I took what you did as a starting point.


anymaniax:
@aodinok you have custom validation or using something like `yup`?

aodinok:
We are using yup at the moment

anymaniax:
If I do the same for yup would it work for you?

aodinok:
I think the problem with this schemas would be that it is hard to modify/extend them. 
In most of the cases you probably will want to make validation more strict in some places or add different custom messages to the schemas (e.g. intl) but if we export the schema per model without exporting basics (e.g. minLength value or even schema per prop) it will be pretty hard to extend the schemas.

anymaniax:
Probably a mix of both would be great

aodinok:
Yeah exactly. Basics plus model schemas - and then you can use whole model schema when needed or basics with custom schema.

No need to build yup specifically for us. I'm happy to contribute using your example for zod when I will have some spare time. Thanks a lot for adding this 👍

stijnvanhulle:
@anymaniax Sure, no problem. Started with this a couple of months ago but that only included Zod schemas for **component.schemas** and not for the different API calls. Thanks for finishing this :)

anymaniax:
@stijnvanhulle from what I saw it's based on pathItem. Which is great for a first version. I would like to split by schema but before the v7 it's not possible I need to refactor the config and the output system by package first.

anymaniax:
@aodinok I have something like [this](https://gist.github.com/anymaniax/a6fdefee67e69a3f4b9f81b966ed6035) right now

aodinok:
This is cool! Thanks for letting me know. I think we will be able to use this for our use case.

packages/zod/src/index.ts
```diff
@@ -0,0 +1,341 @@
...
+        syntheticDefaultImport: true,
+      },
+    ],
+    dependency: 'zod',
+  },
+];
+
+export const getZodDependencies = () => ZOD_DEPENDENCIES;
+
+const resolveZodType = (schemaTypeValue: SchemaObject['type']) => {
```
aodinok, line 38: type can be array in OpenAPI, e.g. `type: ["string", null]` is this being handled in some internal lib?

packages/zod/src/index.ts
```diff
@@ -0,0 +1,341 @@
...
+    schema?.minimum ??
+    schema?.exclusiveMinimum ??
+    schema?.minLength ??
+    undefined;
+  const max =
+    schema?.maximum ??
+    schema?.exclusiveMaximum ??
+    schema?.maxLength ??
+    undefined;
+  const matches = schema?.pattern ?? undefined;
```
aodinok, line 73: There is also a format that can be set in the schema that controls different things but I guess it may be hard to transform that to a validation function. Maybe good to add as a simple export string const though.

packages/zod/src/index.ts
```diff
@@ -0,0 +1,341 @@
...
+  }
+  if (max !== undefined) {
+    validationFunctions.push(['max', max]);
+  }
+  if (matches) {
+    const isStartWithSlash = matches.startsWith('/');
+    const isEndWithSlash = matches.endsWith('/');
+    validationFunctions.push([
+      'matches',
+      `new RegExp('${matches.slice(
```
aodinok: We may need to escape "'" in the matches string? 🤔

aodinok (approved):

packages/zod/src/index.ts
sirtimbly: warning it looks like this logic might fall victim to type coercion bugs. minimum and maximum might be `0` which would be coerced to false incorrectly here.

also, specifically these  openapi string `format`s should work with zod:
```
    "email", () => "email()"
   "hostname", () => "url()"
    "uri", () => "url()"
   "uuid", () => "uuid()"
   "datetime", () => "datetime()"
```

anymaniax:
I think we have a good enough first version I will merge this

stijnvanhulle:
@anymaniax Is it possible to create a new version so we can start playing around with the Zod integration?

sirtimbly:
just git clone the branch from github directly, good idea to check it out. I ran it and had a few issues, I'll try to open an issue.

anymaniax:
Hello guys, @stijnvanhulle I did an alpha version you can try it with `@orval/alpha`. @sirtimbly I did a lot of fixes for the zod client yesterday and will try to release a patch version today

anymaniax:
@stijnvanhulle @sirtimbly  just published Orval 6.13.1

PR #806, opened 2023-03-15
feat: Add specific type to query key function

maikdijkstra:
## Status

**READY**

## Description

The generated query key function would generate a function with an inferred return type (string | ParamType)[]. We can be more specific by defining [string, ParamType] instead. This is a small change, but more strict types improves the usability and cleanliness of code.

See: https://github.com/anymaniax/orval/issues/500

## Todos

- [ ] Tests
- [x] Documentation

## Steps to Test or Reproduce

1. Update a sample
- The query key function now has a strict return type of the following possible forms:
- [ ] [string] when only the route is returned
- [ ] [string, ParamsType] when the params are not optional
- [ ] [string, ParamsType | undefined] when the params are optional

### DIFF
--- a/packages/query/src/index.ts
+++ b/packages/query/src/index.ts
@@ -906,7 +906,13 @@ const generateQueryHook = async (
 
     const queryKeyFn = `export const ${queryKeyFnName} = (${queryKeyProps}) => [\`${route}\`${
       queryParams ? ', ...(params ? [params]: [])' : ''
-    }${body.implementation ? `, ${body.implementation}` : ''}];`;
+    }${body.implementation ? `, ${body.implementation}` : ''}] as ${
+      queryParams
+        ? `[string, ${queryParams.schema.name}${
+            queryParams.isOptional ? '| undefined' : ''
+          }]`
+        : '[string]'
+    };`;
 
     const implementation = `${!queryKeyMutator ? queryKeyFn : ''}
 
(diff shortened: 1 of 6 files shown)


packages/query/src/index.ts
```diff
@@ -906,7 +906,13 @@ const generateQueryHook = async (
 
     const queryKeyFn = `export const ${queryKeyFnName} = (${queryKeyProps}) => [\`${route}\`${
       queryParams ? ', ...(params ? [params]: [])' : ''
-    }${body.implementation ? `, ${body.implementation}` : ''}];`;
+    }${body.implementation ? `, ${body.implementation}` : ''}] as ${
```
anymaniax, line 909: You don't check to add the body type?

anymaniax: Would be easier to do as const no?

maikdijkstra: > You don't check to add the body type?

Yes, I missed that one. However, I cannot seem to create valid specification that has this ```body.implementation``` . OpenAPI 3.0 and Swagger 2.0 both seem to not support ```requestBody``` on GET requests which create the query key function. Do you have an example which does generate a ```body.implementation``` for a query key function?

> Would be easier to do as const no?

Yes, I agree that would make it a bit more clean. I will make this return type string a separate const based on the ```queryParams``` and ```body``` for now

maikdijkstra:
@anymaniax I see that I misunderstood your second comment. I also see that you fixed the issue, so I will close this

Issue #739, opened 2023-01-23
Add option to filter out generated endpoints and dtos


mat-mik:
## What are the steps to reproduce this issue?

1. Have legacy system with hundreds of endpoints
2. Generate schemas

## What happens?

Hundreds/Thousands of files are generated.

## What were you expecting to happen?

Generate only a subset of endpoints and dtos. Maybe by OpenAPI `tags`? Configuration could look like:
```ts
input: {
      target: 'xxx',
      filter: {
            containsTags: ['abc']
      }
},
```

## What versions are you using?

**Package Version:** 6.11.1

melloware:
Closing as this has been implemented.

PR #810, opened 2023-03-22
feat(packages): add option to filter generated endpoints

yesyesufcurs:
In the case orval is run on a system with a lot of endpoints, it can be useful to limit which endpoints are actually generated.

## Status

**READY**

## Description

In the case orval is run on a system with a lot of endpoints, 
it can be useful to limit which endpoints are actually generated.

See: #739 

## Todos

- [ ] Tests
- [x] Documentation

## Steps to Test or Reproduce

1. Add a tag that should be filtered out to a orval.config for instance as follows:
```
output: {
      filterTags: ["pets"]
},
```
2. Only end points that contain tag "pets" will be generated.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -137,6 +137,9 @@ export type NormalizedInputOptions = {
   override: OverrideInput;
   converterOptions: swagger2openapi.Options;
   parserOptions: SwaggerParserOptions;
+  filters?: {
+    tags?: string[];
+  };
 };
 
 export type OutputClientFunc = (
@@ -169,6 +172,9 @@ export type InputOptions = {
   override?: OverrideInput;
   converterOptions?: swagger2openapi.Options;
   parserOptions?: SwaggerParserOptions;
+  filters?: {
+    tags?: string[];
+  };
 };
 
 export type OutputClient =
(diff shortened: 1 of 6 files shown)


jordanrigo (approved):

jordanrigo (approved):

packages/core/src/types.ts
```diff
@@ -51,6 +51,7 @@ export type NormalizedOutputOptions = {
   tsconfig?: Tsconfig;
   packageJson?: PackageJson;
   headers: boolean;
+  filterTag?: string[];
```
anymaniax: What do you think about having this.

```ts
filters?: {
  tags?: string[]
}
```

Like that we can add other filters like on the operation id.

What do you think about adding it in the `input` section and not the `output` one?

yesyesufcurs: I agree with changing it to have a filters object, that would make more sense. I will be updating this asap.

About the `input`  section I'm not sure. I think it is more of a question how we see it:

1. We only "read in" the relevant filtered elements.
2. We only "output" the relevant filtered elements.

The first one would imply adding it in the `input` section, the second one the `output` section.
The main reason I chose to add it to the output section is since the function that  I changed in the `api.ts` file only receives `NormalizedOutputOptions` as parameter. I could of course add the `NormalizedInputOptions` as parameter to get the desired change, or do you think there would be a smarter way to go about this?

anymaniax: I would do the filter [here](https://github.com/anymaniax/orval/blob/master/packages/core/src/generators/verbs-options.ts#L202) and so we don't even read the operation. And in that case, you can also pass the `NormalizedInputOptions ` as a parameter.

yesyesufcurs: I have fixed it as you proposed :)

anymaniax: Really nice works! Thanks for the pull request

Issue #616, opened 2022-10-10
Incorrect code generated when `readOnly: true` and also included in `required` 


localrobot:
When same schema is used for `get` and `post` request, some fields are marked as `readOnly` but the same field is also listed under `required`. In this situation the expected behaviour as per [Read-Only and Write-Only Properties](https://swagger.io/docs/specification/data-models/data-types/) section on Swagger docs is quoted below:

> If a `readOnly` or `writeOnly` property is included in the `required` list, `required` affects just the relevant scope – responses only or requests only. That is, read-only required properties apply to responses only, and write-only required properties – to requests only.

## What are the steps to reproduce this issue?

Here is an example (this is [`petstore.yaml`](https://github.com/OAI/OpenAPI-Specification/blob/main/examples/v3.0/petstore.yaml) with only relevant information:

```yaml
openapi: "3.0.0"
paths:
  /pets:
    get:
      operationId: listPets
      parameters:
        - name: limit
          in: query
          required: false
          schema:
            type: integer
      responses:
        '200':
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Pets"
    post:
      operationId: createPets
      requestBody:  # this is an addition
        required: true
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/Pet"
      responses:
        '201':
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Pet"

components:
  schemas:
    Pet:
      type: object
      required:
        - id
        - name
      properties:
        id:
          type: integer
          readOnly: true  # this is an addition
        name:
          type: string
        tag:
          type: string
    Pets:
      type: array
      items:
        $ref: "#/components/schemas/Pet"
```

and the `orval` config used:

```ts
export default defineConfig({
  mcaas: {
    input: {
      target: './petstore.yaml',
    },
    output: {
      clean: true,
      mode: 'split',
      workspace: 'src/generated/',
      target: './petstore.ts',
      schemas: './model',
      client: 'react-query',
      override: {
        requestOptions: false,
        query: {
          signal: false,
        },
      },
    },
  },
});
```

## What happens?

This is a snippet of generated code:

```ts
export interface Pet {
  readonly id: number;
  name: string;
  tag?: string;
}

export const createPets = (pet: Pet): Promise<AxiosResponse<void>> => {
  return axios.post(`/pets`, pet);
};

export type CreatePetsMutationResult = NonNullable<Awaited<ReturnType<typeof createPets>>>;
[...]

PR #813, opened 2023-03-29
feat(core): Only consider non-readonly properties in request bodies

maikdijkstra:
## Status

**READY**

## Description

Whenever a readonly property is defined in a request body then these may be sent as part of a response but should not be sent as part of a request according to the openapi 3.0 specification. To solve this using types, we can only allow the properties that are non-readonly since these are the actual request body.

Should fix: #616

## Steps to Test or Reproduce

1. Use a readonly property in a requestBody
2. Check that the readonly properties are discarded from the generated body type in the defined hooks.


packages/orval/src/client.ts
```diff
@@ -35,6 +36,13 @@ export const GENERATOR_CLIENT: GeneratorClients = {
   zod: zod()(),
 };
 
+const ORVAL_DEPENDENCIES: GeneratorDependency[] = [
```
anymaniax: I don't know about importing orval as a dependency of the generated code. You create then a direct dependence on orval. Would make more sense to generate it automatically with all the other schemas no?

maikdijkstra: Yes, I think that would be better specifically if Orval is only a dev dependency. The only question that I then have how should we distinguish this "orval defined" schema from "specification defined" schemas? Because I don't know whether it would be acceptable to mix these types among all other schemas?

What do you think would be a proper place to put these types @anymaniax?

anymaniax: Good question because in that case, we could have some conflicts with the generated models from the specification. Do you know that you can already achieve that with a mutator and exporting a BodyType in it? 

``` ts
export type BodyType<Body> = NonReadonly<Body>;
```

maikdijkstra: I did not know this then technically speaking the issue can be solved by simply setting the mutator correctly.
```typescript
// https://stackoverflow.com/questions/49579094/typescript-conditional-types-filter-out-readonly-properties-pick-only-requir/49579497#49579497
type IfEquals<X, Y, A = X, B = never> = (<T>() => T extends X ? 1 : 2) extends <
  T,
>() => T extends Y ? 1 : 2
  ? A
  : B;

type WritableKeys<T> = {
  [P in keyof T]-?: IfEquals<
    { [Q in P]: T[P] },
    { -readonly [Q in P]: T[P] },
    P
  >;
}[keyof T];

type Writable<T> = Pick<T, WritableKeys<T>>;
export type NonReadonly<T> = {
  [P in keyof Writable<T>]: T[P] extends number | string | boolean
    ? T[P]
    : NonReadonly<NonNullable<T[P]>>;
};

export type BodyType<Body> = NonReadonly<Body>;
```

maikdijkstra: Is this small use-case then important enough to change Orval to adhere to this OpenAPI specification? Or should this be the solution to issue #616 ?

@anymaniax

maikdijkstra:
@anymaniax I updated the PR, so now we keep track of whether the type contains any readonly properties, if it contains a readonly property anywhere in the object tree then the ```NonReadonly``` type is added to the imports at the request body. So, when readonly is not there than we also dont import it. We add the orval generated types in the index.ts of the schemas. Let me know whether this would be an appropriate solution.

anymaniax:
@maikdijkstra thanks a lot. Seems great. I found it weird to inject it in the index.ts. What do you think about inject it in each body file that need it? I was thinking would this need to be deactivated for some case? I don’t think so but what is your thoughts about this? I thought also about another solution. Be able to do the same thing as with a mutator but without the need to define all the mutator.

maikdijkstra:
@anymaniax My reason to inject it into the index.ts is that it will not show up with any other schemas in a separate file but is importable from the schemas folder and therefore only generate this type once. Does the body file mean the file with the endpoints? If so, then yes that is also an option, you will get quite some duplication if you generate an endpoint per file and all of them need it.

I think this case should not be deactivatable, since then the specification is simply wrong according to the OpenApi specification. 

I don't really follow your proposed solution using the mutator. Could you explain how that would work?

maikdijkstra:
@anymaniax Good to hear! I added a test case to the default config. I also updated the react-query/basic sample to have a readonly property in the request body

anymaniax:
Super @maikdijkstra I plan to checkout and test the branch tomorrow and merge it after that. I will create a new release with it during the week.

anymaniax:
@maikdijkstra for me all is good when you have time can you check the comment of @localrobot

michaelslec:
Will we get the same functionality for `writeonly` fields?

PR #820, opened 2023-04-07
fix: change min max and regex to not reuse a const

### LINKED ISSUE #819: Zod generator produced unnecessary consts for min and max values

Zod 6.13.1

Create a zod schema. In the OAS have a number property in a schema that specifies a minimum or maximum. Generate a zod schema client.

## What happens?

A `const` representing the value of the `min` or `max` arguments is exported and re-exported multiple times. The same variable name string can be re-generated in the same file multiple times because they are deterministic, and not guaranteed to be unique.

```ts
export const postFooServiceHeader = zod.object({
  "x-operation": zod.string().uuid()
})

export const postFooServicePurchasedAreaMin = 0;

export const postFooServiceBody = zod.object({
  "purchasedArea": zod.number().min(postFooServicePurchasedAreaMin),"cYear": zod.string(),"active": zod.boolean(),"enrollmentId": zod.string().uuid().optional(),"endDate": zod.string().datetime(),"startDate": zod.string().datetime(),"cOrgId": zod.string()
})

export const postFooServicePurchasedAreaMin = 0;export const postFooServiceTotalPriceMin = 0;

export const postFooServiceResponse = zod.object({
  "purchasedArea": zod.number().min(postFooServicePurchasedAreaMin),"active": zod.boolean(),"creatorGId": zod.string().uuid().optional(),"enrollmentId": zod.string().uuid(),"totalPrice": zod.number().min(postFooServiceTotalPriceMin),"endDate": zod.string().datetime(),"startDate": zod.string().datetime(),"connexOrgId": zod.string().optional()
})
```

The typescript error looks something like

`Cannot redeclare block-scoped variable 'postFooServicePurchasedAreaMin'.ts(2451)
const postFooServicePurchasedAreaMin: 0`

Just put scalar values directly in the function ... I don't think the consts are improving things.

## Any other comments?

I think I can submit a PR to fix this... but I'm not sure what to do about regex ... a scalar value of a number doesn't really need to be optimized, but re-creating the regex multiple times might be a real performance impact?

See https://github.com/anymaniax/orval/blob/master/packages/zod/src/index.ts#L188

**Operating System:** MacOS
**Package Version:** 6.13.1

sirtimbly:
## Status

**READY**

## Description

Fixes #819 

## Todos

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -56,7 +56,7 @@ const generateZodValidationSchemaDefinition = (
 ): { functions: [string, any][]; consts: string[] } => {
   if (!schema) return { functions: [], consts: [] };
 
-  const consts = [];
+  const consts: string[] = [];
   const functions: [string, any][] = [];
   const type = resolveZodType(schema.type);
   const required = schema.default !== undefined ? false : _required ?? false;
@@ -182,16 +182,10 @@ const generateZodValidationSchemaDefinition = (
   }
 
   if (min !== undefined) {
-    if (min === 1) {
-      functions.push(['min', `${min}`]);
-    } else {
-      consts.push(`export const ${name}Min = ${min};`);
-      functions.push(['min', `${name}Min`]);
-    }
+    functions.push(['min', `${min}`]);
   }
   if (max !== undefined) {
-    consts.push(`export const ${name}Max = ${max};`);
-    functions.push(['max', `${name}Max`]);
+    functions.push(['max', `${max}`]);
   }
   if (matches) {
     const isStartWithSlash = matches.startsWith('/');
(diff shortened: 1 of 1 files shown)


melloware:
There is a merge conflict

sirtimbly:
I'm sorry I forgot about this PR. I still think this is the correct choice, using consts for simple literal values isn't going to improve perf in a meaningful way, and it's easier to use literals than to write logic that properly created different consts for different min/max values.

melloware:
@sirtimbly did you see AnyMananiax comment "Just pushed a fix" in April?  Have you tried 6.19.1 and do you still get the same errors you were getting before?

melloware:
also @sirtimbly can you check out this issue with Zod and two dimensional array?  https://github.com/anymaniax/orval/issues/1027

sirtimbly:
> @sirtimbly did you see AnyMananiax comment "Just pushed a fix" in April? Have you tried 6.19.1 and do you still get the same errors you were getting before?

Tested, it seems to be fixed because the name calculation is more complete.

PR #835, opened 2023-04-19
feat(packages): add regex option to filter generated endpoints

ta1m1kam:
## Status

**READY**

## Description

Issue: https://github.com/anymaniax/orval/issues/834

Allow Regex to filter OpenAPI yaml tags.

Add a regex option to filter that accepts a boolean and tags will be able to receive string[] or RegExp.
If regex option is true, tags receive RegExp
If regex option is false or undefined, tags receive string[].

## Related PRs

| [yesyesufcurs:issues/739](https://github.com/yesyesufcurs/orval/tree/issues/739) | [link](https://github.com/anymaniax/orval/pull/810) |

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

1. Add a tag that should be filtered out to a orval.config for instance as follows:

```
input: {
  filters: {
     tags: [/pets/]
  }
},
```

Only end points that contain tag "pets" will be generated.


packages/core/src/types.ts
```diff
@@ -131,15 +131,24 @@ export type NormalizedOperationOptions = {
...
 export type NormalizedInputOptions = {
   target: string | Record<string, unknown> | OpenAPIObject;
   validation: boolean;
   override: OverrideInput;
   converterOptions: swagger2openapi.Options;
   parserOptions: SwaggerParserOptions;
-  filters?: {
-    tags?: string[];
-  };
+  filters?: {} & InputFilterTags;
```
anymaniax: would be simpler to have an array of string and regex at the same time no?

ta1m1kam: @anymaniax 
Thank you for your comment.
I fixed the filter array can receive string and regex.

anymaniax:
Thanks!

PR #841, opened 2023-04-24
feat(core): add optional flag to force number type instead of bigint

GBDev13:
## Status

**READY**

## Description

The recent change from to add support for bigints caused issues in frontend code, as commented [here](https://github.com/anymaniax/orval/issues/824)

So, in this PR, I just added an optional field called `useNumberTypeForBigInt` that when true, forces the typing to always be number, instead of bigint.

## Todos

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -100,7 +100,7 @@ export type NormalizedOverrideOutput = {
   useDates?: boolean;
   useTypeOverInterfaces?: boolean;
   useDeprecatedOperations?: boolean;
-  useNumberTypeForBigInt?: boolean;
+  useBigInt?: boolean;
 };
 
 export type NormalizedMutator = {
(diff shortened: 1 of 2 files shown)


packages/core/src/types.ts
```diff
@@ -100,6 +100,7 @@ export type NormalizedOverrideOutput = {
   useDates?: boolean;
   useTypeOverInterfaces?: boolean;
   useDeprecatedOperations?: boolean;
+  useNumberTypeForBigInt?: boolean;
```
anymaniax: I would do useBigInt and only generate bigInt when it's set to true. What do you think? Cause indeed right now it's causing a bit a breaking change with this

GBDev13: Yeah, I agree, that way it will support both needs and not break all the frontends that are currently using it, haha. I pushed this change!

PR #848, opened 2023-05-03
fix(docs): add scrollbar to toc

### LINKED ISSUE #847: [docs] toc can only be scrolled once you scroll to the bottom of the page
viewing https://orval.dev/reference/configuration/output. when I want to scroll the toc, I need first to get to the bottom of the page and then scroll more.

need to add a height property, like there is on the sidebar

chiptus:
fix #847


anymaniax:
thanks @chiptus

PR #852, opened 2023-05-06
Vue query improve reactive support

gkweb:
## Status

**READY/WIP/HOLD**

READY

## Description

Addresses: https://github.com/anymaniax/orval/issues/851

A few sentences describing the overall goals of the pull request's commits.

Improves vue reactivity support.

## Related PRs

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

See the sample update, I've expanded on the sample by dynamically updating the list to load in a single pet.

```bash
> git pull --prune
> git checkout <branch>
> grunt jasmine
```

1.


gkweb:
@anymaniax  - Now ready for review, remaining type fixes added. Thank you for your consideration.

Maxim-Mazurok:
This seems to work quite well for me

anymaniax:
I see that the tests are not passing. There is a problem for `react-query` and `svelte-query`

gkweb:
> I see that the tests are not passing. There is a problem for `react-query` and `svelte-query`

@anymaniax  - Are you available on a chat anywhere? Or can you share the output of failure here? I can take a look then! Many Thanks

anymaniax:
I try to be available on the Orval discord if you need to talk to me. You will see the problem if in the project you run the tests in the tests folder you just need to install the dependencies in it and then `yarn generate && yarn build`

packages/query/src/index.ts
```diff
@@ -945,11 +977,15 @@ const generateQueryHook = async (
     ];
 
     const queryKeyFnName = camel(`get-${operationName}-queryKey`);
-    const queryKeyProps = toObjectString(
+    let queryKeyProps = toObjectString(
       props.filter((prop) => prop.type !== GetterPropType.HEADER),
       'implementation',
     );
 
+    if (OutputClient.VUE_QUERY) {
```
gkweb: @anymaniax is`OutputClient.VUE_QUERY` a reliable way to detect if the runner is vue only? 

Ta

gkweb:
> > I see that the tests are not passing. There is a problem for `react-query` and `svelte-query`
> 
> @anymaniax - Are you available on a chat anywhere? Or can you share the output of failure here? I can take a look then! Many Thanks

All good I see this now! 

Left a comment above. Will join discord. Ta

packages/query/src/index.ts
anymaniax: Yes it's the only way to know it

Maxim-Mazurok:
As I've demonstrated in https://github.com/anymaniax/orval/issues/851#issuecomment-1539428961 the query isn't being prefetched when selected petId changes.
This seems to fix the issue but I hope there's a better way:
```ts
import { watch } from 'vue';
watch(petId, () => {
  petQuery.refetch();
});
```

Also linking the issue, this PR fixes https://github.com/anymaniax/orval/issues/851

Maxim-Mazurok:
I think I figured it out.
In `getShowPetByIdQueryKey` you're doing `unref()` so that you can return a string.
If we instead return an array (which is also a good key option for Vue Query as far as I understand) - then array values will still be reactive.
So this works:

```patch
export const getShowPetByIdQueryKey = (petId: MaybeRef<string>, version = 1) =>
-  [`/v${unref(version)}/pets/${unref(petId)}`] as const;
+  [`v`, version, `/pets/`, petId] as const;
```

gkweb:
> I think I figured it out.
> 
> In `getShowPetByIdQueryKey` you're doing `unref()` so that you can return a string.
> 
> If we instead return an array (which is also a good key option for Vue Query as far as I understand) - then array values will still be reactive.
> 
> So this works without manual watcher:
> 
> 
> 
> ```patch
> 
> export const getShowPetByIdQueryKey = (petId: MaybeRef<string>, version = 1) =>
> 
> -  [`/v${unref(version)}/pets/${unref(petId)}`] as const;
> 
> +  [`/v`, version, `/pets/`, petId] as const;
> 
> ```

Thanks @Maxim-Mazurok 

I agree with the query key solution as Vue-query handles it and unwraps automatically. Let me take a look when I get some time. My first suggestion was to update queryKey also. Cheers

Maxim-Mazurok:
Found another small issue:
image

This seems to fix that:
```patch

**Comment by Maxim-Mazurok** [comment]
Found another option which also works:

```ts
export const getShowPetByIdQueryKey = (petId: MaybeRef<string>, version = 1) =>
  computed(() => [`/v${unref(version)}/pets/${unref(petId)}`]);
```

Maxim-Mazurok:
Implemented this option based on this PR
```ts
[`/v`, version, `/pets`]
```
https://github.com/gkweb/orval/pull/1

gkweb:
@anymaniax tests are now passing for this, so awaiting your review.

Maxim-Mazurok:
> [`/v`, version, `/pets`]

I did it this way because I wasn't sure how Vue Query hashes the key, I thought that if it does `[].join("")` for example - then `/v1/air/plane` would have the same key as `/v1/airplane` which would be a mistake.

Checking out their code, I've created a test to confirm that we can drop `/` from the keys and not get collisions. Keys are hashed using `JSON.stringify()` so commas will serve as delimiters, no need for `/` indeed. Here's the test:

```js
function isPlainObject(o) {
  if (!hasObjectPrototype(o)) {
    return false
  }

  // If has modified constructor
  const ctor = o.constructor
  if (typeof ctor === 'undefined') {
    return true
  }

  // If has modified prototype
  const prot = ctor.prototype
  if (!hasObjectPrototype(prot)) {
    return false
  }

  // If constructor does not have an Object-specific method
  if (!prot.hasOwnProperty('isPrototypeOf')) {
    return false
  }

  // Most likely a plain Object
  return true
}

function hasObjectPrototype(o) {
  return Object.prototype.toString.call(o) === '[object Object]'
}

function hashQueryKey(queryKey) {
  return JSON.stringify(queryKey, (_, val) =>
    isPlainObject(val)
      ? Object.keys(val)
          .sort()
          .reduce((result, key) => {
            result[key] = val[key]
            return result
          }, {})
      : val,
  )
}

console.log(hashQueryKey(["v", 1, "air", "plane"])) // prints ["v",1,"air","plane"]
console.log(hashQueryKey(["v", 1, "airplane"])) // prints ["v",1,"airplane"]
```

samples/vue-query/src/api/endpoints/petstoreFromFileSpecWithTransformer.ts
```diff
@@ -49,7 +49,7 @@ export const listPets = (
 export const getListPetsQueryKey = (
   params?: MaybeRef<ListPetsParams>,
   version = 1,
-) => [`/v`, version, `/pets`, ...(params ? [params] : [])] as const;
+) => ['v', version, 'pets', ...(params ? [params] : [])] as const;
```
gkweb, line 52: @Maxim-Mazurok 

> RE: I did it this way because I wasn't sure how Vue Query hashes the key, I thought that if it does [].join("") for example - then /v1/air/plane would have the same key as /v1/airplane which would be a mistake.

I've changed what functionality you had here previously. Is this what you're describing as a requirement? (forward slash removed). Any string concatenation with variables even in the pathname should be cut up and recognised as a var. 
eg: `/something${var}/pets/${petId}` will explode to `['something',var,'pets',petId]`

gkweb:
> Just tested this on my project and it generates everything well and works well, so LGTM.
> 
> There's another reactivity issue tho, now with `enabled` option.
> 
> Basically, we have this now: `enabled: !!myArgument` and since `myArgument` is a ref/proxy - `enabled` is always true.
> 
> I've solved it by doing this:
> 
> ```diff
> - return { queryKey, queryFn, enabled: !!myArgument, ...queryOptions };
> + return { queryKey, queryFn, enabled: computed(() => !!unref(myArgument)), ...queryOptions };
> ```
> 
> Seems to work fine. I'll submit PR

vue-query supports ref here I'm pretty sure, So this only needs to be changed to: `enabled: myArgument` right? - If you need to expand on that just pass in your own _enabled_ implementation?

packages/query/src/index.ts
```diff
@@ -849,7 +878,7 @@ export const ${camel(
     `${operationPrefix}-${type}`,
   )}(${queryOptionsVarName}) as ${returnType};
 
-  query.queryKey = ${queryOptionsVarName}.queryKey;
+  query.queryKey = ${queryOptionsVarName}.queryKey as QueryKey;
```
Maxim-Mazurok: This seems to apply to `samples/react-query/hook-mutator/endpoints.ts` as well as vue one, not sure if that was your intention.

samples/vue-query/src/api/endpoints/petstoreFromFileSpecWithTransformer.ts
Maxim-Mazurok: It's all good now, thank you. I was just confirming that we can get rid of slashes and keep arrays just like you did, all good currently 👍

Maxim-Mazurok:
> vue-query supports ref here I'm pretty sure

It doesn't seem to be the case...

> So this only needs to be changed to: enabled: myArgument right?

Nope, that doesn't work, it's being evaluated to `true` when I pass `ref(undefined)`.

> If you need to expand on that just pass in your own enabled implementation?

I guess I could... But I rather patch it up in the generator, see https://github.com/gkweb/orval/pull/2 We don't have to include it in this PR tho if you'd like to do it separately. Seems like a small fix to me tho.

anymaniax:
Hello guys, thanks for the work can I do something here? is the pull request ready?

gkweb:
> Hello guys, thanks for the work can I do something here? is the pull request ready?

Hey @anymaniax - yeah any subsequent issue can be fixed in additional pull requests. 
There are some more improvements which can still be done here but this has paved the way to improving reactivity.

anymaniax:
Just did a first tour. Insane work thanks guys ❤️

samples/vue-query/src/components/pet.vue
```diff
@@ -0,0 +1,23 @@
...
+    {{ pet }}
+  </div>
+</template>
+
+<script lang="ts" setup>
+import { computed, unref } from 'vue';
+import { useShowPetById } from '../api/endpoints/petstoreFromFileSpecWithTransformer';
+
+const props = defineProps<{ petId: string }>();
+const petId = computed(() => props.petId);
```
anymaniax, line 13: the computed is not really need here right?

gkweb: The props is a reactive object. Wrapping in the computed can help to demonstrate reactivity of the individual prop though?

https://vuejs.org/guide/extras/reactivity-transform.html#reactive-props-destructure

Let me know if you want this changed?

anymaniax: no it was to be sure

anymaniax (approved):
Thanks 🎉

AlmarAubel:
Great work, any idea when you will release this to NPM?

anymaniax:
This week probably. I want to add one more pull request before the release.

anymaniax:
Changed my mind it's released right now

PR #861, opened 2023-05-17
feat: anymaniax#352 support allOf

### LINKED ISSUE #352: Required field ignored in case of allOf

yokoyama-takuya:
## Status

**READY**

Fix #352

## Description

fix allOf type definition
e.g. https://github.com/anymaniax/orval/issues/352

## Related PRs
https://github.com/anymaniax/orval/pull/503

### before

```ts
export type GetPets200PetsItemAllOf = {
  /** required name */
  name?: unknown;
  /** @deprecated */
  owner?: unknown;
};

export type GetPets200PetsItem = Pet & unknown & GetPets200PetsItemAllOf;

export type GetPets200 = {
  pets?: GetPets200PetsItem[];
};

/**
 * default Person description
 */
export interface Person {
  id: string;
  /** default Person name */
  name?: string;
}

export type PetOwnerAllOf = {
  /** required name */
  name?: unknown;
};

export type PetOwner = Person & PetOwnerAllOf;

/**
 * default Pet description
 */
export interface Pet {
  id?: string;
  /** default Pet name */
  name?: string;
  owner: PetOwner;
}
```

### after

```ts
/**
 * default Person description
 * @deprecated
 */
export type GetPets200PetsItemOwner = {
  id: string;
  /** required name */
  name: string;
};

/**
 * default Pet description
 */
export type GetPets200PetsItem = {
  id?: string;
  /** required name */
  name: string;
  /**
   * default Person description
   * @deprecated
   */
  owner: GetPets200PetsItemOwner;
};

export type GetPets200 = {
  pets?: GetPets200PetsItem[];
};

/**
 * default Person description
 */
export interface Person {
  id: string;
  /** default Person name */
  name?: string;
}

/**
 * default Person description
 */
export type PetOwner = {
  id: string;
  /** required name */
  name: string;
};

/**
 * default Pet description
 */
export interface Pet {
  id?: string;
  /** default Pet name */
  name?: string;
  /** default Person description */
  owner: PetOwner;
}
```


yokoyama-takuya:
@anymaniax  Resubmitted a fix pull request for allOf. I have rewritten it to a simpler code, so please consider it.

packages/core/src/getters/object.ts
```diff
@@ -101,7 +101,7 @@ export const getObject = ({
           acc.value += '{';
         }
 
-        const doc = jsDoc(schema as SchemaObject, true);
+        const doc = jsDoc(resolvedValue.originalSchema, true);
```
anymaniax: We don't need both here?

packages/core/src/resolvers/object.ts
```diff
@@ -15,6 +17,34 @@ export const resolveObject = ({
   combined?: boolean;
   context: ContextSpecs;
 }): ResolverValue => {
+  if (typeof schema === 'object' && (schema as SchemaObject)?.allOf) {
```
anymaniax: is it causing a breaking change in the generated files?

packages/core/src/getters/object.ts
yokoyama-takuya: Does it mean that there are cases where `originalSchema` is empty?

packages/core/src/resolvers/object.ts
yokoyama-takuya: I think `Change's before after` is easy to understand.
https://github.com/anymaniax/orval/pull/861#issue-1713057019

`allOf` works as a mix in and is different from Typescript's intersection type. Therefore, it was necessary to merge the schema first.

yokoyama-takuya (commented):
@anymaniax Thanks for the review. I replied.

packages/core/src/getters/object.ts
anymaniax: No but here we take the jsDoc only for the deeper ref. We probably also or only need the jsDoc of the parent no?

packages/core/src/resolvers/object.ts
anymaniax: my bad didn't check all the comment of the pr

packages/core/src/getters/object.ts
yokoyama-takuya: For example, originalSchema is appropriate when rewriting ref's parent description with allOf.
Also, it has the same behavior as Swagger UI.

スクリーンショット 2023-05-17 20 20 06

yokoyama-takuya: The screenshot above is correct. It was from the schema.
Fixed here.

https://github.com/anymaniax/orval/pull/861/commits/0fac1cc37421c7e3bbebaa380ed8e9f0af7edc60

packages/core/src/resolvers/object.ts
anymaniax: Do you think it can be moved in the combineSchemas function? And maybe refactor that one only? It's executed in getters/object.ts and line `41`

yokoyama-takuya: I tried but nested allOf is not correct.  Because `mix in` is not working at the following line.
https://github.com/anymaniax/orval/blob/master/packages/core/src/getters/object.ts#L93

anymaniax: Ok then for schemas we could have a problem no? Since [here](https://github.com/anymaniax/orval/blob/master/packages/core/src/generators/schema-definition.ts) it’s calling resolveValue only

yokoyama-takuya: It's possible, but I can't get to that line in the OpenAPI description I know. Which will be the sample?

```yaml
schema:
  allOf:
    - $ref: 'schemas/pet.yaml#/components/schemas/Pet'
    - required:
        - name
    - type: object
      properties:
        name:
          description: required name
        owner:
          deprecated: true
```

yokoyama-takuya: I got it. Case by components top level allOf. I will add test case.

yokoyama-takuya: Converted to getters/object.ts. The components side also worked correctly.

packages/core/src/getters/object.ts
```diff
@@ -90,6 +95,8 @@ export const getObject = ({
           propName = propName + 'Property';
         }
 
+        schema = mergeAllOf(schema, context);
```
anymaniax, line 98: During your testing you discover why you obsoletely need to do it here too? Just for my understanding

anymaniax: Otherwise it's pretty great I am gonna merge. I will need also to cleanup the combineSchemas function

yokoyama-takuya: Thank you for review.
This line is required. Case by using allOf at properties name, generated `unknown` type.

スクリーンショット 2023-05-23 20 52 22

anymaniax:
I did other test with more complex specification. I got a RangeError: Maximum call stack size exceeded.

yokoyama-takuya:
@anymaniax Can I have the Swagger file that I tested here?

anymaniax:
I cannot share it but will try to do a reproduction of the issue

yokoyama-takuya:
@anymaniax In specs where the elements of objects and arrays are cyclically referenced, as expected. I wonder if the recursive function never ends. Is that API realistic?

PR #864, opened 2023-05-17
chore: dereference whole API on importing to avoid extra definitions

hugoArregui:
My solution to https://github.com/anymaniax/orval/issues/863  I'm not quite sure if this is a proper solution for the problem, but it works for me.

Basically, I'm dereferencing the whole API at the beginning so there are no extra schemas that could generate types.

### DIFF
--- a/packages/orval/src/import-specs.ts
+++ b/packages/orval/src/import-specs.ts
@@ -35,16 +35,8 @@ const resolveSpecs = async (
       }
     }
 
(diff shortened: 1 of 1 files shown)


anymaniax:
We can add the deferencing but as an option in the input options of orval.

soartec-lab:
@hugoArregui @anymaniax 
This PR is too old. Also, since a conflict has occurred, I will close it once. I would be happy to open it again if necessary.

PR #887, opened 2023-07-03
feat(axios): use named parameters

Marcel-G:
## Status

**READY**

## Description

Often times routes have many path parameters. As the number grows, type-safety is less effective as arguments (commonly string or number types) can easily get used in the wrong order.
Using named parameters helps this, as you get the added safety of needing to specify the appropriate name for the parameter.

This PR adds a `useNamedParameters` option for the axios client which creates path parameters as a single object.

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout Marcel-G:feat/named-parameters
> yarn build
> cd /tests
> yarn generate
> vi generated/axios/named-parameters/endpoints.ts
```

Heres an example output:

```ts
/**
 * @summary Info for a specific pet
 */
export const showPetById = <TData = AxiosResponse<Pet>>(
  { petId, version = 1 }: ShowPetByIdPathParameters, options?: AxiosRequestConfig
): Promise<TData> => {
  return axios.get(
    `/v${version}/pets/${petId}`, options
  );
}
type ShowPetByIdPathParameters = {
  petId: string,
  version?: number,
}
```


docs/src/pages/reference/configuration/output.md
```diff
@@ -671,6 +671,36 @@ Default Value: `'root'`.
 
 Can be used to set the value of `providedIn` on the generated Angular services. If `false`, no `providedIn` will be set. If `true` or not specified, it will fall back to the default value: `root`.
 
+#### axios
+
+Type: `Object`.
+
+Give you specific options for the axios client
```
Marcel-G:
```suggestion
Give specific options for the axios client
```

docs/src/pages/reference/configuration/output.md
```diff
@@ -671,6 +671,36 @@ Default Value: `'root'`.
...
+
+```js
+module.exports = {
+  petstore: {
+    output: {
+      ...
+      override: {
+        angular: {
```
Marcel-G:
```suggestion
        axios: {
```

anymaniax:
it's doesn't seems to work properly when there is no path params

Marcel-G:
@anymaniax this change seems to have some overlap with https://github.com/anymaniax/orval/pull/396. Do you have a preference between the two implementations?

anymaniax:
Hello @Marcel-G and thanks a lot. Seems great! Would you be interested to add a global config for that and do it for all client instead?

anymaniax:
@Marcel-G do you think we can merge the other one directly?

Marcel-G:
> @Marcel-G do you think we can merge the other one directly?

Actually this PR goes with a slightly different approach so the other PR is redundant if we go this way.

anymaniax:
@Marcel-G for me other one make more sense to have a more global way of doing it. What do you think?

anymaniax:
@Marcel-G can you also contact me on discord when you have the time?

PR #894, opened 2023-07-12
feat: make all params optional #858

### LINKED ISSUE #858: Make all params optional
This is realted to https://github.com/gkweb/orval/pull/2

Basically there are checks in place to `enable` query only when query parameters are not falsy, but when I try to pass something like `Ref<string | undefined>` to `useBla()` - it doesn't let me because it expects parameter to be `MaybeRef<string>`.

Here's a potential solution:

```diff
diff --git a/packages/query/src/utils.ts b/packages/query/src/utils.ts
index b5a9424..51e5e9a 100644
--- a/packages/query/src/utils.ts
+++ b/packages/query/src/utils.ts
@@ -87,7 +87,7 @@ export function vueWrapTypeWithMaybeRef(input: string): string {
     .map((param) => {
       const [paramName, paramType] = param.split(':');
       if (paramType) {
-        return `${paramName}: MaybeRef<${paramType.trim()}>,`;
+        return `${paramName}: MaybeRef<${paramType.trim()} | undefined | null>,`;
       } else {
         return `${param},`;
       }
```

I didn't want to include `null`, but since our API is in C# it's kinda hard to avoid dealing with nulls in responses, etc.

Maxim-Mazurok:
## Status

**READY**

## Description

Makes all query params optional to resolve #858 

## Todos

- [x] Tests
- [x] Documentation - not needed IMO
- [ ] Changelog Entry (unreleased) - not sure what to do about it
- [x] Resolve conflicts/cleanup
- [x] Make it a configurable non-default option so not to break other people

## Steps to Test or Reproduce

See tests https://github.com/Maxim-Mazurok/orval/blob/858/samples/vue-query/src/tests/all-params-optional.test-d.ts


anymaniax:
Hello @Maxim-Mazurok, could it cause breaking chance for others?

WickyNilliams:
commented on the issue, before seeing the PR. so re-posting here to keep closer to the code:

why make them optional? shouldn't you follow whatever the schema says? if you want them optional, you can wrap the generated composable in your own which handles this, no?

Maxim-Mazurok:
> Hello @Maxim-Mazurok, could it cause breaking change for others?

Yes, I will make this configurable

> why make them optional?

Because otherwise, Orval has a feature that doesn't make sense.
See https://github.com/anymaniax/orval/blob/ed1261af7ff773ca1338302e313ac8f02c3da00c/samples/vue-query/src/api/endpoints/petstoreFromFileSpecWithTransformer.ts#L319
The query is only enabled when all parameters are not falsy.
It is quite handy, especially when you chain multiple queries. Then you can do something like:
```ts
const getPetsQuery = useGetPetsQuery();
const firstPetId = computed(() => getPetsQuery.data.value?.data?.[0]);
const petInfoQuery = useGetPetInfoQuery({ petId: firstPetId });
```
This will magically fetch pets, and then as soon as they are fetched - it'll fetch the first pet info.

> shouldn't you follow whatever the schema says?

Well, that's a good point, I guess you might end up with invalid queries this way.
For example, you could have 
```ts
const firstPetId = computed(() => undefined);
const petInfoQuery = useGetPetInfoQuery({ petId: firstPetId });
```
In that example `firstPetId` will always be `undefined` and your query will never be enabled.
I guess that's a downside.

However, there's another interesting thing: if you make `petId` not required in schema - you will end up with `petId: MaybeRef<string> | undefined` type. This kinda breaks your reactivity, you can either pass a reactive string, or a non-reactive undefined. You can't pass reactive undefined that will become a string. And this kinda kills the beauty of Vue Query for me.

> if you want them optional, you can wrap the generated composable in your own which handles this, no?

Yeah, I guess that could be an option. However, this kinda defeats the purpose of using Orval... I don't want to manually wrap generated composables, I want to reduce manual to a minimum by generating composables that work for me.

Maxim-Mazurok:
I think most of the arguments are fair, I found it a bit unexpected that some queries were disabled due to falsy values.

However, we've been using this approach for a while now, including this PR as a patch in our project. So I would like to keep it for now.

Perhaps you could create a PR that adds a non-default configurable option to disable the "only enable when values are truthy" feature? This way it won't break existing users. And I would be interested in giving that PR a try. 

---

Also, I just remembered that this feature and this PR are only about route/url params. I'd say it's quite unlikely you'll have boolean type in the route, like `/api/pet/false/details` or something like that. You can still pass falsy values without issues as query params like `?limit=false`, etc.

---

My plan for this PR:
- [ ] Resolve conflicts/cleanup
- [ ] Make it a configurable non-default option so not to break other people
- [ ] Add tests

Would you guys agree to this plan, or would you still insist on removing this feature altogether?
@anymaniax @WickyNilliams

melloware:
Looks like there are conflicts with this PR.

Maxim-Mazurok:
> Looks like there are conflicts with this PR.

Hello, yes, I'm back from the holidays and going to continue work on my PRs for Orval this or next week, will also connect with @anymaniax on discord, cheers :)

packages/core/src/getters/params.ts
```diff
@@ -83,6 +90,11 @@ export const getParams = ({
       },
     });
 
+    let paramType = resolvedValue.value;
+    if (output.allParamsOptional) {
+      paramType = `${paramType} | undefined | null`; // TODO: maybe check that `paramType` isn't already undefined or null
```
Maxim-Mazurok, line 95: This is the core change directly related to the issue

Maxim-Mazurok:
@WickyNilliams this is ready for review now if you'd like to have a look

melloware (approved):

Maxim-Mazurok:
It looks like @WickyNilliams is more involved in [other](https://github.com/drwpow/openapi-typescript) [projects](https://github.com/kubb-project/kubb), so I'll merge it for now to avoid conflicts, but still open to feedback, cheers!

Issue #679, opened 2022-11-23
Path Parameters Are Not Encoded


lew:
## What are the steps to reproduce this issue?

1. Create a minimal OpenAPI example like    
```json
{
  "info": {
    "title": "My API",
    "version": "1.0.0"
  },
  "openapi": "3.0.0",
  "paths": {
    "/test/{parameter}": {
      "get": {
        "parameters": [
          {
            "in": "path",
            "name": "parameter",
            "required": true,
            "schema": {
              "type": "string"
            }
          }
        ],
        "responses": {
          "200": {
            "description": "OK"
          }
        }
      }
    }
  }
}

```
2. Generate the functions.
3. Use a parameter with special characters in above example.

## What happens?

The parameter `parameter` is just string concatenated into the URL:
```ts
export const getTestParameter = <TData = AxiosResponse<void>>(
  parameter: string,
  options?: AxiosRequestConfig
): Promise<TData> => {
  return axios.get(`/test/${parameter}`, options)
}
```

As a result, requests fail. 

## What were you expecting to happen?

Strings in the path fragment must be encoded 

```ts
export const getTestParameter = <TData = AxiosResponse<void>>(
  parameter: string,
  options?: AxiosRequestConfig
): Promise<TData> => {
  return axios.get(`/test/${encodeURIComponent(parameter)}`, options)
}
```

## Any other comments?

URL encoding should take place here because neither the rest of the app should care about encoding nor is it possible to later encode a fully constructed URL. 

A programmer would typically prefer to use a method like this to avoid having to worry about internals:
```ts
const someParam = 'special[s]/?test'
const results = useGetTestParameter(someParam)
```

This bug happens at different places in the generated code.

## What versions are you using?

    "orval": "6.11.0-alpha.6"

lew:
I believe it is a good idea not to allow the function's user to provide input that breaks the execution.
When a user enters a path segment with a character like `/`, it completely changes the resulting URI; when they enter a path segment with a reserved character, it can result in invalid HTTP requests. 

The [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986#section-2.4) explicitly mentions when to encode the URI:
> Under normal circumstances, the only time when octets within a URI
   are percent-encoded is during the process of producing the URI from
   its component parts.  This is when an implementation determines which
   of the reserved characters are to be used as subcomponent delimiters
   and which can be safely used as data.  Once produced, a URI is always
   in its percent-encoded form.

[...]

PR #895, opened 2023-07-13
fix: path parameters are not encoded #679

Maxim-Mazurok:
## Status

**READY**

## Description

Addresses https://github.com/anymaniax/orval/issues/679 issue, based on my patch https://github.com/anymaniax/orval/issues/679#issuecomment-1583652672

Also updates vue to the latest, so that we can use `MaybeRef` from it's indirect dependency `@vue/reactivity`. I need this because the old import was unresolved when using `"moduleResolution": "bundler"` which is the default now for vue in in `node_modules/@vue/tsconfig/tsconfig.json`

I think some people might have implemented encoding on their end, so this probably should be a major release, otherwise people will get double-encoding.

## Todos

- [x] Tests
- [x] Documentation
- [ ] Changelog Entry (unreleased) - not sure what to do about this


anymaniax:
@Maxim-Mazurok thanks for the help. It's encode every parameters? If wouldn't be better to add that as a config?

Maxim-Mazurok:
> @Maxim-Mazurok thanks for the help. It's encode every parameters? If wouldn't be better to add that as a config?

I would argue that it should be encoded by default, for example `/api?song=Me&you` will unexpectedly be parsed as `song=Me; you=`

But I guess it'd be easier to make a non-default option, then no need for major release, I'll do that.

melloware:
I would think we want all parameters encoded by default no?

Maxim-Mazurok:
> I would think we want all parameters encoded by default no?

I would want that as well, but as @anymaniax mentioned this will be a breaking change and might require a major release. I think that would be a good idea overall, what do you think?

melloware:
I mean I guess I don't see the harm in encoding URLs it feels like the safe thing to do. I wonder what @anymaniax concern is?

Maxim-Mazurok:
> I mean I guess I don't see the harm in encoding URLs it feels like the safe thing to do. I wonder what @anymaniax concern is?

Well, I'm pretty sure that we agree that it's a good thing, but changing default behaviour will be a breaking change.

For example, if someone had `/api?song=Me&you` in their project - their server would receive `{song: "Me", you: undefined}`. Even thought this is a bug, some people might be fine with this bug and actually want it to work this way. If we make the change and encode all by default - their server will now get `{song: "Me & you"}` which is technically correct, but imagine if they already have some logic like `if (song === "Me")` to handle this request - then it'll break, potentially introducing bugs.

So I think the best thing to do is to encode all params by default and make a major release (major version bump) so that people know to expect breaking changes.

The second best thing to do is to make encoding params a non-default option.

packages/query/src/index.ts
```diff
@@ -368,13 +368,22 @@ const generateQueryRequestFunction = (
...
   let props = _props;
+  let route = _route;
+
   if (isVue(outputClient)) {
     props = vueWrapTypeWithMaybeRef(_props);
+    route = vueMakeRouteReactive(_route);
   }
+
+  if (output?.urlEncodeParameters) {
+    route = makeRouteSafe(route);
```
Maxim-Mazurok, line 382: This is the core change that addresses the original issue

Maxim-Mazurok:
This is ready for review now

Summary:
- Added `urlEncodeParameters` option, `false` by default
- Enable `tsc --noEmit` for `packages/orval` (ignored most of the pre-existing issues for now)
- Added docs for `urlEncodeParameters` and `allParamsOptional` new output options
- Added `zod` option for `OutputClient`
- Passing `output: NormalizedOutputOptions` all the way into `packages/query`'s `builder` (now I realise that it probably makes sense to have `urlEncodeParameters` for all clients and I could probably have done it somewhere else... perhaps we can refactor in the future, still I think it's useful to pass config options to the builder)

anymaniax (approved):
Seems really good to me! Thanks @Maxim-Mazurok

packages/query/src/utils.ts
```diff
@@ -109,5 +109,14 @@ export const vueUnRefParams = (props: GetterProps): string => {
     .join('\n');
 };
 
+export const wrapRouteParameters = (
+  route: string,
+  prepend: string,
+  append: string,
+): string => route.replaceAll(/\${(.+?)}/g, `\${${prepend}$1${append}}`);
```
lukaw3d, line 116: getRouteAsArray uses a slightly different regex

Maxim-Mazurok: Ah, interesting. I think mine is better because it seems to be more aligned with OpenAPI specs by allowing special characters and an empty string: https://github.com/OAI/OpenAPI-Specification/issues/1969#issuecomment-509016671

But yeah, it's probably better to refactor it to avoid duplication

Maxim-Mazurok: I've created https://github.com/anymaniax/orval/issues/1084 to address that

PR #905, opened 2023-08-02
fix(core): capitalize http verb patch

hkfi:
## Status

READY

## Description

Changed the HTTP verb 'patch' to 'PATCH'. 
PATCH is the only HTTP method that is case sensitive. It needs to be capitalized to work.

## Related PRs

None

## Todos

- [x] Tests
- [x] Documentation

### DIFF
--- a/docs/src/pages/guides/custom-client.md
+++ b/docs/src/pages/guides/custom-client.md
@@ -31,7 +31,7 @@ export const customInstance = async <T>({
   data,
 }: {
   url: string;
-  method: 'get' | 'post' | 'put' | 'delete' | 'PATCH';
+  method: 'GET' | 'POST' | 'PUT' | 'DELETE' | 'PATCH';
   params?: any;
   data?: BodyType<unknown>;
   responseType?: string;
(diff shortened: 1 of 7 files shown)


anymaniax:
I never had that issue. In which case do have that? An what http client do you use and if it’s a dependency which version?

hkfi:
When using JavaScript's fetch.

Another person described here in the issues as well:
https://github.com/anymaniax/orval/issues/888

anymaniax:
@hkfi the problem it's that currently done like that It would be a breaking change and you can uppercase all method in your custom instance. Would be better to create a fetch client no?

anymaniax:
Then would be better to pass everything to uppercase no?

Jacob-Roberts:
The only verb where uppercase is required is PATCH, so it doesn't need to be every verb updated as part of this PR, but yes it could be every verb

melloware (approved):

melloware:
It should be every verb. They all should be uppercase

melloware:
Looks like this is failing the ZOD compilation because `PathItemObject` from openapi-ts

https://github.com/drwpow/openapi-typescript/blob/43b1b69bc26ee333d0fed1f4aa702496c485dc01/packages/openapi-typescript/src/types.ts#L166-L187

PR #915, opened 2023-08-18
feat(core): option to disable creating index files

codebutler:
## Status

**READY**

## Description

Adds an option to disable creating `index.ts` files. 

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

Set option:

```ts
{
    output: {
        indexFiles: false
    }
}
```

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -51,7 +51,7 @@ export type NormalizedOutputOptions = {
   tsconfig?: Tsconfig;
   packageJson?: PackageJson;
   headers: boolean;
-  index: boolean;
+  indexFiles: boolean;
 };
 
 export type NormalizedOverrideOutput = {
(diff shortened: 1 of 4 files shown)


codebutler:
My motivation for this change: When using vite there's no tree-shaking during development, so having an index barrel file causes every model to be fetched before the app can load. In a big app this is thousands of requests.

anymaniax:
@codebutler seems really great I am just not sure about the name of the options do you have other idea then “index”? Personally nothing come to my mind

codebutler:
> @codebutler seems really great I am just not sure about the name of the options do you have other idea then “index”? Personally nothing come to my mind

It's also known as a [barrel file](https://basarat.gitbook.io/typescript/main-1/barrel).  Something like `createIndex`, `createIndexFile`, or `createBarrelFile`?

anymaniax (approved):
Thanks @codebutler !

PR #923, opened 2023-08-27
feat(core): include originalOperation in VerbOptions

codebutler:
## Status

**READY**

## Description

Provides a way for generators to access the original OpenAPI schema. 

Immediate use case is a need for the `security` prop.

## Todos

- [ ] Tests
- [ ] Documentation


packages/core/src/types.ts
```diff
@@ -525,6 +525,7 @@ export type GeneratorVerbOptions = {
   formUrlEncoded?: GeneratorMutator;
   override: NormalizedOverrideOutput;
   deprecated?: boolean;
+  originalSchema: OperationObject;
```
anymaniax: originalSchema already exist in the code. What do you think about originalOperation?

codebutler: sure sounds good, updated.

PR #925, opened 2023-08-31
Returns an AxiosResponse on return value

### LINKED ISSUE #924: About the return value of AxiosCustomInstance
orval is a very nice solution.
I have a question about the following issue and a request for improvement.
There is no `AxiosResponse` type, so status etc. are not returned.

1. see https://orval.dev/guides/custom-axios

## What happens?

```ts
export const customInstance = <T>(
   config: AxiosRequestConfig, options?
   options?: AxiosRequestConfig, options?
 ): Promise<T> => { // can't get status because it's not an Promise<AxiosResponse<T>>
   const source = Axios.CancelToken.source();
   const promise = AXIOS_INSTANCE({
     ... .config,.
     ... .options, .
     cancelToken: source.token, ...
   }).then(({ data }) => data);
 
   // @ts-ignore
   promise.cancel = () => {
     source.cancel('Query was cancelled');
   }; }
```

To be `Promise<AxiosResponse<T>>`.

motio:
## Status

READY

## Description

Fix #924.
The custom-axios guide returns `Promise<T>`,
But There is no AxiosResponse type, so status etc. are not returned.
Change to "Promise<AxiosResponse<T>" to get status code, etc. in the return value.

see https://orval.dev/guides/custom-axios

## Related PRs

No related PR.

## Steps to Test or Reproduce

see https://orval.dev/guides/custom-axios

### DIFF
--- a/docs/src/pages/guides/custom-axios.md
+++ b/docs/src/pages/guides/custom-axios.md
@@ -22,7 +22,7 @@ module.exports = {
 ```ts
 // custom-instance.ts
 
-import Axios, { AxiosRequestConfig } from 'axios';
+import Axios, { AxiosRequestConfig, AxiosResponse } from 'axios';
 
 export const AXIOS_INSTANCE = Axios.create({ baseURL: '<BACKEND URL>' }); // use your own URL here or environment variable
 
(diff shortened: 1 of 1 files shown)


melloware (approved):

docs/src/pages/guides/custom-axios.md
```diff
@@ -22,15 +22,15 @@ module.exports = {
...
+import Axios, { AxiosRequestConfig, AxiosResponse } from 'axios';
 
 export const AXIOS_INSTANCE = Axios.create({ baseURL: '<BACKEND URL>' }); // use your own URL here or environment variable
 
 // add a second `options` argument here if you want to pass extra options to each generated query
 export const customInstance = <T>(
   config: AxiosRequestConfig,
   options?: AxiosRequestConfig,
-): Promise<T> => {
+): Promise<AxiosResponse<T>> => {
```
anymaniax, line 33: it's not the case with the following code `then(({ data }) => data)`

anymaniax (changes requested):

PR #928, opened 2023-09-05
fix: filter null from enum values

alanpoulain:
## Status

**READY**

## Description

Following the [specification](https://swagger.io/docs/specification/data-models/enums/), a nullable enum needs to have `nullable: true` **and** `null` in its values.
When it's the case, Orval generates a bad TS enum. For instance:
```ts
export const NullEnum = {
  null: 'null',
  null: null,
} as const;
```
This PR fixes it.

## Todos

- [ ] Tests
- [ ] Documentation


anymaniax (approved):

anymaniax:
thanks @alanpoulain

PR #914, opened 2023-08-16
feat(namedParameters): Add global named parameters option

Marcel-G:
## Status

**WIP**

## Description

Often times routes have many path parameters. As the number grows, type-safety is less effective as arguments (commonly string or number types) can easily get used in the wrong order.
Using named parameters helps this, as you get the added safety of needing to specify the appropriate name for the parameter.

This PR adds a global `useNamedParameters` option which creates path parameters as a single object.

## Related PRs

List related PRs against other branches:

- https://github.com/anymaniax/orval/pull/887
- https://github.com/anymaniax/orval/pull/396

## Todos

- [x] Tests
- [x] Documentation
- [ ] Clients
  - [x] angular
  - [x] axios
  - [x] react-query
  - [x] svelte-query
  - [x] vue-query
  - [x] swr
  - [ ] zod

## Steps to Test or Reproduce

Outline the steps to test or reproduce the PR here.

```bash
> git pull --prune
> git checkout <branch>
> yarn install
> yarn prerelease
```

View the `named-parameters` builds under each client in `tests/generated`

PR #931, opened 2023-09-08
Fix Query parameters reactivity in Vue

### LINKED ISSUE #930: Vue: Query parameters are not reactive in Vue 3
Sorry for rushed issue, but the gist is that we have the following:
We pass a reactive (computed) object with some URL query parameters into the useBla()
And then what ends up happening is that we get the whole ref destructured in the URL, including all the internals of the ref object, like isref, value, etc.
In order to avoid that we need to `unref(params)`
I will create a patch and publish it here soon

Maxim-Mazurok:
## Status

**READY/WIP**

Fix #930 

## Description

Among other minor fixes this addresses #930 by unwrapping `params` mayberef

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

In the `vue-query` sample, open dev tools, refresh the page, see first request with filter=2 and then a few seconds later request for filter=3. This confirms that reactivity works. If we don't unwrap params - we'll get this error:

```
query.ts:444 TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'ReactiveEffect'
    |     property 'computed' -> object with constructor 'ComputedRefImpl'
    --- property 'effect' closes the circle
    at JSON.stringify (<anonymous>)
    at parseValue (buildURL.js:51:20)
    at Object.forEach (utils.js:245:10)
    at serialize (buildURL.js:47:13)
    at Object.forEach (utils.js:251:12)
    at buildURL (buildURL.js:36:11)
    at dispatchXhrRequest (xhr.js:44:47)
    at new Promise (<anonymous>)
    at xhrAdapter (xhr.js:15:10)
    at dispatchRequest (dispatchRequest.js:58:10)
```


packages/core/src/generators/options.ts
```diff
@@ -195,7 +196,11 @@ export const generateQueryParamsAxiosConfig = (
   let value = '';
 
   if (queryParams) {
-    value += ',\n        params';
+    if (isVue) {
+      value += ',\n        params: unref(params)';
```
Maxim-Mazurok, line 230: This is the core change

Maxim-Mazurok:
Was updating and cleaning up the branch, decided to add some tests as well, and couldn't make them work. Turned out that tests are fine, but the functionality actually never worked :D Kinda note to self, basically when filter changes - query doesn't refetch because of that. It was refetching because of 404 retry or something like that.

Maxim-Mazurok:
@anymaniax 
Here's an update on this PR:
- Resolve conflicts
    - [x] Merge master
    - [x] Fix regressions introduced in https://github.com/anymaniax/orval/pull/914, see https://github.com/anymaniax/orval/pull/931#issuecomment-1752355686
    - [ ] Double-check #914 for other regressions
    - [ ] Deal with `useNamedParameters: true`: I think there's no easy and good way to deal with it for Vue reactivity, also we don't use it, and it requires to pass version explicitly instead of using default, I would suggest to disable it for `vue-query` completely, probably throw error if found in config. If @Marcel-G or someone else wants to add it later - they will need to make sure it passes vue reactivity tests. I've added [vue-query-useNamedParameters](https://github.com/Maxim-Mazurok/orval/tree/930/samples/vue-query-useNamedParameters) sample that works for references.
- Add tests
  - [x] Testing query parameter reactivity
  - [ ] Testing path parameter reactivity (for the `routeString` thing)
  
  Please let me know what you think guys, cheers!

packages/query/src/index.ts
```diff
@@ -1126,10 +1129,12 @@ const generateQueryHook = async (
...
     );
 
-    const routeString = `\`${route}\``;
+    const routeString = isVue(outputClient)
+      ? getRouteAsArray(route) // Note: this is required for reactivity to work, we will lose it if route params are converted into string, only as array they will be tracked // TODO: add tests for this
+      : `\`${route}\``;
 
+    // Note: do not unref() params in Vue - this will make key lose reactivity
     const queryKeyFn = `export const ${queryKeyFnName} = (${queryKeyProps}) => {
-    ${isVue(outputClient) ? vueUnRefParams(props) : ''}
```
Marcel-G, line 1214: @Maxim-Mazurok, thanks for the detailed description of the issue here. Sorry for the regression added to main. 
The `unref` in query key was a consequence of accidentally reverting to a string based route key. 

Preserving the refs does pose a challenge for the useNamedParameters feature. I'm onboard with your proposal to disable this feature for vue since there was never explicitly a request for it. We added it just to have uniform support across all clients. Do you think it's generally not useful in the case of vue or just a challenge to implement right now?


Would it potentially work to accept `MaybeRefDeep` for pathParams?
We would also need to build in the defaults to each unref statement

```ts
export const getShowPetByIdQueryKey = (
  pathParams: MaybeRefDeep<ShowPetByIdPathParameters>,
  params?: MaybeRef<ListPetsParams>
) => {
  const { version = 1, petId } = unref(pathParams);
  return ['v', version, 'pets', petId, ...(params ? [params] : [])] as const;
};
```

```ts
const petId = computed(() => props.petId);
const petQuery = useShowPetById({ petId });
```

Would it be more idiomatic to have a `computed` statement for each path parameter or one for all the parameters?

Maxim-Mazurok: Got it, all good :) We should've added tests in the first place :)

Thanks for being on board with disabling named params for Vue for now. I think it might be equally useful for Vue as for any other framework, it's just a technical challenge I believe.

`MaybeRefDeep` seems to work type-wise, however, it didn't pass the test I've added to verify path params reactivity. You can check it out in this commit: [`a36a767` (#931)](https://github.com/anymaniax/orval/pull/931/commits/a36a767e4300399d4cad65081ffbc7f14b9d8c9f)

> Would it be more idiomatic to have a `computed` statement for each path parameter or one for all the parameters?

I'm not sure... I'd expect both to work. It's pretty annoying to deal with all that reactivity stuff, you literally have to check all the options...

---

I will disable named params for vue-query client for now in this PR

anymaniax:
@Maxim-Mazurok I am a bit lost now. Can we merge de pull request or you need to still do the two points above

Maxim-Mazurok:
@anymaniax yes, I still need to disable named params for vue-query, sorry for the delay, got a bit distracted by other work, I'll mark it as WIP for now

anymaniax:
@Maxim-Mazurok are you on discord and if yes can you ping me pls?

Maxim-Mazurok:
@anymaniax sure, just let me know how to find you there, I'm currently on a holiday for a week, will reach out when I get back! My email is maxim@mazurok.com if that helps

anymaniax:
@Maxim-Mazurok no emergency take care. Holidays are really important! If you join the orval discord that you can find on the doc. You will find me with the name of anymaniax there.

Maxim-Mazurok:
An update on this PR:
- [x] Resolved conflicts with master (also common guys, why do I see outdated snapshots and unformatted code every time, surely we can handle this in CI)
- [x] Dealt with `useNamedParameters` for `vue-query`, not it throws error if you try to use it with either `vue-query` builder type or `vue-query` output client, added [tests](https://github.com/Maxim-Mazurok/orval/blob/930/packages/query/src/index.test.ts) for that
- [x] Disabled `namedParameters` config test for `vue-query` - it doesn't even compile, added [comment](https://github.com/Maxim-Mazurok/orval/blob/4bcfe7f8e36eec612e4a63096aaa2dbf6d93f3a8/tests/configs/vue-query.config.ts#L50)

What's left to do in this PR:
(as per https://github.com/anymaniax/orval/pull/931#issuecomment-1752549954)
- [ ] Double-check #914 for other regressions
- [ ] Testing path parameter reactivity (for the routeString thing)

Most likely will be done tomorrow

Considering that it's a pretty big PR (I'm sorry, but I just had to fix a bunch of loosely related stuff) - would be great to start reviewing it now. Thanks @anymaniax and @melloware 
I will also ask my colleague @BenGane to have a look

melloware (dismissed):
LGTM.  Can you tell me all what ticket #'s this fixes so I can link those tickets to this PR?

Maxim-Mazurok:
> LGTM.  Can you tell me all what ticket #'s this fixes so I can link those tickets to this PR?

Thanks for having a look!
This was originally started for #930 and then other stuff got added that didn't have tickets I believe. There are at least 3 more issues that have tickets that I will address separately and probably already have PRs, not sure.

melloware:
Thanks I linked the ticket!  also looks like you have a conflict probably because of my latest fixes.

packages/query/src/index.test.ts
```diff
@@ -0,0 +1,36 @@
+import { it, expect, describe } from 'vitest';
+import { builder } from './index';
+import {
+  GeneratorOptions,
+  GeneratorVerbOptions,
+  NormalizedOverrideOutput,
+} from '@orval/core';
+
+describe('throws when trying to use named parameters with vue-query client', () => {
```
Marcel-G, line 9: Considering we're not permitting the use of `useNamedParameters` with `vue-query` do we still get value from adding `samples/vue-query-useNamedParameters` in this PR?

Maxim-Mazurok: Yeah good point, I removed the sample to avoid confusion in users. And added some comments for future developments: https://github.com/Maxim-Mazurok/orval/blob/13ec0680778217fd54608903e34d17be6b17fdee/scripts/update-samples.mjs#L17-L31

Maxim-Mazurok:
Hey guys, so since you all had a look already at this PR, let's merge it as is.

Two remaining tasks are optional and can be done later.
I've created these for myself:
- https://github.com/anymaniax/orval/issues/1078
- https://github.com/anymaniax/orval/issues/1079

I addressed feedback, resolved conflicts with master and hopefully tests will pass, will merge it when get approval review, thanks!

melloware (approved):
Now that 6.21.0 is out I am good with merging this

melloware:
Looks like there is some merge conflicts?

melloware (approved):

PR #945, opened 2023-09-21
feat(msw): allow mock delay override via function

andersonvom:
## Status

**READY**

## Description

This adds the possibility to override MSW mocks with a dynamic delay using a function that returns a number, making API calls more similar to real-world ones, where responses can potentially come back at different times:

```js
{
  // ... other options ...
  override: {
    mock: {
      delay: () => { ... }
    }
  }
}
```

This change is backwards compatible and the previous default value has been maintained.

## Todos

- [ ] Tests
- [X] Documentation

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -745,9 +745,9 @@ module.exports = {
 
 ##### delay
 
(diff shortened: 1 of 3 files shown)


anymaniax (approved):
Seems good thanks!

PR #962, opened 2023-10-09
7x performance improvement

EwenQuim:
Hi ! Two weeks ago I made a PR improving the `pascal` function, because it was a bottleneck. I continued investigating and I found that it was often called with the same values. So this PR adds a very simple cache to this function, with a tremendous impact on performance.

## Obvious duplicated calls to the pascal function
image

## Hidden duplicated calls
Some calls are made from different places of the codebase with the same input.

## Screenshots of the result
Before
image

After
image

99%+ hit ratio, this is huge ! 

## Going further

The default `petstore.yml` really is not a good OpenAPI doc to test on, because we can see that it does not have the size of a real project. Maybe a good idea for the future might be to test on a bigger document ?


anymaniax (approved):
Thanks a lot!

PR #963, opened 2023-10-09
feat(msw): support using OpenAPI example/examples fields as MSW values

### LINKED ISSUE #922: Support using OpenAPI example field as MSW value
It'd be nice if we could get the ability to prefer using an OpenAPI field's `example` as the value MSW uses to make deterministic testing a little smoother.

GMierzwa:
## Status

**READY**

## Description

Resolves #922.

Adds support for OpenAPI's example and examples fields to be used as MSW values. Handles example refs, as well as every variation of the example/examples fields described in [the docs](https://swagger.io/docs/specification/adding-examples/).

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -807,6 +807,26 @@ module.exports = {
 };
 ```
 
+##### useExamples
+
+Type: `Boolean`.
+
+Give you the possibility to use the `example`/`examples` fields from your OpenAPI specification as mock values.
+
+```js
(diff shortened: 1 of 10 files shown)


GMierzwa:
Rebased to fix a conflict.

melloware (approved):
This looks great.

xaviertholot:
Hello, from my tests and what I understand from this PR, it seems like we only mock the first example of a list. I think it would be an amazing feature to mock all examples in order to avoid having to create custom mocks when performing tests. What do you think ?

PR #968, opened 2023-10-16
feat(orval): allow custom baseUrl for each OpenAPI specification

### LINKED ISSUE #865: Customize baseUrl based on input
If I have multiple inputs, it would be great if there's a way to configure the baseUrl for each one:

```
module.exports = {
   petstore: {
     output: {
       target: 'src/petstore.ts',
       schemas: 'src/model',
       client: 'react-query',
       baseUrl: "/api/petstore"
     },
     input: {
       target: './petstore.yaml',
     },
   },
 };
```

GMierzwa:
## Status

**READY**

## Description

Resolves #865.

Allows `baseUrl` configuration directly via Orval. Previous configuration methods still work, so no breaking changes.

`baseUrl` can be defined under `output` for each OpenAPI specification, allowing urls to be different without the need for multiple custom client instances.

### DIFF
--- a/docs/src/pages/guides/set-base-url.md
+++ b/docs/src/pages/guides/set-base-url.md
@@ -1,16 +1,30 @@
 ## Use your own base url
 
-orval doesn't set any base url by default but you have multiple possibility to add it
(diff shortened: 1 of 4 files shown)


anymaniax (approved):
Seems good thanks!

Issue #975, opened 2023-10-20
TanStack react query v5 marks `queryKey` as required


leniency:
This makes adding additional query options annoying as it forces overriding the query key that gets set in the generated code.

This only appears to affect queries and not mutations.

melloware:
This was released in 6.19.1 can you test @leniency please?

leniency:
Yup all good now, thanks!

ruiaraujo012:
Still happens, any fix?

melloware:
Not many changes since 6.22.1 so look through the commits and see if you can spot it: https://github.com/anymaniax/orval/compare/v6.22.1...master

ruiaraujo012:
The genereted code is different
```ts
(6.30.0) query?: UseQueryOptions<Awaited<ReturnType<typeof getDomainValues>>, TError, TData>;
(6.22.1) query?: Partial<UseQueryOptions<Awaited<ReturnType<typeof getDomainValues>>, TError, TData>>;
```

melloware:
@anymaniax I don't see any commit between 6.22.1 and 6.23.0 that would affect this `Partial` not being generated?  @Maxim-Mazurok @codebutler @soartec-lab

soartec-lab:
@melloware 

So you want to know the reason why `Partial` is no longer generated?
if me question is correct, I'll search into it. and, I will also make corrections if possible.

I also read [this](https://github.com/anymaniax/orval/issues/899) as well.

melloware:
Yes I don't see what changed that would cause this between 6.22.1 and 6.23.0.

soartec-lab:
@maapteh @ruiaraujo012 

Hi, I checked the operation using `PetStore`, and the source code was generated with `Partial` existing as intended.

```
export const getListPetsQueryOptions = <TData = Awaited<ReturnType<typeof listPets>>, TError = AxiosError<Error>>(params?: ListPetsParams, options?: { query?:Partial<UseQueryOptions<Awaited<ReturnType<typeof listPets>>, TError, TData>>, axios?: AxiosRequestConfig}
) => {
```

using `orval` version is bellow:

```
$ npm list --depth=0 | grep orval
+-- orval@6.23.0
```

I also checked the source code, but could not find any part that would reproduce the problem.
If possible, could you please share the minimal OpenAPI definition file?

soartec-lab:
@maapteh @ruiaraujo012 @melloware 

[...]

PR #976, opened 2023-10-20
Make react query's `queryKey` optional.

leniency:
#975

This changes the query options to mark `queryKey` as optional in function arguments. This allows the `use___` hooks to add query options but not force overriding the query key as well. The query key is set internally in the generated code if not present so should be safe to make optional

Example - this no longer yields a typescript error about the `queryKey` option:
```tsx
const { data: pets, refetch } = useListPets(undefined, 1, {
    query: {
      staleTime: 5000
    }
  });
```

### DIFF
--- a/packages/query/src/index.ts
+++ b/packages/query/src/index.ts
@@ -538,6 +538,7 @@ const getQueryOptionsDefinition = ({
   hasQueryV5,
   queryParams,
   queryParam,
+  isReturnType
 }: {
   operationName: string;
   definitions: string;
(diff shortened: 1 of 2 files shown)


melloware:
Looks like tests are failing with this change?

packages/query/src/index.ts
```diff
@@ -558,14 +561,14 @@ const getQueryOptionsDefinition = ({
         : `typeof ${operationName}`
     }>>`;
 
-    return `${prefix}${pascal(type)}Options<${funcReturnType}, TError, TData${
+    return `${partialOptions ? 'WithOptional<':''}${prefix}${pascal(type)}Options<${funcReturnType}, TError, TData${
```
anymaniax, line 564: Why add something new and not just use Partial?

anymaniax:
I will close this one because I opened another one with Partial. I will go forward with that for now to be able to do a release asap. Let me know if it's an issue for you. Or if you see any other problems with that. Thanks a lot @leniency

packages/query/src/index.ts
leniency: My thought here was to just target the one troublesome property only and let any similar future changes get caught early rather than fail at query time. If react-query made a similar change in the future, typescript checking would error the same as `queryKey` did here, rather than allow a required property be passed undefined and cause errors down the pipeline.

PR #982, opened 2023-10-26
fix(react-query): do not partialize required query options

kpoelhekke:
## Status

READY

## Description

The fix in https://github.com/anymaniax/orval/pull/977 also omits required query options such as `initialPageParam` and `getNextPageParam` for inifinite queries. This PR makes sure only the `queryKey` is partialized.

### DIFF
--- a/packages/query/src/index.ts
+++ b/packages/query/src/index.ts
@@ -561,7 +561,7 @@ const getQueryOptionsDefinition = ({
         : `typeof ${operationName}`
     }>>`;
 
-    return `${partialOptions ? 'Partial<' : ''}${prefix}${pascal(
+    return `${partialOptions ? 'WithOptional<' : ''}${prefix}${pascal(
       type,
     )}Options<${funcReturnType}, TError, TData${
       hasQueryV5 &&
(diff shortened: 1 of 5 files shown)


anymaniax:
@kpoelhekke infinite it's still a bit broken right now because the queryOptions are not required currently

anymaniax:
Also there is probably a simpler way to do it no?

kpoelhekke:
You're right indeed, `query` should be required for infinite queries. I can fix that in the MR. 

What do you mean by simpler? Should we create a generic type globally, something like: `type PartializeQueryKey<T> = Omit<T, "queryKey"> & Partial<Pick<T, "queryKey">>`?

anymaniax:
maybe do like here #976

anymaniax:
and also not make query optional doesn't change the issue that options is optional and without transforming everything to object cannot be required

kpoelhekke:
🤦  that `options` problem is a bummer indeed. Guess that's beyond the scope of this PR. I would say this ia at last an improvement, but if it's worth merging is up to you.

Regarding the `WithOptional` generic type. Would it be a problem to add it like this? 

```tsx
// eslint-disable-next-line
type WithOptional<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>;
```

it will also be included into files that do not actually need it.

kpoelhekke:
In my opinion it's worth to merge, since currently with only `Partial` the types are lying for infinite queries. During my refactor to react-query 5 i solved all the types, just to find out that my app would crash because `getNextPageParam` was missing.

melloware:
@kpoelhekke where did we end up with this?  And it looks like the tests are failing?

kpoelhekke:
@melloware I'm not sure if it's worth investing more time, since it's not a perfect fix because of the optional options param?

I can improve it a bit more but please let me know first if this is the way forward.

melloware:
Ok if it's not perfect then let's close this PR.

PR #986, opened 2023-10-27
fix: Nullable Object type

### LINKED ISSUE #981: Nullable objects are missing ` | null`

Have an Open API 3.0 API spec with such key:
```yaml
          "example": {
            "type": "object",
            "description": "example of nullable object",
            "nullable": true
          },
```

## What happens?

final result has type with required key:
```
...
 example: SomeType;
...
```

```
...
 example: SomeType | null;
...
```

I managed to narrow down issue to this location:
https://github.com/anymaniax/orval/blob/master/packages/core/src/resolvers/object.ts#L7

**Package Version:** 6.19.1

karlismelderis-mckinsey:
## Status

Fix #981

**READY**

## Description

`| null` was missing for nullable objects without properties

```yaml
    NullableAnyObject:
      type: object
      nullable: true
```

got translated to
```ts
export type NullableAnyObject = { [key: string]: any }; // <<-- missing ` | null`
```

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

I added new test specification with described case - `null-type-v3-0.yaml`

### DIFF
--- a/packages/core/src/getters/object.ts
+++ b/packages/core/src/getters/object.ts
@@ -178,7 +178,7 @@ export const getObject = ({
 
   return {
     value:
-      item.type === 'object' ? '{ [key: string]: any }' : 'unknown' + nullable,
+        (item.type === 'object' ? '{ [key: string]: any }' : 'unknown') + nullable,
     imports: [],
     schemas: [],
     isEnum: false,
--- a/tests/configs/default.config.ts
+++ b/tests/configs/default.config.ts
@@ -57,6 +57,13 @@ export default defineConfig({
       target: '../generated/default/null-type/endpoints.ts',
     },
   },
+  'null-type-v-3-0': {
+    input: '../specifications/null-type-v3-0.yaml',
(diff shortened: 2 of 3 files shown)


melloware (approved):

melloware:
Does this fix both issues the one about Nullable Enums too?

karlismelderis-mckinsey:
> Does this fix both issues the one about Nullable Enums too?

Just one

I will take a look at Enums + strict node next week.

karlismelderis-mckinsey:
@melloware and @anymaniax any luck to release this fix? 🙏

we would fix couple bugs in our generated clients

davidpcaldwell:
> > Does this fix both issues the one about Nullable Enums too?
> 
> Just one
> 
> I will take a look at Enums + strict node next week.

We seem to be using version 7.5.0 per our `package.json` and this still seems to be an issue for `enum`s.

We're on OpenAPI 3.0.x.

PR #999, opened 2023-10-30
feat: support params serializer

### LINKED ISSUE #296: String array in query appends [] to the query variable name
1. Describe a GET request with query parameter that is an array of strings
2. Inspect Orval client will send the request with `name[]=value` instead of `name=value` in the query

## What happens?
Request will be with incorrect query param.

The query param name shouldn't be appended with []. 

**Operating System:** macOS Big Sur 11.6
**Package Version:** Always latest, running in CI pipeline
**Browser Version:** Safari 15.0

GMierzwa:
## Status

**READY**

## Description

Resolves #296 .

Adds support for axios' `paramsSerializer`. This can be used to adapt the format of arrays and other properties to the API's requirements.

Adds a `paramsSerializer` property, which behaves as other mutators, allowing for any custom params serializer. Also adds a `paramsSerializerOptions`, to use `qs` directly, where users only have to define the `options` that will be passed.

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -989,6 +989,71 @@ export const customFormUrlEncodedFn = <Body>(body: Body): URLSearchParams => {
 };
 ```
 
+#### paramsSerializer
+
+Type: `String` or `Object`.
+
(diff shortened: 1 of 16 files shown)


melloware (approved):
Excellent PR.

PR #1000, opened 2023-10-30
feat(faker): update deprecated methods

GMierzwa:
## Status

**READY**

## Description

Faker v8 has [deprecated a few methods](https://fakerjs.dev/guide/upgrading.html#deprecations-and-other-changes). Most of them were already updated in #978 , but arrays were still using `faker.datatype.number`. This PR fixes that, as well as deprecated uses found in the documentation and samples.


melloware (approved):
Thank you!

PR #1004, opened 2023-10-31
feat: disable minification

### LINKED ISSUE #964: Disable minification for output package
This is rather suggestion that would help to attract contributes. 
Currently, published package of orval contains minified JS code which is impossible to debug if you find an issue. If package was not minified it would be easier to investigate issues locally.

And other nodejs tools e.g.: gulp, webpack, or typescript publish non-minified packages.

GMierzwa:
## Status

**READY**

## Description

Resolves #964 

There's no need to minify the code, since Orval is only used as a dev dependency and never at runtime. I've also disabled `splitting` as it generates extra files that constantly change between versions, which can make debugging harder.

### DIFF
--- a/packages/angular/package.json
+++ b/packages/angular/package.json
@@ -8,7 +8,7 @@
     "dist"
   ],
   "scripts": {
-    "build": "tsup ./src/index.ts --target node12 --minify --clean --dts --splitting",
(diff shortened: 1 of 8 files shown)


melloware:
I like this idea but leaving this one up to @anymaniax

melloware (approved):

Marcel-G (approved):

PR #920, opened 2023-08-24
fix: use require_tld instead of regex

alanpoulain:
## Status

**READY**

## Description

In this [commit](https://github.com/anymaniax/orval/commit/11e5dc96901f7a17634f1abc7ae0bafe0407dd0b), `require_tld` usage was removed in favor of a regex. I'm not sure why, but it creates issue with URLs like this one: `http://my-docker-service/docs.json`, which are not recognized.

## Related PRs

List related PRs against other branches: N/A

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Changelog Entry (unreleased)

## Steps to Test or Reproduce

N/A

PR #1032, opened 2023-11-10
fix: improve isurl regex

melloware:
Improves on #920 and adds unit tests

Ensures that new scenario and old scenarios are still all working so we don't keep breaking this.

according to Snyk this is the correct way to detect a URL: https://snyk.io/blog/secure-javascript-url-validation/

### DIFF
--- a/packages/core/src/utils/assertion.ts
+++ b/packages/core/src/utils/assertion.ts
@@ -1,5 +1,4 @@
 import { ReferenceObject, SchemaObject } from 'openapi3-ts';
-import validatorIsUrl from 'validator/lib/isURL';
 import { SchemaType, Verbs } from '../types';
 import { extname } from './path';
 
@@ -81,5 +80,11 @@ export const isRootKey = (specKey: string, target: string) => {
 };
 
 export const isUrl = (str: string) => {
-  return validatorIsUrl(str, { require_tld: false });
+  let givenURL;
+  try {
+    givenURL = new URL(str);
+  } catch (error) {
+    return false;
+  }
+  return givenURL.protocol === 'http:' || givenURL.protocol === 'https:';
 };
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -14,17 +14,16 @@
     "test": "vitest --global test.ts"
   },
   "devDependencies": {
-    "@types/debug": "^4.1.10",
-    "@types/fs-extra": "^11.0.3",
-    "@types/inquirer": "^9.0.6",
-    "@types/lodash.get": "^4.4.8",
-    "@types/lodash.isempty": "^4.4.8",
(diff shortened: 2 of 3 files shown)


packages/core/src/utils/assertion.ts
```diff
@@ -80,6 +80,8 @@ export const isRootKey = (specKey: string, target: string) => {
   return specKey === target;
 };
 
+const LOCALHOST_REGEX = /^(https?|ftp):\/\/[^\s/$.?#].[^\s]*$/i;
+
 export const isUrl = (str: string) => {
-  return validatorIsUrl(str, { require_tld: false });
+  return validatorIsUrl(str) || LOCALHOST_REGEX.test(str);
```
alanpoulain: Too bad to not use only the validator. Which case is not handled by it?

melloware: So with your `require_tld` fix it broke the files so this test fails `expect(isUrl('D:/a/test.txt')).toBeFalsy();` it says that is a URL.

melloware: My new fix catches your scenario as well as the legacy localhost scenarios and all tests are passing on Windows.

alanpoulain: That's weird for sure. It should be investigated in the validator lib. I'll try to do it soon.

melloware: Yeah i went and looked and there a tons of bugs in validator about isUrl and the real issue was that your `http://my-docker-service/docs.json` was not being reported as a URL when it clearly is a URL is because of a bug about dashes in `url`.  If you add `.com` to your URL it works but to me that validator lib has lots of bugs.

here is  Regex tester you can use for my Regex: https://regex101.com/r/aB2hcG/1

melloware:
Can someone review this. The build is currently broken until this fix is implemented.

packages/core/src/utils/assertion.ts
alanpoulain: Why using the validator if the regex handles all the cases?

melloware: That is a great question I am more than happy to get rid of the validator? But that is a question for @anymaniax ?

melloware: The package `validator` is only brought it for this single check so it would allow us to remove another dependency.

melloware: Updated my solution using Snyk's suggestion: https://snyk.io/blog/secure-javascript-url-validation/

Will-Mann-16 (approved):

alanpoulain (approved):
Much better!

anymaniax (approved):

PR #1043, opened 2023-11-13
fix: uppercase method verbs

### LINKED ISSUE #888: HTTP PATCH case sensitivity

1. Have an openAPI spec that has a PATCH object
2. Don't use axios (because axios auto-solves this issue)
3. Generate a client

## What happens?

The client will be generated with a lowercase "patch" method. ie:

```js
{
      url: `...`,
      method: "patch",
      data: patchBody,
}
```

Since patch is the only standard HTTP method that is case sensitive, it should result in a method of `PATCH`.

N/A

## Any other comments?

For more details on this, see https://blog.k-nut.eu/patch-capitalization and the [HTTP standard normalization section](https://fetch.spec.whatwg.org/#concept-method-normalize)

**Operating System:** Mac OS Ventura 13.4
**Package Version:** 6.16.0
**Browser Version:** Chrome 114.0.5735.198

melloware:
## Status

**READY**

## Description

Fix #888

At the last possible second uppercase the verb so `method: 'PATCH'`

### DIFF
--- a/packages/axios/src/index.ts
+++ b/packages/axios/src/index.ts
@@ -181,7 +181,7 @@ const generateAxiosImplementation = (
   } ): Promise<TData> => {${bodyForm}
     return axios${
       !isSyntheticDefaultImportsAllowed ? '.default' : ''
-    }.${verb}(${options});
+    }.${verb.toLowerCase()}(${options});
   }
 `;
 };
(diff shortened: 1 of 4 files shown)


melloware:
For more details on this, see https://blog.k-nut.eu/patch-capitalization and the [HTTP standard normalization section](https://fetch.spec.whatwg.org/#concept-method-normalize)

PR #1044, opened 2023-11-13
fix: uppercase method verbs

### LINKED ISSUE #888: HTTP PATCH case sensitivity

melloware:
## Status

Fix #888 

Bare minimum change and least invasive to uppercase the Axios method like...

```js
method: 'GET'
```

### DIFF
--- a/packages/core/src/utils/assertion.ts
+++ b/packages/core/src/utils/assertion.ts
@@ -1,5 +1,4 @@
 import { ReferenceObject, SchemaObject } from 'openapi3-ts';
-import validatorIsUrl from 'validator/lib/isURL';
 import { SchemaType, Verbs } from '../types';
 import { extname } from './path';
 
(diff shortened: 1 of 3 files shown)


packages/core/src/utils/assertion.ts
```diff
@@ -81,5 +80,11 @@ export const isRootKey = (specKey: string, target: string) => {
...
 
 export const isUrl = (str: string) => {
-  return validatorIsUrl(str, { require_tld: false });
+  let givenURL;
+  try {
+    givenURL = new URL(str);
+  } catch (error) {
+    return false;
+  }
+  return givenURL.protocol === 'http:' || givenURL.protocol === 'https:';
```
melloware, line 89: This is in another PR but without this right now the tests won't pass.

anymaniax (approved):

PR #1045, opened 2023-11-14
Add option to enable Zod type coercion for query params

### LINKED ISSUE #961: Zod: Query Params Validator is unusable with URLSearchParams for booleans/numbers
The issue is that query string values are only ever represented as `string`s within `URLSearchParams`, so they can never be parsed effectively by a `zod` schema like `z.boolean()` or `z.number()` (they _can_ be parsed by `z.coerce.boolean()` or `z.coerce.number()`).

1. Generate a some zod schema based on a swagger file with an endpoint that has `parameters` that include a boolean field named `active`
2. Construct a `URLSearchParams` from a query string like `?active=true`, `?active=1`, `?active`
3. Try to parse this with the generated zod schema `generatedZodSchemaFn.parse(Object.fromEntries(new URLSearchParams('?active=true')))`

## What happens?

The schema will throw an error, because `active` will be `"true"`, `"1"`, or `""` (respective to the examples listed above), which are strings and not booleans.

If I were to write this zod schema myself, expecting to use it to parse query params, I would use `z.coerce.boolean()` (for the simple case) or `z.transform(..).boolean()` if I needed more flexibility in interpreting a parameter as a boolean value.

Just the typical zod parse error for "expected boolean".

## Any other comments?

Casting data is difficult to get right in the general case. I was hoping there might be a way for me to customize this via the config but I didn't find anything. It _could_ be sensible to use the `z.coerce` variants specifically for the `XQueryParams` schemas.

**Operating System:** Mac/OSX 12.4
**Package Version:** `orval` 6.18.1, `zod` 3.22.4
**Browser Version:** Chrome Version 117.0.5938.149 (Official Build) (arm64)

solomonhawk:
## Status

**READY**

## Description

Fix #961 I expressed a desire to make the generated zod schema for endpoint query parameters more useful for parsing and validating `URLSearchParams` by making use of zod's type-coercion. It turns out that part of this problem isn't easily solvable within this library's generated code (more on that below), but the part that is ended up being relatively simple to add.

This PR is one possible (relatively naive) approach to how that might work.

#### **Key points**:
- adds a new `override.coerceTypes` boolean flag to the output config
- when enabled, the zod codegen will emit e.g. `zod.coerce.string()` instead of `zod.string()`
- this only applies only to coerce-able types (string, number, boolean, bigint, date)
- this _will_ apply to nested schemas, and I'm not sure that's _always_ the correct thing to do (e.g. `zod.array(zod.coerce.number())`) - I expect this is useful in some cases

#### **Additional considerations**:
- zod schemas can be arbitrarily complex, and assuming users want simply `zod.coerce.<type>()` is a naive/opinionated approach
- there are merits to taking different approaches, e.g.:
  - `zod.string().pipe(zod.coerce.date())` or 
  - `zod.transform(v => new Date(v)).date()` (full example not shown here, but you get the idea - do something with input data before sending it to a parser/schema)
- another approach we could take would be exposing to users something like mutators for individual zod schemas
  - I didn't opt to attempt to solve for that, which is going to be a lot more complex than this simple and hopefully useful approach

---

#### **Deeper dive on zod + URLSearchParams**
It has been discussed before and there are some lingering challenges that aren't really in the scope of the zod package to solve.
Ref: https://github.com/colinhacks/zod/discussions/1763

**Assumptions:**
- you are using the zod query params schema to both cast AND validate `URLSearchParams`

**Problems:**
- you still need to collect the entries from `URLSearchParams`, since it can contain multiple values for the same key
- you can't just do `Array.from(urlSearchParams.entries())` because it will return multiple entries with the same key, whereas the zod schema will expect those to be grouped as an array
- you instead need to collect the values from the `URLSearchParams` first, only converting them to an array if there are multiple values for the same key
- a field may be an array field despite having 0 or 1 values, so you can't only rely on the presence of multiple same keys to determine whether it should be an array (you _can_ write the zod schema in a way that handles this more intelligently using `zod.transform()` before passing that data into `zod.array()` or another schema)
- you could introspect the schema to see if the field is expected to be an array, but that code can be a little hairy
  - the approach [zodix](https://github.com/rileytomasek/zodix/blob/master/src/parsers.ts#L218C73-L231) takes is to assume it's an array if the key appears more than once
  - [remix-params-helper](https://github.com/kiliman/remix-params-helper/blob/main/src/helper.ts#L217) uses schema introspection

---

**Closing thoughts**:

I found it a bit challenging to figure out the best way to add test coverage for these changes. There are some reasonably well-defined boundaries among the source files here but still it felt heavy-handed to test at the `generateZod` level. Spinning up the necessary input data wasn't super straightforward.

Right now the tests cover the main codegen emission which is conditional on `override.coerceTypes` but does not test the code that pipes the generator options through to the internal function (`parseZodValidationSchemaDefinition`) that uses it.

I know there has been some discussion in other places that touches on some of the overall design decisions within the library such as using string building vs ASTs for the codegen.

I simply wanted to share my perspective as a new contributor wishing it were easier to contribute test coverage to new features.

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

1. set up an openapi document that uses query parameters with types other than string (e.g. integer, boolean)
2. configure orval with:

```ts

export default defineConfig({
  validators: {
    output: {
      target: '<target>', // fill in
      client: 'zod',
      override: {
        coerceTypes: true,
      },
    },
    input: {
      target: '<input specfile>', // fill in
    },
  },
});
```

3. run orval
4. the emitted zod schemas should have e.g. `zod.coerce.number()` rather than `zod.number()`

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -691,7 +691,7 @@ module.exports = {
         mock: {
           properties: {
             '/tag|name/': 'jon', // Matches every property named 'tag' or 'name', including nested ones
-            '/.*\.user\.id/': faker.string.uuid(), // Matches every property named 'id', inside an object named 'user', including nested ones
+            '/.*.user.id/': faker.string.uuid(), // Matches every property named 'id', inside an object named 'user', including nested ones
             email: () => faker.internet.email(), // Matches only the property 'email'
             'user.id': () => faker.string.uuid(), // Matches only the full path 'user.id'
           },
@@ -1153,6 +1153,30 @@ module.exports = {
     output: {
       override: {
         useBigInt: true,
+      },
+    },
+  },
+};
+```
+
+#### coerceTypes
+
+Type: `Boolean`
+
+Valid values: true or false. Defaults to false.
+
+Use this property to enable [type coercion](https://zod.dev/?id=coercion-for-primitives) for [Zod](https://zod.dev/) schemas (only applies to query parameters schemas).
+
+This is helpful if you want to use the zod schema to coerce (likely string-serialized) query parameters into the correct type before validation.
+
+Example:
+
+```js
+module.exports = {
+  petstore: {
+    output: {
+      override: {
+        coerceTypes: true,
       },
     },
   },
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -107,6 +107,7 @@ export type NormalizedOverrideOutput = {
   ) => string;
   requestOptions: Record<string, any> | boolean;
   useDates?: boolean;
+  coerceTypes?: boolean;
   useTypeOverInterfaces?: boolean;
   useDeprecatedOperations?: boolean;
   useBigInt?: boolean;
--- a/packages/zod/package.json
+++ b/packages/zod/package.json
@@ -10,7 +10,8 @@
   "scripts": {
     "build": "tsup ./src/index.ts --target node12 --minify --clean --dts --splitting",
     "dev": "tsup ./src/index.ts --target node12 --clean --watch src",
-    "lint": "eslint src/**/*.ts"
+    "lint": "eslint src/**/*.ts",
+    "test": "vitest --global test.ts"
   },
   "dependencies": {
     "@orval/core": "6.19.1",
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -49,6 +49,9 @@ const resolveZodType = (schemaTypeValue: SchemaObject['type']) => {
   }
 };
 
+// https://github.com/colinhacks/zod#coercion-for-primitives
+const COERCEABLE_TYPES = ['string', 'number', 'boolean', 'bigint', 'date'];
+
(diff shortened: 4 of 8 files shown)


melloware:
@solomonhawk this is great!  Can you check a couple of other open issues and see if your PR addresses them?

Particularly these 2:

https://github.com/anymaniax/orval/issues/971

https://github.com/anymaniax/orval/issues/889

solomonhawk:
I only took a quick glance, but I don't have any reason to believe that these changes would have any impact on those 2 issues.

melloware:
well the date one you mentioned above about `zod.date()` and that ticket is saying it currently generates a REGEXP so I thought this PR was related?

melloware (approved):

melloware:
@solomonhawk mind submitting another PR that moves the property to `zod` specific node?.

```
override: {
	useDates: false,
	mutator: {
		path: "src/service/AxiosMutator.ts",
		name: "useAxiosMutator",
		},
	zod: {
		oerceTypes: true,
	},
}
````

Issue #942, opened 2023-09-17
MSW: set the language of faker.js


shooooori0128:
I want to be able to set languages in the faker.js imported by the msw resolver.
Are there any problems?

shooooori0128:
Currently, the file is rewritten directly in Hooks as follows. What can be done about it?

- replaceFakerForMswResolver.ts

```ts
import fs from 'fs';
import path from 'path';

const replaceFakerImport = (filePath: string) => {
  const fileContent = fs.readFileSync(filePath, 'utf8');

  const replacedFileContent = fileContent.replace(
    `import { faker } from '@faker-js/faker';`,
    `import { fakerJA as faker } from '@faker-js/faker';`,
  );

  fs.writeFileSync(filePath, replacedFileContent);
};

const searchMswResolver = (dirPath: string) => {
  fs.readdir(dirPath, (err, files) => {
    if (err) {
      // eslint-disable-next-line no-console
      return console.error(`Unable to scan directory: ${err.message}`);
    }

    return files.forEach((file) => {
      const filePath = path.join(dirPath, file);

      if (fs.statSync(filePath).isDirectory()) {
        searchMswResolver(filePath);
      } else if (file.endsWith('.msw.ts')) {
        replaceFakerImport(filePath);
      }
    });
  });
};

export default searchMswResolver;
```

- orval.config.ts

```ts
import searchMswResolver from './scripts/replaceFakerForMswResolver';

const OPENAPI_DIR = 'src/types/openapi';

export default defineConfig({
  all: {
    output: {
      ...
    hooks: {
      afterAllFilesWrite: ['yarn prettier --write', () => searchMswResolver(OPENAPI_DIR)],
    },
  },
});
```

melloware:
So what are you proposing?

melloware:
@Will-Mann-16 another one for you

PR #1053, opened 2023-11-14
feat(mock): moved msw folder to 'mock' to account for future work wit…

Will-Mann-16:
…h other mocking solutions

## Status

**READY**

## Description

The goal behind this PR was to move `msw` to `mock` and categorise `msw` as a tool of the mocking solution. This opens up work where we can use cypress/playwright interceptions instead of msw for user preferences. Also added configurable and overridable options to the mock handler instead of the boolean flag, and passed those options around (a lot of property drilling between layers - I think a piece of tech debt work would be to move the MSW stuff into a class-based system where the properties and methods of the classes can be used without the unnecessary amount of data being passed around and to clean up the code base a little. This would also allow us to add more tests). 

Also as an extension worked on #942 as it was easy to setup with the new configuration options I had setup.

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -1303,3 +1303,25 @@ module.exports = {
   },
 };
 ```
+
+#### useNativeEnums
+
+Type: `Boolean`
+
+Valid values: true or false. Defaults to false.
+
+Use this property to generate native Typescript `enum` instead of `type` and `const` combo.
+
+Example:
+
+```js
+module.exports = {
+  petstore: {
+    output: {
+      override: {
+        useNativeEnums: true,
+      },
+    },
+  },
+};
+```
--- a/packages/core/src/generators/schema-definition.ts
+++ b/packages/core/src/generators/schema-definition.ts
@@ -59,6 +59,7 @@ export const generateSchemasDefinition = (
             resolvedValue.value,
             schemaName,
             resolvedValue.originalSchema?.['x-enumNames'],
+            context.override.useNativeEnums,
           );
         } else if (schemaName === resolvedValue.value && resolvedValue.isRef) {
           // Don't add type if schema has same name and the referred schema will be an interface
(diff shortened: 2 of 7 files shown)


melloware:
Looks like there are conflicts?

packages/mock/src/msw/index.ts
```diff
@@ -89,7 +83,10 @@ export const generateMSW = (
           ? `export const ${functionName} = () => (${value})\n\n`
           : '',
       handler: `http.${verb}('${route}', async () => {
-        await delay(${getDelay(override)});
+        await delay(${getDelay(
+          override,
+          !isFunction(mock) ? mock : undefined,
+        )});
         return new HttpResponse(${
```
aitor-segura: The format of the handler response must be different:
`HttpResponse` should be used as static method:
- [json response](https://mswjs.io/docs/api/http-response#httpresponsejsonbody-init)
- [text response](https://mswjs.io/docs/api/http-response#httpresponsetextbody-init)

An example of this would be something like this:
```
return isTextPlan ? HttpResponse.text(`${functionName}()`) : HttpResponse.json(`${functionName}()`);
```

In the msw version, using the HttpResponse class constructor, it returns a type incompatibility bug:
```
Argument of type '() => Promise<HttpResponse>' is not assignable to parameter of type 'ResponseResolver<HttpRequestResolverExtras<PathParams<string>>, DefaultBodyType, undefined>'.
  Type 'Promise<HttpResponse>' is not assignable to type 'AsyncResponseResolverReturnType<undefined>'.
```

Will-Mann-16: Yes you're entirely right - thanks for catching. Fix coming in now, changes the msw output to this:

```ts
export const getDefaultMock = () => [
  http.get("*/pets/:petId", async () => {
    await delay(1000);
    return HttpResponse.json(getShowPetByIdMock(), {
      status: 200,
    });
  }),
  http.get("*/pets", async () => {
    await delay(1000);
    return HttpResponse.text(getShowPetByIdBasicMock(), {
      status: 200,
    });
  }),
];
```

PR #1058, opened 2023-11-19
feat: add optional configuration parameter to generate native Typescript enums

### LINKED ISSUE #907: Generate enum instead of type + const

Why are type + const generated instead of enum?
```
    MyEnum:
      type: string
      enum:
        - testA
        - testB
        - testC
```

## What happens?
generating
```ts
export type MyEnum = typeof MyEnum[keyof typeof MyEnum];

// eslint-disable-next-line @typescript-eslint/no-redeclare
export const MyEnum = {
    testA: 'testA',
    testB: 'testB',
    testC: 'testC',
} as const;
```

generating
```ts
export enum MyEnum {
    testA = 'testA',
    testB = 'testB',
    testC = 'testC',
}
```

## Any other comments?
typescript supports enums https://www.typescriptlang.org/docs/handbook/enums.html
I do not understand why it is impossible to generate it?

why do I need this:
```ts
export type MyEnum = typeof MyEnum[keyof typeof MyEnum];
export const MyEnum = {
    testA: 'testA',
    testB: 'testB',
    testC: 'testC',
} as const;

const a = MyEnum.testA;
const arr = [MyEnum.testB, MyEnum.testC];
arr.includes(a); // TS2345: Argument of type '"testA"' is not assignable to parameter of type '"testB" | "testC"'. 
```
how to fix:
1.
```ts
...
const arr: MyEnum[] = [MyEnum.testB, MyEnum.testC];
arr.includes(a); // ok, but I don't want to put down the types manually
```
2. (best way with enums!)
```ts
export enum MyEnum {
    testA = 'testA',
    testB = 'testB',
    testC = 'testC',
}

const a = MyEnum.testA;
const arr = [MyEnum.testB, MyEnum.testC];
arr.includes(a); // ok! I don't need to put array types manually!
```

6.13.1
**Operating System:** macOs ventura 13.4.1
**typescript version:** 4.7.4
**Package Version:** …
**Browser Version:** …

dsthode:
## Status

**READY**

## Description
Fix #907

As requested in issue #907 this PR adds a new optional configuration parameter to generate native Typescript enums.

The new parameter `useNativeEnums` is set in the `output.override` configuration block and when set to `true` (defaults to `false`) Orval will generate a native Typescript `enum` instead of the `type` and `const` combo.

### DIFF
--- a/packages/core/src/getters/query-params.ts
+++ b/packages/core/src/getters/query-params.ts
@@ -8,7 +8,7 @@ import {
   GetterQueryParam,
 } from '../types';
 import { jsDoc, pascal, sanitize } from '../utils';
-import { getEnum } from './enum';
+import { getEnum, getNativeEnum } from './enum';
 import { getKey } from './keys';
 
 type QueryParamsType = {
@@ -68,12 +68,21 @@ const getQueryParamsTypes = (
 
     if (resolvedValue.isEnum && !resolvedValue.isRef) {
       const enumName = queryName;
+      let enumValue: string;
 
-      const enumValue = getEnum(
-        resolvedValue.value,
-        enumName,
-        resolvedValue.originalSchema?.['x-enumNames'],
-      );
+      if (context.override.useNativeEnums) {
+        enumValue = getNativeEnum(
+          resolvedValue.value,
+          enumName,
+          resolvedValue.originalSchema?.['x-enumNames'],
+        );
+      } else {
+        enumValue = getEnum(
+          resolvedValue.value,
+          enumName,
+          resolvedValue.originalSchema?.['x-enumNames'],
+        );
+      }
 
       return {
         definition: `${doc}${key}${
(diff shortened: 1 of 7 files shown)


packages/core/src/getters/query-params.ts
```diff
@@ -68,12 +68,21 @@ const getQueryParamsTypes = (
...
       const enumName = queryName;
-
-      const enumValue = getEnum(
-        resolvedValue.value,
-        enumName,
-        resolvedValue.originalSchema?.['x-enumNames'],
-      );
+      let enumValue: string;
+
+      if (context.override.useNativeEnums) {
```
melloware: This looks pretty good but I am wondering if the getEnum method can just be passed this config so that getEnum returns either current or the native enum?  It would make all these if else's cleaned up?

melloware (changes requested):

packages/core/src/getters/query-params.ts
dsthode: You are absolutely right, the ifs look terribly ugly. I've performed the changes requested.

melloware:
Thanks letting the build run now!

melloware (approved):

melloware:
@anymaniax handles releases I pinged him.

melloware:
@DiRover 6.21.0 is out if you want to try this now.

PR #1068, opened 2023-11-20
fix: support es6 modules

### LINKED ISSUE #886: Orval: Support ES6 Modules
orval is only supporting if the project type is commonJs.

My project package.json contains type module. Unable to generate api clients. However If I remove type then it is working. I can't remove type as it might affect other things

melloware:
## Status

**READY/WIP/HOLD**

## Description

Fix so projects using ES6 modules can use Orval.

Fix #886

fix: support es6 modules

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "packages/*"
   ],
   "private": true,
+  "type": "module",
   "scripts": {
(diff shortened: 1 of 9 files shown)


melloware:
I need some help understanding this one..

PR #1069, opened 2023-11-21
feat(mock): moved msw folder to 'mock' to account for future work wit…

### LINKED ISSUE #942: MSW: set the language of faker.js

Will-Mann-16:
…h other mocking solutions

## Status

**READY**

## Description

The goal behind this PR was to move `msw` to `mock` and categorise `msw` as a tool of the mocking solution. This opens up work where we can use cypress/playwright interceptions instead of msw for user preferences. Also added configurable and overridable options to the mock handler instead of the boolean flag, and passed those options around (a lot of property drilling between layers - I think a piece of tech debt work would be to move the MSW stuff into a class-based system where the properties and methods of the classes can be used without the unnecessary amount of data being passed around and to clean up the code base a little. This would also allow us to add more tests). 

Also as an extension worked on #942 as it was easy to setup with the new configuration options I had setup.

Version 2 of the PR, in attempt to sync with main to prevent merge conflicts I accidentally closed the original PR https://github.com/anymaniax/orval/pull/1053

Fix #942


melloware:
Looks like tests are failing?

Will-Mann-16:
Yes they are, so the reason I put my previous fix in way because the typescript wouldn't build using `HttpResponse.json` for some reason. I'll investigate again as that's the way we should do it - not the previous work around. However @anymaniax I can't replicate the build issues with the text/plain on the previous version - this has no TS issues that I can see.
CleanShot 2023-11-21 at 12 31 09

Will-Mann-16:
I'm reverting back to the previous implementation as the text/plain approach does indeed work as we have to manually override. This HTTPResponse class doesn't work with the unioned types/constants to combine to one type, which is why the pipelines were failing.

melloware (approved):

PR #1092, opened 2023-12-04
Dependabot switch to monthly

melloware:
switch to monthly dependabot checks from weekly

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,7 +8,7 @@ updates:
   - package-ecosystem: 'npm'
     directory: '/'
     schedule:
-      interval: 'weekly'
+      interval: 'monthly'
   - package-ecosystem: 'github-actions'
     directory: '/'
     schedule:


Maxim-Mazurok:
I guess it's fine to do

I would suggest setting up automatic management of updates. In my https://github.com/Maxim-Mazurok/google-api-typings-generator project, I have the following setup:
- weekly dependabot PRs: https://github.com/Maxim-Mazurok/google-api-typings-generator/blob/master/.github/dependabot.yml
- kodiak automatically approves/rebases/merges them: https://github.com/Maxim-Mazurok/google-api-typings-generator/blob/master/.kodiak.toml
- I also have these two actions, not sure if they needed: https://github.com/Maxim-Mazurok/google-api-typings-generator/blob/master/.github/workflows/rebase-pull-requests.yml and https://github.com/Maxim-Mazurok/google-api-typings-generator/blob/master/.github/workflows/add-automerge-label.yml
- I also have email filters in gmail to delete all notifications from dependabot on that repo

As a result I have automatic ever-green dependencies, and I can be pretty sure nothing is broken badly because in that case tests wouldn't pass. As long as tests are passing - I'm fine with updates.

This means I might miss some deprecations sometimes, or new recommended ways of doing stuff, but it's probably for the best tbh, as my goal is not to spend much time on the project maintenance.

Maxim-Mazurok (approved):

melloware:
@Maxim-Mazurok would be happy if you want to implement that here.

Maxim-Mazurok:
> @Maxim-Mazurok would be happy if you want to implement that here.

I don't have bandwidth at the moment, but happy to assist, should be relatively easy to do given my working config examples.

PR #1105, opened 2023-12-08
RFC: feat(query): add shouldExportMutatorHooks option

ezequiel:
## Status

**HOLD**

## Description

When client is set as react-query, 3 different functions are generated for each  query/mutation. For example:

```ts
export const useCreateUserHook = () => {
  ... 
};

export const useCreateUserMutationOptions = <TError = ErrorType<unknown>, TContext = unknown>(...): ... => {
  ...
};


export const useCreateUser = <TError = ErrorType<unknown>, TContext = unknown>(...) => {
  ....
};
```
**GOAL**: I don't want to export `useCreateUserHook`. This causes confusion for my team and I on which function to use: `useCreateUserHook`, or `useCreateUser`? Ideally we always use the react-query version (`useCreateUser`), but it's easier if we just don't export it at all, so no one can make the mistake

(I arguably don't want `useCreateUserMutationOptions` exported either, but that one is less problematic)

I made this an option to preserve backwards compatibility, but let me know your thoughts

### orval.config.js
```ts
module.exports = {
  api: {
    input: {
      target: '../docs/swagger.json',
      override: {
        transformer,
      },
    },
    validation: true,
    output: {
      target: 'api.generated.ts',
      client: 'react-query',
      workspace: 'src/api',
      override: {
        query: {
          useQuery: false,
          useSuspenseQuery: true,
          shouldExportMutatorHooks: false,
        },
      },
    },
  },
};
```
Verified the mutator hook is no longer exported

### DIFF
--- a/packages/orval/src/utils/options.ts
+++ b/packages/orval/src/utils/options.ts
@@ -201,6 +201,7 @@ export const normalizeOptions = async (
           useQuery: true,
           useMutation: true,
           signal: true,
+          shouldExportMutatorHooks: true,
           ...normalizeQueryOptions(outputOptions.override?.query, workspace),
         },
         swr: {
@@ -439,6 +440,9 @@ const normalizeQueryOptions = (
             queryOptions?.mutationOptions,
           ),
         }
+      : {}),
+    ...(!isUndefined(queryOptions.shouldExportMutatorHooks)
+      ? { shouldExportMutatorHooks: queryOptions.shouldExportMutatorHooks }
       : {}),
     ...(!isUndefined(queryOptions.signal)
       ? { signal: queryOptions.signal }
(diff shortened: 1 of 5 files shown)


ezequiel:
Do we think this is an actual problem, and worth adding an option for? Thinking a little more here, it'd be nice if there was a granular export configuration capability to control exactly how much gets `export`ed, though that seems a little difficult here given all the different clients

packages/orval/src/utils/options.ts
```diff
@@ -201,6 +201,7 @@ export const normalizeOptions = async (
           useQuery: true,
           useMutation: true,
           signal: true,
+          shouldExportMutatorHooks: true,
```
melloware, line 204: I wonder if for consistency it should be `useMutationHooks` ?

melloware (changes requested):

packages/orval/src/utils/options.ts
anymaniax: Not sure because it's still generated but just not exported anymore

anymaniax (approved):

melloware (approved):

PR #1116, opened 2023-12-19
Revert "fix(zod): allow blank value for non-required enums"

### LINKED ISSUE #1114: 6.23.0 adds blank values to required enums

1. Upgrade to orval 6.23.0
2. Regenerate zod schema for given openapi spec
3. Blank values get added to my enums

I would not expect the blank strings in my case as `deviceType` is part of the `required` array (see last line)

```
      "DeviceType": {
        "type": "string",
        "enum": ["RingLight", "LightStrip", "Unknown"],
        "required": true // <- this breaks swagger schema validation
      },
      "ElgatoDeviceDetailsResponseDto": {
        "type": "object",
        "properties": {
          "productName": {
            "type": "string"
          },
          "deviceType": {
            "$ref": "#/components/schemas/DeviceType"
          },
          "displayName": {
            "type": "string"
          }
        },
        "required": ["productName", "deviceType", "displayName"]
      },
```

Hey,

maybe I'm not getting something but given this snippet from my openapi spec seems not to behave as I would expect.

Without the `required` in L4 (which breaks schema validation) I get the blank values added to my enum even though `deviceType` is part of the `required` array in the last line.

## Any other comments?

The blank values are not being added when I (manually) add `"required": true` in L4 of the snippet. But adding it results in the schema validation to be broken. So it seems to me it's checking for the wrong "required" field?

**Operating System:** macos
**Package Version:** 6.23.0

codebutler:
Reverts anymaniax/orval#1113

Fixes #1114

I'll work out the proper fix for optional enums in a follow up PR.

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -214,13 +214,14 @@ const generateZodValidationSchemaDefinition = (
   }
 
   if (schema.enum && type !== 'number') {
-    const enumValues = schema.enum.map((value) =>
-      isString(value) ? `'${escape(value)}'` : `${value}`,
-    );
-    if (!schema.required) {
-      enumValues.push(`""`);
-    }
-    functions.push(['enum', [`[${enumValues.join(', ')}]`]]);
+    functions.push([
+      'enum',
+      [
+        `[${schema.enum
+          .map((value) => (isString(value) ? `'${escape(value)}'` : `${value}`))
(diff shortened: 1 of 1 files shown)


o-alexandrov:
cc @melloware 
thank you for the PR, just fyi, formatting issue in CI will be fixed when #1117 is merged

melloware (approved):

PR #1125, opened 2023-12-28
fix: adjust SecondParameter utility type

victorboucher:
## Status

**READY**

## Description

Fixes the poorly typed SecondParameter utility type.

### DIFF
--- a/packages/axios/src/index.ts
+++ b/packages/axios/src/index.ts
@@ -199,13 +199,7 @@ export const generateAxiosHeader: ClientHeaderBuilder = ({
 }) => `
 ${
   isRequestOptions && isMutator
-    ? `// eslint-disable-next-line
-  type SecondParameter<T extends (...args: any) => any> = T extends (
-  config: any,
-  args: infer P,
-) => any
-  ? P
-  : never;\n\n`
+    ? `type SecondParameter<T extends (...args: any) => any> = Parameters<T>[1];\n\n`
     : ''
 }
   ${!noFunction ? `export const ${title} = () => {\n` : ''}`;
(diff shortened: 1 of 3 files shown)


melloware:
Can you please open an Issue ticket explaining what is poorly typed and show us how this ticket makes it better?

melloware:
@soartec-lab since this affects SWR can you review this change?

melloware (approved):

soartec-lab:
@melloware 

Of cource 👍 I'll review this, so could you assign me reviewer for this PR ?

soartec-lab (commented):
@victorboucher @melloware 

This change will affect the case where `global mutator` is used, and I don't think it will affect the generated `swr` custom hook alone.
At the same time, we thought it would have the following advantages:

- Can be expressed in a simpler and more understandable form
- Being able to remove eslint disable

However, this change forces a fixed order of arguments in the `global mutator` function definition. This will result in unintentionally breaking changes for existing users, so it may be wise to avoid making this change unless it is a more important change.

soartec-lab (commented):
I removed the approve because I was concerned about the content being different from what I was getting the mention for. Let me confirm about `global mutator`.
this change forces a fixed order of arguments in the `global mutator` function definition.
This would result in an unintentionally breaking change for existing users, so it might be wise not to make this change unless it's a more significant change. What do you think?

melloware:
No problem just let me know what your final decision is.

soartec-lab:
@melloware 
Previously, `never` was returned if the second argument did not exist, but we investigated the impact of changing it to `undefined`.

In the first place, there will be no mismatch due to the case where the second argument does not exist, and even if it were to happen, there would be no problem because the type was specified as `never | undefined` in the function used. so, i agree with merging PR.

PR #1130, opened 2023-12-31
fix: avoid duplicate variable name in `swr` function

soartec-lab:
## Status

**READY**

## Description

fix https://github.com/anymaniax/orval/issues/800

When using `query` for an API parameter defined in `OpenAPI`, the name of the argument in the automatically generated function and the name of the argument within the function collide.
In that case, we changed the variable name `query` to `_query`.
Variables other than `query` are not considered because they are rarely used as parameters.

Similar problems occur with `client` other than `swr`, so we will deal with them in order once this PR is merged.

## Related PRs

none

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

1. Specify `query` in the parameter name as below:

```yaml
openapi: '3.0.0'
info:
  version: 1.0.0
  title: Swagger
  license:
    name: MIT
paths:
  /pets/{query}:
    get:
      summary: Info for a specific pet
      operationId: showPetById
      tags:
        - pets
      parameters:
        - name: query
          in: path
          required: true
          description: The id of the pet to retrieve
          schema:
            type: string
      responses:
        '200':
          description: Expected response to a valid request
          content:
            application/json:
              schema:
        default:
          description: unexpected error
          content:
            application/json:
              schema:
components:
  schemas:
    Pet:
      type: object
      required:
        - id
        - name
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
        tag:
          type: string
    Error:
      type: object
      required:
        - code
        - message
      properties:
        code:
          type: integer
          format: int32
        message:
          type: string
```

2. Specify `swr` for `client` in `orval.config.js` as bellow:

```javascript
module.exports = {
  'petstore-file': {
    input: {
      target: './petstore.yaml',
    },
    output: {
      mode: 'tags-split',
      client: 'swr',
      target: 'src/gen/endpoints',
      schemas: 'src/gen/model',
      indexFiles: false,
    },
  },
};
```

3. `orval` execute

```bash
orval
```

4. The function is output where the name of the variable is replaced from `query` to `_query`

```typescript
export const useShowPetById = <TError = AxiosError<Error>>(
  query: string,
  options?: {
    swr?: SWRConfiguration<Awaited<ReturnType<typeof showPetById>>, TError> & {
      swrKey?: Key;
      enabled?: boolean;
    };
    axios?: AxiosRequestConfig;
  },
) => {
  const { swr: swrOptions, axios: axiosOptions } = options ?? {};

  const isEnabled = swrOptions?.enabled !== false && !!query;
  const swrKey =
    swrOptions?.swrKey ?? (() => (isEnabled ? getShowPetByIdKey(query) : null));
  const swrFn = () => showPetById(query, axiosOptions);

  const _query = useSwr<Awaited<ReturnType<typeof swrFn>>, TError>(
    swrKey,
    swrFn,
    swrOptions,
  );

  return {
    swrKey,
    ..._query,
  };
};
```

### DIFF
--- a/packages/swr/src/index.ts
+++ b/packages/swr/src/index.ts
@@ -14,6 +14,7 @@ import {
   GeneratorOptions,
   GeneratorVerbOptions,
   GetterParams,
+  GetterProp,
   GetterProps,
   GetterPropType,
   GetterResponse,
@@ -235,6 +236,12 @@ const generateSwrImplementation = ({
   doc?: string;
 }) => {
   const swrProps = toObjectString(props, 'implementation');
+
+  const hasParamReservedWord = props.some(
+    (prop: GetterProp) => prop.name === 'query',
+  );
+  const queryResultVarName = hasParamReservedWord ? '_query' : 'query';
+
   const httpFunctionProps = swrProperties;
 
   const swrKeyImplementation = `const isEnabled = swrOptions?.enabled !== false${
@@ -291,7 +298,7 @@ ${doc}export const ${camel(
       : ''
   });
 
-  const query = useSwr<Awaited<ReturnType<typeof swrFn>>, TError>(swrKey, swrFn, ${
+  const ${queryResultVarName} = useSwr<Awaited<ReturnType<typeof swrFn>>, TError>(swrKey, swrFn, ${
     swrOptions.options
       ? `{
     ${stringify(swrOptions.options)?.slice(1, -1)}
@@ -302,7 +309,7 @@ ${doc}export const ${camel(
 
   return {
     swrKey,
-    ...query
+    ...${queryResultVarName}
   }
 }\n`;
 };


packages/swr/src/index.ts
```diff
@@ -235,6 +236,12 @@ const generateSwrImplementation = ({
   doc?: string;
 }) => {
   const swrProps = toObjectString(props, 'implementation');
+
+  const hasParamNameQuery = props.some(
```
melloware: I wonder if we should have a CONST array of reservedWords and right now it would only be `['query']` that you could check but that way it would make this more flexible in the future if we found another reserved word we want to underscore?

melloware (changes requested):

packages/swr/src/index.ts
soartec-lab: indeed. I'm not sure if the list will increase in the future, but I think it's a good idea in terms of emphasizing it. I'll deal with it.

soartec-lab: @melloware 
thank you for review. I thought about the comment for a while.

If new reservations are added in the future, you will need to validate `prop` for each variable name and give `_` to that variable. In the end, it is necessary to process in units of reserved words, and it may not be possible to be flexible. I think it's great to emphasize what happens after a reservation is made, but it actually increases the complexity at this point, so I thought it would be better to leave it as is for now.

What do you think after hearing my opinion?

melloware: OK then rename `hasParamNamedQuery` to `hasParamReservedWord` so people know why you are looking specifically for the word `query` they will know its a reserved word.

melloware (changes requested):

packages/swr/src/index.ts
soartec-lab: @melloware 

Please allow me to add to the explanation.

The purpose of using `hasParamNameQuery` in the variable name is not whether it is a reserved word or not, but simply to check whether `query` is used as a parameter.

Let's consider a case where another reserved word has increased.
For example, suppose you want to treat `swrFn` as a reserved word and add `_` to it.
In that case, I think the following will be added:

```typescript
const hasParamNameSwrFn = props.some(
  (param: GetterProp) => param.name === 'swrFn',
);
const queryResultVarName = hasParamNameSwrFn ? '_swrFn' : 'swrFn';
```

Because, variables used within functions with the same name are marked with `_` depending on the existence of individual specified parameters, not whether the parameters contain reserved words. However, please note that this is just an example and there are other options if you actually want to respond.

From this, I think that the original `hasParamNameQuery` represents the original intention of the variable name rather than `hasParamReservedWord` as commented.

And I thought that the reason was expressed in the code, so it was sufficient, but what do you think?

melloware: I disagree the word `query` is to generic in terms of React Query, Vue Query, SWR?   So this code tells me no reason wht you are looking for `hasParamNamedQuery` its too specific with no reasoning why you are doing is.

*"looking for a reserved word, preventing its use by appending an underscore so it does not collide with other usages"*  From looking at your code I can't tell any of that and no where does it mention `query` is a reserved word you are looking for.

So someone looking at this code in 1 year will have the spend the time to think "why are they doing this" where as naming things properly like "checkReservedWord" or "cleanReservedWords" tells the developer exactly what you are doing without thinking about it.

melloware (changes requested):
See comment

packages/swr/src/index.ts
melloware: At the very least put a comment above the line.

```js
// #800 Convert param named query to _query to prevent collision
```

soartec-lab: @melloware 
Got it. I respect your opinion and change the function name from "hasParamNamedQuery" to "hasParamReservedWord".

soartec-lab: Fixed as bellow:

https://github.com/anymaniax/orval/compare/5693ebe1387492d5c25e8bc0c512fd4e8db63148..146f9bb90da54fbb012f2948c65280a445ca70bf

melloware (approved):

PR #1146, opened 2024-01-06
fix: strictly check dependencies import judgment using regular expressions

soartec-lab:
## Status

**READY**

## Description

fix https://github.com/anymaniax/orval/issues/1145

>The current judgment logic uses a simple condition of whether the string exists.
Therefore, for example, if the variable useKey is used, Key will be imported, but this may be an unnecessary import.

I achieved this by using regular expressions to determine.

## Related PRs

none

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

1. prepare OpenAPI definition for pet store
2. execute `orval`

```
orval
```

3. We have confirmed that the existing generation process is not destroyed using the pet store.

### DIFF
--- a/packages/core/src/generators/imports.ts
+++ b/packages/core/src/generators/imports.ts
@@ -181,9 +181,12 @@ export const addDependency = ({
   hasSchemaDir: boolean;
   isAllowSyntheticDefaultImports: boolean;
 }) => {
-  const toAdds = exports.filter((e) =>
-    implementation.includes(e.alias || e.name),
-  );
+  const toAdds = exports.filter((e) => {
+    const searchWords = [e.alias, e.name].filter((p) => p?.length).join('|');
+    const pattern = new RegExp(`\\b(${searchWords})\\b`, 'g');
+
+    return implementation.match(pattern);
+  });
 
   if (!toAdds.length) {
     return undefined;


packages/core/src/generators/imports.ts
```diff
@@ -181,9 +181,12 @@ export const addDependency = ({
   hasSchemaDir: boolean;
   isAllowSyntheticDefaultImports: boolean;
 }) => {
-  const toAdds = exports.filter((e) =>
-    implementation.includes(e.alias || e.name),
-  );
+  const toAdds = exports.filter((e) => {
+    const petternValues = [e.alias, e.name].filter((p) => p?.length).join('|');
```
melloware: PatternValues not PetternValues

melloware (changes requested):

packages/core/src/generators/imports.ts
soartec-lab: @melloware 
Thank you for review.
I interpreted the name of the variable to be inappropriate. Do you think this is `searchWords` in the sense of the words you want to search for in a regular expression?

melloware: Yes searchWords would be a good name

soartec-lab: Thanks! I fixed it with bellow:

https://github.com/anymaniax/orval/compare/9f349c41c88d6e17162d81c087ff6dc59f2d6718..5d01571013e61fc45575874279201c3c9736746a

melloware (approved):

Issue #1156, opened 2024-01-15
Bug: Nordic characters are ignored


orhels:
## What are the steps to reproduce this issue?

1. Use Orval to generate typescript files with the following schema:
```yaml
openapi: '3.0.0'
info:
  version: 1.0.0
  title: Swagger ÆrligØnskeÅpen
  license:
    name: MIT
paths:
  /ærligØnskeÅpen:
    get:
      summary: List all ærlig-ønske-åpen
      operationId: listÆrligØnskeÅpen
      responses:
        '200':
          description: A paged array of ærlig-ønske-åpen
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ÆrligØnskeåpen'
components:
  schemas:
    ÆrligØnskeåpen:
      type: object
      required:
        - ærlig
        - ønske
        - åpen
      properties:
        ærlig:
          type: string
        ønske:
          type: string
        åpen:
          type: string
```

## What happens?

It generates the file `rligNskePen.ts`:
```typescript
export interface RligNskePen {
  ærlig: string;
  åpen: string;
  ønske: string;
}
```

## What were you expecting to happen?

It should generate the file `ærligØnskeåpen.ts`:
```typescript
export interface ÆrligØnskeåpen {
  ærlig: string;
  åpen: string;
  ønske: string;
}

```
## Any logs, error output, etc?

Nope.

## Any other comments?

I've been testing with different versions of Orval, and the issue starts with 6.18.0.

## What versions are you using?

**Operating System:** MacOs 13.6.3
**Package Version:** 6.23.0 
**Browser Version:** Chrome 119.0.6045.123

melloware:
My gut feeling is this is the ticket that broke it: https://github.com/anymaniax/orval/pull/909

orhels:
I have a PR comming where I belive I've fixed the issues. My testing found that #951 was probably the cause.

PR #1157, opened 2024-01-15
fix: handle nordic characters

orhels:
## Status

**READY**

## Description

Solves #1156 

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout fix-handle-nordic-characters
> yarn build
> yarn test
```

1. Use below schema to generate:
```yaml
openapi: '3.0.0'
info:
  version: 1.0.0
  title: Swagger ÆrligØnskeÅpen
  license:
    name: MIT
paths:
  /ærligØnskeÅpen:
    get:
      summary: List all ærlig-ønske-åpen
      operationId: listÆrligØnskeÅpen
      responses:
        '200':
          description: A paged array of ærlig-ønske-åpen
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ÆrligØnskeåpen'
components:
  schemas:
    ÆrligØnskeåpen:
      type: object
      required:
        - ærlig
        - ønske
        - åpen
      properties:
        ærlig:
          type: string
        ønske:
          type: string
        åpen:
          type: string
```
2. Check that the output respects nordic characters.

### DIFF
--- a/packages/core/src/utils/case.ts
+++ b/packages/core/src/utils/case.ts
@@ -102,7 +102,7 @@ export const pascal = (s: string) => {
     s = low.call(s);
   }
 
-  const pascalString = (s?.match(/[a-zA-Z0-9]+/g) || [])
+  const pascalString = (s?.match(/[a-zA-Z0-9\u00C0-\u017F]+/g) || [])
     .map((w) => w.charAt(0).toUpperCase() + w.slice(1))
     .join('');
 
--- a/packages/core/src/utils/case.test.ts
+++ b/packages/core/src/utils/case.test.ts
@@ -31,6 +31,18 @@ describe('pascal case testing', () => {
     expect(pascal('')).toBe('');
     expect(pascal(undefined)).toBe('');
   });
+
+  it('should handle nordic characters', () => {
+    // norwegian
+    expect(pascal('ærlig-ønske-åpen')).toBe('ÆrligØnskeÅpen');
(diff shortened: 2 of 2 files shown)


orhels:
Let me know if there are any issues or if more info is needed.

packages/core/src/utils/case.ts
```diff
@@ -102,7 +102,7 @@ export const pascal = (s: string) => {
     s = low.call(s);
   }
 
-  const pascalString = (s?.match(/[a-zA-Z0-9]+/g) || [])
+  const pascalString = (s?.match(/[a-zæøåäöA-ZÆØÅÄÖ0-9]+/g) || [])
```
melloware: Can you try this instead...

```
/[a-zA-Z0-9\u00C0-\u017F]+/g
```

melloware (changes requested):

packages/core/src/utils/case.ts
melloware: This should catch ANY character in most languages.

melloware:
This looks great running the tests now....

packages/core/src/utils/case.ts
orhels: Thanks, that worked. The test should probably be updated as well, but im unsure which characters to test with.

melloware: You could throw in a French word like...

`Égal à`

melloware (approved):

PR #1165, opened 2024-01-20
feat(msw): possible to overwrite value when creating an `object` with `msw` mock

soartec-lab:
## Status

**READY**

## Description

I responded to part of https://github.com/anymaniax/orval/issues/822

If there is only one response case and the type is `object`, the returned value can be overwritten externally by specifying `overrideResponse` as an argument.
In order to handle the case where it is an `array` of `object` such as `Pets`, after calculating the return value, we determine whether overwriting is possible based on whether `overrideResponse` is included in it.

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

1. Specify basic pet schema as input
2. execute `orval`

```
orval
```

3. generated mock function bellow: 

```typescript
export const getListPetsMock = (overrideResponse?: any) => (Array.from({
  length: faker.number.int({ min: 1, max: 10 }) }, (_, i) => i + 1).map(() => ({
    id: faker.number.int({min: undefined, max: undefined}),
    name: faker.word.sample(),
    tag: faker.helpers.arrayElement([faker.word.sample(), undefined]),
    ...overrideResponse
  }))
)

export const getShowPetByIdMock = (overrideResponse?: any) => ({
  id: faker.number.int({min: undefined, max: undefined}),
  name: faker.word.sample(),
  tag: faker.helpers.arrayElement([faker.word.sample(), undefined]), ...overrideResponse
})
```

4. usage

```typescript

import { getListPetsMock, getShowPetByIdMock } from 'gen/pets.msw'

const pets = getListPetsMock({name: 'orverride'})
console.log(pets)
// => [{ id: 645061935104000, name: "orverride", tag: undefined }, { id: 83169910980608, name: "orverride", tag: undefined }]
​​
const pet = getShowPetByIdMock({name: 'orverride'})
console.log(pet)
// => { id: 7272122785202176, ​name: "orverride", tag: undefined }
```

### DIFF
--- a/packages/mock/src/msw/index.ts
+++ b/packages/mock/src/msw/index.ts
@@ -7,7 +7,7 @@ import {
   isFunction,
   pascal,
 } from '@orval/core';
-import { getRouteMSW } from '../faker/getters';
+import { getRouteMSW, overrideVarName } from '../faker/getters';
 import { getMockDefinition, getMockOptionsDataOverride } from './mocks';
 import { getDelay } from '../delay';
 
@@ -72,7 +72,7 @@ export const generateMSW = (
     value = definitions[0];
   }
 
-  const isResponseOverridable = value.includes('...overrideResponse');
+  const isResponseOverridable = value.includes(overrideVarName);
 
   const isTextPlain = response.contentTypes.includes('text/plain');
 
--- a/packages/mock/src/msw/index.ts
+++ b/packages/mock/src/msw/index.ts
@@ -63,6 +63,7 @@ export const generateMSW = (
   const mockData = getMockOptionsDataOverride(operationId, override);
 
   let value = '';
+  let isResponseOverridable = false;
 
   if (mockData) {
     value = mockData;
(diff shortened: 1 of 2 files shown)


melloware:
@Will-Mann-16 can you review since you recently did a lot of work in Mock?

packages/mock/src/msw/index.ts
```diff
@@ -80,7 +82,7 @@ export const generateMSW = (
     implementation: {
       function:
         value && value !== 'undefined'
-          ? `export const ${functionName} = () => (${value})\n\n`
+          ? `export const ${functionName} = (${isResponseOverridable ? `overrideResponse?: any` : ''}) => (${value})\n\n`
```
Will-Mann-16, line 85: Some ESLint config setups would throw errors if you try to spread an undefined value, is it worth setting this to a default param of `{}`, empty object?

soartec-lab:
@Will-Mann-16 
Hi, could you review this as well?

melloware (approved):

melloware:
It looks good to me if @Will-Mann-16 has an issue we can re-open.

Will-Mann-16 (commented):
This was the only thing I saw - if it's tested and works I'm happy. Otherwise might need a separate issue but would be an easy fix.

packages/mock/src/msw/index.ts
soartec-lab: @Will-Mann-16 

Thank you for your comment. That's good. I didn't have that perspective. I'll fix it in another PR.

PR #1181, opened 2024-01-26
feat: don't inject generated files into hook command

Arkanii:
## Status

**WIP OR READY IF NO TEST NEEDED**

## Description

`afterAllFilesWrite` hook inject generated files automatically for each command. 

In my case I use : 

```typescript

export default defineConfig({
  api: {
    input: {
      target: "http://php/api/docs.jsonopenapi",
      validation: false,
    },
    output: {
      mode: "tags-split",
      client: "react-query",
      baseUrl: "/api",
      target: "./api/endpoints.ts",
      schemas: "./types/models",
      override: {
        useDates: true,
        useNativeEnums: true,
      },
    },
    hooks: {
      afterAllFilesWrite: async () => {
        const { execa } = await import("execa");

        await execa("pnpm", ["run", "lint:orval"]);
      },
    },
  },
});
```
My `afterAllFilesWrite` function is a bit ugly. 😅 

I've added here the possibility to use `afterAllFilesWrite` as an object with two fields : 
- `command`: the command to run
- `injectGeneratedDirsAndFiles`: a boolean that indicates whether the generated files and folders should be injected into the command or not

I've keep the previous implementation to not create a BC.

## Todos

- [ ] Tests
- [x] Documentation

## Steps to Test or Reproduce

`orval.config.ts`

```typescript

export default defineConfig({
  petstore: {
    hooks: {
      afterAllFilesWrite: {
        command: 'prettier --write .',
        injectGeneratedDirsAndFiles: false,
      },
    },
    output: {
      mode: 'split',
      target: 'src/api/endpoints/petstoreFromFileSpecWithTransformer.ts',
      schemas: 'src/api/model',
      client: 'react-query',
      mock: true,
      override: {
        mutator: {
          path: './src/api/mutator/custom-instance.ts',
          name: 'customInstance',
        },
        operations: {
          listPets: {
            mock: {
              properties: () => ({
                '[].id': () => faker.number.int({ min: 1, max: 99999 }),
              }),
            },
            query: {
              useQuery: true,
              useSuspenseQuery: true,
              useSuspenseInfiniteQuery: true,
              useInfinite: true,
              useInfiniteQueryParam: 'limit',
            },
          },
          showPetById: {
            mock: {
              data: () => ({
                id: faker.number.int({ min: 1, max: 99 }),
                name: faker.person.firstName(),
                tag: faker.helpers.arrayElement([
                  faker.word.sample(),
                  undefined,
                ]),
              }),
            },
          },
        },
        mock: {
          properties: {
            '/tag|name/': () => faker.person.lastName(),
          },
        },
      },
    },
    input: {
      target: './petstore.yaml',
      override: {
        transformer: './src/api/transformer/add-version.js',
      },
    },
  },
});
```

I'm new to open source so every comment or help is welcome.

Thanks !


melloware:
I like where you are going but I assigned @anymaniax as he is the man in charge of this repo and I want to make sure this change aligns with what he typically expects.

docs/src/pages/reference/configuration/hooks.md
```diff
@@ -20,3 +20,18 @@ module.exports = {
...
 };
 ```
+
+If you don't want to inject the generated files into the command, you can use `afterAllFilesWrite` with an object:
+
+```js
+module.exports = {
+  petstore: {
+    hooks: {
+      afterAllFilesWrite: {
```
melloware: Shouldn't it be `afterAllFilesWritten` ?

melloware (changes requested):

docs/src/pages/reference/configuration/hooks.md
anymaniax: I think too

anymaniax (dismissed):

melloware (approved):

docs/src/pages/reference/configuration/hooks.md
melloware: I took care of it

Arkanii: @melloware @anymaniax 

Into code I saw `afterAllFilesWrite` and not `afterAllFilesWritten`

See :
- https://github.com/anymaniax/orval/blob/master/packages/orval/src/write-specs.ts#L150
- https://github.com/anymaniax/orval/blob/f529b9606165d9e5775ed9b65fc05b4187d6c327/packages/core/src/types.ts#L396

melloware: Good catch looks like something got messed up with the PR.

melloware: @Arkanii I just submitted a new PR if you want to review!

PR #1193, opened 2024-02-01
feat: don't inject generated files into hook command

melloware:
this fixes 1 spot that `afterAllFilesWritten` was missed

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -393,7 +393,7 @@ export type OperationOptions = {
   requestOptions?: object | boolean;
 };
 
-export type Hook = 'afterAllFilesWrite';
+export type Hook = 'afterAllFilesWritten';
 
 export type HookFunction = (...args: any[]) => void | Promise<void>;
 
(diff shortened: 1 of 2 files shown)


Arkanii (changes requested):
For me this PR is a BC, this will generate an error for each Orval user if this hook is implemented.

You need to accept both functions names, and deprecate `afterAllFilesWrite`.

`afterAllFilesWrite` needs to be removed in 7.0.0 version.



And don't forget to update documentation:

- https://github.com/anymaniax/orval/blob/master/docs/src/pages/reference/configuration/overview.md?plain=1#L63
- https://github.com/anymaniax/orval/blob/master/docs/src/pages/reference/configuration/full-example.md?plain=1#L59

melloware:
i will revert it back to allFilesWrite then even though it is grammatically incorrect.  ITs an easier fix in the short term

Arkanii:
Just revert the documentation in https://github.com/anymaniax/orval/pull/1181, I'll accept the PR quickly.

melloware:
Yep PR submitted

PR #1199, opened 2024-02-04
feat: enhance overrideResponse assignment

DevSDK:
## Status

**READY**

### 1. Apply loadsh/merge

According to the ECMAScript specification [1], The spread syntax does not guarantee deep assignment.

Because the properties will overwrite per each key by CreateDataProperty.

```
1. Let newDesc be the PropertyDescriptor { [[Value]]: V, [[Writable]]: true, [[Enumerable]]: true, [[Configurable]]: true }.
2. Return ? O.[[DefineOwnProperty]](P, newDesc).
```

(Also Object.assign [2] - [[Set]] in this case operation.)

For example:

```javascript
const override = { 
   data: {
          content: "John Doe"
   }
};


const result = { 
   data: {
      type: "NAME",
      content: "Seokho"
      // and the other properties mocked by fackerjs
   },
    ...override
};
```

The `result` will be:

```javascript
{
   data: {
          content: "John Doe"
   }
}

```

The user expectation that presents given properties is only valid on 1depth of the result.

Therefore, apply `loadsh/merge` [3] to assign property recursively.

### 2. Deterministic array length by overrideResponse

The overrideResponse can contain an array property. Currently, the override target generates a random length of it.

The length of the generated array can be shorter than the length in overrideResponse.
That possibly leads to unexpected overriding behavior.

For example:

```javascript
  const generated = [
     {
        content: "Seokho",
        id: "test-id", 
       // and many other properties... include required type
     },
  ]
  // The second one can present a non-deterministic result 
  // between only one property and added other properties per each call.
  const mocked = merge(generated, [{ content: "John Doe"}, { content: "Apple" }])

  
```

Hence, get length information from the array of the given overrideResponse to ensure the length of the generated array is greater than the given parameter.

The result will be looks like:

```javascript

   // ...
    {
      foo: Array.from(
        {
          length: Math.max(
            overrideResponse?.['foo']?.length ?? 0,
            faker.number.int({ min: 1, max: 10 }),
          ),
        },
        (_, i) => i + 1,
      ).map((_, index_0) =>  faker.helpers.arrayElement([
       // ...
            set: faker.helpers.arrayElement([
              {
                 testSet: Array.from(
                  {
                    length: Math.max(
                      overrideResponse?.['foo']?.[index_0]?.['set']?.['testSet']
                        ?.length ?? 0,
                      faker.number.int({ min: 1, max: 10 }),
                    ),
                  },
       // ...
```

[1] https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-runtime-semantics-propertydefinitionevaluation

[2] https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-object.assign

FYR: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#spread_in_object_literals

[3] https://lodash.com/docs/4.17.15#merge

## Related PRs

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout <branch>
> grunt jasmine
```

### DIFF
--- a/packages/mock/src/msw/index.ts
+++ b/packages/mock/src/msw/index.ts
@@ -108,10 +108,9 @@ export const ${handlerName} = (${isReturnHttpResponse && !isTextPlain ? `overrid
 
   return {
     implementation: {
-      function:
-        value && value !== 'undefined'
-          ? `export const ${functionName} = (${isResponseOverridable ? `${overrideVarName}: any = {}` : ''}): ${returnType} => ${isResponseOverridable ? `merge((${value}), overrideResponse)` : `(${value}) as any`}\n\n`
-          : '',
+      function: isReturnHttpResponse
+        ? `export const ${functionName} = (${isResponseOverridable ? `${overrideVarName}: any = {}` : ''}): ${returnType} => ${isResponseOverridable ? `merge((${value}), overrideResponse)` : `(${value}) as any`}\n\n`
+        : '',
       handlerName: handlerName,
       handler: handlerImplementation,
     },
--- a/packages/mock/src/msw/index.ts
+++ b/packages/mock/src/msw/index.ts
@@ -7,9 +7,10 @@ import {
   isFunction,
   pascal,
 } from '@orval/core';
-import { getRouteMSW, overrideVarName } from '../faker/getters';
+import { getRouteMSW } from '../faker/getters';
 import { getMockDefinition, getMockOptionsDataOverride } from './mocks';
 import { getDelay } from '../delay';
+import { overrideVarName } from '../createOverrideVariable';
 
 const getMSWDependencies = (locale?: string): GeneratorDependency[] => [
   {
@@ -23,6 +24,10 @@ const getMSWDependencies = (locale?: string): GeneratorDependency[] => [
   {
     exports: [{ name: 'faker', values: true }],
     dependency: locale ? `@faker-js/faker/locale/${locale}` : '@faker-js/faker',
+  },
+  {
+    exports: [{ name: 'merge', default: true, values: true }],
+    dependency: 'lodash/merge',
   },
 ];
 
@@ -103,9 +108,10 @@ export const ${handlerName} = (${isReturnHttpResponse && !isTextPlain ? `overrid
 
   return {
     implementation: {
-      function: isReturnHttpResponse
-        ? `export const ${functionName} = (${isResponseOverridable ? `overrideResponse: any = {}` : ''}): ${returnType} => (${value})\n\n`
-        : '',
+      function:
+        value && value !== 'undefined'
+          ? `export const ${functionName} = (${isResponseOverridable ? `${overrideVarName}: any = {}` : ''}): ${returnType} => ${isResponseOverridable ? `merge((${value}), overrideResponse)` : `(${value}) as any`}\n\n`
+          : '',
       handlerName: handlerName,
       handler: handlerImplementation,
     },
--- a/packages/mock/src/msw/index.ts
+++ b/packages/mock/src/msw/index.ts
@@ -109,7 +109,8 @@ export const ${handlerName} = (${isReturnHttpResponse && !isTextPlain ? `overrid
   return {
     implementation: {
       function: isReturnHttpResponse
-        ? `export const ${functionName} = (${isResponseOverridable ? `${overrideVarName}: any = {}` : ''}): ${returnType} => ${isResponseOverridable ? `merge((${value}), overrideResponse)` : `(${value}) as any`}\n\n`
+        ? // TODO(https://github.com/anymaniax/orval/issues/1210): 'as any' is used to avoid type errors until #1210 is fixed
+          `export const ${functionName} = (${isResponseOverridable ? `${overrideVarName}: any = {}` : ''}): ${returnType} => ${isResponseOverridable ? `merge((${value}), overrideResponse)` : `(${value}) as any`}\n\n`
         : '',
       handlerName: handlerName,
       handler: handlerImplementation,
--- a/packages/mock/src/createOverrideVariable.ts
+++ b/packages/mock/src/createOverrideVariable.ts
@@ -1,0 +1,23 @@
+export const overrideVarName = 'overrideResponse';
+
+export function createOverrideVariable(path?: string) {
+  let index = 0;
+
+  if (!path) {
+    return overrideVarName;
+  }
+
+  return (
+    overrideVarName +
+    path
+      ?.replace('#.', '')
+      .split('.')
+      .reduce((acc, key) => {
+        if (key !== '[]') {
+          return acc + `?.['${key}']`;
+        } else {
+          return acc + `?.[index_${index++}]`;
+        }
+      }, '')
+  );
+}
--- a/packages/mock/src/faker/getters/object.ts
+++ b/packages/mock/src/faker/getters/object.ts
@@ -13,8 +13,6 @@ import { MockDefinition, MockSchemaObject } from '../../types';
 import { combineSchemasMock } from './combine';
 import { DEFAULT_OBJECT_KEY_MOCK } from '../constants';
 
-export const overrideVarName = 'overrideResponse';
-
 export const getMockObject = ({
   item,
   mockOptions,
@@ -127,8 +125,6 @@ export const getMockObject = ({
         return `${keyDefinition}: ${resolvedValue.value}`;
       })
       .filter(Boolean);
-
-    properyScalars.push(`...${overrideVarName}`);
 
     value += properyScalars.join(', ');
     value +=
--- a/packages/mock/src/faker/getters/scalar.ts
+++ b/packages/mock/src/faker/getters/scalar.ts
@@ -15,6 +15,7 @@ import {
 import { MockDefinition, MockSchemaObject } from '../../types';
 import { getMockObject } from './object';
 import { DEFAULT_FORMAT_MOCK } from '../constants';
+import { createOverrideVariable } from '../../createOverrideVariable';
 
 export const getMockScalar = ({
   item,
@@ -195,10 +196,10 @@ export const getMockScalar = ({
 
       return {
         value:
-          `Array.from({ length: faker.number.int({ ` +
+          `Array.from({ length: Math.max(${createOverrideVariable(item.path)}?.length ?? 0,  faker.number.int({ ` +
           `min: ${mockOptions?.arrayMin}, ` +
-          `max: ${mockOptions?.arrayMax} }) ` +
-          `}, (_, i) => i + 1).map(() => (${mapValue}))`,
+          `max: ${mockOptions?.arrayMax} })) ` +
+          `}, (_, i) => i + 1).map((_, index_${item.path?.match(/\[\]/g)?.length ?? 0}) => (${mapValue}))`,
         imports: resolvedImports,
         name: item.name,
       };


DevSDK:
@melloware  Oops sorry. I didn't expect the Pull request to close when I renamed the branch on Github.

packages/mock/src/msw/index.ts
```diff
@@ -103,9 +107,10 @@ export const ${handlerName} = (${isReturnHttpResponse && !isTextPlain ? `overrid
 
   return {
     implementation: {
-      function: isReturnHttpResponse
-        ? `export const ${functionName} = (${isResponseOverridable ? `overrideResponse: any = {}` : ''}): ${returnType} => (${value})\n\n`
-        : '',
+      function:
+        value && value !== 'undefined'
+          ? `export const ${functionName} = (${isResponseOverridable ? `overrideResponse: any = {}` : ''}): ${returnType} => ${isResponseOverridable ? `merge((${value}), overrideResponse)` : `(${value}) as any`}\n\n`
```
DevSDK, line 112: It seems we have a type error already between the mocking result and the return type.

So far, the spread syntax for objects has resulted in an 'any' type that is broader than the intended return type, making it assignable to the return type.

This issue may handled separately.
Therefore, append ` as any ` type assertion temporarily.

FYI: https://www.typescriptlang.org/play?ts=5.2.2#code/FAFwngDgpgBAClEMC8MDeMB2BDAtrALhgGcQAnAS0wHMYBfYYAYwHtNSYA3bAGwFcoRBElRoc+IgCIQUXBEn1GrdkmyYwANV4CiasCnSLmbDjNJDEBjDAB0dvVv6wGwIA

packages/mock/src/msw/index.ts
```diff
@@ -103,9 +108,10 @@ export const ${handlerName} = (${isReturnHttpResponse && !isTextPlain ? `overrid
 
   return {
     implementation: {
-      function: isReturnHttpResponse
-        ? `export const ${functionName} = (${isResponseOverridable ? `overrideResponse: any = {}` : ''}): ${returnType} => (${value})\n\n`
-        : '',
+      function:
+        value && value !== 'undefined'
+          ? `export const ${functionName} = (${isResponseOverridable ? `${overrideVarName}: any = {}` : ''}): ${returnType} => ${isResponseOverridable ? `merge((${value}), overrideResponse)` : `(${value}) as any`}\n\n`
```
DevSDK: It seems we have a type error already between the mocking result and the return type.

So far, the spread syntax for objects has resulted in an 'any' type that is broader than the intended return type, making it assignable to the return type.

This issue may handled separately.
Therefore, append `as any` type assertion temporarily.

FYI: https://www.typescriptlang.org/play?ts=5.2.2#code/FAFwngDgpgBAClEMC8MDeMB2BDAtrALhgGcQAnAS0wHMYBfYYAYwHtNSYA3bAGwFcoRBElRoc+IgCIQUXBEn1GrdkmyYwANV4CiasCnSLmbDjNJDEBjDAB0dvVv6wGwIA

soartec-lab:
@DevSDK 
Thank you for made this PR, I'll review this.

packages/mock/src/msw/index.ts
```diff
@@ -103,9 +108,10 @@ export const ${handlerName} = (${isReturnHttpResponse && !isTextPlain ? `overrid
 
   return {
     implementation: {
-      function: isReturnHttpResponse
-        ? `export const ${functionName} = (${isResponseOverridable ? `overrideResponse: any = {}` : ''}): ${returnType} => (${value})\n\n`
-        : '',
+      function:
+        value && value !== 'undefined'
```
soartec-lab: common methods can be used.

```suggestion
        isReturnHttpResponse
```

packages/mock/src/msw/index.ts
soartec-lab: Please provide details about the error written here. What kind of error occurs when a type mismatch occurs?

soartec-lab (commented):
@DevSDK

I reviewed this PR. This PR includes several issues and feature additions that we would like to address.

I want to deeply understand how each problem occurs and how it can be improved.
Please consider creating an issue as it will help us proceed smoothly.

packages/mock/src/msw/index.ts
DevSDK: Sure :)

You can reproduce it by the steps:

At the current master 9f16e37f84a7747914007f54f88106de11927f1e

1. remove spread syntax in `packages/mock/src/faker/getters/object.ts` like this PR.

2. Enter `/test` directory and run `yarn generate`

3. Find the generated file like `tests/generated/angular/tags-split/pets/pets.msw.ts`

You can find the error message like:

`
Type '{ breed: "Labradoodle"; cuteness: number; barksPerMinute: number | undefined; type: "dog"; length?: undefined; '@id'?: undefined; callingCode?: undefined; country?: undefined; email?: undefined; id?: undefined; name?: undefined; tag?: undefined; petsRequested?: undefined; } | { ...; } | { ...; }' is not assignable to type 'Pet'.
...
`

Because the generated elements in `faker.helpers.arrayElement` missing for the required type like id.

Full-generated function is:

```typescript
export const getCreatePetsMock = (): Pet =>
  faker.helpers.arrayElement([
    {
      // This element doesn't have 'id' which is required type.
      breed: faker.helpers.arrayElement(['Labradoodle'] as const),
      cuteness: faker.number.int({ min: undefined, max: undefined }),
      barksPerMinute: faker.helpers.arrayElement([
        faker.number.int({ min: undefined, max: undefined }),
        undefined,
      ]),
      type: faker.helpers.arrayElement(['dog'] as const),
    },
    {
      breed: faker.helpers.arrayElement(['Dachshund'] as const),
      length: faker.number.int({ min: undefined, max: undefined }),
      barksPerMinute: faker.helpers.arrayElement([
        faker.number.int({ min: undefined, max: undefined }),
        undefined,
      ]),
      type: faker.helpers.arrayElement(['dog'] as const),
      '@id': faker.helpers.arrayElement([faker.word.sample(), undefined]),
      callingCode: faker.helpers.arrayElement([
        faker.helpers.arrayElement(['+33', '+420', '+33'] as const),
        undefined,
      ]),
      country: faker.helpers.arrayElement([
        faker.helpers.arrayElement([
          "People's Republic of China",
          'Uruguay',
        ] as const),
        undefined,
      ]),
      email: faker.helpers.arrayElement([faker.internet.email(), undefined]),
      id: faker.number.int({ min: undefined, max: undefined }),
      name: faker.word.sample(),
      tag: faker.helpers.arrayElement([faker.word.sample(), undefined]),
    },
    {
      petsRequested: faker.helpers.arrayElement([
        faker.number.int({ min: undefined, max: undefined }),
        undefined,
      ]),
      type: faker.helpers.arrayElement(['cat'] as const),
      '@id': faker.helpers.arrayElement([faker.word.sample(), undefined]),
      callingCode: faker.helpers.arrayElement([
        faker.helpers.arrayElement(['+33', '+420', '+33'] as const),
        undefined,
      ]),
      country: faker.helpers.arrayElement([
        faker.helpers.arrayElement([
          "People's Republic of China",
          'Uruguay',
        ] as const),
        undefined,
      ]),
      email: faker.helpers.arrayElement([faker.internet.email(), undefined]),
      id: faker.number.int({ min: undefined, max: undefined }),
      name: faker.word.sample(),
      tag: faker.helpers.arrayElement([faker.word.sample(), undefined]),
    },
  ]);
```


This can lead incomplete type between the return type and the infferred type.

See 'test' variable in [1].

It seems like a bug.

I'll file an issue on github then and add `// TODO(Issue number)` comment to track that. 

How about this way?

[1] https://www.typescriptlang.org/play?ssl=2&ssc=1&pln=3&pc=13#code/CYUwxgNghgTiAEAzArgOzAFwJYHtX1higE8BRCEAWxFQwB4AVAPgApCSAueAJRCmDwRiAQRhFijJgEouDeACh5GYgAcEABRAZ4AXngBvefGPwswLgGcMMLKgDmAGiMnoAIxARL123efGcKth4UBAA-F429vIAvopgeFbwGCBWXJraeuxkFNS0LADafgZmXADkUK5gpdEOBm4eZWBN1fIAulJAA

packages/mock/src/msw/index.ts
DevSDK: Oh it seems changed by missing conflict-resolving sorry!

Fixed in 3d152bbef1389566d37331370a3fdf2bf6588cb1

PTAL

packages/mock/src/msw/index.ts
soartec-lab: Thank you let me known.
I think it is better not to change the fact that the returned value is `Pet`, which is the same as the API response.

The problem here is that the generated fake object does not have an `id` and `name`, so the fake object generation itself needs to be fixed.

soartec-lab: @DevSDK 
Can you use a separate issue to fix the Fake object?

soartec-lab:
For example, there are some functions that can no longer be overridden, such as `tests/generated/angular/tags-split/pets/pets.msw.ts`. Can you check this?

```typescript
export const getListPetsMock = (overrideResponse: any = {}): Pets => (Array.from({ length: faker.number.int({ min: 1, max: 10 }) }, (_, i) => i + 1).map(() => (faker.helpers.arrayElement([{breed: faker.helpers.arrayElement(['Labradoodle'] as const), cuteness: faker.number.int({min: undefined, max: undefined}), ...overrideResponse,barksPerMinute: faker.helpers.arrayElement([faker.number.int({min: undefined, max: undefined}), undefined]), type: faker.helpers.arrayElement(['dog'] as const), ...overrideResponse},{breed: faker.helpers.arrayElement(['Dachshund'] as const), length: faker.number.int({min: undefined, max: undefined}), ...overrideResponse,barksPerMinute: faker.helpers.arrayElement([faker.number.int({min: undefined, max: undefined}), undefined]), type: faker.helpers.arrayElement(['dog'] as const), ...overrideResponse,'@id': faker.helpers.arrayElement([faker.word.sample(), undefined]), callingCode: faker.helpers.arrayElement([faker.helpers.arrayElement(['+33','+420','+33'] as const), undefined]), country: faker.helpers.arrayElement([faker.helpers.arrayElement(['People\'s Republic of China','Uruguay'] as const), undefined]), email: faker.helpers.arrayElement([faker.internet.email(), undefined]), id: faker.number.int({min: undefined, max: undefined}), name: faker.word.sample(), tag: faker.helpers.arrayElement([faker.word.sample(), undefined]), ...overrideResponse},{petsRequested: faker.helpers.arrayElement([faker.number.int({min: undefined, max: undefined}), undefined]), type: faker.helpers.arrayElement(['cat'] as const), ...overrideResponse,'@id': faker.helpers.arrayElement([faker.word.sample(), undefined]), callingCode: faker.helpers.arrayElement([faker.helpers.arrayElement(['+33','+420','+33'] as const), undefined]), country: faker.helpers.arrayElement([faker.helpers.arrayElement(['People\'s Republic of China','Uruguay'] as const), undefined]), email: faker.helpers.arrayElement([faker.internet.email(), undefined]), id: faker.number.int({min: undefined, max: undefined}), name: faker.word.sample(), tag: faker.helpers.arrayElement([faker.word.sample(), undefined]), ...overrideResponse}]))))

export const getCreatePetsMock = (overrideResponse: any = {}): Pet => (faker.helpers.arrayElement([{breed: faker.helpers.arrayElement(['Labradoodle'] as const), cuteness: faker.number.int({min: undefined, max: undefined}), ...overrideResponse,barksPerMinute: faker.helpers.arrayElement([faker.number.int({min: undefined, max: undefined}), undefined]), type: faker.helpers.arrayElement(['dog'] as const), ...overrideResponse},{breed: faker.helpers.arrayElement(['Dachshund'] as const), length: faker.number.int({min: undefined, max: undefined}), ...overrideResponse,barksPerMinute: faker.helpers.arrayElement([faker.number.int({min: undefined, max: undefined}), undefined]), type: faker.helpers.arrayElement(['dog'] as const), ...overrideResponse,'@id': faker.helpers.arrayElement([faker.word.sample(), undefined]), callingCode: faker.helpers.arrayElement([faker.helpers.arrayElement(['+33','+420','+33'] as const), undefined]), country: faker.helpers.arrayElement([faker.helpers.arrayElement(['People\'s Republic of China','Uruguay'] as const), undefined]), email: faker.helpers.arrayElement([faker.internet.email(), undefined]), id: faker.number.int({min: undefined, max: undefined}), name: faker.word.sample(), tag: faker.helpers.arrayElement([faker.word.sample(), undefined]), ...overrideResponse},{petsRequested: faker.helpers.arrayElement([faker.number.int({min: undefined, max: undefined}), undefined]), type: faker.helpers.arrayElement(['cat'] as const), ...overrideResponse,'@id': faker.helpers.arrayElement([faker.word.sample(), undefined]), callingCode: faker.helpers.arrayElement([faker.helpers.arrayElement(['+33','+420','+33'] as const), undefined]), country: faker.helpers.arrayElement([faker.helpers.arrayElement(['People\'s Republic of China','Uruguay'] as const), undefined]), email: faker.helpers.arrayElement([faker.internet.email(), undefined]), id: faker.number.int({min: undefined, max: undefined}), name: faker.word.sample(), tag: faker.helpers.arrayElement([faker.word.sample(), undefined]), ...overrideResponse}]))

export const getShowPetByIdMock = (overrideResponse: any = {}): Pet => (faker.helpers.arrayElement([{breed: faker.helpers.arrayElement(['Labradoodle'] as const), cuteness: faker.number.int({min: undefined, max: undefined}), ...overrideResponse,barksPerMinute: faker.helpers.arrayElement([faker.number.int({min: undefined, max: undefined}), undefined]), type: faker.helpers.arrayElement(['dog'] as const), ...overrideResponse},{breed: faker.helpers.arrayElement(['Dachshund'] as const), length: faker.number.int({min: undefined, max: undefined}), ...overrideResponse,barksPerMinute: faker.helpers.arrayElement([faker.number.int({min: undefined, max: undefined}), undefined]), type: faker.helpers.arrayElement(['dog'] as const), ...overrideResponse,'@id': faker.helpers.arrayElement([faker.word.sample(), undefined]), callingCode: faker.helpers.arrayElement([faker.helpers.arrayElement(['+33','+420','+33'] as const), undefined]), country: faker.helpers.arrayElement([faker.helpers.arrayElement(['People\'s Republic of China','Uruguay'] as const), undefined]), email: faker.helpers.arrayElement([faker.internet.email(), undefined]), id: faker.number.int({min: undefined, max: undefined}), name: faker.word.sample(), tag: faker.helpers.arrayElement([faker.word.sample(), undefined]), ...overrideResponse},{petsRequested: faker.helpers.arrayElement([faker.number.int({min: undefined, max: undefined}), undefined]), type: faker.helpers.arrayElement(['cat'] as const), ...overrideResponse,'@id': faker.helpers.arrayElement([faker.word.sample(), undefined]), callingCode: faker.helpers.arrayElement([faker.helpers.arrayElement(['+33','+420','+33'] as const), undefined]), country: faker.helpers.arrayElement([faker.helpers.arrayElement(['People\'s Republic of China','Uruguay'] as const), undefined]), email: faker.helpers.arrayElement([faker.internet.email(), undefined]), id: faker.number.int({min: undefined, max: undefined}), name: faker.word.sample(), tag: faker.helpers.arrayElement([faker.word.sample(), undefined]), ...overrideResponse}]))
```

anymaniax:
I am not sure that it’s a good idea to force the user to install lodash merge no?

melloware:
I agree I am not a fan of forced dependencies.

soartec-lab:
I don't want to increase the number of dependent libraries if possible. It is convenient because there are no dependent libraries, and `lodash` is a general-purpose library, which may be a reason for some users to avoid it.

DevSDK:
Make sense. I respect that.

We can choose the alternative ways:

Implement recursive assignments like lodash.merge, and then

1. Insert on top of every .msw file 
    This way requires some mechanism to inject only one function on top of the msw file. (Does a mechanism  already exist to handle it?)
2. Or add like libs/ directory and import that. 
   It needs to copy or generate a file that we want to import. (AFAIK, The /mock package is unable to handle this right?) 
Perhaps we should create an initializer on /mock or add another place other than /mock

Which way would you like prefer to approach? 

Let me know prefer way and it would be great to advise me on some consideration points of each way if you don't mind.

packages/mock/src/msw/index.ts
DevSDK: Yep. I filed on https://github.com/anymaniax/orval/issues/1210

And add a TODO comment [3a7e95f](https://github.com/anymaniax/orval/pull/1199/commits/3a7e95fefb4f3fa8598be85783c65affa90442ee)

PTAL.

DevSDK:
@soartec-lab 

Related: https://github.com/anymaniax/orval/pull/1199#issuecomment-1931733511

Yes. It can be unable to override.

Because it will be decided by whether the 'overrideResponse' string exists or not in value.

It happened by removing  '...overrideResponse' string from the value string by removing the push statement in the object.js.

The simplest way to resolve this is adding `/* overrideResponse */` comment right before place. That would work with the previous logic.

I assume a bit complex way will be to return overridable information from `getMockDefinition()` or store that information somewhere.

How about a simplest way? Or can you suggest fixing this case?

soartec-lab:
Regarding the override variable being lost, if it was implemented correctly without adding comments, the current string match judgment would work fine. And, alternative implementations of `lodash`, i may need to discuss more whether the ideas you want to improve are compatible with ours.

DevSDK:
@soartec-lab 

The overridable information is only checked by string match.
If we're considering removing the spread syntax, which seems to be the primary mechanism for handling this, we'll need to find an alternative solution. One possibility could be just adding comments or exploring a different approach altogether.

I understand that there are some blocking tasks related to separate issues that need to be addressed first. I'm happy to wait until those are completed before diving into this further.
So feel free to notify me when my task is ready to progress.

And sure, the alternative way requires the maintainer's review :).
So any suggestion or feedback to fit yours will be welcome.

packages/mock/src/msw/index.ts
soartec-lab: fixed #1210 so i mark resolve this

karlismelderis-mckinsey:
I don't think merging arrays is an obvious, easy to grasp option

array should be overwritten with new array

soartec-lab:
Hi, @DevSDK 

We have considered handling mock overrides, but are trying to avoid nested overrides, as suggested in the PR below:

https://github.com/anymaniax/orval/pull/1338

It conflicts with the idea you suggested, so if you want to discuss it, please comment.

DevSDK:
@soartec-lab oh wow. That change is huge.

For now, I haven't caught the context yet so please understand if I say weird below.

We should handle nested.
The change seems to handle only Partial<Response> type my thought is we should handle [DeepPartial](https://github.com/ts-essentials/ts-essentials/tree/master/lib/deep-partial).


use case of our product was (a bit overstated)
* array
    * object - A
       * array
         * object -B
           * singleProperty - I want to test this value!

If only the root is overridable with Partial<T>, we should make whole Object A and B. 



In our company, we created a fixture creator to resolve the same problem with that nested interface.
For example
```typescript
 function createAwesomeThingFixtures(props: DeepPartial<Type>) {
    ...
}

const fixture =  createAwesomeThingFixtures(
    [
      {
        id: '1',
        type,
        roleReference:  {
            name: "Seokho"
            // There are more than 20 properties in here! 
           // but I want to test only "name"
        }
       // and here, also have more properties
	
      },
      {
        id: '2',
        type,
        roleReference: {
            name: "Ph"
            // There are more than 20 properties in here! 
           // but I want to test only "name"
        }
      },
    ]
)
```

What if we write the whole 20 properties in there, I think there can be significant developer experience damage and it will be hard to change

I'll check and participate in it but now I'm a bit busy.. So... I may slow.

soartec-lab:
@DevSDK 
Thank you for your response. Please take care of your body.
I understand that. For objects with many properties, there may be another solution, such as making the object easier to create. At least #1338 solves the current `orval` having problem and I think it will be useful for many users, so I would like to merge it.

Do you have an opinion on that?
If your response is slow, I'll pass it on first, but I welcome discussion, so please comment again.

DevSDK:
@soartec-lab 

Yeah. I think #1338 will cause a breaking change because the exposed function argument type changed: any -> Partial<T>

But after that, Partial<T> -> DeepPartial<T> seems safe from the breaking changes at the interface level.

So.. I agree that we can move on gradually. I'll pull that change and make supporting DeepPartial<T> that will enhance dx for nested.

soartec-lab:
@DevSDK 

Thank you for confirming.

soartec-lab:
Hey @DevSDK , it's been a while since this PR was opened. Mocks are still being improved, so conflicts still occur.
Therefore, is it okay to close this PR?

If there is an issue, it would be a good idea to open the PR again, and once the policy is organized in an issue, I think someone else can continue to deal with it. and if you would like to come back to development, I would always welcome you back.

DevSDK:
Yep. If you prefer so.

Actually, I resolved the conflict on the local repo. (It was not as much as I thought)

So we can decide to push the merge commit with resolving the conflicts and continue gradually or simply reopen when I'm ready. 

However, I might be slow due to the other things to do... So If we decide to close this PR, yeah, I'll do it again someday ;)

soartec-lab:
@DevSDK 

Thank you for contacting me. I have determined that the issue and approach have changed from the original purpose of creating this PR, so if it is possible to continue, could you please open the PR again?
If there is no problem, I will close this PR once.

DevSDK:
@soartec-lab Yep :) I'll do it soon. Feel free to close this

PR #1214, opened 2024-02-10
fix(msw): refactor to include duplication of mock imports in mocks in `split` and `split-tags` modes

soartec-lab:
## Status

**READY**

## Description

In `split` and `split-tags` modes, a hooks import is included to import the return type into `mock`.
However, we can get around that by getting the return type from `response.imports` and including it in the `mock` import, so we fixed it.

## Related PRs

none

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

pass test

### DIFF
--- a/packages/core/src/writers/split-tags-mode.ts
+++ b/packages/core/src/writers/split-tags-mode.ts
@@ -75,10 +75,7 @@ export const writeSplitTagsMode = async ({
         });
         mockData += builder.importsMock({
           implementation: implementationMock,
-          imports: [
-            ...importsForBuilder,
-            { exports: importsMock, dependency: relativeSchemasPath },
-          ],
+          imports: [{ exports: importsMock, dependency: relativeSchemasPath }],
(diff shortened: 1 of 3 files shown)


packages/core/src/writers/split-tags-mode.ts
```diff
@@ -75,10 +75,7 @@ export const writeSplitTagsMode = async ({
         });
         mockData += builder.importsMock({
           implementation: implementationMock,
-          imports: [
-            ...importsForBuilder,
```
melloware, line 79: So this `...importsForBuilder` was not necessary or not used?

soartec-lab: No, until now this was certainly necessary. For example, because we were using it to import response return types like `Pet` or `Pets`. With this change, we can now be covered by mock imports, so I removed them.

melloware (approved):

PR #1215, opened 2024-02-11
fix(query): type return of useMutation query

### LINKED ISSUE #1195: Node16 or bundler module resolution causes TS2742 error
1. Use orval in a repo where the tsconfig `moduleResolution` is set to `node16` or `bundler`
2. Generate react-query client where there are mutation queries

## What happens?

The following gets reported in IDE and also when running `npx tsc`.

```
The inferred type of 'X' cannot be named without a reference to '../../node_modules/msw/lib/core/handlers/HttpHandler'. This is likely not portable. A type annotation is necessary.

The inferred type of 'X' cannot be named without a reference to '../../node_modules/@tanstack/react-query/build/modern/types'. This is likely not portable. A type annotation is necessary.
```

Declaration files where these errors happen do not get generated.

TS2742 should not happen.

Please see above

## Any other comments?

Possibly linked to:
- https://github.com/microsoft/TypeScript/issues/42873
- https://github.com/TanStack/query/discussions/6670

- I think if we strictly type the return of the `useMutation` on the following line, it fixes this error. https://github.com/anymaniax/orval/blob/9b28acd87b88dee1b5556c9af7b446b429c92e79/packages/query/src/index.ts#L1451
- Same thing should be done for the mock files.

Strictly typing the returns would be a workaround until this is actually fixed upstream in TypeScript.

I can look into creating a PR for this. Please advise if you would like me to address this.

david10sing:
## Status

**READY**

## Description

Fix #1195 

Using `orval` in a repository where the `tsconfig` `moduleResolution` property is not `node` results in https://github.com/anymaniax/orval/issues/1195 when producing `useMutation` queries.

As per the issue, looks like it needs to be fixed properly upstream but in the meantime, this PR puts in a fix which should silence that error for users using `orval` with `moduleResolution` set to something newer than `node`.

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout <branch>
> grunt jasmine
```

1. Run `yarn build`


melloware:
This LGTM but going to let @anymaniax review.

david10sing:
@anymaniax could I get your eyes on this please? Thank you 🙏

anymaniax:
Seems good. Why only for react query?

david10sing:
> Seems good. Why only for react query?

Hi @anymaniax,

Sorry for the delay in getting back.

I am using orval with react-query that's why. I am not sure about the other libraries.

david10sing:
@anymaniax,

Updated the mutation queries for `react-query`, `svelte-query` and `vue-query`

melloware (approved):

PR #1229, opened 2024-02-19
fix: TypeError custom override

### LINKED ISSUE #1202: Mock TypeError custom override
## When using the override mock option the TypeCheck is too strict

With the release of 6.24 orval introduced the return type for the mws files, see #1171. 
When using the override option for certain opperations the TypeCheck is too strict, without any options to set the type.
As orval uses `js` to parse the document we can not add extra type declarations to the config. This breaks our 

1. Define a override operation in the `orval.config.js`
2. Set the option `mock` to `true`
3. Get a type error when override Enums or literals

## What happens?

TypeErrors in the `.msw.ts` file

Not sure what the best solution is. The return type is a nice feature, but maybe we should be able to override the return value, or not set it in case of the override function in the config. As orval uses JavaScript and not TypeScript we can not strictly type the return value in the config.

Not at the moment, if you need more details I can add it. 

## Any other comments?

Open for suggestions.

dipsaus9:
## Status

Ready

Fix #1202 

## Description

If a custom mock is implemented we don't want a return type. This would be ideal to have. But, in the scenario of an enum, strict literal it is not possible to follow these strict types. Turning them off would be better in this scenario.

## Todos

- [x] Tests (only added an empty object that should not give a type error)
- [ ] Changelog Entry (unreleased) - not sure what to do?

## Steps to Test or Reproduce

I added a extra test. Without the change the test would break


melloware (approved):

soartec-lab:
@dipsaus9 
It seems like I didn't understand what you meant correctly, but after seeing this PR I understood. Thank you for the great changes.

PR #1231, opened 2024-02-21
feat: jsdoc block tags

Marfyy:
## Status

READY

## Description

Added support to generate block tags to JSDoc comments for some of the common Schema object properties. Did some simple refactoring in the `jsDoc` function to avoid duplication of code.

Here is the OpenAPI documentation for the properties [here](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#properties)

They are not official JSDoc tags, but VSCode picks them up as if they were in Intellisense, which was my goal.

I added all the new supported tags in `sample/basic/petstore.yaml`, which is where I tested that it worked. I looked through the other samples, but none of them used any of these properties, so no need to update them.

## Todos

- [X] Tests
- [ ] Documentation

## Steps to Test or Reproduce

Add any of the following properties to a schema object:

- minLength
- maxLength
- minimum
- maximum
- exclusiveMinimum
- exclusiveMaximum
- minItems
- maxItems
- nullable
- pattern

### DIFF
--- a/packages/core/src/utils/doc.ts
+++ b/packages/core/src/utils/doc.ts
@@ -8,10 +8,30 @@ export function jsDoc(
     description,
     deprecated,
     summary,
+    minLength,
+    maxLength,
+    minimum,
+    maximum,
+    exclusiveMinimum,
+    exclusiveMaximum,
+    minItems,
+    maxItems,
+    nullable,
+    pattern,
   }: {
     description?: string[] | string;
     deprecated?: boolean;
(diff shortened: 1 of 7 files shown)


melloware:
I like it. Running the build now.

soartec-lab (approved):
Thank you. I like it, but i wait review by @melloware and @anymaniax .

melloware (approved):

PR #1235, opened 2024-02-25
fix(query): usePrefetch should only work for Query/Infinite types

### LINKED ISSUE #1234: usePrefetch+useSuspenseQuery incorrectly generates prefetchSuspenseQuery calls

- Enable both `usePrefetch: true` and `useSuspenseQuery: true` 

## What happens?

- `prefetchSuspenseQuery` calls are generated. `prefetchSuspenseQuery` actually doesn't exist in react-query. The correct method to call is `prefetchQuery`

- Strictly `prefetchQuery` generated calls

**Package Version:** 6.25.0

ezequiel:
## Status

**READY**

## Description

Make sure we only generate prefetch calls for `QUERY` and `INFINITE` types, as it isn't possible to prefetch any other types.

Fixes #1234

## Steps to Test or Reproduce

```ts
export default defineConfig({
  client: {
    input: '...',
    output: {
      client: 'react-query',
      override: {
        query: {
          usePrefetch: true,
          useQuery: true,
          useSuspenseQuery: true,
        },
      },
      ...output,
    },
  },
});
```

Verified strictly `await queryClient.prefetchQuery(queryOptions)` calls were generated

### DIFF
--- a/packages/query/src/index.ts
+++ b/packages/query/src/index.ts
@@ -1075,7 +1075,7 @@ ${doc}export const ${camel(
   return ${queryResultVarName};
 }\n
 ${
-  usePrefetch
+  usePrefetch && (type === QueryType.QUERY || type === QueryType.INFINITE)
     ? `${doc}export const ${camel(
         `prefetch-${name}`,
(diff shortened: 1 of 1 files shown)


ezequiel:
cc @anymaniax @melloware

melloware (approved):

PR #1237, opened 2024-02-25
fix(core): improve NonReadonly type

ezequiel:
## Status

**READY**

## Description

Fixes https://github.com/anymaniax/orval/issues/1236

Improves the `NonReadonly<T>` type so it actually works now. Note: this implementation was taken from ts-essentials' DeepWritable<T>, which seem to have the most accurate type as of now

## Before
Screenshot 2024-02-25 at 2 58 28 AM

## After
image


ezequiel:
cc @anymaniax @melloware

ezequiel:
Not sure if this affects https://github.com/anymaniax/orval/issues/616. I can't really make sense of the original issue at the moment. It seems the goal was to turn readonly fields into optional fields? Usually we do this as part of the Swagger model definition itself by making the input body field optional

melloware (approved):

PR #1238, opened 2024-02-26
fix(core): NPE check in validator

### LINKED ISSUE #1139: Updating from 6.19.1 to 6.20.0 leads to TypeError: Cannot read properties of undefined (reading 'length')

0. Update from 6.19.1 to 6.20.0 (still broken in 6.23.0)
1. Given the following config:
```ts
import path = require('path');

export default defineConfig({
  api: {
    output: {
      mode: 'tags',
      workspace: path.join(__dirname, '../../libs/api-interfaces/src'),
      target: './index.ts',
      schemas: './models',
      client: 'react-query',
      mock: false,
      override: {
        mutator: {
          path: './custom-client.ts',
          name: 'customClient',
        },
        query: {
          options: {
            throwOnError: false,
          },
        },
      },
    },
    input: {
      target: path.join(__dirname, '../../dist/apps/api/spec.json'),
      validation: true,
    },
  },
});
```
2. Running orval `orval --config apps/api/orval.config.ts`

## What happens?

```
🛑  api - TypeError: Cannot read properties of undefined (reading 'length')
```

```
🎉 api - Your OpenAPI spec has been converted into ready to use orval!
```

## Any other comments?

Works in version `6.19.1`. Also works with `validation: false` 🤔

melloware:
fix #1139 fix(core): NPE check in validator

### DIFF
--- a/packages/core/src/utils/validator.ts
+++ b/packages/core/src/utils/validator.ts
@@ -17,11 +17,11 @@ export const ibmOpenapiValidator = async (specs: OpenAPIObject) => {
   spectral.setRuleset(ibmOpenapiRuleset);
   const { errors, warnings } = await spectral.run(specs);
 
-  if (warnings.length) {
+  if (warnings && warnings.length) {
     ibmOpenapiValidatorWarnings(warnings);
   }
 
-  if (errors.length) {
+  if (errors && errors.length) {
     ibmOpenapiValidatorErrors(errors);
   }
 };


packages/core/src/utils/validator.ts
```diff
@@ -17,11 +17,11 @@ export const ibmOpenapiValidator = async (specs: OpenAPIObject) => {
   spectral.setRuleset(ibmOpenapiRuleset);
   const { errors, warnings } = await spectral.run(specs);
 
-  if (warnings.length) {
+  if (warnings && warnings.length) {
     ibmOpenapiValidatorWarnings(warnings);
   }
 
-  if (errors.length) {
+  if (errors && errors.length) {
```
ezequiel, line 24: I'm curious why this was returning undefined in the first place -- is that normal, or is it because `spectral.run(...)` failed?

melloware: It is not for me i have no issues with Orval so I am not sure if its related to your environment or settings or something?

soartec-lab (approved):
Thanks ! I checked it.

PR #1265, opened 2024-03-17
Fix/zod entry array

anymaniax:
## Status

**READY**

## Description

Zod client handle schema with directly an array as a response or body


soartec-lab (approved):
So you made it possible to handle request and response arrays right? I think it's great!

PR #1266, opened 2024-03-17
Feature/hono client

anymaniax:
## Status

**READY**

## Description

Setup hono as new client

Also add a new mode by operation


samples/hono/src/handlers/createPets.ts
```diff
@@ -0,0 +1,12 @@
+import { createFactory } from 'hono/factory';
```
soartec-lab, line 1: The handlers is  sccafold and i have to implement it myself after generation right?
In that case, I understand that regenerating will roll back the implementation, is this my correct understanding?

samples/hono/src/petstore.ts
```diff
@@ -0,0 +1,153 @@
...
+  "@id": zod.string().optional(),
+  "id": zod.number(),
+  "name": zod.string(),
+  "tag": zod.string().optional(),
+  "email": zod.string().email().optional(),
+  "callingCode": zod.enum(['+33', '+420', '+33']).optional(),
+  "country": zod.enum(['People\'s Republic of China', 'Uruguay']).optional()
+})
+
+export type CreatePetsContext<E extends Env = any> = Context<E, '/pets', { in: { json: CreatePetsBodyItem[], }}>;
```
soartec-lab: IMO, I think it is a good idea to move `Contenxt` and zod validate objects to the `handler` file for two reasons.

1. Because they are used only within the handler.
2. To avoid circular references between modules in `handler` and `petstore`

packages/hono/README.md
```diff
@@ -0,0 +1,28 @@
+[![npm version](https://badge.fury.io/js/orval.svg)](https://badge.fury.io/js/orval)
```
soartec-lab, line 1: We should also add a link to the `hono` example in the `README` of other packages. but, It's not a direct effect, so you can deal with it later.

soartec-lab (commented):
@anymaniax 

It is wonderful. I took a look at it right away, and there were some points I wanted to ask at this point, so I commented.

samples/hono/src/handlers/createPets.ts
anymaniax: Indeed you will need to implement it your self and I added some verification to check if the handler name is in the file or not like that I don't scrap your implementation if it's already in the handler file

samples/hono/src/petstore.ts
anymaniax: Mmmh didn't thought about that indeed. And also the context definition. It's a bit annoying because that's not something that the user will touch. I will try to add a zod file instead then.

samples/hono/src/handlers/createPets.ts
soartec-lab: > I don't scrap your implementation if it's already in the handler file

that's great 👍  I had missed that implementation.

samples/hono/src/petstore.ts
soartec-lab: Indeed, I agree that it noisy when the handler contains definitions that the user does not need to edit. And considering that the handler file is not regenerated, it doesn't seem like a good idea to bundle these with the handler file. It might be ideal to move the `Context` or `zod` objects or both to a new separate file.

melloware (approved):
This looks great and great feedback @soartec-lab !

samples/hono/src/petstore.ts
anymaniax: Just added the implementation to generate the two new files

packages/hono/README.md
anymaniax: Just updated the readme too

samples/hono/src/handlers/createPets.ts
```diff
@@ -1,12 +1,14 @@
...
 
 const factory = createFactory();
 
+
 export const createPetsHandlers = factory.createHandlers(
-  zValidator('json', createPetsBody),
-  (c: CreatePetsContext) => {
-    return c.json({ message: 'createPets handler' });
+zValidator('json', createPetsBody),
+(c: CreatePetsContext) => {
```
soartec-lab: Since there is no longer a default implementation, `c` is now unused. Depending on the user's formatter settings, a warning will occur.
Is this an intended change?

soartec-lab (commented):
@anymaniax 

Thank you for the update. I think it's great. I commented because there was one additional place I wanted to check.

Also, please consider adding a guide to the `hono` client as it will be useful to many users and will encourage its use. However, we can add documents later, so we will leave it up to you to include them in this PR.

samples/hono/src/petstore.ts
soartec-lab: thank you. We have confirmed that this issue is resolved 👍

packages/core/src/types.ts
```diff
@@ -209,6 +211,7 @@ export const OutputMode = {
   SPLIT: 'split',
   TAGS: 'tags',
   TAGS_SPLIT: 'tags-split',
+  OPERATION: 'operation',
```
soartec-lab: Please allow me to ask a question to include my understanding
Why do we need an operational mode?
I think other modes such as `split` will work usefully in `hono` as well. If manual updating is necessary, I thought it would be sufficient to clearly state in the guide a rule that only the `builder` of `hono` should not update the file. Or it may be a good idea to specify it with another option such as override. I thought we shouldn't add new modes if we can do it without reducing complexity.

What do you think?

anymaniax: The operation mode is not even used by the hono client. I did the development of it at the beginning. It didn't work how I wanted for it. I thought it was a good feature to keep too. To be able on every client to have a operation generation by file. But if you think it's not useful I can remove it or move it to another pull request

samples/hono/src/handlers/createPets.ts
```diff
@@ -0,0 +1,11 @@
+import { createFactory } from 'hono/factory';
+import { zValidator } from '@hono/zod-validator';
+import { CreatePetsContext } from '../petstore.context';
+import { createPetsBody } from '../petstore.zod';
+
+const factory = createFactory();
+
+export const createPetsHandlers = factory.createHandlers(
+  zValidator('json', createPetsBody),
+  (c: CreatePetsContext) => {},
```
soartec-lab, line 10: @anymaniax 

I wasn't able to comment properly...

Since there is no longer a default implementation, c is now unused. Depending on the user's formatter settings, a warning will occur. Is this an intended change?

packages/core/src/types.ts
soartec-lab: > The operation mode is not even used by the hono client. 

thank you for teaching me. 
Currently, other clients output corrupted files and do not work, so since this mode is common to all clients, I would suggest moving to another PR.

samples/hono/src/handlers/createPets.ts
anymaniax: It’s normally always need to be implemented so shouldn’t be an issue

soartec-lab: Sounds good 👍

packages/core/src/types.ts
anymaniax: Just removed the code

soartec-lab: thank you. I have confirmed the changes !

PR #1268, opened 2024-03-18
fix(vue): update samples to v5 and fix type issue

Maxim-Mazurok:
## Status

**READY**

## Description

Updated vue-query to v5 in `samples`, and had to make a `queryOptions.queryKey` -> `unref(queryOptions).queryKey` fix for it to compile.

Without it:
image

---

Also I tried to update to v5 in `tests`, but I'm getting errors like this:
```
generated/vue-query/tags/pets.ts:270:14 - error TS2742: The inferred type of 'useDeletePetById' cannot be named without a reference to '@tanstack/vue-query/node_modules/@tanstack/query-core/build/legacy/queryClient-6vLRMq5C'. This is likely not portable. A type annotation is necessary.

270 export const useDeletePetById = <TError = AxiosError<Error>,
```
and I have no idea how to fix it.
I feel like the whole idea is to use inferred type instead of re-defining it... Anyway, we can leave it for later I guess.

### DIFF
--- a/packages/query/src/index.ts
+++ b/packages/query/src/index.ts
@@ -1110,9 +1110,9 @@ ${doc}export const ${camel(
     `${operationPrefix}-${type}`,
   )}(${queryOptionsVarName}) as ${returnType};
 
-  ${queryResultVarName}.queryKey = ${queryOptionsVarName}.queryKey ${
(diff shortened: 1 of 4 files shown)


anymaniax (approved):
Thanks seems good

PR #1274, opened 2024-03-20
fix(msw): make sure to not try to include error type in mock implementation

### LINKED ISSUE #1273: SWR/React Query: Custom error type model is not imported when mock: true

1. Change petstore.yaml and rename `components/schemas/Error` to `components/schemas/ErrorX` (make sure to also change the name of the component in the bottom of the file)
2. yarn generate:swr
3. See that generated code is `export type ListPetsQueryError = AxiosError<ErrorX>`, which is correct, but ErrorX is not imported from ./model

## What happens?

Compile error, it seems like it used to import the ts default Error interface instead of the actual error model.

The error model should be imported

image

**Package Version:** 6.25.0

AffeJonsson:
## Status

**READY**

## Description

Fixes #1273 
Mock tried to import the error type, which caused the non-mock implementation to not try and import it. The mock code then filtered out all imports that weren't used in the mock, thus leaving out the error type.

## Todos

- [x] Tests
- [ ] Documentation

## Steps to Test or Reproduce

1. cd tests
2. yarn generate:swr
3. go to generated/swr/petstore/endpoints.ts
4. see that the Error type is the generated Error model, and not the default js Error
```ts
export type ListPetsQueryError = AxiosError<Error>
//                                            ^ Go to definition leads to generated Error model
```
5. compare to previous version, where go to definition leads you to default js Error type.

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -712,7 +712,6 @@ module.exports = {
 };
 ```
 
-
 ##### queryOptions
 
 Type: `String` or `Object`.
--- a/packages/mock/src/msw/index.ts
+++ b/packages/mock/src/msw/index.ts
@@ -107,7 +107,16 @@ export const ${handlerName} = (${isReturnHttpResponse && !isTextPlain ? `overrid
 
   const includeResponseImports =
     isReturnHttpResponse && !isTextPlain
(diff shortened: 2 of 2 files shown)


melloware:
Looks like the tests are failing?

melloware (approved):

PR #1276, opened 2024-03-21
fix(swr): error type is correct even when it's the same type as success type

### LINKED ISSUE #1256: Unknown error type when using duplicate reference

1. Update swagger to use same schema for 200 as another response code or default in the example app like: https://github.com/anymaniax/orval/compare/master...vonkanehoffen:orval:master
2. Run Orval to regenerate React Query handlers

## What happens?

The generated code now types React Query errors as `unknown`

It should type them as `Pets` right?

No errors. Output committed to forked repo above.
https://github.com/vonkanehoffen/orval/commit/fb4a59f98b80c7c6bd97f0a0e87c058778e6c3c1

## Any other comments?

I wish I could figure out why this is happening myself 😭 

**Operating System:** MacOS 14.2.1
**Package Version:** 6.25.0
**Browser Version:** N/A

AffeJonsson:
## Status

**READY**

## Description

Fixes #1256 
Problem was that the error type was getting filtered out as it was equal to the success type, resulting in error being unknown.

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

1. cd tests
2. yarn generate:swr
3. check generated/swr/errors/endpoints.ts
4. `export type CreateItemsQueryError = AxiosError<Item>`
5. compared to previous build, where it would be `export type CreateItemsQueryError = AxiosError<unknown>`

### DIFF
--- a/packages/core/src/getters/response.ts
+++ b/packages/core/src/getters/response.ts
@@ -37,6 +37,7 @@ export const getResponse = ({
     operationName,
     context,
     'void',
+    (type) => type.key.startsWith('2') + type.value,
   );
 
   const filteredTypes = contentType
--- a/packages/core/src/getters/res-req-types.ts
+++ b/packages/core/src/getters/res-req-types.ts
@@ -1,5 +1,7 @@
 import { keyword } from 'esutils';
+import type { ValueIteratee } from 'lodash';
 import uniqBy from 'lodash.uniqby';
+
 import {
   MediaTypeObject,
   ReferenceObject,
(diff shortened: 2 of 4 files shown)


packages/core/src/getters/response.ts
```diff
@@ -37,6 +37,7 @@ export const getResponse = ({
     operationName,
     context,
     'void',
+    (type) => type.key.startsWith('2') + type.value,
```
melloware, line 40: Can `type.key` ever be null?

melloware (approved):

packages/core/src/getters/response.ts
AllieJonsson: It shouldn't, it comes from `[statuscode: string]` in `ResponsesObject`, which gets picked up from 
```yaml
paths:
  /error:
    post:
      responses:
        '200': 
        # ^
```
So I think this is safe

PR #1277, opened 2024-03-22
fix: request bodies should take required property into account

### LINKED ISSUE #1111: Orval does not respect the "optional by default" `requestBody` for OpenAPI 3.0
`requestBody` is optional by default in OpenAPI 3.0[[1]](https://swagger.io/docs/specification/describing-request-body/). Orval does not respect that; it also does not respect `required: false`.

[BUG REPRO REPO](https://github.com/xbc5/orval-bug-repro)

1. exclude `required: true` for a `requestBody`;
2. generate the client;

## What happens?

The request body is not optional.

The request body to be optional.

No

## Any other comments?

The new Star Wars is overrated.

**Operating System:** Fedora 38
**Package Version:** 6.22.1
**Browser Version:** N/A

AffeJonsson:
## Status

**READY**

## Description

~~Fixes #1111~~

Takes the required property of request bodies into account. We till make it required by default, but setting it to required: false will make it optional.

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

1. yarn generate:swr
2. check tests/generated/swr/optional-request-body/endpoints.ts. Endpoints without a required property has the request body required. Endpoints with a required: false property has the request body as optional.

### DIFF
--- a/packages/core/src/getters/body.ts
+++ b/packages/core/src/getters/body.ts
@@ -1,7 +1,8 @@
 import { ReferenceObject, RequestBodyObject } from 'openapi3-ts/oas30';
 import { generalJSTypesWithArray } from '../constants';
+import { resolveRef } from '../resolvers';
 import { ContextSpecs, GetterBody, OverrideOutputContentType } from '../types';
-import { camel, sanitize } from '../utils';
+import { camel, isReference, sanitize } from '../utils';
 import { getResReqTypes } from './res-req-types';
 
 export const getBody = ({
@@ -53,6 +54,7 @@ export const getBody = ({
         context.output.override.components.requestBodies.suffix
       : camel(definition);
 
+  let isOptional = true;
   if (implementation) {
     implementation = sanitize(implementation, {
       underscore: '_',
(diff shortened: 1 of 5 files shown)


melloware (approved):

tests/specifications/optional-request-body.yaml
```diff
@@ -0,0 +1,223 @@
+openapi: '3.0.0'
```
soartec-lab, line 1: We will continue to maintain the tests, so we would like to keep them to a minimum as much as possible. Is it possible to slim down to a minimized case that only checks the request body?

soartec-lab (commented):
@AffeJonsson 
Hi, Thank you for great PR, I commented on one consultation, so please check it.

tests/specifications/optional-request-body.yaml
AllieJonsson: Absolutely, I'll fix!

anymaniax:
I understand that in the specification it’s optional per default but 99% if you have a request body you wanted required no?

AffeJonsson:
That's probably true. Maybe it would be better if it's required as default. I'll change that, but at least with this change it'll take the required prop into account

packages/core/src/getters/body.ts
```diff
@@ -61,6 +63,17 @@ export const getBody = ({
...
       es5IdentifierName: true,
     });
+    if (isReference(requestBody)) {
+      const { schema: bodySchema } = resolveRef<RequestBodyObject>(
+        requestBody,
+        context,
+      );
+      if (bodySchema.required !== undefined) {
+        isOptional = !bodySchema.required;
+      }
```
anymaniax, line 73: What is the value of required if we don’t set it?

anymaniax (approved):

soartec-lab (approved):
Thank you.

Issue #1279, opened 2024-03-24
Allow delay to be configurable for tests


huwshimi:
## What are the steps to reproduce this issue?

1. Generate the API with mocks.
2. Create tests using the mocks.

## What happens?

Depending on the delay you set in your config, either your tests are very slow or you have to have a very short delay when simulating the API in a browser.

## What were you expecting to happen?

Have a way to either set the delay separately for tests, or set the delay at runtime.

## Any logs, error output, etc?

…

## Any other comments?

…

## What versions are you using?

**Operating System:** …
**Package Version:** …
**Browser Version:** …

PR #1281, opened 2024-03-25
feat: support runtime delay functions

huwshimi:
## Status

**READY**

## Description

Add support for generating mock delay times at runtime.

Fixes: #1279

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

1. In root run `yarn build`.
2. cd samples/react-query/basic.
3. Open orval.config.ts and replace `mock: true,` with:
```typescript
      mock: {
        type: 'msw',
        delay: () => (process.env.NODE_ENV === 'test' ? 10 : 1000),
        delayCalcRuntime: true,
      },
```
4. Run `yarn generate-api`.
5. Open `src/api/endpoints/petstoreFromFileSpecWithTransformer.msw.ts`.
6. See that the delays are now `await delay((() => (process.env.NODE_ENV === 'test' ? 10 : 1e3))());`

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -300,6 +300,13 @@ Default Value: `1000`.
 
 Use to specify the delay time for the mock. It can either be a fixed number or a function that returns a number.
 
+#### delayCalcRuntime
+
+Type: `boolean`.
+
+Gives you the possibility to have functions that are passed to `delay` to be
+executed at runtime rather than at build time.
+
 #### useExamples
 
 Type: `Boolean`.
@@ -1009,6 +1016,13 @@ module.exports = {
   },
 };
 ```
+
+#### delayCalcRuntime
+
+Type: `boolean`.
+
+Gives you the possibility to have functions that are passed to `delay` to be
+executed at runtime rather than at build time.
 
 ##### arrayMin
 
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -300,7 +300,7 @@ Default Value: `1000`.
 
 Use to specify the delay time for the mock. It can either be a fixed number or a function that returns a number.
 
-#### delayMockFunction
+#### delayFunctionLazyExecute
 
 Type: `boolean`.
 
@@ -1017,7 +1017,7 @@ module.exports = {
 };
 ```
 
-#### delayMockFunction
+#### delayFunctionLazyExecute
 
 Type: `boolean`.
 
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -300,12 +300,12 @@ Default Value: `1000`.
 
 Use to specify the delay time for the mock. It can either be a fixed number or a function that returns a number.
 
-#### delayCalcRuntime
+#### delayMockFunction
 
 Type: `boolean`.
 
 Gives you the possibility to have functions that are passed to `delay` to be
-executed at runtime rather than at build time.
+executed at runtime rather than when the mocks are generated.
 
 #### useExamples
 
@@ -1017,12 +1017,12 @@ module.exports = {
 };
 ```
 
-#### delayCalcRuntime
+#### delayMockFunction
 
 Type: `boolean`.
 
 Gives you the possibility to have functions that are passed to `delay` to be
-executed at runtime rather than at build time.
+executed at runtime rather than when the mocks are generated.
 
 ##### arrayMin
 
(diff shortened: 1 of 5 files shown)


soartec-lab:
Thank you for made this PR. I'll review it.

docs/src/pages/reference/configuration/output.md
```diff
@@ -300,6 +300,13 @@ Default Value: `1000`.
 
 Use to specify the delay time for the mock. It can either be a fixed number or a function that returns a number.
 
+#### delayCalcRuntime
+
+Type: `boolean`.
+
+Gives you the possibility to have functions that are passed to `delay` to be
+executed at runtime rather than at build time.
```
soartec-lab: To be more precise, is it "generated timing" rather than build time?

```suggestion
executed at runtime rather than at code generated timing.
```

packages/core/src/types.ts
```diff
@@ -230,6 +230,9 @@ export type GlobalMockOptions = {
   useExamples?: boolean;
   // This is used to set the delay to your own custom value
   delay?: number | (() => number);
+  // This is used to execute functions that are passed to the 'delay' argument
+  // at runtime rather than build time.
+  delayCalcRuntime?: boolean;
```
soartec-lab: Could you please add a new regression test case to the following file:

https://github.com/anymaniax/orval/blob/master/tests/configs/default.config.ts

soartec-lab (commented):
@huwshimi

Good job. I have made some comments so please check them.

docs/src/pages/reference/configuration/output.md
huwshimi: Or would `executed at runtime rather than when the mocks are generated.` be better?

packages/core/src/types.ts
huwshimi: Done!

docs/src/pages/reference/configuration/output.md
melloware: I like the second one `executed at runtime rather than when the mocks are generated.`

melloware (changes requested):

docs/src/pages/reference/configuration/output.md
melloware: i am not sure about this name `delayCalcRuntime` for a name shouldn't it be `delayMockFunction` or something like that?

melloware (changes requested):

docs/src/pages/reference/configuration/output.md
huwshimi: I've updated the comment and the variable name.

soartec-lab: Even before the change, I was able to specify a function, but the function would be executed when it was automatically generated, so I would like to have the delay function executed when a mock is used, right?
`Runtime` is certainly misleading, so it is inappropriate to use it in the sense of "when a function is executed." Since the execution is delayed relative to the existing automatic generation, what about `delayFunctionLazyExecute`? I think it will become more clear.

Or how about changing it so that it doesn't run at the timing of automatic generation, and instead runs the delay function when mock is used in all cases? I think this idea is good because it eliminates the complexity of adding options and does not affect existing users.

melloware: I like that idea one less configuration.  Really wouldn't it be fine to always run at runtime instead of buildtime?

huwshimi: I'm happy to change this to remove the extra option and always call the function at runtime. However, this would be a breaking change as the previous implementation allows for functions which reference external variables etc. whereas inserting the function into the mock requires it to be self contained.

Note: I chose the name runtime as this is how the docs refer to [other functions](https://orval.dev/reference/configuration/output#operations) that delay the execution of the function e.g. "Which can take a function or the value directly. The function will be executed at runtime."

melloware: Ok leave both functions.  It would be a breaking change.

soartec-lab: I hadn't thought of breaking changes yet. thank you for let me know. What about the name?

> what about delayFunctionLazyExecute? I think it will become more clear.

huwshimi: Happy to change it to that, or even `delayFunctionLazy` might be descriptive enough (I'm happy either way)?

soartec-lab: Hmm. I prefer `delayFunctionLazyExecute`.

huwshimi: > Hmm. I prefer `delayFunctionLazyExecute`.

Done!

soartec-lab (approved):
Thanks!

docs/src/pages/reference/configuration/output.md
soartec-lab: thank you for your great work 👍

soartec-lab:
@melloware 
I think it's good, but could you review it again?

melloware (approved):

Issue #1280, opened 2024-03-24
Only 2xx mock responses are generated


huwshimi:
## What are the steps to reproduce this issue?

1. Create an openapi spec with 400 or 500 responses e.g.

```yaml
openapi: '3.0.0'
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
servers:
  - url: http://petstore.swagger.io/v1
paths:
  /pets:
    get:
      summary: List all pets
      operationId: listPets
      tags:
        - pets
      parameters:
        - name: limit
          in: query
          description: How many items to return at one time (max 100)
          required: false
          schema:
            type: string
      responses:
        '200':
          description: A paged array of pets
          headers:
            x-next:
              description: A link to the next page of responses
              schema:
                type: string
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Labradoodle'
        '201':
          description: A paged array of pets
          headers:
            x-next:
              description: A link to the next page of responses
              schema:
                type: string
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Dachshund'
        '400':
          description: A paged array of pets
          headers:
            x-next:
              description: A link to the next page of responses
              schema:
                type: string
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Cat'
        default:
          description: unexpected error
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Error'
components:
  schemas:
    Labradoodle:
      type: object
      required: ['cuteness']
      properties:
        cuteness:
          type: integer
    Dachshund:
      type: object
      required: ['length']
      properties:
        length:
          type: integer
    Cat:
      type: object
      required: ['type']
      properties:
        petsRequested:
          type: integer
        type:
          type: string
          enum:
            - cat
    Error:
      type: object
      required:
        - code
        - message
      properties:
        code:
          type: integer
          format: int32
        message:
          type: string
```

3. Generate mocks using orval

## What happens?

Orval only generates an http response with a 200 status (which picks from an array of all 2xx so these can't be used independently) and does not generate any code for error responses:

```array element
export const getListPetsMock = () =>
  faker.helpers.arrayElement([
    { cuteness: faker.number.int({ min: undefined, max: undefined }) },
    { length: faker.number.int({ min: undefined, max: undefined }) },
  ]);

export const getSwaggerPetstoreMSW = () => [
  http.get("*/pets", async () => {
    await delay(1000);
[...]

PR #1284, opened 2024-03-26
feat: generate mocks for all responses

huwshimi:
## Status

**READY**

## Description

Generate mocks for all responses defined in an openapi spec.

Fixes: #1280.

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

1. cd to `samples/react-query/status-codes`
2. Run `yarn generate-api`.
3. Open `samples/react-query/status-codes/src/api/endpoints/petstoreFromFileSpecWithTransformer.msw.ts`
4. See that mocks and http handlers are generated for each response.
5. At the bottom see that only `getListPetsMockHandler()` is included in the array of handlers (this is so that when passing the handlers to the worker there is only one http response for each path).


samples/react-query/status-codes/petstore.yaml
huwshimi, line 1: I wasn't sure if I should commit this sample with multiple responses defined.

melloware: Well I think adding samples/tests is always a good idea. Just my two cents.

melloware (dismissed):

soartec-lab:
I'll check it in order 👍

soartec-lab:
@huwshimi 
Hi, this PR is nice idea.
I thought it was overkill if we only want to get a simple mock.
For example, is it okay to output all statuses only when `true` is specified with an option like `generateEachHttpStatus: boolean` ?

samples/react-query/status-codes/petstore.yaml
soartec-lab: I think a sample project is unnecessary. As I commented elsewhere, I want to control this response as an option, and by default I think it is sufficient to output only the minimum success status, so it is sufficient to check it in the test rather than in the sample project.

huwshimi: I've removed the sample project and added a config test.

huwshimi:
> @huwshimi Hi, this PR is nice idea. I thought it was overkill if we only want to get a simple mock. For example, is it okay to output all statuses only when `true` is specified with an option like `generateEachHttpStatus: boolean` ?

Making it optional sounds like a good idea. I've modified this to add that config option.

melloware (dismissed):

packages/mock/src/msw/index.ts
```diff
@@ -129,3 +137,73 @@ export const ${handlerName} = (${isReturnHttpResponse && !isTextPlain ? `overrid
...
+    response.contentTypes,
+  );
+
+  const mockImplementations = [baseDefinition.implementation.function];
+  const handlerImplementations = [baseDefinition.implementation.handler];
+  const imports = [...baseDefinition.imports];
+
+  if (
+    generatorOptions.mock &&
+    'generateEachHttpStatus' in generatorOptions.mock &&
```
soartec-lab: Is this existence check unnecessary?

packages/mock/src/msw/index.ts
```diff
@@ -129,3 +137,73 @@ export const ${handlerName} = (${isReturnHttpResponse && !isTextPlain ? `overrid
...
+
+  const route = getRouteMSW(
+    pathRoute,
+    override?.mock?.baseUrl ?? (!isFunction(mock) ? mock?.baseUrl : undefined),
+  );
+
+  const handlerName = `get${pascal(operationId)}MockHandler`;
+  const getResponseMockFunctionName = `get${pascal(operationId)}ResponseMock`;
+
+  const baseDefinition = generateDefinition(
```
soartec-lab, line 157: This is a refactoring consultation.
Is it possible to increase the generated http status with options and then loop the generation targets as shown below?

```typescript
const isGenerateEachHttpStatus = generatorOptions.mock && 'generateEachHttpStatus' in generatorOptions.mock && generatorOptions.mock.generateEachHttpStatus;
generateStatuses = isGenerateEachHttpStatus ? [...response.types.success, ...response.types.errors] : [response.types.success]

generateStatuses.forEach(
  (statusResponse) => {}
)
```

With the current implementation, duplicate handlers and mock definition for success will be created.

soartec-lab (commented):
Thaks! I made a comment, please check it.

packages/mock/src/msw/index.ts
huwshimi: `.mock` can be an object or a function, this is probably more readable if it uses `isObject(generatorOptions.mock)` so I'll change it to that.

packages/mock/src/msw/index.ts
huwshimi: There's a difference between the "base" and other success responses as I didn't want to introduce a breaking change. The "base" response behaves like the response generated before my change, which picks a random success response each time e.g.

```typescript
export const getListPetsResponseMock = (
  overrideResponse: any = {},
): Okay | Accepted | Success =>
  faker.helpers.arrayElement([ // pick a random "success" response
    Array.from( // the "Okay" type (an array of pets)
      { length: faker.number.int({ min: 1, max: 10 }) },
      (_, i) => i + 1,
    ).map(() =>
      faker.helpers.arrayElement([
        {
          breed: faker.helpers.arrayElement(['Labradoodle'] as const),
         ...
        },
        ...
      ]),
    ),
    { // The "Accepted" response
      code: faker.number.int({ min: undefined, max: undefined }),
      message: faker.word.sample(),
      ...overrideResponse,
    },
    { // The "Success" response
      code: faker.number.int({ min: undefined, max: undefined }),
      message: faker.word.sample(),
      ...overrideResponse,
    },
  ]);
```

Whereas the other mocks that are generated only return the specific response for a status:

```typescript
export const getListPetsResponseMock200 = (overrideResponse: any = {}): Okay =>
  Array.from(
    { length: faker.number.int({ min: 1, max: 10 }) },
    (_, i) => i + 1,
  ).map(() =>
    faker.helpers.arrayElement([
      {
        breed: faker.helpers.arrayElement(['Labradoodle'] as const),
        ...
      },
      ...
    ]),
  );

export const getListPetsResponseMock202 = (
  overrideResponse: any = {},
): Accepted => ({
  code: faker.number.int({ min: undefined, max: undefined }),
  message: faker.word.sample(),
  ...overrideResponse,
});

export const getListPetsResponseMock2xx = (
  overrideResponse: any = {},
): Success => ({
  code: faker.number.int({ min: undefined, max: undefined }),
  message: faker.word.sample(),
  ...overrideResponse,
});
```

I could change it to return a single success response for the "base" response if you would be happy with a breaking change, though it could be difficult to know which response to return (I guess it could be the lowest 2xx number defined in the responses).

soartec-lab: @huwshimi 

that's sounds good. I had missed this.
In other words, does this mean that when multiple success responses are defined, the mocked types are different, so it is not a duplicate, right?

huwshimi: > @huwshimi
> 
> that's sounds good. I had missed this. In other words, does this mean that when multiple success responses are defined, the mocked types are different, so it is not a duplicate, right?

That's correct.

soartec-lab (approved):
@huwshimi 
Thank you for great job !

PR #1303, opened 2024-04-11
feat(zod): add option to setup strict mode

anymaniax:
## Status

**READY**

## Description

Zod add strict option to each object

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -96,6 +96,7 @@ export type NormalizedOverrideOutput = {
   query: NormalizedQueryOptions;
   angular: Required<AngularOptions>;
   swr: SwrOptions;
+  zod: NormalizedZodOptions;
   operationName?: (
     operation: OperationObject,
     route: string,
@@ -129,6 +130,7 @@ export type NormalizedOperationOptions = {
   query?: NormalizedQueryOptions;
   angular?: Required<AngularOptions>;
   swr?: SwrOptions;
+  zod?: NormalizedZodOptions;
   operationName?: (
     operation: OperationObject,
     route: string,
@@ -329,6 +331,7 @@ export type OverrideOutput = {
   query?: QueryOptions;
   swr?: SwrOptions;
   angular?: AngularOptions;
+  zod?: ZodOptions;
   operationName?: (
     operation: OperationObject,
     route: string,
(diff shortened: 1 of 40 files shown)


anymaniax:
I don't understand why formatting doesn't work. Locally all seems ok

anymaniax:
I needed to update deps 😅

soartec-lab:
Regarding the formatting failure, a formatting error occurred in my environment as well. Perhaps the version of prettier in your environment was outdated?
And now, this time the `zod` test is failing. This is an intended test failure, so it seems that the test case needs to be updated.

packages/core/src/types.ts
```diff
@@ -352,6 +355,26 @@ export type NormalizedHonoOptions = {
   handlers?: string;
 };
 
+export type ZodOptions = {
+  strict?: {
```
soartec-lab, line 359: Although it is not required for this PR, I think it would be better if the guide updates are done together so that users can recognize them.

packages/zod/src/zod.test.ts
```diff
@@ -34,7 +34,11 @@ const queryParams: ZodValidationSchemaDefinitionInput = {
 describe('parseZodValidationSchemaDefinition', () => {
   describe('with `override.coerceTypes = false` (default)', () => {
     it('does not emit coerced zod property schemas', () => {
-      const parseResult = parseZodValidationSchemaDefinition(queryParams);
+      const parseResult = parseZodValidationSchemaDefinition(
```
soartec-lab, line 37: I am also updating the unit tests, but at the same time, I thought it would be a good idea to add a test that the options added this time work using the automatically generated source code of `zod` in the `tests` directory.

soartec-lab (commented):
@anymaniax 

it's nice update 👍
CI seems to have passed all checks. And, please check as I have given some opinions.

packages/zod/src/zod.test.ts
anymaniax: I am not sure to understand. I added a test for the new option too. And you want to test the generated files also right? If so indeed would be great

packages/core/src/types.ts
anymaniax: You mean update the doc?

soartec-lab: Yes, It seems good to add the `output.override.zod` item.

https://orval.dev/reference/configuration/output#override

anymaniax: I updated the doc to add the output reference

packages/zod/src/zod.test.ts
soartec-lab: Yes, it is a good idea to continuously test that the automatically generated source code does not break when the options added this time are specified. And it's now in the `tests` directory.

packages/core/src/types.ts
soartec-lab: Thanks. That's great !

soartec-lab: However, it seems like there is a formatting error again.

packages/zod/src/zod.test.ts
anymaniax: I added a test with the option

melloware (approved):

packages/zod/src/zod.test.ts
soartec-lab: Thanks!

soartec-lab (approved):
LGTM!
If all CI is successful I'll merge this.

PR #1310, opened 2024-04-15
feat(zod): add option coerce

anymaniax:
## Status

**READY**

## Description

Add option coerce to zod client

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -947,6 +947,13 @@ module.exports = {
             header: true,
             body: true
           },
+          coerce: {
+            response: true,
+              query: true,
+              param: true,
+              header: true,
+              body: true
+          },
         },
       },
     },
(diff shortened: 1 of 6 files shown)


melloware (approved):

soartec-lab:
@anymaniax @melloware 

Although it is an undocumented specification, it is actually possible to configure it so that `coerce` is used even when the `coerceTypes` option is specified.

https://github.com/anymaniax/orval/blob/master/packages/zod/src/index.ts#L327-L329

However, you cannot specify parameters and it will apply to all objects. I think the options supported this time are good because they can be set in more detail.
From an overall perspective, there are multiple options with the same settings, so I would like to unify them. My suggestion is to deprecate `coerceTypes` and remove it in the next two versions, what do you think?

anymaniax:
@soartec-lab I already added the deprecated warning on `coerceTypes`. if you use it or in the documentation directly

soartec-lab:
@anymaniax 
Oops, I deprecated it on this PC. I am careful to remove it in the next two versions or later.

Issue #1327, opened 2024-04-23
Zod: add to @orval/zod support of anyOf/allOf/oneOf in requestBody


danyadanch:
## What are the steps to reproduce this issue?

1. Use schema like this:
```json
{
    "openapi": "3.1.0",
    "info": {
        "title": "Luca API",
        "version": "0.0.1",
        "description": ""
    },
    "paths": {
        "/api/sheets/transactions/": {
            "post": {
                "operationId": "createTransaction",
                "summary": "Create",
                "parameters": [],
                "responses": {
                    "200": {
                        "description": "OK"
                    }
                },
                "tags": [
                    "sheets/transactions"
                ],
                "requestBody": {
                    "content": {
                        "application/json": {
                            "schema": {
                                "anyOf": [
                                    {
                                        "$ref": "#/components/schemas/CreateTransactionSchema"
                                    },
                                    {
                                        "$ref": "#/components/schemas/CreateTransactionTransferSchema"
                                    }
                                ],
                                "title": "Schema"
                            }
                        }
                    },
                    "required": true
                }
            }
        }
    },
    "components": {
        "schemas": {
            "CreateTransactionSchema": {
                "properties": {
                    "type": {
                        "$ref": "#/components/schemas/TransactionType"
                    },
                    "amount": {
                        "anyOf": [
                            {
                                "type": "number"
                            },
                            {
                                "type": "string"
                            }
                        ],
                        "title": "Amount"
                    },
                    "account_id": {
                        "format": "uuid",
                        "title": "Account Id",
                        "type": "string"
                    },
                    "accounting_item": {
                        "anyOf": [
                            {
                                "format": "uuid",
                                "type": "string"
                            },
                            {
                                "type": "null"
                            }
                        ],
                        "title": "Accounting Item"
                    },
                    "counterparty": {
                        "anyOf": [
                            {
                                "format": "uuid",
                                "type": "string"
                            },
                            {
                                "type": "null"
                            }
                        ],
                        "title": "Counterparty"
                    },
                    "commentary": {
                        "anyOf": [
                            {
                                "type": "string"
                            },
                            {
                                "type": "null"
                            }
                        ],
                        "default": "",
                        "title": "Commentary"
                    },
                    "payment_date": {
                        "format": "date",
                        "title": "Payment Date",
                        "type": "string"
                    },
                    "accrual_date": {
                        "format": "date",
                        "title": "Accrual Date",
                        "type": "string"
                    }
                },
                "required": [
[...]

PR #1330, opened 2024-04-25
refactor(zod): unify implementation

anymaniax:
## Status

**READY**

## Description

I did a refactor of the zod client implementation to simplify everything and unify the way we generate the code to also use the `generateZodValidationSchemaDefinition` function even for the root schema.

I improved the way duplicate was handled for consts but will cause a `breaking change` for the name since for now we always add a number at the end. If think it's a needed breaking change since it's a bit weird to add always a number in the name. I also transformed that number to a named number to unify with other clients.

it should also solve this issue #1327

it's also a first step to be able to generate a zod schema for each components and not only based on paths. Would be a good added value for this client I think

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -161,8 +161,8 @@ const generateZodValidationSchemaDefinition = (
         const separator = schema.allOf
           ? 'allOf'
           : schema.oneOf
-          ? 'oneOf'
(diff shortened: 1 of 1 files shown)


melloware (approved):
Nice Refactor!

soartec-lab (approved):
Thanks. It's a good improvement.

PR #1333, opened 2024-04-26
orval/query - override shouldExportHttpClient / shouldExportQueryKey flags introduced

mironbalcerzak:
## Status

**READY**

## Description

**Feature**
New overrides for react-query:
- shouldExportQueryKey - flag enabling/disabling QueryKey export
- shouldExportHttpClient - flag enabling/disabling "raw" Axios client export

**Bugfix**
seems like react-query/overrides did not work - they were defaulting to "true" :)

## Todos

- [ ] Tests
- [x] Documentation
- [X] Changelog Entry (unreleased)

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -410,6 +410,8 @@ export type NormalizedQueryOptions = {
   queryOptions?: NormalizedMutator;
   mutationOptions?: NormalizedMutator;
   shouldExportMutatorHooks?: boolean;
+  shouldExportHttpClient?: boolean;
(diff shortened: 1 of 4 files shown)


anymaniax (approved):
Seems good thanks

tlqvst:
After this change, any overriden query will not export its query key unless this is explicitly set to true. Is this intended? This is a breaking change for us, and we have a lot of overriden queries where we want query keys to be exported by default.

melloware:
@mironbalcerzak see discussion: https://github.com/anymaniax/orval/discussions/1436

PR #1334, opened 2024-04-27
orval/core - generating factory method for schema interfaces

### LINKED ISSUE #1331: Feature Request: create factory method for instantiating OpenAPI models

Using any schema, do generate models, we get following (example):
```
export interface User {
  name: string;
  age: number;
  address: Address;
}

export interface Address {
  street: string;
  city: string;
}
```

Beside interface, getting a factory method (naive approach)
```
export function createUser() {
  return {
    name: '',
    age: 0,
    address: createAddress(),
  };
}

export function createAddress() {
  return {
    street: '',
    city: '',
  };
}
```
Naturally, all primitive types should be handled, re complex (other schema types) - we should discuss on how to handle "relationship"

## Motivation

Initializing state in react components is very tedious work

**Package Version:** 6.27.1

mironbalcerzak:
## Status

**WIP**

## Description
Fix  https://github.com/anymaniax/orval/issues/1331

This PR is providing simple implementation.
Opening for discussion.
Bugs? Possible - definitely "stackoverflow" in case of self referencing or cycle

## Todos

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/core/src/generators/imports.ts
+++ b/packages/core/src/generators/imports.ts
@@ -5,6 +5,7 @@ import {
   GeneratorMutator,
   GeneratorVerbOptions,
   GetterPropType,
+  OutputModelFactoryMethodsMode,
 } from '../types';
 import { camel, upath } from '../utils';
 
@@ -14,12 +15,16 @@ export const generateImports = ({
   isRootKey,
   specsName,
   specKey: currentSpecKey,
+  factoryMethodOutput,
+  factoryMethodPrefix
 }: {
   imports: GeneratorImport[];
   target: string;
   isRootKey: boolean;
   specsName: Record<string, string>;
   specKey: string;
+  factoryMethodOutput?: (typeof OutputModelFactoryMethodsMode)[keyof typeof OutputModelFactoryMethodsMode];
+  factoryMethodPrefix?: string;
 }) => {
   if (!imports.length) {
     return '';
@@ -47,9 +52,14 @@ export const generateImports = ({
         } } from \'./${upath.join(path, camel(name))}\';`;
       }
 
-      return `import { create${name}, ${name}${
+      let mainImport = `import ${!values ? 'type ' : ''}{ ${name}${
         alias ? ` as ${alias}` : ''
-      } } from \'./${camel(name)}\';`;
+      }${factoryMethodOutput == OutputModelFactoryMethodsMode.SINGLE ? `${factoryMethodPrefix}${name}` : ''} } from \'./${camel(name)}\';`;
+      if (factoryMethodOutput == OutputModelFactoryMethodsMode.SPLIT) {
+        let factoryMethodImport = `import { ${factoryMethodPrefix}${name} } from \'./${camel(name)}.factory\';`;
+        mainImport += '\n' + factoryMethodImport;
+      }
+      return mainImport;
     })
     .join('\n');
 };
--- a/packages/core/src/generators/imports.ts
+++ b/packages/core/src/generators/imports.ts
@@ -47,7 +47,7 @@ export const generateImports = ({
         } } from \'./${upath.join(path, camel(name))}\';`;
       }
 
-      return `import ${!values ? 'type ' : ''}{ ${name}${
+      return `import { create${name}, ${name}${
         alias ? ` as ${alias}` : ''
       } } from \'./${camel(name)}\';`;
     })
--- a/packages/core/src/generators/interface.ts
+++ b/packages/core/src/generators/interface.ts
@@ -53,6 +53,7 @@ export const generateInterface = ({
   } else {
     model += `export type ${name} = ${scalar.value};\n`;
   }
+  model += `export function create${name}(): ${name} ${scalar.factoryMethodValue}\n`;
 
   // Filter out imports that refer to the type defined in current file (OpenAPI recursive schema definitions)
   const externalModulesImportsOnly = scalar.imports.filter(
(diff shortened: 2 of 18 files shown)


packages/core/src/generators/interface.ts
```diff
@@ -53,6 +53,7 @@ export const generateInterface = ({
   } else {
     model += `export type ${name} = ${scalar.value};\n`;
   }
+  model += `export function create${name}(): ${name} ${scalar.factoryMethodValue}\n`;
```
anymaniax: You will add this in the model file here. It's a bit weird no?

mironbalcerzak: Yeah... was thinking whether it should be in separate file or not. I've decided to leave it with model. Anyone who has imported a model, can either instantiate it through "new", or right away call a factory method - which increases the chances that random dev will learn about it and use it.

People are lazy and do not read the code - especially the one that's generated.

soartec-lab:
Hi, @mironbalcerzak 
I'll review it. Let me know when you change the status to Ready.

mironbalcerzak:
@soartec-lab  - ive been using this branch in few projects of mine over last months - seems ok.
How can we progress so it ends in "master"?

melloware:
It looks like the tests are failing?

mironbalcerzak:
@melloware, yeah, i am aware and that will be fixed shortly - i meant on a broader scale - should i implement/change something in PR?

melloware:
i think once its fixed and passing we will re-review.

soartec-lab:
@mironbalcerzak 
Thank you for mada good enhance. I'll review this !

packages/core/src/generators/imports.ts
```diff
@@ -47,7 +47,7 @@ export const generateImports = ({
         } } from \'./${upath.join(path, camel(name))}\';`;
       }
 
-      return `import ${!values ? 'type ' : ''}{ ${name}${
+      return `import { create${name}, ${name}${
```
soartec-lab: Since `create` exists in the http status, I want to change the function name to the `new` prefix to avoid confusion.

```suggestion
      return `import { new${name}, ${name}${
```

packages/core/src/generators/interface.ts
soartec-lab: Factory methods are just noise for people who don't need them, so I'd like to be able to choose whether to generate them with a property.
For example, `generateFactoryMethod: boolean`.

And by making it a property, people who want to use it will be aware that it will be generated. Therefore, you will be able to notice it even if you separate the file like `.factory.ts`, right?

soartec-lab (commented):
@mironbalcerzak 
I made some comments 👍

melloware:
Where are we at with this PR?

mironbalcerzak:
@melloware  - sorry for late reply, very hectic time. keep it open if you may, i am slowly working on it to make it good enough and eventually i will notify you guys that it's ready for proper review

melloware:
No problem.

melloware:
Any update on this?

mironbalcerzak:
hey @melloware @soartec-lab @anymaniax 
been a while! ive recreated the factory methods in latest codebase. took me a little bit longer than i had planned.

ive pushed changes to the branch from this PR.

please review and let me know your thoughts :)

hopefully we will merge it this time.

------

### 1. Configuration

The feature is controlled via the `factoryMethods` object in the `output` section of the Orval configuration.

- **generate** (`boolean`): Enables or disables factory generation. Defaults to `false`.
- **prefix** (`string`): Defines the prefix for the function name. Defaults to `'create'`.
- **location** (`'inline-with-model' | 'separate-file' | 'combined-separate-file'`):
  - `inline-with-model`: Appends the factory function to the model file.
  - `separate-file`: Creates a `.factory.ts` file next to each model.
  - `combined-separate-file`: Aggregates all factories into a single `factoryMethods.ts` file.
- **optionalPropertyStrategy** (`'include' | 'omit'`): Determines whether optional schema properties are included in the default factory output. Defaults to `'include'`.

### 2. Cycle Detection

To prevent infinite recursion during object instantiation, we implemented a generation-time circular reference detector.

- The logic tracks the lineage of schemas during the build process.
- If a schema is detected as a descendant of itself, the factory breaks the chain by injecting a type-cast empty object: `{} as <Type>`.
- This ensures that complex graphs (e.g., `Parent -> Child -> Parent`) generate valid code without runtime stack overflows.

### 3. Property Visibility (readOnly / writeOnly)

The generator intelligently handles OpenAPI `readOnly` and `writeOnly` flags to generate appropriate payload structures:

- **Required properties:** Always included in the factory output, regardless of their visibility flags.
- **Optional `readOnly` properties:** Always omitted from the factory output, as they are typically generated by the server (e.g., database IDs, creation timestamps).
- **Optional `writeOnly` properties:** Always included in the factory output (even if `optionalPropertyStrategy` is set to `'omit'`), as they are typically necessary for client payloads (e.g., passwords).

soartec-lab:
It's been a while @mironbalcerzak. I've been eagerly awaiting this response. Thank you. I have one suggestion. Is it possible to include `mode` and `outputDirectory` in the `location` of `factoryMethods` so that output can be placed in an arbitrary directory?

```ts
factoryMethods: {
  mode: 'inline-with-model' | 'separate-file' | 'combined-separate-file',
  prefix: 'create',
  outputDirectory: 'src/gen/test/factory',
}
```

For example, there are cases where you want to use a factory only for generating test data and keep it separate from the production code.

mironbalcerzak:
@soartec-lab sorry it took so long. ive opened new PR and ive added (your very valid) suggestions

Issue #845, opened 2023-04-30
TS Error: argument of type 'number' is not assignable to parameter of type 'string | Blob'


chiptus:
Wow, this tool is amazing, it has everything we need to make types generation automatic (until now we did by hand when needed)

## What are the steps to reproduce this issue?
I have this schema:

```yaml
/endpoints:
    post:
      consumes:
      - multipart/form-data
      parameters:
      - collectionFormat: csv
        description: List of tag identifiers to which this environment(endpoint) is
          associated
        in: formData
        items:
          type: integer
        name: TagIds
        type: array
```
## What happens?
![Screenshot_20230430_184400](https://user-images.githubusercontent.com/1381655/235351169-1267883e-658c-484b-a909-715843f4cf72.png)

…

## What were you expecting to happen?
no error. so code should probably be:

```ts
if (endpointCreateBody.TagIds !== undefined) {
    endpointCreateBody.TagIds.forEach((value) =>
      formData.append('TagIds', String(value))
    );
  }
```

## Any logs, error output, etc?
`Argument of type 'number' is not assignable to parameter of type 'string | Blob'`

## What versions are you using?

**Operating System:** ubuntu
**Package Version:** v6.15.0
**Browser Version:** not in browser

jamalsoueidan:
I changed all boolean to string and integer or number to string, I convert them back to number when it hits the endpoint.
I don't know if their is any solution to the problem.

```js
export const NumberOrStringType = z
  .union([z.number(), z.string()])
  .transform((value) =>
    typeof value === "string" ? parseInt(value, 10) : value
  );
```

PR #1336, opened 2024-04-29
fix(form-data): handling array items

BijanRegmi:
## Status

**READY**

## Description

Fixes: #845 

Formdata only supports either string or blob types. So items of an array should be converted to string before adding them to the form.

## Todos

- [x] Tests
- [ ] Documentation


melloware (approved):

hariacharya80 (approved):
Nice catch!

melloware:
Looks like a formatting issue. Can you run `npm run format` please?

BijanRegmi:
> I see that the approved changes sets the data as a JSON string when `itemSchema.type === 'object'`, This gets rid of the error, but doesn't realy make it form-data. Shouldn't each of the properties of the object also be iterated? And also add an array suffix to the key?
> 
> So that the end result looks something like this for the form data key: `<arrayFieldName>[<arrayIndex>].<objectPropertyName>`

I think it's a personal opinion kind of thing. Surely `<key1>.<key2>` makes more sense for formdata but in most of the cases when transferring json data in formdata most tend to `JSON.stringify` the data and set `content-type` of the data to `application/json`

melloware (approved):

melloware:
@Ultre00 do you want to review the latest changes?

Ultre00:
Isn't this change only related to formdata anyway? I don't see how setting `content-type` to `application/json` solves that issue as there was never a problem with that `content-type`. In my case, I couldn't use `application/json` as I want to send some files in the same request, so I am using `multipart/form-data`.  I know there isn't really a standard defined on how to handle nested objects/arrays. What I suggested is just what .NET expects, I guess.

I don't think there is anything else to review. Only Main is merged into this branch.

melloware:
@Ultre00 @BijanRegmi where are we with this PR?  Is there more work to be done?

BijanRegmi:
It's ready to be merged

PR #1341, opened 2024-04-30
fix: valid syntax when anyOf is present on querystring

arthurfiorette:
## Status

**READY**

## Description

Fix https://github.com/anymaniax/orval/issues/884.

Tests were also added. I'm not familiar enough with the orval codebase (for now haha), but I think this fix is enough. Also tested on our production codebase and can confirm the problem was fixed.

## Todos

- [x] Tests
- [x] Documentation


arthurfiorette:
What should I do to create add a changelog entry?

melloware:
That I don't know I have never done one so I think your PR is ok for now.

anymaniax (approved):

PR #1353, opened 2024-05-03
feat: add new client `fetch`

soartec-lab:
## Status

**READY**

## Description
Follow up for #1387

I added a new `fetch`client.
Becouse, since the `fetch` API is mature, it has the advantage of reducing the bundle size of the application compared to using `Axios`, and I feel that this is sufficient in some cases.
Furthermore, we think it can be used to select an HTTP client with `tanstack query` or `swr`, or when calling an API from a server-side framework.

In this PR, i implemented a client using the minimal `fetch` API. For example, customization by specifying mutator is not yet supported. However, I still think it is practical enough and would like to submmit it on first. I'll also submit a sample application in a subsequent PR.

Ref: https://developer.mozilla.org/en-US/docs/Web/API/fetch

## Related PRs

none

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

I added a test case, so it should pass.


anymaniax:
the params are passed to the body. Is it intended?

soartec-lab:
thanks. I'll check later 👍

anymaniax:
@soartec-lab it's what I use in my project if you need https://gist.github.com/anymaniax/44c1331a5643081a82da070e45f405f0

soartec-lab:
@anymaniax 
wonderful. Thank you for sharing 🙌 

> the params are passed to the body. Is it intended?

This is by design. Referring to this, I interpret that the request parameters are to be specified in the body.

https://developer.mozilla.org/en-US/docs/Web/API/fetch

anymaniax:
@soartec-lab If you pass the query params into the body it wont be a query params anymore or I am missing something?

soartec-lab:
@anymaniax 
I see, the sample you shared with me adds query parameters to the URL. I will refer to it and consider it again.

packages/fetch/src/index.ts
```diff
@@ -0,0 +1,160 @@
...
+${
+  queryParams
+    ? `
+  const normalizedParams = new URLSearchParams();
+
+  Object.entries(params || {}).forEach(([key, value]) => {
+    if (value !== null && value !== undefined) {
+      normalizedParams.append(key, value.toString());
+    }
+  });`
```
soartec-lab: normalize it by excluding `undefined` and `null`, because `query params` join to `fetch url`.

soartec-lab:
Hi, @anymaniax 

I took your advice into consideration and changed the `URL` used by `fetch` to be obtained from a  `get URL function`. This function combines `path parameters` and `query parameters`. And they only use the `request body` for `body parameters`.

Cloud you review again ?

packages/fetch/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+  toObjectString,
+  generateBodyOptions,
+  isObject,
+} from '@orval/core';
+
+const PARAMS_SERIALIZER_DEPENDENCIES: GeneratorDependency[] = [
+  {
+    exports: [
+      {
+        name: 'qs',
```
anymaniax: Do we really need qs? I would prefer not having a dependencies to add for that personally. What do you think?

packages/fetch/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+    if (value === null) {
+      normalizedParams.append(key, 'null');
+    } else if (value !== undefined) {
+      normalizedParams.append(key, value.toString());
+    }
+  });`
+    : ''
+}
+
+  return \`${route}${queryParams ? '?${new URLSearchParams(normalizedParams).toString()}' : ''}\`
```
anymaniax: You could just toString the normalizedParams I think

packages/fetch/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+    : '';
+  const fetchMethodOption = `method: '${verb.toUpperCase()}'`;
+
+  const requestBodyParams = generateBodyOptions(
+    body,
+    isFormData,
+    isFormUrlEncoded,
+  );
+  const fetchBodyOption = requestBodyParams
+    ? `body: JSON.stringify(${requestBodyParams})`
```
anymaniax, line 94: In this case we only handle object body no?

packages/fetch/src/index.ts
soartec-lab: That's right. I left it there with the assumption that it would be used in later development, but I want to minimize dependencies on `fetch`, so I agree with the design without it.

soartec-lab: Fixed by https://github.com/anymaniax/orval/pull/1353/commits/805fbde24c78bc43903d604674aee5a718e199dc

packages/fetch/src/index.ts
soartec-lab: Oops, `normalizedParams` is indeed generated from `URLSearchParams` at the time of initialization. This is an unnecessary process.

soartec-lab: I removed it.

https://github.com/anymaniax/orval/pull/1353/commits/cc654b3d28b45d3d7cc9ba4ef33418c33dacca70

packages/fetch/src/index.ts
soartec-lab: Yes, Here we process and request only the body parameters.

soartec-lab:
@anymaniax 
Thank you for the review. I have fixed all responses and commented.

I wanna add the `fetch` client incrementally and improve it little by little. First of all, I'm adding a minimal working package, so additional behaviors can be added in subsequent PRs. Detailed plans are also managed in this issue.

https://github.com/anymaniax/orval/issues/1387

soartec-lab:
@melloware 

Even if i merge this, I don't wanna close #1387, so I deleted `fix #1387` in description, is there any problem?

melloware:
No problem!  We can always reopen if it accidentally closes.  I always prefer to have PR linked to ticket so we can reference later.

packages/fetch/src/index.ts
anymaniax: What I wanted to say is that for now if we have a binary in body it won't work. Is it what you want?

soartec-lab: @anymaniax 

Support for binary was not included in the scope.
For the functionality I want at this point, just being able to send a simple json request parameter is sufficient.
After that, I think it would be good if other users, including other users, could receive support when they have a case or improvement idea.

anymaniax (approved):

soartec-lab:
Thanks !

PR #1357, opened 2024-05-04
fix: allow making second axios, etc., param `required` instead of `optional`

barrownicholas:
### TLDR

I turn this existing behaviour:
```ts
`options?: SecondParameter<ReturnType<typeof ${mutator.name}>>`
```
into:
```ts
`options${context.output.optionsParamRequired ? "" : "?"}: SecondParameter<ReturnType<typeof ${mutator.name}>>`
```

## Status

**READY**

## Description

When generating a client using Axios, and some other generators, an optional, second parameter can be passed. In my case, I want this to be required instead of optional, based on a type of my choosing. For example:

### mutator.ts
```ts
import Axios, { AxiosRequestConfig } from 'axios';

// add a second `options` argument here if you want to pass extra options to each generated query
export const axiosInstance = <T>(
  config: AxiosRequestConfig,
  options: {
    baseURL: string;
  },
): Promise<T> => {
  const axios = Axios.create({ baseURL: options.baseURL });
  const source = Axios.CancelToken.source();
  const promise: Promise<any> | any = axios({
    ...config,
    cancelToken: source.token,
  }).then(({ data }) => data);


  promise.cancel = () => {
    source.cancel('Query was cancelled');
  };

  return promise;
};
```

The issue with the above, is that the current behavior always makes the second param as optional (leading to typescript errors, and I want to enforce the second parameter being included); example:
```ts
const listPayables = (
    params?: ListPayablesParams,
 options?: SecondParameter<typeof axiosInstance>,) => {
      return axiosInstance<ItemsPayablesDiscriminatedPayableList>(
      {url: `/v1/payables`, method: 'GET',
        params
    },
      options);
    }
```

This PR adds an output option `optionsParamRequired` (that defaults to `false` to maintain existing behavior). It customizes the output typescript file to accomodate this option's inclusion. 

## Related PRs

N/A

## Todos

*Not sure what to do here, help wanted!!*

- [x] Tests
- [X] Documentation

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -1725,8 +1725,9 @@ module.exports = {
 
 Type: `Boolean`
 
-Valid Values: true or false. Defaults to false.
+Valid Values: `true` or `false`.
 
+Default Value: `false`.
 Use this property to make the second `options` parameter required (such as when using a custom axios instance)
 
 ```js
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -1720,3 +1720,21 @@ module.exports = {
   },
 };
 ```
+
+### optionsParamRequired
+
+Type: `Boolean`
+
+Valid Values: true or false. Defaults to false. 
+
+Use this property to make the second `options` parameter required (such as when using a custom axios instance)
+
+```js
+module.exports = {
+  petstore: {
+    output: {
+      optionsParamRequired: true,
+    },
+  },
+};
+```
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -1725,7 +1725,7 @@ module.exports = {
 
 Type: `Boolean`
 
-Valid Values: true or false. Defaults to false. 
+Valid Values: true or false. Defaults to false.
 
 Use this property to make the second `options` parameter required (such as when using a custom axios instance)
 
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "orval-workspaces",
   "namespace": "@orval",
-  "version": "6.28.2",
+  "version": "6.28.3",
   "license": "MIT",
   "workspaces": [
     "packages/*"
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "orval-workspaces",
   "namespace": "@orval",
-  "version": "6.28.3",
+  "version": "6.28.2",
   "license": "MIT",
   "workspaces": [
     "packages/*"
--- a/yarn.lock
+++ b/yarn.lock
@@ -4,12 +4,12 @@
 
 "@aashutoshrathi/word-wrap@^1.2.3":
   version "1.2.6"
-  resolved "https://registry.yarnpkg.com/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz#bd9154aec9983f77b3a034ecaa015c2e4201f6cf"
+  resolved "https://registry.npmjs.org/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz"
   integrity sha512-1Yjs2SvM8TflER/OD3cOjhWWOZb58A2t7wpE2S9XfBYTiIl+XFhQG2bjy4Pu1I+EAlCNUzRDYDdFwFYUKvXcIA==
 
 "@apidevtools/json-schema-ref-parser@9.0.6":
   version "9.0.6"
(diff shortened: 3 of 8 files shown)


melloware (changes requested):
Looks like tests are failing

barrownicholas:
@melloware it looks like I forgot to run prettier; ran it and updated the PR. Looks like I just need approval for the tests to re-run.

melloware:
Re running

barrownicholas:
looks like it passed this time @melloware

melloware (dismissed):

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "orval-workspaces",
   "namespace": "@orval",
-  "version": "6.28.2",
```
soartec-lab: Version updates are performed at the time of release, so there is no need to update each PR.

docs/src/pages/reference/configuration/output.md
```diff
@@ -1720,3 +1720,21 @@ module.exports = {
   },
 };
 ```
+
+### optionsParamRequired
```
soartec-lab, line 1724: I thought it would be easier to understand if you specify what option it is.

```suggestion
### axiosOptionsParamRequired
```

docs/src/pages/reference/configuration/output.md
```diff
@@ -1720,3 +1720,21 @@ module.exports = {
...
+
+Type: `Boolean`
+
+Valid Values: true or false. Defaults to false.
+
```
soartec-lab: Separated valid and default values.

```suggestion
Type: `Boolean`

Valid Values: `true` or `false`.

Default Value: `false`.
```

soartec-lab (changes requested):
thank you for a good change. I have made some detailed suggestions, so please check them out.

docs/src/pages/reference/configuration/output.md
barrownicholas: I wouldn't call it only for `axios...` since it affects renderers other than axios, like `swr`

package.json
barrownicholas: gotcha, I ditched this change

barrownicholas:
@soartec-lab I pulled all your changes except using `axiosOptionsParamRequired` since this param is used for more than just axios (I only used `axios` as the example since that's what I was using, but this change affects clients like `swr`, so I think it's better to keep it more generic)

docs/src/pages/reference/configuration/output.md
soartec-lab: What also affects `swr` is the instance of `axios` as `fetcher`, right?
Therefore, I think there is no problem in explicitly specifying `Axios` here, but what do you think?

yarn.lock
```diff
@@ -4,12 +4,12 @@
 
 "@aashutoshrathi/word-wrap@^1.2.3":
   version "1.2.6"
-  resolved "https://registry.yarnpkg.com/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz#bd9154aec9983f77b3a034ecaa015c2e4201f6cf"
+  resolved "https://registry.npmjs.org/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz"
```
soartec-lab: `yarn.lock` is not related to this PR change, so I think it can be deleted. Could you check it out?

docs/src/pages/reference/configuration/output.md
barrownicholas: It looks like the `swr`, `query`, and `axios` modules would all share this change (I'm honestly not familiar enough with `query` and `swr` to know if they should just take the `axios` prefix so I'll defer to your judgement on that one)

yarn.lock
barrownicholas: good catch, removing now

docs/src/pages/reference/configuration/output.md
soartec-lab: OK, I'll think about it for a moment.

soartec-lab: Actually, I suggest adding a `fetch API` client, so I was concerned about the balance with that.

https://github.com/anymaniax/orval/pull/1353/files#diff-bb3fc1dad2ae5a09c4f6c065a987d0d6719e469a95aaad147885938e1bbdcddeR60

For now, we can coexist, and if necessary, we can separate the options in the future, so i agree to accept this option.

soartec-lab (approved):
Thanks!

docs/src/pages/reference/configuration/output.md
barrownicholas: @soartec-lab makes total sense; once the `fetch` client drops, I'm happy to integrate support for the same setup introduced here as well

anymaniax (approved):

soartec-lab:
@barrownicholas 
Thank you for nice update. I merge it.

PR #1360, opened 2024-05-08
fix(msw): make delay optional

AffeJonsson:
## Status

**READY**

## Description

This allows disabling the `delay` call in mocked api calls, which makes the MockHandler sync instead of async.
As per request of Nate on discord

## Todos

- [ ] Tests
- [x] Documentation

## Steps to Test or Reproduce

Make sure delay still can be set by both number and function and that the MockHandler is async.
Try changing the `delay` settings to `false` and make sure the MockHandlers become synchronous and no call to `delay` is made.


packages/mock/src/delay.ts
```diff
@@ -5,7 +5,7 @@ export const getDelay = (
   options?: GlobalMockOptions,
 ): GlobalMockOptions['delay'] => {
   const overrideDelay =
-    typeof override?.mock?.delay === 'number'
+    override?.mock?.delay !== undefined
```
AllieJonsson, line 8: A bit unsure about this, but it feels like override always should override when present, and not only when it's overridden by a number?

melloware: Good question. Not sure why this decision was made?

AllieJonsson: Maybe it was missed when support for functions was added?

packages/mock/src/msw/index.ts
```diff
@@ -94,11 +94,15 @@ const generateDefinition = (
...
   const handlerImplementation = `
 export const ${handlerName} = (${isReturnHttpResponse && !isTextPlain ? `overrideResponse?: ${returnType}` : ''}) => {
-  return http.${verb}('${route}', async () => {
-    await delay(${isFunction(delayTime) ? `(${delayTime})()` : delayTime});
+  return http.${verb}('${route}', ${
+    delay === false
+      ? '() => {'
+      : `async () => {
+    await delay(${isFunction(delay) ? `(${delay})()` : delay});`
+  }
```
soartec-lab, line 105: I thought that the definition itself is unnecessary if `delay` is `false`, but what do you think?

```suggestion
  return http.${verb}('${route}', ${
    delay === isFunction(delay) || isNumber(delay)
      ? `async () => {
    await delay(${isFunction(delay) ? `(${delay})()` : delay});`
      : ''
  }
```

soartec-lab (commented):
@AffeJonsson 
Thank you for the nice update. I made a comment in one place, so please check it.

packages/mock/src/msw/index.ts
AllieJonsson: No, the second argument to http.get must be a function, so we cannot remove the `() => {`

AllieJonsson: This is how the mock handler looks with `delay: false`
image

soartec-lab: Oh, that's nice. That's what I wanted.

soartec-lab (approved):
Thanks!

PR #1361, opened 2024-05-08
fix: allParamsOptional implemented in params and props

### LINKED ISSUE #1322: Build setting allParamsOptional is not working as expected

1. Generate with [allParamsOptional](https://orval.dev/reference/configuration/output#allparamsoptional) set to true.

## What happens?

Not all params are optional.

All params are optional.

No

## Any other comments?

```ts
// orval.config.ts

export default defineConfig({
  petstore: {
    input: {
      target: "https://petstore.swagger.io/v2/swagger.json",
    },
    output: {
      clean: true,
      target: "./generated",
      allParamsOptional: true,
    },
  },
});
```

Generated output

```ts
// generated/swaggerPetstore.ts

export const deleteOrder = <TData = AxiosResponse<unknown>>(
  orderId: number | undefined | null, // ✅ orderId is optional
  options?: AxiosRequestConfig
): Promise<TData> => {
  return axios.delete(`/store/order/${orderId}`, options);
};

export const findPetsByTags = <TData = AxiosResponse<Pet[]>>(
  params: FindPetsByTagsParams, // ❌ FindPetsByTagsParams is not optional
  options?: AxiosRequestConfig
): Promise<TData> => {
  return axios.get(`/pet/findByTags`, {
    ...options,
  });
};

export const updateUser = <TData = AxiosResponse<unknown>>(
  username: string | undefined | null,
  user: User, // ❌ user is not optional
  options?: AxiosRequestConfig
): Promise<TData> => {
  return axios.put(`/user/${username}`, user, options);
};
```
My guess is that params of type Object (and Array?) are not handled correctly.

**Operating System:** Windows

**Package Version:** 6.27.1

**Browser Version:** n/a

wouterkroes:
## Status

Ready

Fix #1322 

## Description

- Enhanced the current build setting allParamsOptional (refer to https://github.com/anymaniax/orval/issues/1322).
- Fixed/removed the TODO comment (`Maybe ensure that 'paramType' isn't already undefined or null`). In my view, optional signifies a property marked with a ?, rather than extending the type with `| undefined | null`.

I experimented with these adjustments locally in a playground project using Orval clients: default (axios), vue-query, and Angular. 
I have created a draft PR to ensure that my changes pass all the pipeline tests. 
These adjustments are to the core, so I hope that by testing them with all the mentioned setups, I have thoroughly validated my changes. 

Please feel free to double-check before approving my PR

## Related PRs

n/a

## Todos

n/a

## Steps to Test or Reproduce

n/a

### DIFF
--- a/packages/core/src/generators/verbs-options.ts
+++ b/packages/core/src/generators/verbs-options.ts
@@ -136,6 +136,7 @@ const generateVerbOptions = async ({
     headers,
     operationName,
     context,
+    output,
   });
 
   const mutator = await generateMutator({
--- a/packages/core/src/getters/params.ts
+++ b/packages/core/src/getters/params.ts
@@ -91,16 +91,20 @@ export const getParams = ({
     });
 
     let paramType = resolvedValue.value;
-    if (output.allParamsOptional) {
-      paramType = `${paramType} | undefined | null`; // TODO: maybe check that `paramType` isn't already undefined or null
-    }
 
     const definition = `${name}${
-      !required || resolvedValue.originalSchema!.default ? '?' : ''
+      !required ||
+      resolvedValue.originalSchema!.default ||
+      output.allParamsOptional
+        ? '?'
+        : ''
     }: ${resolvedValue.value}`;
 
     const implementation = `${name}${
-      !required && !resolvedValue.originalSchema!.default ? '?' : ''
+      (!required && !resolvedValue.originalSchema!.default) ||
+      output.allParamsOptional
+        ? '?'
+        : ''
     }${
       !resolvedValue.originalSchema!.default
         ? `: ${paramType}`
--- a/packages/core/src/getters/props.ts
+++ b/packages/core/src/getters/props.ts
@@ -5,6 +5,7 @@ import {
   GetterProps,
   GetterPropType,
   GetterQueryParam,
+  NormalizedOutputOptions,
 } from '../types';
 import { isUndefined, pascal, sortByPriority } from '../utils';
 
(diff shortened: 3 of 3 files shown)


melloware (approved):

melloware:
@Maxim-Mazurok you were recently in there as well want to give these a review?

Maxim-Mazurok (changes requested):
This flag was working as intended, consider updating docs and/or creating a new config flag. See https://github.com/anymaniax/orval/issues/1322#issuecomment-2100634682

soartec-lab:
@wouterkroes 
I agree with they. I think this change is a workaround and essentially requires revising the definition of `OpenAPI` itself for the relevant use case. What do you think?

wouterkroes:
First of all, thanks to everyone for the responses.

This entire discussion has made me rethink what I wanted to achieve and the implementation I had in mind. As a result, I've changed my plan and no longer need the changes proposed in my pull request. Therefore, I suggest we close this pull request.

It would be beneficial if we could enhance the Orval documentation. @Maxim-Mazurok suggestion to "fix this issue by renaming allParamsOptional to allPathParamsOptional and/or updating the documentation" makes sense.

soartec-lab:
@wouterkroes 

I got it. Thanks to you for suggesting this, I found a way to improve it with another approach, so I appreciate it.
I'd prefer documenting the options instead of destructively renaming them, as it's a minor improvement, but would you update the documentation in a separate PR?

soartec-lab:
I updated docs at #1453

PR #1375, opened 2024-05-15
feat: allow override for mock handlers to be a function

### LINKED ISSUE #1329: MSW: overrideResponse can be a function
Currently, a msw mock handler would be:

```ts
import { HttpResponse, HttpResponseResolver, delay, http } from 'msw'

export const getSomethingMockHandler = (overrideResponse?: SomethingResponse) => {
  return http.get('/something', async () => {
    await delay(1000);
    return new HttpResponse(JSON.stringify(overrideResponse ? overrideResponse : getSomethingResponseMock()),
      {
        status: 200,
        headers: {
          'Content-Type': 'application/json',
        }
      }
    )
  })
}

```

But it would be nice if overrideResponse is a function. So that we can produce response programmatically.

```ts
export const getSomethingMockHandler = (overrideResponse?: (resolver: HttpResponseResolver) => SomethingResponse | Promise<SomethingResponse>) => {
  return http.get('/something', async (resolver) => {
    await delay(1000);
    const response = overrideResponse ? overrideResponse(resolver) : getSomethingResponseMock()
    return new HttpResponse(JSON.stringify(response)),
      {
        status: 200,
        headers: {
          'Content-Type': 'application/json',
        }
      }
    )
  })
}

```

AffeJonsson:
## Status

**READY**

## Description

Fixes #1329 
Adds the ability to pass a function into mock handlers as override

## Steps to Test or Reproduce

yarn generate with a config that has mock: true
check mock handlers, the override parameter should be able to be a function as well

### DIFF
--- a/packages/mock/src/msw/index.ts
+++ b/packages/mock/src/msw/index.ts
@@ -95,19 +95,23 @@ const generateDefinition = (
     : '';
 
   const delay = getDelay(override, !isFunction(mock) ? mock : undefined);
+  const handlerHasOverride = isReturnHttpResponse && !isTextPlain;
+  const infoParam = handlerHasOverride ? 'info' : '';
   const handlerImplementation = `
-export const ${handlerName} = (${isReturnHttpResponse && !isTextPlain ? `overrideResponse?: ${returnType}` : ''}) => {
+export const ${handlerName} = (${handlerHasOverride ? `overrideResponse?: ${returnType} | ((${infoParam}: Parameters<Parameters<typeof http.${verb}>[1]>[0]) => ${returnType})` : ''}) => {
   return http.${verb}('${route}', ${
     delay === false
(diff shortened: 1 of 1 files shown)


packages/mock/src/msw/index.ts
```diff
@@ -95,19 +95,23 @@ const generateDefinition = (
     : '';
 
   const delay = getDelay(override, !isFunction(mock) ? mock : undefined);
+  const handlerHasOverride = isReturnHttpResponse && !isTextPlain;
```
melloware:
```suggestion
  const isHandlerOverridden = isReturnHttpResponse && !isTextPlain;
```

 I prefer `is` to `has`

melloware (changes requested):

melloware (approved):

PR #1376, opened 2024-05-15
feat: use string literals for const properties

### LINKED ISSUE #1326: Support for `const` values in string parameter types

Given a schema like this:
```yaml
     responses:
        "200":
          content:
            text/event-stream:
              schema:
                description: Each oneOf object in the array represents one possible Server Sent Events (SSE) message, serialized as UTF-8 text according to the SSE specification.
                items:
                  oneOf:
                    - properties:
                        data:
                          $ref: "#/components/schemas/ResponseError"
                        event:
                          const: error
                          description: The event name.
                          type: string
                        id:
                          description: The event ID.
                          type: integer
                        retry:
                          description: The retry time in milliseconds.
                          type: integer
                      required:
                        - data
                        - event
                      title: Event error
                      type: object
                    - properties:
                        data:
                          $ref: "#/components/schemas/ResponseData"
                        event:
                          const: data
                          description: The event name.
                          type: string
                        id:
                          description: The event ID.
                          type: integer
                        retry:
                          description: The retry time in milliseconds.
                          type: integer
                      required:
                        - data
                        - event
                      title: Event data
                      type: object
```

## What happens?

My schema types actually look like this:

```ts
export type Response200ItemOneOf = {
  data: ResponseError;
  /** The event name. */
  event: string;
  /** The event ID. */
  id?: number;
  /** The retry time in milliseconds. */
  retry?: number;
};

export type Response200ItemOneOfTwo = {
  data: ResponseData;
  /** The event name. */
  event: string;
  /** The event ID. */
  id?: number;
  /** The retry time in milliseconds. */
  retry?: number;
};
```

For my schema types to look like this. Note that the `event` property is restricted to only the const value (`error` or `data`).

```ts
export type Response200ItemOneOf = {
  data: ResponseError;
  /** The event name. */
  event: "error";
  /** The event ID. */
  id?: number;
  /** The retry time in milliseconds. */
  retry?: number;
};

export type Response200ItemOneOfTwo = {
  data: ResponseData;
  /** The event name. */
  event: "data";
  /** The event ID. */
  id?: number;
  /** The retry time in milliseconds. */
  retry?: number;
};
```

## Any other comments?

I guess that either [`const` isn't supported](https://json-schema.org/draft/2020-12/json-schema-validation#name-const) or maybe I'm just doing something wrong?

**Operating System:** MacOS 14.4.1
**Package Version:** 6.27.1
**Browser Version:** N/A

RinseV:
## Status

**WIP**

## Description

Fix #1326 to use [string literal types](https://www.typescriptlang.org/docs/handbook/literal-types.html#string-literal-types) for `const` properties defined in an OAS 3.1 schema. OAS 3.1 [introduced the `const` property](https://www.apimatic.io/blog/2021/09/migrating-to-and-from-openapi-3-1) as an alternative to single-valued enums. Previously, Orval did not generate string literal types for constant strings but instead typed them as just `string`. With this change, properties with `type: string` and a `const` keyword are transformed to their literal values as their type.

I've included a new specification that has different values and nullable versions of `const` properties. However, the `example-v3-1.yaml` specification also has a `const` property which is now also generated correctly. I'll be happy to remove the new spec if it's not needed.

## Related PRs

[A previous PR](https://github.com/anymaniax/orval/pull/1180) fixed a problem with the `const` keyword where the type was being interpreted as `unknown`. However, this fix does not make it so string constants have their value as their type.

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout <branch>
> grunt jasmine
```

1. Generate the test specifications using `yarn generate` from the `tests` folder
2. Observe the generated types in `tests/generated/default/example-v3-1/model/test.ts` and `tests/generated/swr/constants/model/*.ts`
3. The properties marked with `const` should have a string literal type instead of just `string`

### DIFF
--- a/packages/core/src/getters/object.ts
+++ b/packages/core/src/getters/object.ts
@@ -1,14 +1,15 @@
 import { ReferenceObject, SchemaObject } from 'openapi3-ts/oas30';
 import { resolveExampleRefs, resolveObject, resolveValue } from '../resolvers';
-import { ContextSpecs, ScalarValue, SchemaType } from '../types';
+import {
+  ContextSpecs,
+  ScalarValue,
+  SchemaType,
+  SchemaWithConst,
+} from '../types';
 import { isBoolean, isReference, jsDoc, pascal } from '../utils';
 import { combineSchemas } from './combine';
 import { getKey } from './keys';
 import { getRefInfo } from './ref';
-
-interface SchemaWithConst extends SchemaObject {
-  const: string;
-}
 
 /**
  * Return the output type from an object
--- a/packages/core/src/getters/scalar.ts
+++ b/packages/core/src/getters/scalar.ts
@@ -1,5 +1,10 @@
 import { SchemaObject } from 'openapi3-ts/oas30';
-import { ContextSpecs, ScalarValue, OutputClient } from '../types';
+import {
+  ContextSpecs,
+  ScalarValue,
+  OutputClient,
+  SchemaWithConst,
+} from '../types';
 import { escape, isString } from '../utils';
 import { getArray } from './array';
 import { getObject } from './object';
@@ -25,6 +30,9 @@ export const getScalar = ({
   const nullable = item.nullable && !isAngularClient ? ' | null' : '';
 
   const enumItems = item.enum?.filter((enumItem) => enumItem !== null);
+
+  const itemWithConst = item as SchemaWithConst;
+  const isConst = itemWithConst.const !== undefined;
 
   if (!item.type && item.items) {
     item.type = 'array';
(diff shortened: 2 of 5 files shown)


RinseV:
I'm actually not sure if this is the right fix. It seems like the `const` property is evaluated properly, but is ignored when you have a property with:
```
type: string
const: something
```

The `const` keyword does generate correctly if you don't add `type: string` to your property.

RinseV:
Closed as the the workaround by not adding `type: string` is sufficient

Issue #1206, opened 2024-02-06
MSW: Allow overriding entire handler function in generated MSW handlers, not just response object


severinh:
This is a feature request, not a bug.

## What are the steps to reproduce this issue?

#1186 made it possible to override response objects in MSW handlers. Thank you for that! Unfortunately, we often cannot use them in our project in many tests because we're missing two things:

1. In unit tests, we often override the default handlers to simulate a HTTP 500 errors to test the request failure path.
2. We sometimes create handlers where the output depends on the request.

Here's an example of a handler generated by Orval:

```typescript
export const getProjectQueryMockHandler = (overrideResponse?: Result) => {
  return http.post('*/projects/:projectId', async () => {
    await delay(200);
    return new HttpResponse(JSON.stringify(overrideResponse ? overrideResponse : getProjectQueryMock()),
      {
        status: 200,
        headers: {
          'Content-Type': 'application/json',
        }
      }
    )
  })
}
```

### Simulating HTTP errors in tests

Then, for example, a unit test needs to simulate a HTTP 500 error. It currently cannot use `getProjectQueryMockHandler` for that since that always uses HTTP 200. So we instead don't use any code generated by Orval, and write something like this. Unfortunately, this is a lot less type-safe, since there's nothing checking that the path `projects/:projectId` is correct.

```typescript
  server.use(
    http.post('*/projects/:projectId', async () => new HttpResponse('', { status: 500 }))
  );
```

What we'd love to see is `getProjectQueryMockHandler` allowing callers to either override the response, or the entire handler function. Something like:

```typescript
  server.use(
    getProjectQueryMockHandler(async () => new HttpResponse('', { status: 500 }))
  );
```

Alternative solution: Generate a function to get the path of the endpoint, such as:

```typescript
  server.use(
    http.post(getProjectQueryPath(), async () => new HttpResponse('', { status: 500 }))
  );
```

### Changing request based on the response

Similarly, we have tests where the MSW handlers need to return a response that depends on what is in the request. So we currently write something like:

```typescript
  server.use(
    http.post('*/projects/:projectId', async ({ request }) => {
      const input = (await request.json()) as Input;
      // Return a response based on input.
  })),
```

Ideally, we could write this as:

```typescript
  server.use(
    getProjectQueryMockHandler(async ({ request }) => {
      const input = (await request.json()) as Input;
      // Return a response based on input.
  }));
```
## Any other comments?

[...]

PR #1390, opened 2024-05-20
fix: allow async overrideResponse to be passed to msw handler (issue 1389)

severinh:
## Status

**WIP**

## Description

This fixes issue 1389.

In MSW, a mock handler can access the response resolver info. For example, it can access the request body to change the response based on the request. Access to that request body is only possible through an `async` function call to `json()`.

```typescript
http.post("/some/path", async ({ request }) => {
  const requestJson = await request.json();
  // Do something with the request, such as producing a response based on it.
  return ...
});
```

In Orval, PR #1375 made it possible for the override in a MSW mock handler to be a function, and Orval passes the response resolver info to that function. However, Orval requires a synchronous method that returns a response object. It does not allow `async` functions.

Before this PR, it was not possible to migrate the above mock handler to Orval.

```typescript
getSomePathMockHandler(async ({ request }) => {  // This does not type-check. Orval does not allow a Promise<SomeResponse> return type.
  const requestJson = await request.json();
  // Do something with the request, such as producing a response based on it.
  return ...
});
```

This PR changes the code generation such that both `Promise<ResponseType> | ResponseType` is allowed as a return type for the override, analogous to what MSW does.

## Related PRs

#1375

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

_I'm not sure what's expected of me in this section._

### DIFF
--- a/packages/mock/src/msw/index.ts
+++ b/packages/mock/src/msw/index.ts
@@ -109,7 +109,7 @@ const generateDefinition = (
   const isHandlerOverridden = isReturnHttpResponse && !isTextPlain;
   const infoParam = isHandlerOverridden ? 'info' : '';
   const handlerImplementation = `
-export const ${handlerName} = (${isHandlerOverridden ? `overrideResponse?: ${returnType} | ((${infoParam}: Parameters<Parameters<typeof http.${verb}>[1]>[0]) => ${returnType})` : ''}) => {
+export const ${handlerName} = (${isHandlerOverridden ? `overrideResponse?: ${returnType} | ((${infoParam}: Parameters<Parameters<typeof http.${verb}>[1]>[0]) => Promise<${returnType}> | ${returnType})` : ''}) => {
   return http.${verb}('${route}', ${
     delay === false
       ? `(${infoParam}) => {`
@@ -121,7 +121,7 @@ export const ${handlerName} = (${isHandlerOverridden ? `overrideResponse?: ${ret
         ? isTextPlain
           ? `${getResponseMockFunctionName}()`
           : `JSON.stringify(overrideResponse !== undefined 
-            ? (typeof overrideResponse === "function" ? overrideResponse(${infoParam}) : overrideResponse) 
+            ? (typeof overrideResponse === "function" ? await overrideResponse(${infoParam}) : overrideResponse)
             : ${getResponseMockFunctionName}())`
         : null
     },


soartec-lab:
Hi, @severinh 
I would like to understand the background a little more, so please tell me specifically when the problem occurs when it is not a `Promise`. Is it just that type checking doesn't work?

severinh:
> Hi, @severinh I would like to understand the background a little more, so please tell me specifically when the problem occurs when it is not a `Promise`. Is it just that type checking doesn't work?

It's not just the type-checking that does not pass. Let's assume we ignore the type checking error. The whole mock handler does not work correctly when passing an `async` method that returns a `Promise`. The problem is: When passing an `async` method, the `overrideResponse` function returns a `Promise` of a response, not a response object. Orval then directly passes that response object into `JSON.stringify`.

Orval only supports passing a non-async method right now. The following currently works in Orval:
```typescript
// Suppose the user passes the following overrideResponse function:
const overrideResponse = (info) => {
  // Cannot get the request body from info, because that would require the method to be async
  return {"field": "value"}
}
// Orval then does the following call
JSON.stringify(overrideResponse())
which outputs "{'field': 'value'}"
```

However, suppose the user passes an async function:
```typescript
const overrideResponse = async (info) => {
  // Do something async with info, such as reading the request body.
  return {"field": "value"}
}
// Orval (before this PR) then does the following call
JSON.stringify(overrideResponse())
// This outputs "{}", which is wrong, because we're stringifying the Promise itself, not the awaited object.
```

Hope that helps.

soartec-lab:
@severinh 

Thank you for your kindness. I understand that much, but what is the specific background of why I want to use `Promise`?
In the current state, we are only dealing with simple objects or functions that return objects, but adding more `Promises' here will also increase complexity. I would also like to understand the benefits I can receive from this, so please let me know.

severinh:
> @severinh
> 
> Thank you for your kindness. I understand that much, but what is the specific background of why I want to use `Promise`? In the current state, we are only dealing with simple objects or functions that return objects, but adding more `Promises' here will also increase complexity. I would also like to understand the benefits I can receive from this, so please let me know.

Happy to clarify.

There are two concrete use cases:

1. **Returning a response based on the request**:

Here's what you can do in MSW:

```typescript
http.post("/greeting", async ({ request }) => {
  const { firstName, lastName } = await request.json();
  return { greeting: `Hello ${firstName} ${lastName}` }
});
```

The request body is only available via `request.json`, which returns a `Promise`. And hence can only be called in an `async` method.

2. **Making assertions in a mock handler on the request in a test**:

Here's what you can do in MSW (though this use case is generally [discouraged](https://mswjs.io/docs/best-practices/avoid-request-assertions/) by MSW):

```typescript
http.post("/greeting", async ({ request }) => {
  const { firstName, lastName } = await request.json();
  expect(firstName).toEqual("Test");
  expect(lastName).toEqual("Name");
});
```

Neither of these is possible if you want to use an Orval mock handler. That means you have to fall back to hard-coding the URLs and don't get any type safety - which is the whole point of using Orval.

This is related to issue #1206 that I filed some time ago.

Personally, I would have have advised against merging #1375, because it only allows partial flexibility. With the current method signature of Orval's mock handlers (introduced in #1375), you can only override the response body, but you cannot override the entire `HttpResponse` (for example to simulate a HTTP 500 error, or add some more response headers).

My preferred solution would have been for Orval mock handlers to look like this:

```typescript
export const getGreetingMockHandler = (override?: GreetingResponse | Parameters<typeof http.post>[1]) => {
  return http.post('/greeting', async (info) => {
    return typeof override === 'function'
      ? await override(info) // This override can return anything it wants, such as a HTTP 500 error.
      : new HttpResponse(JSON.stringify(override !== undefined ? override : getGreetingResponseMock()), {
          status: 200,
          headers: {
            'Content-Type': 'application/json',
          },
        });
  });
};

// One could then use this Orval mock handler like this:
server.use(getGreetingMockHandler(async (resolverInfo) => {
  return new HttpResponse(/* anything can go here */);
});
```

This would be both simpler and more flexible than what #1375 introduced. But the problem is that since #1375 was already released in 0.29, this would be a backwards-incompatible change. Would that be a possibility? If yes, I'm happy to prepare a PR to make the mock handlers look like the above, solving #1206 too.

soartec-lab:
@severinh 

First of all, regarding 1, you want to perform calculations in `request` and calculate the value of the `return type` object, right? And, on the side that uses mocks, you can do the same thing by creating and using an object that is not a `Promise` from the beginning, right?

severinh:
> @severinh
> 
> First of all, regarding 1, you want to perform calculations in `request` and calculate the value of the `return type` object, right? And, on the side that uses mocks, you can do the same thing by creating and using an object that is not a `Promise` from the beginning, right?

Thanks for the patience. I'm afraid I don't fully understand the question. What do you mean by "on the side that uses mocks"? Could you explain how you would do 1) with an Orval-provided MSW handler like `getGreetingMockHandler`?

soartec-lab:
OK. In short, it's not clear whether you want to use `Promise` to optimize it individually for your complex project, or if it will benefit you as a whole.
This change itself is minor, so I would like to merge it if there is any benefit.
However, when using the original handler, if a response `object` or calculation is required, wouldn't it be sufficient to input the `function`?
For example, in the case of `getGreetingMockHandler`, wouldn't it be sufficient to just pass the function without doing any asynchronous processing?

```
const fullname = (firstName: string, lastName: string) => {
  return { fullname: `${firstName} ${lastName}`
}

getGreetingMockHandler(fullname)
```

severinh:
Ah, now I understand the confusion.

> In short, it's not clear whether you want to use Promise to optimize it individually for your complex project, or if it will benefit you as a whole.

To clarify, this PR:

* has nothing to do with optimization.
* is **not** about using `async` because something is slow or complex.
* is **not** about building a mock response. You're right that to build a mock response, we can just use `getGreetingResponse({greeting: "Hello world"})`. That works perfectly.

This PR about accessing the **HTTP request** that the application made to the MSW request handler, which is currently not possible with an Orval-created request handler.

Probably my toy example 1) is too simple, so it's not clear why this would be useful.

Right now, with Orval request handlers you can only use a static mock response object. No matter what request the application sends, the request handler will return the same response object.

In the real server-side implementation of an API, the response will be produced based on the request. Now, it does not make sense to reimplement the real server implementation in a request handler. However, in specific cases, it can help to make the request handler just small bit more realistic, by changing the response based on what request the application made. But that requires access to the request that the application sent, and producing a HTTP response object based on that. But Orval currently does not allow that kind of access.

Here's one [example in the MSW documentation](https://mswjs.io/docs/best-practices/avoid-request-assertions#request-validity), which is not possible when using an Orval-generated MSW handler (notice the `await request`).

Maybe what I could suggest: I'm fine with leaving this PR open for now, since it only addresses half of my problem. It only partially resolves #1206. If it's okay for you, I can try to write a new PR to solve #1206. I believe the benefit of that more complex fix will be more clear than with this PR. About to go on vacation though, so will not respond for a week. :)

soartec-lab:
@severinh 

Hmm, I understand the scene you want to use. You want to reproduce the case where the `http response` changes dynamically depending on the `http request`, so you want to execute that logic, right?

Also, for example, if you want to test whether the `request body` of an `http request` is correct, `request.json()` can only be handled by `Promise`.

well understood. I agree with this so I'll merge it. Thank you for teaching me so carefully.

Let's discuss #1206 in an issue.

soartec-lab (approved):

PR #1397, opened 2024-05-23
fix(override): add deep merge with operations and tag options

anymaniax:
## Status

**READY**

## Description

Currently when you provide operation or tag options it completely override the global one. This will solve the problem with a deep merge


melloware (approved):
This is clever!

soartec-lab (approved):
great 👍

PR #1412, opened 2024-05-29
fix(swr): correct second mutator parameter when using custom mutator

### LINKED ISSUE #1233: Pass an additional option object in fetcher function when using custom axios instance

1. Follow de docs to create an axios custom instance, with an optional 2nd parameter
2. for the client, choose `swr`

## What happens?

The **second optional parameter** is not added to the generated fetchers.

### orval.config.js

```js
module.exports = {
  test: {
    output: {
      ....
      client: 'swr',
      override: {
        mutator: {
          path: './custom-instance.ts',
         name: 'customInstance',
       },
       ...
      },
    },
  }
}
```

### The custom instance:

```js
import Axios, { AxiosRequestConfig } from 'axios';
 
 export const AXIOS_INSTANCE = Axios.create(); 
 
 export const customInstance = <T>(
   config: AxiosRequestConfig,
   options?: AxiosRequestConfig,
 ): Promise<T> => {
 // ...
}
```

### The generated fecher:

```js
export const getMyEndpointMutationFetcher = ( ) => {
  return (_: string, { arg }: { arg: Arguments }): Promise<MyEndpointResponse> => {
    return myEndpoint(arg as MyEndpointRequest);
  }
}
```

The generated fetcher should allow for optional options and the options should be passed as a second parameter of the returning function call:

```js
export const getMyEndpointMutationFetcher = (options?: AxiosRequestConfig) => {
  return (_: string, { arg }: { arg: Arguments }): Promise<MyEndpointResponse> => {
    return myEndpoint(arg as MyEndpointRequest, options);
  }
}
```

No

## Any other comments?

-

**Operating System:** MacOS 14.2.1
**Package Version:** 6.25.0
**Browser Version:** The problem is not the browser, but... Chrome Version 121.0.6167.184 (Official Build) (arm64)

AffeJonsson:
## Status

**READY**

## Description

Ran into this issue in my project and noticed there was an issue describing it:
Fixes #1233 

## Todos

- [x] Tests
- [ ] Documentation

## Steps to Test or Reproduce

1. Use the changed swr config (without the changed code) and run generate:swr
2. Notice mutation fetchers are missing the options parameter
3. Build with changed code
4. Run generate:swr
5. Notice mutation fetchers are now correctly taking an options parameter and passing it on to the api call

### DIFF
--- a/packages/swr/src/index.ts
+++ b/packages/swr/src/index.ts
@@ -552,7 +552,7 @@ const generateSwrHook = (
     summary,
     deprecated,
   }: GeneratorVerbOptions,
-  { route }: GeneratorOptions,
+  { route, context }: GeneratorOptions,
 ) => {
   const isRequestOptions = override?.requestOptions !== false;
   const doc = jsDoc({ summary, deprecated });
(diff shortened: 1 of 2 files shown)


soartec-lab (approved):
Thank you for handling this 🙌

PR #1415, opened 2024-05-29
chore: removed redundant file from react-query `hook-mutator` sample

hallucinogenizer:
## Status

**READY**

## Description
Successor to 
- #1405 

If you 
1. go to `/samples/react-query/hook-mutator/`,
2. delete all the files that already exist in `/models`,
3. generate them anew by running `npm run generate-api`

then the new files that are generated using petstore.yaml are different from the old ones.
The `pets` file in /api/models doesn't seem to be generated from petstore.yaml. The output in these folders is not exactly the output that would be generated by orval on consuming `petstore.yaml`.

I believe this file was generated while contributors were testing locally. They made changes to `petstore.yaml`, the new file was generated, they reverted their changes to `petstore.yaml` but forgot to remove this file.

This file does nothing and is causing confusion. I have simply removed it.

## Related PRs
- #1405 
- #1409 
- #1410 
- #1411 
- #1413 
- #1414 

## Todos

- [X] Tests
- [X] Documentation

## Steps to Test or Reproduce
1. Go to `/samples/react-query/hook-mutator/`,
2. Delete all the files that already exist in `/models`,
3. Generate them anew by running `npm run generate-api`


hallucinogenizer:
CI job failed due to an error with the worker's internet connection during the package install step.
Should be restarted. I don't have access to do it.

melloware (approved):

PR #972, opened 2023-10-19
fix: zod additional properties error

CxGarcia:
## Status

**READY**

## Description
in zod generator, the `additionalProperties` that are appended to functions array are either an `array` or a `boolean` (see [code](https://github.com/anymaniax/orval/blob/8f1682ce72d721d8992b4c613513191b31029c7c/packages/zod/src/index.ts#L164)). however, `parseProperty` function is treating them all as arrays and using array methods which causes a runtime error. you can reproduce the error by running zod generator on [this file](https://github.com/anymaniax/orval/pull/972/files#diff-2f9624a4143cd200d5a2962c3b6b5e0d0049997d4d2e02ea6d63891ce2716da2)

this PR fixes the error and uses the `passthrough` method when `additionalProperties` are set to `true` to avoid unknown keys being stripped from the object by zod

## Todos

- [x] Tests
- [ ] Documentation (N/A)
- [ ] Changelog Entry (unreleased)

## Steps to Test or Reproduce

```bash
git pull https://github.com/CxGarcia/orval
cd orval
git checkout fix/zod-additional-properties
yarn build
(cd tests && yarn generate:zod)
```

PR #1443, opened 2024-06-08
fix(zod): treat `additionalProperties` keyword 

ixth:
## Status

**READY**

## Description

`additionalProperties: true` isn't treated the right way during schema parsing, which results in runtime error during generation. Since #972 is stalled, decided to dig into it myself.

## Related PRs

Previous attempt to fix this:

| [CxGarcia:fix/zod-additional-properties](https://github.com/CxGarcia/orval/tree/fix/zod-additional-properties) | [link](https://github.com/anymaniax/orval/pull/972) |

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -72,13 +72,18 @@ let constsUniqueCounter: Record<string, number> = {};
 // https://github.com/colinhacks/zod#coercion-for-primitives
 const COERCIBLE_TYPES = ['string', 'number', 'boolean', 'bigint', 'date'];
 
+export type ZodValidationSchemaDefinition = {
+  functions: [string, any][];
+  consts: string[];
+};
+
 export const generateZodValidationSchemaDefinition = (
   schema: SchemaObject | undefined,
   context: ContextSpecs,
   _required: boolean | undefined,
   name: string,
   strict: boolean,
-): { functions: [string, any][]; consts: string[] } => {
+): ZodValidationSchemaDefinition => {
   if (!schema) return { functions: [], consts: [] };
 
   const consts: string[] = [];
@@ -292,13 +297,8 @@ export const generateZodValidationSchemaDefinition = (
   return { functions, consts: uniq(consts) };
 };
 
-export type ZodValidationSchemaDefinitionInput = {
-  functions: [string, any][];
-  consts: string[];
-};
-
 export const parseZodValidationSchemaDefinition = (
-  input: ZodValidationSchemaDefinitionInput,
+  input: ZodValidationSchemaDefinition,
   context: ContextSpecs,
   coerceTypes: boolean | ZodCoerceType[] = false,
   preprocessResponse?: GeneratorMutator,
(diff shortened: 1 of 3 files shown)


ixth:
Also, couldn't help but fix few typos here and there.

packages/zod/src/index.ts
```diff
@@ -224,15 +229,15 @@ export const generateZodValidationSchemaDefinition = (
...
-            ? schema.additionalProperties
-            : generateZodValidationSchemaDefinition(
-                schema.additionalProperties as SchemaObject,
-                context,
-                true,
-                name,
-                strict,
-              ),
+          generateZodValidationSchemaDefinition(
+            isBoolean(schema.additionalProperties)
```
ixth, line 233: Actual fix which I'm not sure about: since `additionalProperties: true` is just an alias for `additionalProperties: {}`, I decided to treat it as latter. Should definition be faithful to original scheme?

melloware:
@TommoLeedsy can you review this also? Its related to other recent zod changes

packages/zod/src/index.ts
soartec-lab: Is it correct to understand that if it is simply `true`, it is determined that the additional property does not exist?

soartec-lab (commented):
@ixth 
I have asked a simple question, please answer.

packages/zod/src/index.ts
ixth-ind: As far as I understood from the OpenAPI docs, it's the opposite: 

> If the dictionary values can be of any type (aka free-form object), use additionalProperties: true:

https://swagger.io/docs/specification/data-models/dictionaries/#free-form

So, `additionalProperties: true` is equivalent to `additionalProperties: {}`, which means that any type of property values allowed. Basically, it means it's `Record<string, any>`. `additionalProperties: {}` works fine, btw, it's just `additionalProperties: true` that's broken.

UPD. Oops, used work account. Comment is still relevant, though.

melloware:
@ixth in this other PR that has been open for a long time: https://github.com/anymaniax/orval/pull/972 he used `zod.passthrough()`?  not sure if your solution is better than what he was trying to do?

ixth:
@melloware 
Can't tell right away, have to check it more thoroughly, but I assume these two points:
1. `additionalProperties: {}` works just fine, `additionalProperties: true` is an alias for it, and must behave the same way, i.e. generate `z.record(z.string(), z.any())`
2. `z.passthrough()` is a parsing directive, and it doesn't affect inferred types, so it won't work in cases when I want to use `z.infer` on generated type. I might add a test case for that.

melloware:
Awesome thanks for the answers @ixth

packages/zod/src/index.ts
soartec-lab: Thanks, I understood.

soartec-lab:
@melloware @ixth 

Hey, it seemed to me that this response wasn't exactly the same as what you're trying to do with #972.
In other words, it is possible to do both. Here we will modify it so that `additionalProperties` works, but in #972 we will use `zod` to make it easier to handle properties, right? What do you guys think?

ixth:
> In other words, it is possible to do both. 
As soon as `additionalProperties: true` will result in `z.record(z.string(), z.any())`, there will be no need to add `.passthrough()`: any fields with string key will be allowed and passed through.

I might not get the whole picture. What cases am I missing?

soartec-lab:
@ixth 
I got it. It looks good. There is no specific case, it's just that I have doubts, so it's okay.

soartec-lab (approved):
Thank you for your response. I think it's a good change, but I'd like to wait for some comments by #972, so I'll merge it in a few days.

melloware:
@soartec-lab i am going to merge as this is the correct solution and @ixth pointed out why the previous PR was actually incorrect.

soartec-lab:
Gotcha!

PR #1466, opened 2024-06-19
fix: use unknown instead of any for additional properties

jmorel:
## Status

READY

## Description

Fixes https://github.com/anymaniax/orval/issues/1464

The generated type for objects with `additionalProperties: true` becomes `[key: string]: unknown` instead of `[key: string]: any` for better type safety.

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout <branch>
> grunt jasmine
```

1.

### DIFF
--- a/packages/core/src/getters/object.ts
+++ b/packages/core/src/getters/object.ts
@@ -130,7 +130,7 @@ export const getObject = ({
           if (arr.length - 1 === index) {
             if (item.additionalProperties) {
               if (isBoolean(item.additionalProperties)) {
-                acc.value += `\n  [key: string]: any;\n }`;
+                acc.value += `\n  [key: string]: unknown;\n }`;
               } else {
                 const resolvedValue = resolveValue({
                   schema: item.additionalProperties,
@@ -166,7 +166,7 @@ export const getObject = ({
   if (item.additionalProperties) {
     if (isBoolean(item.additionalProperties)) {
       return {
-        value: `{ [key: string]: any }` + nullable,
+        value: `{ [key: string]: unknown }` + nullable,
         imports: [],
         schemas: [],
         isEnum: false,
@@ -206,7 +206,7 @@ export const getObject = ({
 
   return {
     value:
-      (item.type === 'object' ? '{ [key: string]: any }' : 'unknown') +
+      (item.type === 'object' ? '{ [key: string]: unknown }' : 'unknown') +
       nullable,
     imports: [],
     schemas: [],


ezequiel:
Shall we change the `any` on line 209 as well?

https://github.com/anymaniax/orval/blob/82024a4808e57cecf85da2dd2d2befbc033c5da7/packages/core/src/getters/object.ts#L209

jmorel:
Yes definitely, I only concentrated on additionalProperties but it makes sense to fix the issue more broadly.

melloware (approved):

melloware:
The original fix was here in 2021: https://github.com/anymaniax/orval/pull/190  But I agree unknown is safer than any

uykms:
The Typescript checking of my source is invalid once again. If you want to change from `any` to `unknown`, please create a new property in the config file to allow other users to decide whether to apply your change. 
Thank you!
cc @jmorel @ezequiel

ezequiel:
> The Typescript checking of my source is invalid once again. If you want to change from `any` to `unknown`, please create a new property in the config file to allow other users to decide whether to apply your change. Thank you! cc @jmorel @ezequiel

`unknown` should've arguably always been the default.

I'm personally not a big fan of making this configurable, simply because `any` is too dangerous as it will mask potential type errors in your code.

To quickly move forward and upgrade Orval, I suggest adding `// @ts-expect-error` above the offending lines of code, and resolve them later when time permits.

 e.g

```ts
// @ts-expect-error todo: fix this 
const thing = data.response.thing;
```

lessquo:
Hi, I've just noticed this change when I tried the latest version. I agree that `unknown` is much safer than `any`, however this breaking change is a big pain for my huge codebase. It would be really nice if I can incrementally adopt this change.

In addition to that, is there any way to configure each `object` type with typescript interface? because I have a lot of 3rd-party dependencies that provide types with interface in the backend, but they all converted into `object` due to typescript limitation, and then the properties become `unknown` in the frontend. They are actually not 'unknown' because I can provide types for them. Since editing the auto-generated files is not a good idea, an option to configure them would be ideal.

This is just a small feedback. Thanks for the amazing work

FranjoMindek:
Same here, the change broke any location where we explicitly typed a generic object from the back end.
Instead of using:
```js
const data.someObject as SomeType;
```
We now need to use:
```js
const data.someObject as unknown as SomeType;
```

melloware:
Yes this issue has gone back and forth between `any` and `unknown` with both sides arguing why one is better than another.

PR #1478, opened 2024-06-23
fix: punycode deprecation warning

LukeChannings:
## Status

**READY**

## Description

Orval indirectly depends on [`node-fetch`](https://github.com/node-fetch/node-fetch), a shim for the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). 

To handle URLs in a standards-compliant way, it depends on [`whatwg-url`](https://github.com/jsdom/whatwg-url), a Node.js polyfill for the [URL API](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL). 

The version of `whatwg-url` that Orval depends on uses [`punycode`](https://nodejs.org/api/punycode.html) - a built-in (kinda) module that is deprecated in modern versions of Node.js.

Because it is deprecated, users of Orval will be given the following error message:

```
(node:95791) [DEP0040] DeprecationWarning: The 'punycode' module is deprecated. Please use a userland alternative instead.
```

This PR resolved the `whatwg-url` dependency to the latest version, which does not depend on `punycode` and so the message will no longer be shown. The URL API is stable, and swapping out the implementation does cause any breakages.

### Alternative solutions

Orval bundles all dependencies into a single distribution file. This is good for locking dependencies, but causes the package to be bloated and prevents users of Orval from making this change themselves.

Making `node_modules` dependencies external in the build step might be worth considering.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,5 +59,8 @@
     "typescript": "^4.7.4",
     "vitest": "^0.6.3",
     "zx": "^7.2.3"
+  },
+  "resolutions": {
+    "whatwg-url": ">=14"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -5878,13 +5878,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"lodash.sortby@npm:^4.7.0":
-  version: 4.7.0
-  resolution: "lodash.sortby@npm:4.7.0"
-  checksum: 10c0/fc48fb54ff7669f33bb32997cab9460757ee99fafaf72400b261c3e10fde21538e47d8cfcbe6a25a31bcb5b7b727c27d52626386fc2de24eb059a6d64a89cdf5
-  languageName: node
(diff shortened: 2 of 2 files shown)


melloware:
Thanks running the tests now.

melloware (approved):

soartec-lab:
thank you. I'll check it later 👍

soartec-lab (approved):
Thanks !

estenp:
Hey, I'm on `7.2` and still see this warning.. how do I clean that up exactly?

image

melloware:
@LukeChannings any thoughts?

LukeChannings:
I guess either the whatwg-fetch dependency has regressed or another dependency has been introduced that depends on punycode. 

I'll have a closer look when I'm by a computer

melloware:
Thanks @LukeChannings for checking

PR #1482, opened 2024-06-25
fix: handle empty response in MSW mocks

karlismelderis-mckinsey:
## Status

**READY**

## Description

if response body is empty (aka zero bytes) then `Content-Type` cannot be `application/json` as zero bytes cannot be parsed to JSON


karlismelderis-mckinsey:
Please ignore this PR for now
I might have got it wrong and issues is adding `'Content-Type': 'application/json'` to empty response body

Maxim-Mazurok:
Hey @melloware , I checked the source only, it makes sense to me, but I don't use orval for MSW. I also didn't check changes to samples. Hope this helps, cheers!

melloware (approved):

PR #1502, opened 2024-07-05
chore: use `turborepo`script and run `test:ci` script

soartec-lab:
## Status

**READY**

## Description

I'm currently using a pure `yarn` script. All tests are run on every commit, even if there are no associated updates.
It's a bit noisy for development and CI, and reduces productivity during latency.
So I use the `Turborepo` script to update it. By using the cache, tests that are not related to updates will not be executed unnecessarily.

Also, `test:samples:*` has been deleted because it can be replaced by specifying individual scripts.

## Related PRs

none

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

`yarn test:ci`


samples/package.json
```diff
@@ -10,6 +10,7 @@
     "turbo": "^1.13.3"
   },
   "scripts": {
-    "generate-api": "turbo run generate-api"
+    "generate-api": "turbo run generate-api",
+    "test": "turbo test --filter=!react-query-form-data --filter=!react-query-form-url-encoded --filter=!react-query-form-url-encoded-mutator --filter=!react-query-form-data-mutator --filter=!react-query-hook-mutator"
```
soartec-lab, line 14: The test for the package filtered here originally had an error and was not working. Since it has nothing to do with this PR, we maybe should repair it separately.

soartec-lab:
CI is probably failing for reasons unrelated to this change. I will investigate that separately.

melloware (approved):

Issue #1387, opened 2024-05-20
support `fetch` client


soartec-lab:
I'm opening this issue to clarify the support status of the `fetch` client.

# Description

I added a new `fetch` client.
Becouse, since the `fetch` API is mature, it has the advantage of reducing the bundle size of the application compared to using `Axios`, and I feel that this is sufficient in some cases.
Furthermore, we think it can be used to select an `HTTP` client with `tanstack query` or `swr`, or when calling an `API` from a `server-side` framework.

Ref: https://developer.mozilla.org/en-US/docs/Web/API/fetch

# Todo

- [x] add a minimal `fetch` client using `fetch` API.
- [x] add a sample application using the `fetch` client with `Next.js`.
- [x] add tests for the `fetch` client.
- [x] add a document for the `fetch` client.
- [x] add other enhancements. For example, customization by specifying mutator.
- [x] support `fetch` as `http fetcher` in the `swr` client.
- [x] support `fetch` as `http fetcher` in the `query` client.

oferitz:
I think the main benefit of using Axios in this context is the easy creation of a custom instance with interceptors. The most common use case is adding an Auth header in a request interceptor. Do you see this as possible with the Fetch API? If so, could you provide an example as part of this effort? If that is a no-brainer, I would happily replace Axios with Fetch

soartec-lab:
@oferitz 

Thank you for give feed back 🙌 
I agree that `Axios` interceptor is useful.　But I didn't think about that for now.
However, I like your idea and will think about how to use that use0case, such as including a sample app or guide.

jokull:
Axios doesn't run on winter runtimes like Cloudflare, Vercel Edge and Deno. This is why the fetch adapter would be super useful.

n2k3:
Neat! Thank you @soartec-lab for your implementation. We've already switched to this instead of our own `fetchClient.ts` via a mutator override.

I've a question regarding handling different `baseUrl`'s for multiple environments (local, test, prod, etc):

We've made a custom wrapper to handle authentication serverside, it has the following structure `withAuth(fn, fnParams, fetchParams)`. Then in the `orval.config.ts` we've added the following: ``baseUrl: `$\{process.env.API_URL\}`,``.  This works because the value is just being concatenated during code generation. This is so that we can keep using the benefit of having the env variable be evaluated at runtime on the server, instead of on build time.

But this probably isn't the intended way to implement this, do you have any suggestions regarding this?

[...]

PR #1529, opened 2024-07-21
feat(fetch): support form data and form url encoded request

### LINKED ISSUE #1525: Content-Type header with application/json should be assumed if there is a request body
Hello!

It feels very logical to us for a default `Content-Type': 'application/json'` header to be assumed if there are any `requestBodyParams`. Reasoning being that currently, in the fetch template, the body is provided through `JSON.stringify()`, but not all services accept the payload if you don't pass the header. I do realize these can simply be passed through options (either per request/global), but the default seems sensible.

Let me know how you feel about it, happy to provide a PR.

soartec-lab:
## Status

**READY**

## Description

follow up #1387 
fix #1525 

I made the `fetch` client support cases other than `application/json`, such as `multipart/form-data` and `application/x-www-form-urlencoded`. And, if input `OpenAPI` have `content-type`, add `Content-Type` header to fetch request options.

## Related PRs

none

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

You can check from the test case I added

### DIFF
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -94,7 +94,9 @@ ${
     ? `${stringify(override?.requestOptions)?.slice(1, -1)?.trim()}`
     : '';
   const fetchMethodOption = `method: '${verb.toUpperCase()}'`;
-
+  const fetchHeadersOption = body.contentType
+    ? `headers: { 'Content-Type': '${body.contentType}' }`
+    : '';
   const requestBodyParams = generateBodyOptions(
     body,
     isFormData,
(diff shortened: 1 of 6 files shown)


packages/fetch/src/index.ts
```diff
@@ -94,20 +94,25 @@ ${
...
+      ? `body: ${requestBodyParams}`
+      : `body: JSON.stringify(${requestBodyParams})`
     : '';
 
   const fetchFnOptions = `${getUrlFnName}(${getUrlFnProperties}),
   {${globalFetchOptions ? '\n' : ''}      ${globalFetchOptions}
     ${isRequestOptions ? '...options,' : ''}
-    ${fetchMethodOption}${fetchBodyOption ? ',' : ''}
+    ${fetchMethodOption}${fetchHeadersOption ? ',' : ''}
+    ${fetchHeadersOption}${fetchBodyOption ? ',' : ''}
```
anymaniax, line 115: Two time the same?

anymaniax (approved):

melloware (approved):

jamesleeht:
It seems like there is an issue with this - `multipart/form-data` content type header should not be set when using fetch:
https://muffinman.io/blog/uploading-files-using-fetch-multipart-form-data/

If not the boundaries will not be properly set.

PR #1532, opened 2024-07-22
Exclude content type header if multipart/form-data in fetch client

### LINKED ISSUE #1531: Fetch: `multipart/form-data` content type header should not be set in fetch client
`multipart/form-data` content type header should not be set when using fetch:
https://muffinman.io/blog/uploading-files-using-fetch-multipart-form-data/

If not the boundaries will not be properly set.

_Originally posted by @jamesleeht in https://github.com/orval-labs/orval/issues/1529#issuecomment-2243532223_

jamesleeht:
## Status

**READY**

## Description

Excludes form data content header in fetch client. 
Fix #1531

## Related PRs

https://github.com/orval-labs/orval/pull/1529

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

Generate fetch client using multipart/form-data openapi spec

### DIFF
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -94,10 +94,11 @@ ${
     ? `${stringify(override?.requestOptions)?.slice(1, -1)?.trim()}`
     : '';
   const fetchMethodOption = `method: '${verb.toUpperCase()}'`;
-  const ignoreContentTypes = ['multipart/form-data']
-  const fetchHeadersOption = body.contentType && ignoreContentTypes.includes(body.contentType)
-    ? `headers: { 'Content-Type': '${body.contentType}' }`
-    : '';
+  const ignoreContentTypes = ['multipart/form-data'];
+  const fetchHeadersOption =
+    body.contentType && !ignoreContentTypes.includes(body.contentType)
+      ? `headers: { 'Content-Type': '${body.contentType}' }`
+      : '';
   const requestBodyParams = generateBodyOptions(
     body,
     isFormData,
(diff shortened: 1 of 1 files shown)


packages/fetch/src/index.ts
```diff
@@ -94,7 +94,7 @@ ${
     ? `${stringify(override?.requestOptions)?.slice(1, -1)?.trim()}`
     : '';
   const fetchMethodOption = `method: '${verb.toUpperCase()}'`;
-  const fetchHeadersOption = body.contentType
+  const fetchHeadersOption = body.contentType && !isFormData
```
soartec-lab: does this need to all case of ` multipart/form-data`? 
`isFormData` is only valid when you set the `formData` option in config.

jamesleeht: Ah I see. I think I need to change the code to detect if the formData is within the API spec instead of using this boolean then. Didn't realize the boolean is for the override config

soartec-lab: @jamesleeht 

So, what do you think about this update?

```suggestion
  const ignoreContentTypes = ['multipart/form-data']
  const fetchHeadersOption = body.contentType && ignoreContentTypes.includes(body.contentType)
```

This update excludes content-type in all cases using `multipart/form-data`.

jamesleeht: LGTM!

soartec-lab:
@jamesleeht 
Thanks. Also, could you please fix it so that CI is successful?

melloware (approved):

soartec-lab (approved):
thanks for great work!

PR #1560, opened 2024-08-02
fix(query): check for peer dependencies

### LINKED ISSUE #1559: Query: getPackageByQueryClient doesn't respect peer dependencies

1. Clone the [samples/react-query/basic](https://github.com/orval-labs/orval/tree/ccc4ff97eed9fef1eac88ac643c2d69a5c31b2e7/samples/react-query/basic) project, including the `package.json` and `orval.config.js`
1. Edit `package.json` and move `"@tanstack/react-query": "^5.0.0"` from dependencies into peerDependencies
1. Run `orval --config ./orval.config.js`

## What happens?

The generated output does not correctly detect hasQueryV5, and thus fails to generate the appropriate TS interfaces required for v5. For example, the `getListPetsInfiniteQueryOptions` is missing the [`ListPetsParams['limit']`](https://github.com/orval-labs/orval/blob/ccc4ff97eed9fef1eac88ac643c2d69a5c31b2e7/samples/react-query/basic/src/api/endpoints/petstoreFromFileSpecWithTransformer.ts#L102-L103) type [required to make v5 typecheck](https://github.com/orval-labs/orval/blob/ccc4ff97eed9fef1eac88ac643c2d69a5c31b2e7/packages/query/src/index.ts#L480-L486).

The generated output should include the `ListPetsParams['limit']` type.

N/A

## Any other comments?

We use peer dependencies because we have a shared library that generates react query hooks, which is then used by other projects in a monorepo. It looks like a straightforward solution for now is to move the peer dependency back to `devDependencies`, but it seems like an easy fix to add [peerDependencies to the list in `getPackageByQueryClient`](https://github.com/orval-labs/orval/blob/ccc4ff97eed9fef1eac88ac643c2d69a5c31b2e7/packages/query/src/index.ts#L338-L362).

+1 for implementing #1144 which would have made debugging this issue slightly easier :)

```
System:
    OS: macOS 14.5
    CPU: (12) arm64 Apple M3 Pro
    Memory: 783.84 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    @tanstack/react-query: ^5 => 5.51.18
    orval: 6.31.0 => 6.31.0
```

melloware:
Fix #1559

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -622,6 +622,7 @@ export type TsConfigTarget =
 export interface PackageJson {
   dependencies?: Record<string, string>;
   devDependencies?: Record<string, string>;
+  peerDependencies?: Record<string, string>;
 }
 
 export type GeneratorSchema = {
--- a/packages/query/src/index.ts
+++ b/packages/query/src/index.ts
@@ -3,52 +3,52 @@ import {
   ClientBuilder,
   ClientDependenciesBuilder,
   ClientHeaderBuilder,
+  compareVersions,
   generateMutator,
   generateVerbImports,
   GeneratorDependency,
   GeneratorMutator,
   GeneratorOptions,
   GeneratorVerbOptions,
+  getRouteAsArray,
   GetterParams,
   GetterProp,
   GetterProps,
   GetterPropType,
+  GetterQueryParam,
   GetterResponse,
   isObject,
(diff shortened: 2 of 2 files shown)


soartec-lab (approved):
LGTM!

PR #1567, opened 2024-08-07
fix: Always async resolver function

### LINKED ISSUE #1566: [Regression in 7.0.1] Generated MSW mock handlers have type check error

1. Upgrade from 6.31.0 to 7.0.1
2. For routes without response body, the generated MSW mock handler does not type-check. The resolver passed to `http.post` has the signature `(info) => { ... }`, but it should be async `(info) => {`, because the generated body contains an `await`.

## What happens?

Here's a screenshot illustrating the problem. It's the diff in our generated code before and after upgrading to 7.0.1.

The problem only seems to affect routes without a response body. It seems that before 7.0.1, it was not possible to override the handler for such a route. With 7.0.1, the generated code now seems to be more similar to routes with a response body.

Screenshot 2024-08-07 at 10 22 43

Currently, this blocks us from upgrading to orval 7.0.1, because the generated MSW code does not compile.

Either:
1. Add the missing `async` before `(info) => {`, which would be consistent with other generated mock handlers.
2. Revert the behavior to how it was in 6.31.1.

N/A

## Any other comments?

Thanks for all your work on orval! It's a pleasure to work with orval.

7.0.1

karlismelderis-mckinsey:
## Status

**READY**

Fix #1566

## Description

MSW resolver function has to be always async as Orlval is always adding async "override" function

in case `mock.delay = false` Orval failed to add `async` keyword

I also added extra test case to catch the case
🙏 Please do check if I missed anything for tests

🙏 Please release this fix soon as it's blocking us to use Orval v7 and it has couple fixes we're looking for

### DIFF
--- a/packages/mock/src/msw/index.ts
+++ b/packages/mock/src/msw/index.ts
@@ -116,9 +116,7 @@ const generateDefinition = (
   const infoParam = 'info';
   const handlerImplementation = `
 export const ${handlerName} = (overrideResponse?: ${returnType} | ((${infoParam}: Parameters<Parameters<typeof http.${verb}>[1]>[0]) => Promise<${returnType}> | ${returnType})) => {
-  return http.${verb}('${route}', ${
-    (isReturnHttpResponse && !isTextPlain) || delay !== false ? 'async' : ''
-  } (${infoParam}) => {${
+  return http.${verb}('${route}', async (${infoParam}) => {${
     delay !== false
       ? `await delay(${isFunction(delay) ? `(${delay})()` : delay});`
       : ''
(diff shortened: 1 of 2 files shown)


melloware:
@karlismelderis-mckinsey does this fix any open bug ticket?

karlismelderis-mckinsey:
> @karlismelderis-mckinsey does this fix any open bug ticket?

No
I spotted a bug while trying to upgrade to v7 and wen't with a fix as I had an idea where is the issue 😉

Do you require to open a bug?

karlismelderis-mckinsey:
> #1566

spot on
missing `async` is what my PR is fixing

Not sure about `Type check fails for this await` - I didn't see it in our case

melloware (approved):

PR #1578, opened 2024-08-14
fix(client-query): return type can be undefined when config request options is set to false

anymaniax:
## Status

**READY**

## Description

Currently with the client query if you set the config request options to false the return type is always defined which is wrong.


melloware (approved):
LGTM

Issue #1285, opened 2024-03-27
Add support for CSV/explode=false query parameters "styles"


ashb:
There is a feature of OpenAPI 3 that lets you change the style of how repeated parameters are specified https://spec.openapis.org/oas/v3.0.3#parameterStyle and https://spec.openapis.org/oas/v3.0.3#style-examples

The particular use case I have is to set `explode=false` to have CSV parameters rather than multiple query params.

Minimal spec file:

```json
{
  "info": {
    "title": "My API",
    "version": "1.0.0"
  },
  "openapi": "3.0.3",
  "paths": {
    "/test/": {
      "get": {
        "parameters": [
          {
            "in": "query",
            "explode": false,
            "name": "parameter",
            "required": true,
            "schema": {
              "items": {
                "type": "string"
              },
              "type": "array"
            }
          }
        ],
        "responses": {
          "200": {
            "description": "OK"
          }
        }
      }
    }
  }
}
```

When called with multiple params this should result in `?parameter=a,b,c` but it always/only produces  `?parameter=a&parameter=b&parameter=c`. 

Tested with version 6.25.0

programm-ingovals:
Note there also how arrays should be handled according to OpenApi spec

`color=blue&color=black&color=brown`

But **Orval** will do 

`color[]=blue&color[]=black&color[]=brown`

Which I've never seen mentioned in the OpenApi spec.
Which I guess is fine if we could just turn it off.

austin-agronick:
@programm-ingovals that is an [axios feature](https://stackoverflow.com/a/76517213) that can be configured by adding the request key to the options parameter or the output override object:
```
request: {
    paramsSerializer: {
      indexes: null
    }
  }
```

programm-ingovals:
> @programm-ingovals that is an [axios feature](https://stackoverflow.com/a/76517213) that can be configured by adding the request key to the options parameter or the output override object:
> 
> ```
> request: {
>     paramsSerializer: {
>       indexes: null
>     }
>   }
> ```

Thanks, will check that out.

soartec-lab:
i think fix by #1580 so related.

soartec-lab:
I close this issue.
When you use `axios` as a client, use `paramsSerializer`. and when you use `fetch` as a client, it can be controlled by the `explode` property in `OpenAPI`.

melloware:
[...]

PR #1580, opened 2024-08-16
fix(fetch): properly append array values to `normalizedParams` object

n2k3:
## Status

**WIP**

## Description

related https://github.com/orval-labs/orval/issues/1285

I experienced a problem with array values.
For example I want to send this: `Foo: [0, 1]` this will be added by [this line of code](https://github.com/orval-labs/orval/blob/master/packages/fetch/src/index.ts#L55) to the URLSearchParams instance, which in turn will by stringified to `Foo=0,1` and appended to the API url in generated code.

Shouldn't this be done according to [this example in the MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/append#examples)? Resulting in `Foo=0&Foo=1` as stringified value.

This PR addresses that. Let me know if I need to anything else within this PR.

Thanks @soartec-lab for the great work on the fetch client 🎉

## Related PRs

* https://github.com/orval-labs/orval/issues/1387

## Todos

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -52,7 +52,11 @@ ${
     if (value === null) {
       normalizedParams.append(key, 'null');
     } else if (value !== undefined) {
-      normalizedParams.append(key, value.toString());
+      if (value instanceof Array) {
+        value.forEach((v) => normalizedParams.append(key, v.toString()));
+      } else {
+        normalizedParams.append(key, value.toString());
+      }
     }
   });`
     : ''


n2k3:
> Looks like the build is failing?

I can now reproduce locally, I'm working on a fix 👍

soartec-lab (commented):
Thanks!
I'm wondering if I should make it optional and selectable. Because this is currently working.

n2k3:
> Thanks! I'm wondering if I should make it optional and selectable. Because this is currently working.

I've found the cause of the error, it's because the strong typed pet store example `/samples/**/models/listPetsParams.ts` doesn't contain any parameters that have an array type. In the generated example it only is
```ts
  /**
   * How many items to return at one time (max 100)
   */
  limit?: string;
};
```

this is based on the `/samples/**/petstore.yaml`
```yaml
      parameters:
        - name: limit
          in: query
          required: false
          schema:
            type: string
```

So then it would have to be dynamically included based on whether the type is an array or not. I haven't figured out how / where to do this best. If you have any idea on how to do this properly, don't hesitate to do so 😉

soartec-lab:
This means changing the conditions for constructing query parameters depending on the `type` of `OpenAPI`, right?
I will also look for ways to change it 👍

soartec-lab:
Hi, @n2k3 
I looked into this.
A reasonable way to accomplish this is to parse the generated query params string and verify whether it contains an array. Also, consider cases where there are only array-type fields.

soartec-lab:
@n2k3
I created #1604 based on this PR. Does control with `explode` property solve your problem?

soartec-lab:
I'll close this PR. Because, this issue fixed by #1604.

PR #1505, opened 2024-07-06
Feature/react query initial data

lnkarma:
## Status

**READY**

## Description

After generating react-query hooks, When using `useQuery` or `useInfiniteQuery` custom hook generated be Orval, if you profide initial data in the `query` option then the type of `data` property doesn't change it remains `nullable`

```ts
const { data: pets, refetch } = useListPets({}, undefined, {
    query: {
      initialData: [],
    },
  });
```
in this code the type of `pets` variable is `Pet[] | undefined`

where it should be `Pets[]`, because we have provided the initial data, this is the expected behaviour when using `useQuery` directly.

```ts
  const { data: pets } = useQuery({...getListPetsQueryOptions(), initialData: []})
```

in this case the type of `pets` is correct as `Pet[]`

This pr fixes this type issue

## Todos

- [x] Tests
- [x] Documentation
- [ ] Changelog Entry (unreleased)

## Steps to Test or Reproduce

Outline the steps to test or reproduce the PR here.

```bash
> yard dev
```

1.  Open file `samples/react-query/basic/src/App.tsx`
2. replace line no 9: `  const { data: pets, refetch } = useListPets();` with the following code
```ts
const { data: pets, refetch } = useListPets({}, undefined, {
    query: {
      initialData: [],
    },
  });
```
3. Hover over `pets` variable to check it's type
4. The type should be `PetsArray`

PR #1583, opened 2024-08-17
fix: respect the allParamsOptional config option in React query initi…

### LINKED ISSUE #1564: React query initialData overloads don't respect the allParamsOptional config option

1. Enable the [`allParamsOptional`](https://orval.dev/reference/configuration/output#allparamsoptional) option in the config
2. Generate the hooks with the `react-query` client
3. Try to use a hook with an optional parameter

OpenAPI sample spec:
```yaml
openapi: '3.0.0'
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
servers:
  - url: http://petstore.swagger.io/v1
paths:
  /pets/{petId}:
    get:
      summary: Info for a specific pet
      operationId: pet
      tags:
        - pets
      parameters:
        - name: petId
          in: path
          required: true
          description: The id of the pet to retrieve
          schema:
            type: string
      responses:
        '200':
          description: Expected response to a valid request
          content:
            application/json:
              schema:
        default:
          description: unexpected error
          content:
            application/json:
              schema:
components:
  schemas:
    Pet:
      type: object
      required:
        - id
        - name
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
        tag:
          type: string
        email:
          type: string
          format: email
        callingCode:
          type: string
          enum: ['+33', '+420', '+33'] # intentional duplicated value
        country:
          type: string
          enum: ["People's Republic of China", 'Uruguay']
    Error:
      type: object
      required:
        - code
        - message
      properties:
        code:
          type: integer
          format: int32
        message:
          type: string
```

orval.config.ts:
```typescript

export default defineConfig({
  allParamsOptional: {
    output: {
      target: 'src/api/endpoints.ts',
      schemas: 'src/api/model',
      client: 'react-query',
      httpClient: 'fetch',
      allParamsOptional: true,
    },
    input: {
      target: 'petstore.yaml'
    },
  },
})
```

## What happens?

You get a Typescript error:
```
No overload matches this call.
  Overload 1 of 3, '(petId: string, options?: { query?: (Partial<UseQueryOptions<petResponse, Promise<Error>, petResponse, QueryKey>> & Pick<...>) | undefined; fetch?: RequestInit | undefined; } | undefined): UseQueryResult<...> & { ...; }', gave the following error.
    Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
      Type 'undefined' is not assignable to type 'string'.
  Overload 2 of 3, '(petId: string, options?: { query?: Partial<UseQueryOptions<petResponse, Promise<Error>, petResponse, QueryKey>> | undefined; fetch?: RequestInit | undefined; } | undefined): UseQueryResult<...> & { ...; }', gave the following error.
    Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
      Type 'undefined' is not assignable to type 'string'.ts(2769)
endpoints.ts(137, 17): The call would have succeeded against this implementation, but implementation signatures of overloads are not externally visible.
```

Generated overloads:
```typescript
export function usePet<
  TData = Awaited<ReturnType<typeof pet>>,
  TError = Promise<Error>,
>(
  // Should be string | undefined | null
  petId: string,
  options: {
    query: Partial<
      UseQueryOptions<Awaited<ReturnType<typeof pet>>, TError, TData>
    > &
      Pick<
        DefinedInitialDataOptions<
          Awaited<ReturnType<typeof pet>>,
          TError,
          TData
        >,
        "initialData"
      >;
    fetch?: RequestInit;
  },
): DefinedUseQueryResult<TData, TError> & { queryKey: QueryKey };
export function usePet<
  TData = Awaited<ReturnType<typeof pet>>,
  TError = Promise<Error>,
>(
  // Should be string | undefined | null
  petId: string,
  options?: {
    query?: Partial<
      UseQueryOptions<Awaited<ReturnType<typeof pet>>, TError, TData>
    > &
      Pick<
        UndefinedInitialDataOptions<
          Awaited<ReturnType<typeof pet>>,
          TError,
          TData
        >,
        "initialData"
      >;
    fetch?: RequestInit;
  },
): UseQueryResult<TData, TError> & { queryKey: QueryKey };
export function usePet<
  TData = Awaited<ReturnType<typeof pet>>,
  TError = Promise<Error>,
>(
  // Should be string | undefined | null
  petId: string,
  options?: {
    query?: Partial<
      UseQueryOptions<Awaited<ReturnType<typeof pet>>, TError, TData>
    >;
    fetch?: RequestInit;
  },
): UseQueryResult<TData, TError> & { queryKey: QueryKey };
/**
 * @summary Info for a specific pet
 */

export function usePet<
  TData = Awaited<ReturnType<typeof pet>>,
  TError = Promise<Error>,
>(
  // The actual implementation has the correct types
  petId: string | undefined | null,
  options?: {
    query?: Partial<
      UseQueryOptions<Awaited<ReturnType<typeof pet>>, TError, TData>
    >;
    fetch?: RequestInit;
  },
): UseQueryResult<TData, TError> & { queryKey: QueryKey } {
  const queryOptions = getPetQueryOptions(petId, options);

  const query = useQuery(queryOptions) as UseQueryResult<TData, TError> & {
    queryKey: QueryKey;
  };

  query.queryKey = queryOptions.queryKey;

  return query;
}
```

The optional parameter would be allowed in all overloads

## Any other comments?

[This PR](https://github.com/orval-labs/orval/pull/1505) added overloads to all hooks to allow for proper typing of the response when supplying `initialData`. However, these overloads don't consider the `allParamsOptional` option for the path parameters resulting in 3 function signatures that don't allow for `undefined` path parameters while the actual implementation does allow for it.

The problem is that these function overloads use the `definition` of the endpoint rather than the `implementation` and [the `allParamsOptional` config setting only applies to the `implementation` of the endpoint's definition](https://github.com/orval-labs/orval/blob/9ed9ec223e9045e06630b0043b48384987f5ee09/packages/core/src/getters/params.ts#L94). However, the function overloads cannot use the `implementation` because [the `implementation` also includes the default values](https://github.com/orval-labs/orval/blob/9ed9ec223e9045e06630b0043b48384987f5ee09/packages/core/src/getters/params.ts#L105) which you cannot define in function overloads. We would need something that's in between the `definition` and `implementation` that does include the `| undefined | null` from the `allParamsOptional` setting but does not include the default parameters from the `implementation`.

```
System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (16) x64 AMD Ryzen 7 5800X 8-Core Processor
    Memory: 5.14 GB / 7.76 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  npmPackages:
    @tanstack/react-query: ^5.51.21 => 5.51.21 
    orval: ^7.0.1 => 7.0.1
```

lnkarma:
## Status

**READY**

## Description

Fixes #1564

## Related PRs

| feature/react-query-initial-data     | #1505 |

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

Follow the steps mentioned #1564

### DIFF
--- a/packages/core/src/getters/props.ts
+++ b/packages/core/src/getters/props.ts
@@ -26,6 +26,7 @@ export const getProps = ({
   const bodyProp = {
     name: body.implementation,
     definition: `${body.implementation}${body.isOptional ? '?' : ''}: ${body.definition}`,
+    definedDefinition: `${body.implementation}: ${body.isOptional ? 'undefined' : ''} | ${body.definition}`,
     implementation: `${body.implementation}${body.isOptional ? '?' : ''}: ${body.definition}`,
     default: false,
     required: !body.isOptional,
@@ -35,6 +36,9 @@ export const getProps = ({
   const queryParamsProp = {
     name: 'params',
     definition: `params${queryParams?.isOptional ? '?' : ''}: ${
+      queryParams?.schema.name
+    }`,
+    definedDefinition: `params: ${queryParams?.isOptional ? 'undefined' : ''} | ${
       queryParams?.schema.name
     }`,
     implementation: `params${queryParams?.isOptional ? '?' : ''}: ${
@@ -50,6 +54,9 @@ export const getProps = ({
   const headersProp = {
     name: 'headers',
     definition: `headers${headers?.isOptional ? '?' : ''}: ${
+      headers?.schema.name
+    }`,
+    definedDefinition: `headers: ${headers?.isOptional ? 'undefined' : ''} | ${
       headers?.schema.name
     }`,
     implementation: `headers${headers?.isOptional ? '?' : ''}: ${
@@ -90,6 +97,7 @@ export const getProps = ({
         type: GetterPropType.NAMED_PATH_PARAMS,
         name,
         definition: `${name}: ${parameterTypeName}`,
+        definedDefinition: `${name}: ${parameterTypeName}`,
         implementation,
         default: false,
         destructured,
--- a/packages/core/src/getters/params.ts
+++ b/packages/core/src/getters/params.ts
@@ -71,6 +71,7 @@ export const getParams = ({
       return {
         name,
         definition: `${name}${!required ? '?' : ''}: unknown`,
+        definedDefinition: `${name}:${!required ? 'undefined' : ''} | unknown`,
         implementation: `${name}${!required ? '?' : ''}: unknown`,
         default: false,
         required,
@@ -107,9 +108,18 @@ export const getParams = ({
         : `: ${paramType} = ${stringify(resolvedValue.originalSchema!.default)}` // FIXME: in Vue if we have `version: MaybeRef<number | undefined | null> = 1` and we don't pass version, the unref(version) will be `undefined` and not `1`, so we need to handle default value somewhere in implementation and not in the definition
     }`;
 
+    const definedDefinition = `${name}: ${
+      !required &&
+      !resolvedValue.originalSchema!.default &&
+      !output.allParamsOptional
+        ? 'undefined | '
+        : ''
+    }${paramType}`;
+
     return {
       name,
       definition,
+      definedDefinition,
       implementation,
       default: resolvedValue.originalSchema!.default,
       required,
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -882,6 +882,7 @@ export type GetterParameters = {
 export type GetterParam = {
   name: string;
   definition: string;
+  definedDefinition: string;
   implementation: string;
   default: boolean;
   required: boolean;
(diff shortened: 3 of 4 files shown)


melloware:
I will let @anymaniax review but I feel like there is a more clever way to do this. It's so similar to the other branch of code?

packages/core/src/getters/props.ts
```diff
@@ -26,6 +26,7 @@ export const getProps = ({
   const bodyProp = {
     name: body.implementation,
     definition: `${body.implementation}${body.isOptional ? '?' : ''}: ${body.definition}`,
+    definedDefinition: `${body.implementation}: ${body.isOptional ? 'undefined' : ''} | ${body.definition}`,
```
anymaniax, line 29: would be better not modifying the core for that. You could do the same think directly in the react query client no?

soartec-lab:
As there has been no progress for a while, we will close it for now.
The issue remains, so we welcome you to reopen it if necessary.

Issue #1592, opened 2024-08-20
Zod: generation of typed tuples produces illegal zod array syntax


helt:
## What are the steps to reproduce this issue?

1. check out referenced PR
2. in tests: generate and build

## What happens?

zod generator produces `zod.array()`, which is illegal syntax since `error TS2554: Expected 1-2 arguments, but got 0.`

## What were you expecting to happen?

Produce a typed tuple, similar to:

```zod.tuple(zod.string(), zod.object())```

## Any logs, error output, etc?

```
generated/zod/example-v3-1.ts:25:24 - error TS2554: Expected 1-2 arguments, but got 0.

25   "example_tuple": zod.array().min(postApiSampleResponseExampleTupleMin).max(postApiSampleResponseExampleTupleMax).optional(),
                          ~~~~~~~

  node_modules/zod/lib/types.d.ts:1055:49
    1055 declare const arrayType: <T extends ZodTypeAny>(schema: T, params?: RawCreateParams) => ZodArray<T, "many">;
                                                         ~~~~~~~~~
    An argument for 'schema' was not provided.
```

## Any other comments?

more issues pop up which are not *my* main concern, currently.

## What versions are you using?

```
npx envinfo --system --npmPackages orval,zod,axios,msw,swr,@tanstack/react-query,@tanstack/vue-query,react,vue
Need to install the following packages:
envinfo@7.13.0
Ok to proceed? (y) y

  System:
    OS: Linux 5.15 Ubuntu 22.04.4 LTS 22.04.4 LTS (Jammy Jellyfish)
    CPU: (8) x64 Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz
    Memory: 13.61 GB / 31.31 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  npmPackages:
    @tanstack/react-query: ^4.22.0 => 4.22.0 
    @tanstack/vue-query: ^4.22.0 => 4.22.0 
    axios: ^0.26.1 => 0.26.1 
    msw: ^2.0.2 => 2.0.2 
    react: ^18.3.1 => 18.3.1 
    swr: ^2.2.4 => 2.2.4 
    vue: ^3.3.4 => 3.3.4 
    zod: ^3.23.8 => 3.23.8 
```

helt:
reuses test cases of prefixItems added in https://github.com/orval-labs/orval/issues/890

helt:
[...]

PR #1593, opened 2024-08-20
Bugs/typed tuple on zod

helt:
## Status

DONE

## Description

Support for typed tuples as they are produced by pydantic v2.0+

Reproduction for issue https://github.com/orval-labs/orval/issues/1592

* in `generateZodValidationSchemaDefinition`, the `array` switch case has to be altered to adhere `prefixItems` constraints. If `prefixItems` exists,
  * [x] a `tuple` has to be pushed into functions instead of array,
  * [x]  each prefixItem has to be parsed and nested into the tuple
  * [x] if items exists, a rest function has to be pushed after the tuple with the schema in items
* in parseZodValidationSchemaDefinition:
  * [x] logic to add positional parameters to tuple
  * [x] logic for the rest functions has to be added
* somewhere:
  * [x] prevent .min().max() on tuple.
* [x] Tests added


melloware:
Run `npm run format` to format the code.  Also for min() and max() i assume jus tlook at where min and max are being createed in the code and you porbbaly need to ignore that if its a tuple?

helt:
More issues that I would like to know if there is a better solution:
* I have to cast to oas31.SchemaObject and back to oas30.SchemaObject because prefixItems is a oas 3.1.0 thing. Is there a better way? Without casting?

helt:
I think this is done now. Please review and comment (and merge)

melloware:
tests are still failing?

helt:
Ok. I see. Fakerjs got triggered because I accidentially re-used a existing test oas file. I reverted the hijacked file and added an explicit one. Hope thats ok this way.

melloware:
Running build now

helt:
Did huskies job and formatted the code

melloware (approved):

PR #1604, opened 2024-08-31
feat(fetch): support `explode` property

soartec-lab:
## Status

**READY**

## Description

Related #1285

The `fetch` clent now supports `OpenAPI`'s `explode` type query params.

```yaml
parameters:
  - name: limit
    in: query
    description: How many items to return at one time
    explode: true
    schema:
      type: array
      items:
        type: integer
```

### Before

Sets the array as a query parameter regardless of the value of `explode` like a `?parameter=a,b,c`

```ts
  if (value === null) {
    normalizedParams.append(key, 'null');
  } else if (value !== undefined) {
    normalizedParams.append(key, value.toString());
  }
});

// => ?parameter=a,b,c
```

### After

Modify query parameter construction by referencing the value of `explode`.

#### When default or set `false` to `explode`:

The generated processing is as follows:

```ts
  if (value !== undefined) {
    normalizedParams.append(key, value === null ? 'null' : value.toString());
  }
});

// => ?parameter=a,b,c
```
#### When set `true` to `explode`:

The generated processing is as follows:

```diff
Object.entries(params || {}).forEach(([key, value]) => {
+  if (value instanceof Array) {
+    value.forEach((v) => normalizedParams.append(key, v === null ? 'null' : v.toString()));
+  }
    
  if (value !== undefined) {
    normalizedParams.append(key, value === null ? 'null' : value.toString())
  }
});

// => parameter=a&parameter=b&parameter=c
```

## Related PRs

This is a PR derived from #1580 and solves the same problem

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

You can check by i added test case.


soartec-lab:
@n2k3 
I created a new PR based on #1580 that you created. Does control with `explode` property solve your problem?

melloware (approved):

PR #1639, opened 2024-09-23
fix(msw): Incorrect Import Generated for Enum in Mock Files

### LINKED ISSUE #1620: MSW: Incorrect Import Generated for Enum in Mock Files
**Description**:  
There is an issue with generating correct imports for enums in the MSW (Mock Service Worker) files. The problem occurs when processing the following enum in the Swagger specification:

```json
"Domain.Status.Enum": {
  "enum": [
    "Foo",
    "Bar"
  ],
  "type": "string"
}
```

### Current Behavior:
The generated import in the mock file looks like this:
```js
import { Domain.Status.Enum } from '.././model';
```

### Expected Behavior:
The import should follow the correct naming convention:
```js
import { DomainStatusEnum } from '.././model';
```

### Suspected Cause:
This issue might be related to a previously fixed issue in PR #1456  (`fix(msw): correctly add imports for enum references`), which addressed a similar problem with enum imports. It is possible that the current bug is a variation or an edge case that wasn't covered by that fix.

### Steps to Reproduce:
1. Use the given Swagger spec that contains the `Domain.Status.Enum`.
2. Generate the mocks (MSW) from the Swagger file.
3. Check the generated imports in the MSW files.

### Environment:
- Version: 7.1.0

### Additional Information:
Any guidance or fixes would be appreciated, and I'm happy to provide more details if necessary.

AlmarAubel:
## Status
fix #1620

**WIP** 

## Description

Fixes the issue mention in #1620. It now imports like this `import { DomainStatusEnum } from '.././model';` instead of `import { Domain.Status.Enum } from '.././model';`

## Todos
It is realy difficult to check if I didn't broken anything after the change. When i ran  generate api  on vue-query sample it looked good and only added the  new correct import statments. Thats why i marked it as WIP. If you have tips to test it better please let me know.

### DIFF
--- a/packages/mock/src/faker/resolvers/value.ts
+++ b/packages/mock/src/faker/resolvers/value.ts
@@ -89,7 +89,7 @@ export const resolveMockValue = ({
 
     const newSchema = {
       ...schemaRef,
-      name: originalName,
+      name: pascal(originalName),
       path: schema.path,
       isRef: true,
     };
--- a/samples/vue-query/vue-query-basic/petstore.yaml
+++ b/samples/vue-query/vue-query-basic/petstore.yaml
@@ -116,6 +116,12 @@ paths:
           description: Success
 components:
   schemas:
+    Domain.Status.Enum:
+      type: string
+      enum:
+        - new
+        - sold
+        - uknown
     Pet:
       type: object
       required:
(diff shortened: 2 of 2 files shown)


melloware:
cc @dteske25

melloware (approved):

PR #1659, opened 2024-10-09
fix: required of parent schema not considered in allOf #1570

### LINKED ISSUE #1570: Schema allOf generation does not follow Swagger spec

I want to use Orval to generate an API that uses inheritance. But, my subclass' fields are being generated as optional even though I have marked them as required.

1. Use Orval to generate models using the following Yaml:
```
openapi: 3.0.1
info:
  contact:
    email: xxx
    name: xxx
    url: xxx
  description: xxx
  title: xxx
  version: '0.1'
servers:
  - url: http://localhost:8080
    description: Generated server url
paths:
  /api/notifications/mine:
    get:
      operationId: getMyNotifications
      responses:
        '200':
          content:
            '*/*':
              schema:
                type: array
                items:
                  oneOf:
                    - $ref: '#/components/schemas/NewTicketNotification'
          description: OK
      summary: Get my notifications
      tags:
        - Notifications
components:
  schemas:
    NewTicketNotification:
      required:
        - ctime
        - id
        - ticketId
        - type
        - userId
      type: object
      allOf:
        - $ref: '#/components/schemas/Notification'
        - type: object
          properties:
            ticketId:
              type: integer
              format: int32
            userId:
              type: integer
              format: int32
    Notification:
      required:
        - ctime
        - id
        - type
      type: object
      properties:
        ctime:
          type: string
          format: date-time
        id:
          type: integer
          format: int32
        type:
          type: string
      discriminator:
        propertyName: type
```

## What happens?

Orval generates the following:

```
export interface Notification {
  ctime: string;
  id: number;
  type: string;
}

export type NewTicketNotificationAllOf = {
  ticketId?: number;
  userId?: number;
};

export type NewTicketNotification = Notification & NewTicketNotificationAllOf;
```

The fields in the subclass `NewTicketNotificationAllOf` should be required. E.g.
```
export type NewTicketNotificationAllOf = {
  ticketId: number;
  userId: number;
};
```

They actually are generated as required if I manually add `required` in the properties of the allOf. E.g.
```
NewTicketNotification:
      required:
        - ctime
        - id
        - ticketId
        - type
        - userId
      type: object
      allOf:
        - $ref: '#/components/schemas/Notification'
        - type: object
          required:
          - ticketId
          - userId
          properties:
            ticketId:
              type: integer
              format: int32
            userId:
              type: integer
              format: int32
```

But I shouldn't have to do that! `NewTicketNotification` already has `ticketId` declared as required. This should supersede the non-required property included in the allOf.

## Any other comments?
Swagger's editor and codegen tool interpret the YAML correctly so I think I have interpreted the spec correctly! (See pictures)

I presume the parser needs a change to use the oneOf/allOf/anyOf key to make properties required (if needed) or leave them alone if they are already required.

image
image

Latest code built from Github.

System:
    OS: Linux 6.5 Ubuntu 22.04.4 LTS 22.04.4 LTS (Jammy Jellyfish)
    CPU: (12) x64 AMD Ryzen 5 3600X 6-Core Processor
    Memory: 2.69 GB / 15.55 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash

MGabr:
## Status

**READY**

## Description

Fixes #1570

Given a schema like

````yaml
components:
  schemas:
    NewTicketNotification:
      required:
        - ctime
        - id
        - ticketId
        - type
        - userId
      type: object
      allOf:
        - $ref: '#/components/schemas/Notification'
        - type: object
          properties:
            ticketId:
              type: integer
              format: int32
            userId:
              type: integer
              format: int32
````

the `allOf` was considered as schema on its own when resolving the value of the fields, but in fact we want to consider the `required` fields of the parent schema also for the `allOf` sub schema.

This fixes this by including the `required` field of the parent schema in the sub schema.

## Related PRs

There is this existing outdated PR: https://github.com/orval-labs/orval/pull/861

## Todos

- [x] Tests
- [ ] Documentation

## Steps to Test or Reproduce

```bash
> git clone https://github.com/MGabr/orval.git
> yarn build
> cd tests
> yarn generate:react-query
```

The TypeScript model generated for the new `polymorphic-required.yaml` test spec should have the required fields set correctly. 

So for 

````yaml
components:
  schemas:
    DescendantOne:
      required:
        - type
        - count
      type: object
      allOf:
        - $ref: '#/components/schemas/ParentType'
        - type: object
          properties:
            value:
              type: boolean
            count:
              type: integer
              format: int32
            otherCount:
              type: integer
              format: int32
````

it generates

````typescript
export type DescendantOneAllOf = {
  count: number;
  otherCount?: number;
  value?: boolean;
};
````

### CODE REVIEW

**Review by melloware — APPROVED** [review]

**Review by melloware — APPROVED** [review]

**Comment by MGabr** [comment]
I think this still needs to be adjusted to handle edge cases where both the parent and the sub schema have a `required` field. 

Like:

```
components:
  schemas:
    DescendantOne:
      required:
        - type
      type: object
      allOf:
        - $ref: '#/components/schemas/ParentType'
        - type: object
           required:
            - count
          properties:
            value:
              type: boolean
            count:
              type: integer
              format: int32
            otherCount:
              type: integer
              format: int32
```

Currently only the parent `required` would be considered but I think both `required` should be considered in this case. Still to check.

**Review by melloware — APPROVED** [review]

File: tests/specifications/polymorphic-required.yaml [hunk 1]
```diff
@@ -0,0 +1,59 @@
+openapi: 3.0.1
```
soartec-lab: could you merge the yaml of these test files? 
I don't wanna separeate related tests.

soartec-lab (commented):
@MGabr 
Thanks for nice update. i added a comment.

tests/specifications/polymorphic-required.yaml
MGabr: Sure, I merged those two into the existing `polymorphic.yaml` file now

melloware (approved):

soartec-lab (approved):
Thanks!

Issue #1472, opened 2024-06-20
mock does not generate correct bigint type


thegenius:
when field type is bigint, mock will use faker.number.int() to generate number value, which cause error.

thegenius:
and it will be better to generate mock in a separated file

melloware:
PR is welcome i would think this would be easy fix for bigint

thegenius:
As I have read the source code of the mock package, it's not so hard to fix this.
But it's a big problem to create a correct PR, but there is no tests with mock,
I can't find any test code for mock, so if it be possible to create a basic tests suite.
And after we get test suite, then we can create correct PR to help this project.

soartec-lab:
Hi @thegenius, Thanks for making a good issue.

We have tests in the `tests` directory.
In this case, I think you can add a test by adding the `fomat: int64` property to the [`tests/specifications/format](https://github.com/orval-labs/orval/blob/master/tests/specifications/format.yaml).yaml` schema. Would you work on this?

PR #1675, opened 2024-10-25
fix: faker bigint support

Marfyy:
## Status

**READY**

## Description

Fix https://github.com/orval-labs/orval/issues/1472
If orval config uses override useBigInt, faker will be generated with faker.number.bigInt. I also added support for format double/float, to use faker.number.float. if useBigInt is not set, it will default back to faker.number.int

I appended the existing `tests/specifications/format.yaml`, and changed the name in `tests/configs/mock.config.ts` from **useDates**, to **formats**, to better reflect its content with the new test scenarios

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

Running locally against `tests/configs/mock.config.ts`, the config **formats** tests to generate the new number type/format combinations

### DIFF
--- a/packages/mock/src/faker/getters/scalar.ts
+++ b/packages/mock/src/faker/getters/scalar.ts
@@ -122,7 +122,7 @@ export const getMockScalar = ({
     case 'number':
     case 'integer': {
       let value = getNullable(
-        `faker.number.int({min: ${item.minimum}, max: ${item.maximum}})`,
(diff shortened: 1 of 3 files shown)


melloware (approved):
awesome!

soartec-lab:
@Marfyy @melloware 

The `format` and its `faker` function are managed below: 

https://github.com/orval-labs/orval/blob/master/packages/mock/src/faker/constants.ts

Can we change it to be managed by constants instead of function?

melloware:
good question @Marfyy can you take a look?

soartec-lab:
I fixed this by #1697

Issue #773, opened 2023-02-22
Generate Documentation


mwillbanks:
## Feature Request

Provide documentation generation for README.md for all of the generated methods and use to simplify developer documentation.  OpenAPI Generator does this on several of their implementations which is quite useful to the end developer rather than having to install the generated methods and hunt and peck for which methods they need to be utilizing.

This might be made easier by providing a hook for each method that is created.

PR #1705, opened 2024-11-15
feat: generate API as `TypeDoc` docs

huwshimi:
## Status

**READY**

## Description

Generate docs from the output using TypeDoc.

Fixes: #773.

## Related PRs

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout <branch>
> grunt jasmine
```

1. Open the `orval.config.ts` for an example and add `docs: true,` to the `output` object.
2. Run `yarn generate-api`.
3. The markdown docs should get built alongside the API files.

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -394,13 +394,15 @@ Default Value: `false`.
 
 Will generate API docs using [TypeDoc](https://typedoc.org/). by default these docs will be in Markdown format.
 
-TypeDoc can be configured by creating a config file e.g. `typedoc.config.mjs` in your project root (see the [config docs](https://typedoc.org/options/configuration/#options) for a full list of supported file names) or by passing a config filename to the `config` option below.
+TypeDoc can be configured by passing the [options](https://typedoc.org/options/) to the `docs` object or by creating a config file e.g. `typedoc.config.mjs` in your project root (see the [config docs](https://typedoc.org/options/configuration/#options) for a full list of supported file names) or by passing a config filename to the `configPath` option below.
 
 See the TypeDoc [configuration documentation](https://typedoc.org/options/) for more details.
 
 The `docs` option can take some properties to customize the generation if you set it to an object. If you set it to `true`, the default options will be used.
 
-#### config
+When no output directory destination is specified in `config`, the file will be output to the `docs` directory by default.
+
+#### configPath
 
 Type: `String`.
 
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -386,6 +386,26 @@ Default Value: `en`.
 
 Give you the possibility to set the locale for the mock generation. It is used by faker, see the list of available options [here](https://fakerjs.dev/guide/localization.html#available-locales). It should also be strongly typed using `defineConfig`.
 
+### docs
+
+Type: `Boolean | Object`.
+
+Default Value: `false`.
+
+Will generate API docs using [TypeDoc](https://typedoc.org/). by default these docs will be in Markdown format.
+
+TypeDoc can be configured by creating a config file e.g. `typedoc.config.mjs` in your project root (see the [config docs](https://typedoc.org/options/configuration/#options) for a full list of supported file names) or by passing a config filename to the `config` option below.
+
+See the TypeDoc [configuration documentation](https://typedoc.org/options/) for more details.
+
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -49,7 +49,7 @@
     "tsup": "^8.0.2",
     "turbo": "^1.13.3",
     "typescript": "^4.7.4",
-    "vitest": "^0.6.3",
+    "vitest": "0.6.3",
```
huwshimi, line 52: These were getting resolved to a much newer version due to the new dependencies so needed to pin them to prevent breakage..

melloware (approved):
Slick!

soartec-lab:
sounds good! I'll try this later.

packages/orval/src/write-specs.ts
```diff
@@ -198,6 +199,51 @@ export const writeSpecs = async (
...
+          // Use the output location if it has been set in the external config.
+          out = app.options.getValue('out');
+        } else if (output.workspace) {
+          // Generate the docs in the workspace.
+          out = upath.join(output.workspace, 'docs');
+        } else if (output.target) {
+          const base = upath.dirname(output.target);
+          // Generate the docs along side the output target.
+          out = upath.join(base, 'docs');
+        }
```
soartec-lab: I thought that the default specified location should be `docs` regardless of `workspace` or `target`.

These options are not documents, but output destinations for the application's HTTP client, so their nature is different from the document we will be adding this time.
Also, by not supporting those followers, we get the simplicity of the source code.

```suggestion
        const out = app.options.getValue('out') || 'docs';
```

docs/src/pages/reference/configuration/output.md
```diff
@@ -386,6 +386,26 @@ Default Value: `en`.
...
+Default Value: `false`.
+
+Will generate API docs using [TypeDoc](https://typedoc.org/). by default these docs will be in Markdown format.
+
+TypeDoc can be configured by creating a config file e.g. `typedoc.config.mjs` in your project root (see the [config docs](https://typedoc.org/options/configuration/#options) for a full list of supported file names) or by passing a config filename to the `config` option below.
+
+See the TypeDoc [configuration documentation](https://typedoc.org/options/) for more details.
+
+The `docs` option can take some properties to customize the generation if you set it to an object. If you set it to `true`, the default options will be used.
+
```
soartec-lab, line 402: I think it would be easier to understand if I could know in advance where the document would be generated 👍 

```suggestion
When no output directory destination is specified in `config`, the file will be output to the `docs` directory by default.
```

soartec-lab (commented):
@huwshimi 

This is amazing. I have made some comments for improvement, so please check them out 👍

packages/core/src/types.ts
```diff
@@ -239,6 +241,10 @@ export const OutputMode = {
 
 export type OutputMode = (typeof OutputMode)[keyof typeof OutputMode];
 
+export type OutputDocsOptions = {
+  config: string;
```
soartec-lab: I thought that being able to define `TypeDocOptions` directly here instead of in a configuration file might have the advantage of reducing the number of configuration files to manage.
Is it okay to explicitly indicate that it is a file path, considering the expandability of the option in the future?

```suggestion
  configPath: string;
```

packages/orval/src/write-specs.ts
huwshimi: Thanks. `app.options.getValue('out')` defaults to `docs` inside TypeDoc so we can simplify this slightly further and drop the ` || 'docs'`

docs/src/pages/reference/configuration/output.md
huwshimi: Done.

packages/core/src/types.ts
huwshimi: I've renamed the option to `configPath` and also added support for passing the TypeDoc options directly inside the `docs` object.

melloware (dismissed):

packages/core/src/types.ts
soartec-lab: That's nice!

soartec-lab:
@huwshimi 
The test has failed due to the release of v7.3.0. Can I rebase it?
I'll do this as soon as I can see that.

huwshimi:
@soartec-lab I rebased and fixed the merge conflict. Tests are passing now.

soartec-lab (approved):
It's all good, Thanks!

PR #1718, opened 2024-11-28
feat(core): take servers fields into account

### LINKED ISSUE #1689: Support for OpenAPI servers field

Generate from Lichess [api spec](https://lichess.org/api)

## What happens?

There is no way to know what endpoint to use, as the spec has different endpoints for many api calls, specified by the `servers` field.

A way to use the servers field.

-

## Any other comments?

Unsure how this best would be handled? Maybe let there be an optional parameter to be sent to the function doing the api call, which could be the `servers` data?

AllieJonsson:
## Status

**READY**

## Description

Fix #1689 
This allows the user to specify whether or not they want to use the servers field as specified in the openapi spec.
The user can still use the old way of defining the baseUrl as a string for backwards compatibility. 
If the servers field contains a variable to be substituted, Orval tries to substitute it using the value that the user has defined in the configuration file. If the user has not defined that variable, the default value is used.

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

1. Define output.baseUrl in your config as 
```ts
baseUrl: {
  getBaseUrlFromSpecification: true
}
```
2. Define servers in your specification, e.g.
```yaml
servers:
  - url: https://api.example.com
```
3. Generate
4. Notice that the url in the generated getUrl functions (e.g. `getCreatePetUrl`) returns `'https://api.example.com/pets'`

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -296,6 +296,163 @@ module.exports = {
 };
 ```
 
+### baseUrl
+
+Type: `String | Object`.
+
+Default Value: `''`.
+
+Allows you to set the baesUrl used for all API calls. This can either be a constant string or be configured to read from the `servers` field
+in the specification.
+
+Example using constant:
+
+```js
+module.exports = {
+  petstore: {
+    output: {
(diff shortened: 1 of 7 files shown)


melloware (approved):

melloware:
Excellent work!

PR #1722, opened 2024-11-29
feat: handle ndjson response type in fetch client

### LINKED ISSUE #1721: Support for ndjson

1. Generate using the [Lichess api spec](https://lichess.org/api)

## What happens?

Some endpoints are streamed using [NDJson](https://en.wikipedia.org/wiki/JSON_streaming), but Orval does not support streaming this.

Code generated where the stream is read properly and response is typed correctly

## Any other comments?

I'll create a PR with how my thoughts of how this could be implemented

AllieJonsson:
## Status

**READY**

## Description

Fixes #1721 
This makes sure that when the response type is `application/nd-json` or `application/x-ndjson` we return the whole response instead of parsing json.

## Todos

- [x] Tests
- [ ] Documentation


AllieJonsson:
The example in the comment above yields the followin result:
```ts
import type { TeamIdUsers200, TeamIdUsersParams } from '.././model';

const readAggregateStream = <T>(
  response: Response,
  processLine?: (value: T) => void,
) => {
  if (!response.ok || !response.body) return Promise.reject();
  const stream = response.body.getReader();
  const matcher = /\r?\n/;
  const decoder = new TextDecoder();
  let buffer: string = '';
  const aggregate: T[] = [];

  const loop: () => Promise<T[]> = () =>
    stream.read().then(({ done, value }) => {
      if (done) {
        if (buffer.length > 0) {
          const parsed = JSON.parse(buffer);
          aggregate.push(parsed);
          processLine?.(parsed);
        }
        return Promise.resolve(aggregate);
      } else {
        buffer += decoder.decode(value, {
          stream: true,
        });
        const parts = buffer.split(matcher);
        buffer = parts.pop() ?? '';
        for (const part of parts) {
          if (!part) continue;
          const parsed = JSON.parse(part);
          aggregate.push(parsed);
          processLine?.(parsed);
        }

        return loop();
      }
    });

  return loop();
};

/**
 * Members are sorted by reverse chronological order of joining the team (most recent first).
OAuth is only required if the list of members is private.
Up to 5,000 users are streamed as [ndjson](#section/Introduction/Streaming-with-ND-JSON).

 * @summary Get members of a team
 */
export type teamIdUsersResponse = {
  data: Promise<TeamIdUsers200[]>;
  status: number;
  headers: Headers;
};

export const getTeamIdUsersUrl = (
  teamId: string,
  params?: TeamIdUsersParams,
) => {
  const normalizedParams = new URLSearchParams();

    if (value !== undefined) {
      normalizedParams.append(key, value === null ? 'null' : value.toString());
    }
  });

  return normalizedParams.size
    ? `https://lichess.org/api/team/${teamId}/users?${normalizedParams.toString()}`
    : `https://lichess.org/api/team/${teamId}/users`;
};

export const teamIdUsers = async (
  teamId: string,
  params?: TeamIdUsersParams,
  options?: RequestInit,
  onMessage?: (value: TeamIdUsers200) => void,
): Promise<teamIdUsersResponse> => {
  const res = await fetch(getTeamIdUsersUrl(teamId, params), {
    ...options,
    method: 'GET',
  });

  const data = readAggregateStream(res, onMessage);

  return { status: res.status, data, headers: res.headers };
};

```

AllieJonsson:
With `aggregateResult = false`, this is the result:
```ts
import type { TeamIdUsers200, TeamIdUsersParams } from '.././model';

const readStream = <T>(response: Response, processLine: (value: T) => void) => {
  if (!response.ok || !response.body) return Promise.reject();
  const stream = response.body.getReader();
  const matcher = /\r?\n/;
  const decoder = new TextDecoder();
  let buffer: string = '';

  const loop: () => Promise<void> = () =>
    stream.read().then(({ done, value }) => {
      if (done) {
        if (buffer.length > 0) processLine(JSON.parse(buffer));
        return Promise.resolve();
      } else {
        buffer += decoder.decode(value, {
          stream: true,
        });
        const parts = buffer.split(matcher);
        buffer = parts.pop() ?? '';
        for (const part of parts) {
          if (!part) continue;
          processLine(JSON.parse(part));
        }

        return loop();
      }
    });

  return loop();
};

/**
 * Members are sorted by reverse chronological order of joining the team (most recent first).
OAuth is only required if the list of members is private.
Up to 5,000 users are streamed as [ndjson](#section/Introduction/Streaming-with-ND-JSON).

 * @summary Get members of a team
 */
export type teamIdUsersResponse = {
  data: Promise<void>;
  status: number;
  headers: Headers;
};

export const getTeamIdUsersUrl = (
  teamId: string,
  params?: TeamIdUsersParams,
) => {
  const normalizedParams = new URLSearchParams();

    if (value !== undefined) {
      normalizedParams.append(key, value === null ? 'null' : value.toString());
    }
  });

  return normalizedParams.size
    ? `https://lichess.org/api/team/${teamId}/users?${normalizedParams.toString()}`
    : `https://lichess.org/api/team/${teamId}/users`;
};

export const teamIdUsers = async (
  onMessage: (value: TeamIdUsers200) => void,
  teamId: string,
  params?: TeamIdUsersParams,
  options?: RequestInit,
): Promise<teamIdUsersResponse> => {
  const res = await fetch(getTeamIdUsersUrl(teamId, params), {
    ...options,
    method: 'GET',
  });

  const data = readStream(res, onMessage);

  return { status: res.status, data, headers: res.headers };
};

```

melloware:
I know nothing about ND JSON so I will let @anymaniax and @soartec-lab review!

soartec-lab:
Good. I'll handle this later.

AllieJonsson:
I'm looking at axios now, they have a setting `responseType`, which can be set to `stream`. Does anyone have more knowledge about this? Their documentation sucks... Im mostly wondering if we should handle axios the same way as fetch, or if axios has something built in that we can use

AllieJonsson:
Leaving axios for now, maybe it is possible to use interceptors or whatever if the user wants to get information about each line while processing

soartec-lab:
@AllieJonsson 

Thank you for the great suggestion.
I think it makes sense to support NDJSON. However, I do not agree with including the generic function itself in orval.
Becouse that can be processed by calling `Response` of `fetch API`, so you should use it in your project as follows if necessary.

```ts
  const { status, data, headers } = await teamIdUsers(teamId, params, options);
  if (status === 200) {
    // process stream data
  }
````

**Comment by AllieJonsson** [comment]
Thank you @soartec-lab! I think that makes sense, I felt Midway that what i was doing was not perfect. Doing it the way you wrote is better, Ill get on trying to make the datatype more understandable when the response type is ndjson

**Comment by AllieJonsson** [comment]
@soartec-lab Please see changed code, I reverted most of the changes :)

**Comment by soartec-lab** [comment]
thank you. Please give us a few days to see this.

File: docs/src/pages/reference/configuration/output.md [hunk 1]
```diff
@@ -874,6 +874,7 @@ module.exports = {
 ##### includeHttpResponseReturnType
 
 Type: `Boolean`.
+
```
soartec-lab: Is this an unnecessary line break?

packages/core/src/types.ts
```diff
@@ -164,6 +164,11 @@ export type OutputClientFunc = (
   clients: GeneratorClients,
 ) => ClientGeneratorsBuilder;
 
+export type AggregateResultOptions = {
```
soartec-lab: I don't think it's used anywhere.

packages/core/src/writers/split-mode.ts
```diff
@@ -8,9 +8,9 @@ import {
   isSyntheticDefaultImportsAllow,
   upath,
 } from '../utils';
+import { getMockFileExtensionByTypeName } from '../utils/fileExtensions';
```
soartec-lab: Is this an unnecessary difference? It looks like you haven't changed anything

packages/core/src/writers/split-tags-mode.ts
```diff
@@ -1,4 +1,5 @@
 import fs from 'fs-extra';
+import uniqBy from 'lodash.uniqby';
```
soartec-lab: This too

packages/swr/src/index.ts
```diff
@@ -14,23 +14,23 @@ import {
   GetterProps,
   GetterPropType,
   GetterResponse,
+  jsDoc,
```
soartec-lab: This too

tests/configs/fetch.config.ts
```diff
@@ -84,7 +84,7 @@ export default defineConfig({
       client: 'fetch',
       override: {
         fetch: {
-          includeHttpStatusReturnType: false,
+          includeHttpResponseReturnType: false,
```
soartec-lab, line 87: Thanks!

packages/fetch/src/index.ts
```diff
@@ -104,15 +104,19 @@ ${
...
-    response.definition.success,
+    isNdJson ? 'Response' : response.definition.success,
     operationName,
   );
+
   const responseTypeImplementation = override.fetch
     .includeHttpResponseReturnType
     ? `export type ${responseTypeName} = {
-  data: ${response.definition.success || 'unknown'};
+  ${isNdJson ? `stream: Response // type: ${response.definition.success || 'unknown'}` : `data: ${response.definition.success || 'unknown'}`};
```
soartec-lab: I don't want to add type annotation comments because they are implicit and cause confusion, making it difficult to understand what is intended. If you want to know the contents of type, you should add a comment as a new line above the stream definition.

soartec-lab (commented):
@AllieJonsson 

Thanks for the change. Please check that I have added some comments.

packages/fetch/src/index.ts
anymaniax: Completely agree with @soartec-lab

soartec-lab (commented):
@AllieJonsson 
Thanks for the changes.
It looks like all your fixes are done, are there any other changes you have planned?

AllieJonsson:
> @AllieJonsson Thanks for the changes. It looks like all your fixes are done, are there any other changes you have planned?

I have no more changes planned, so this is ready to be merged!

melloware (approved):

PR #1735, opened 2024-12-04
feat: TData generic for mutation hooks

lentyaevpk:
## Status

**READY**

## Description

Right now we have an opportunity to override the type of `data` that is returned by the generated `useQuery` hooks using `TData` generic. It would be great to have the same opportunity with the `useMutation` hooks.

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

You can use samples tests

### DIFF
--- a/packages/query/src/index.ts
+++ b/packages/query/src/index.ts
@@ -481,9 +481,7 @@ const getQueryOptionsDefinition = ({
         > , 'initialData'
       >`
         : '';
-    const optionType = `${prefix}${pascal(
-      type,
-    )}Options<TData, TError, TData${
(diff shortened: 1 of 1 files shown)


anymaniax:
Seems good but it's a breaking change. I don't think it's will break for a lot of people and case. I am gonna do deeper check in the following the days

anymaniax (approved):
it's good for me but it's still a little breaking change for some people

melloware (approved):
LGTM

soartec-lab:
@anymaniax 
I agree with improvements that make it more type safe 👍

PR #1749, opened 2024-12-16
fix/swr infinite key

### LINKED ISSUE #1747: Infinite Pagination Key Loader Missing Page Index in SWRInfinite
**What are the steps to reproduce this issue?**
 - Use Orval to generate an API client for an endpoint with infinite pagination.
 - Implement useSWRInfinite for fetching pages.
 - When attempting to fetch the next page, observe that the same key is used for all pages, causing data to not load properly.

**What happens?**
The same SWR key is used for all pages, which prevents additional pages from being loaded. When modifying the params to include the page number, the data is reset and starts fetching from the first page again.

**What were you expecting to happen?**
Each page should have a unique SWR key that includes the pageIndex, allowing for proper infinite scrolling and page-by-page fetching.

**Any logs, error output, etc?**
No error output, but the issue is related to incorrect SWR key generation for infinite pagination.

**Any other comments?**
The issue only occurs when the pagination key does not consider the pageIndex, causing problems with subsequent page fetching and resetting the data when params are modified.

**What versions are you using?**
 - OS: Windows 10 10.0.19045
-   CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
-   "orval": "^7.2.0"
-   "swr": "^2.2.5"

**Example from orval**

    export const getGetAllStandardsInfiniteKeyLoader = (params?: GetAllStandardsParams) => {
      return (_: number, previousPageData: Awaited<ReturnType<typeof getAllStandards>>) => {
        if (previousPageData && !previousPageData.data) return null;
    
        return ['MY_WEBSITE_PATH', ...(params ? [params] : [])] as const;
      };
    };
**Possible Solution** I’m able to solve this issue by updating the SWR key to consider the `pageIndex` for each page, ensuring that the API properly loads subsequent pages.

AsselAlAssel:
## Status

**READY/WIP/HOLD**

Fix #1747

## Description

A few sentences describing the overall goals of the pull request's commits.

## Related PRs

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout <branch>
> grunt jasmine
```

1.

### DIFF
--- a/packages/swr/src/index.ts
+++ b/packages/swr/src/index.ts
@@ -426,10 +426,10 @@ export const ${swrKeyFnName} = (${queryKeyProps}) => [\`${route}\`${
     );
     const swrKeyLoader = override.swr.useInfinite
       ? `export const ${swrKeyLoaderFnName} = (${queryKeyProps}) => {
-  return (_: number, previousPageData: Awaited<ReturnType<typeof ${operationName}>>) => {
+  return (page: number, previousPageData: Awaited<ReturnType<typeof ${operationName}>>) => {
     if (previousPageData && !previousPageData.data) return null
 
-    return [\`${route}\`${queryParams ? ', ...(params ? [params]: [])' : ''}${
+    return [\`${route}\`${queryParams ? ', ...(params ? [{...params,page}]: [{page}])' : ''}${
       body.implementation ? `, ${body.implementation}` : ''
     }] as const;
   }


soartec-lab:
@AsselAlAssel @melloware 
`pageIndex` is intentionally not implemented. This is to allow for flexible handling by receiving them as query parameters.
For example, for cursors or other names.
Can't you get around this problem by defining the query parameters in the `OpenAPI` definition?

melloware:
@soartec-lab feel free to revert.  I don't use SWR so this seemed like a harmless change but i see the original design was intentional.

PR #1763, opened 2024-12-21
feat(core): property sorting

### LINKED ISSUE #1737: Add `sortFields = true|false` to disable sorting when generating the model
Orval generates `*.ts` model files from a given `openapi.yml`. However, the fields within the `*.ts` model interface/type are always sorted implicitly instead of using the field order defined in `openapi.yml`. For example:

1.`openapi.yml` has a `User` endpoint with the following

```
openapi: 3.1.0
components:
  schemas:
    User:
      type: object
      properties:
        id:
          type: string
        username:
          type: string
        password:
          type: string
```

2. let orval generate the `user.ts` and look at the ordering of the fields:

```
// overal sorts the fields in alphabetical order
export interface User {
  id?: string;
  password?: string;
  username?: string;
}
```

Please add a configuration to https://orval.dev/reference/configuration/output, e.g. `sortFields = true|false`. If set to `false`, the field order should match the order defined in `openapi.yml`. In the above example, it should be:

```
// sortFields=false, i.e. overal uses the sorting order as defined in openapi.yml
export interface User {
  id?: string;
  username?: string;
  password?: string;
}
```

Orval v 7.3.0

AllieJonsson:
## Status

**READY/HOLD**

## Description

Fixes #1737 
Not sure if this is a good way to go, ordering of properties seems irrelevant in the end (no functionality difference), so I'm not sure if this is what we want. I'll let you Maintainers decide :)

## Todos

- [ ] Tests
- [x] Documentation

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -2087,14 +2087,14 @@ module.exports = {
 
 Type: `Alphabetical` | `Specification`
 
-Default Value: `Alphabetical`
+Default Value: `Specification`
 This enables you to specify how properties in the models are sorted, either alphabetically or in the order they appear in the specification.
 
 ```js
 module.exports = {
   petstore: {
     output: {
-      propertySortOrder: 'Specification',
+      propertySortOrder: 'Alphabetical',
     },
   },
 };
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -2082,3 +2082,20 @@ module.exports = {
   },
 };
 ```
+
+### propertySortOrder
+
+Type: `Alphabetical` | `Specification`
+
+Default Value: `Alphabetical`
+This enables you to specify how properties in the models are sorted, either alphabetically or in the order they appear in the specification.
+
+```js
+module.exports = {
+  petstore: {
+    output: {
+      propertySortOrder: 'Specification',
+    },
+  },
+};
+```
--- a/packages/core/src/getters/object.ts
+++ b/packages/core/src/getters/object.ts
@@ -2,6 +2,7 @@ import { ReferenceObject, SchemaObject } from 'openapi3-ts/oas30';
 import { resolveExampleRefs, resolveObject, resolveValue } from '../resolvers';
 import {
   ContextSpecs,
+  PropertySortOrder,
   ScalarValue,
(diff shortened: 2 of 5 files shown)


packages/core/src/getters/object.ts
```diff
@@ -70,98 +71,100 @@ export const getObject = ({
...
-          }
-
-          const allSpecSchemas =
-            context.specs[context.target]?.components?.schemas ?? {};
-
-          const isNameAlreadyTaken = Object.keys(allSpecSchemas).some(
-            (schemaName) => pascal(schemaName) === propName,
+      });
+    }
+    return entries.reduce(
```
AllieJonsson, line 80: This is the only change here, the other is just whitespace changes

packages/core/src/getters/object.ts
```diff
@@ -70,98 +71,100 @@ export const getObject = ({
   }
 
   if (item.properties && Object.entries(item.properties).length > 0) {
-    return Object.entries(item.properties)
-      .sort((a, b) => {
+    const entries = Object.entries(item.properties);
+    if (context.output.propertySortOrder === PropertySortOrder.ALPHABETICAL) {
```
AllieJonsson, line 75: This is the only difference here, the rest is just whitespace changes

melloware (dismissed):
@anymaniax @soartec-lab thoughts?

docs/src/pages/reference/configuration/output.md
```diff
@@ -2082,3 +2082,20 @@ module.exports = {
   },
 };
 ```
+
+### propertySortOrder
+
+Type: `Alphabetical` | `Specification`
+
+Default Value: `Alphabetical`
```
anymaniax: It would make more sense to have specification by default no?

AllieJonsson: I mean, it would, yes, but then why were we sorting alphabetically from the beginning? You get the specification sort by not doing any sorting

soartec-lab:
That's a good improvement.
It's great to make the sorting rules explicit and consistent.
In my opinion, it's better to be consistent with the order defined in the open API.

AllieJonsson:
Ill change the default to be Specification instead!

soartec-lab (approved):
Thanks for great enhance!

melloware (approved):

PR #1768, opened 2024-12-26
fix(fetch): make sure to not double promise for custom fetch mutator

### LINKED ISSUE #1743: Problem with generic type with custom-fetch

1. Create a custom fetch similar to the example in `next-app-with-fetch` [sample](https://github.com/orval-labs/orval/blob/master/samples/next-app-with-fetch/custom-fetch.ts)
2. Generates code, or look at the generated code from the [sample](https://github.com/orval-labs/orval/blob/master/samples/next-app-with-fetch/app/gen/pets/pets.ts)
```
  params?: ListPetsParams,
  options?: RequestInit,
  return customFetch<Promise<listPetsResponse>>(getListPetsUrl(params), {
    ...options,
    method: 'GET',
  });
};
```
3. Look at the type passed for the customFetch in the generated code. If you have vscode, you can hover on your generated function to see the return type of it. 
4. 

## What happens?

Custom-fetch receives a `Promise<listPetsResponse>` in place of the generic `T`. 

Only to pass `listPetsResponse` in place of the generic `T`. Custom-fetch has to await so its expected signature will be `Promise<T>`. If you pass  `Promise<listPetsResponse>`, you end up with functions that have returned type Promise<Promise<listPetsResponse>>`

No errors, I think it is just incorrect in term of type. 

```
  System:
    OS: macOS 15.1.1
    CPU: (8) arm64 Apple M1
    Memory: 76.56 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  npmPackages:
    msw: ^2.6.5 => 2.6.5 
    orval: ^7.1.1 => 7.1.1 
```

AllieJonsson:
## Status

**READY**

## Description

Fixes #1743 
This specific code is only executed when a custom mutator is specified

### DIFF
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -139,7 +139,7 @@ ${
     .join(',');
 
   const args = `${toObjectString(props, 'implementation')} ${isRequestOptions ? `options?: RequestInit` : ''}`;
-  const retrunType = `Promise<${responseTypeName}>`;
+  const returnType = `Promise<${responseTypeName}>`;
 
   const globalFetchOptions = isObject(override?.requestOptions)
     ? `${stringify(override?.requestOptions)?.slice(1, -1)?.trim()}`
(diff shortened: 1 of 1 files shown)


soartec-lab (approved):
Thanks!

PR #1770, opened 2024-12-26
fix(core): take output.indexFiles setting into account

### LINKED ISSUE #1769: Mode single, tags and split generates code that assumes indexFiles is set to true

1. Generate petstore example using this config:
```ts
petstore: {
  output: {
    target: '../generated/fetch/petstore/endpoints.ts',
    schemas: '../generated/fetch/petstore/model',
    indexFiles: false,
    client: 'fetch',
    mode: 'split', // or 'tags' or 'single'
  },
  input: {
    target: '../specifications/petstore.yaml',
  },
},
```

## What happens?

The following code is generated:
```ts
import type {
  CreatePetsBody,
  CreatePetsParams,
  ListPetsParams,
  Pet,
  Pets
} from './model'
```

```ts
import type {
  CreatePetsBody
} from './model/createPetsBody'
import type {
  CreatePetsParams
} from './model/createPetsParams'
import type {
  ListPetsParams
} from './model/listPetsParams'
import type {
  Pet
} from './model/pet'
import type {
  Pets
} from './model/pets'
```

image

## Any other comments?

PR on the way!

AllieJonsson:
## Status

**READY**

## Description

Fixes #1769 
It appears only split-tags mode checked the `indexFiles` flag, but not anymore!
The linked PR below fixed this issue for mock imports for split-tags, but the change is necessary for all modes, both mock imports and regular imports.

## Related PRs

| AlmarAubel:fix-mock-import-indexFiles | [link](https://github.com/orval-labs/orval/pull/1758) |

## Todos

- [x] Tests
- [ ] Documentation


tests/configs/fetch.config.ts
```diff
@@ -178,4 +178,15 @@ export default defineConfig({
       target: '../specifications/parameters.yaml',
     },
   },
+  noIndexFiles: {
```
soartec-lab: Since `indexFiles` is a global property, not limited to `fetch`, I think it is appropriate to write the test case in `default`.

soartec-lab (changes requested):
Thank you for your response. I have requested one minor correction.

tests/configs/fetch.config.ts
AllieJonsson: Oh right, I have moved it to `default`, please see new commit!

melloware (approved):

packages/core/src/writers/single-mode.ts
```diff
@@ -132,3 +138,15 @@ export const writeSingleMode = async ({
     throw `Oups... 🍻. An Error occurred while writing file => ${e}`;
   }
 };
+
+const generateImports = (
```
soartec-lab: This functionality is currently in `split-tags`, but should be common since it is used in all modes.
You can create a file like `generate-imports-for-builder` in the same directory.
There is already a function with the same name in `utils`, so I want to name the function 　`generateImportsForBuilder` to avoid confusion.

soartec-lab (commented):
@AllieJonsson 

Thanks!
I added a comment for refactoring.

melloware (dismissed):

packages/core/src/writers/single-mode.ts
```diff
@@ -1,13 +1,19 @@
 import fs from 'fs-extra';
+import uniqBy from 'lodash.uniqby';
 import { generateModelsInline, generateMutatorImports } from '../generators';
-import { WriteModeProps } from '../types';
+import {
+  GeneratorImport,
```
soartec-lab: There are still some unnecessary imports around here 👀

AllieJonsson: Oops, will fix

AllieJonsson:
Seems like #1780 broke headers:
image

Headers was already added to the fetch call by the property `fetchHeadersOption`:
image

AllieJonsson:
> Seems like #1780 broke headers: image
> 
> Headers was already added to the fetch call by the property `fetchHeadersOption`: image

I included a fix for this in this PR

melloware (approved):

soartec-lab (approved):
Thanks!

PR #1774, opened 2024-12-27
Fix/all of nested

### LINKED ISSUE #1692: Mocks (for MSW) are wrongly generated with deeply nested objects
**YAML Path**
```
  '/api/tenants/{tenantId}':
    get:
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/TenantComposite'
```
**YAML Schema**
```
    TenantComposite:
      allOf:
        - $ref: '#/components/schemas/Tenant'

    TenantComposite:
      allOf:
        - $ref: '#/components/schemas/Tenant'
        
    Tenant:
      allOf:
        - $ref: '#/components/schemas/TenantLight'
```

## What happens?
In the .msw file, it generates the following:
`export const getGetApiTenantsTenantIdResponseMock = (): TenantCompositeDto => ({...enabled: faker.datatype.boolean(), `

In the .msw file, it generated the following:
`export const getGetApiTenantsTenantIdResponseMock = (): TenantCompositeDto => ({enabled: faker.datatype.boolean(),`

7.2.0

AllieJonsson:
## Status

**READY**

## Description

Fixes #1692 
I'm not sure why this extra check was there, it seems like it's legacy and not valid anymore.

## Todos

- [x] Tests
- [ ] Documentation


AllieJonsson:
Seems like tests are failing, looking at it

soartec-lab:
That's the case test in #1635 that I addressed. It seems to affect other nested polymorphic cases 👀

AllieJonsson:
All these nested polymorhpic cases are coming back 😄 Maybe we should investigate if we need to take out the excavator and rewrite some of this 🫨

PR #1778, opened 2024-12-29
feat(fetch): avoid exception when the response body has no content

### LINKED ISSUE #1656: 204 No content generated code throws JSON parsing error

1. Create new project with `fetch` client from openapi scheme which returns 204 (No content).
2. Try calling the generated method.
3. You should get error Unexpected end of JSON input error.

## What happens?

Basically when there is no response in fetch client response there is no content. Which mean the res.json() will throw error.

Generated code looks like this:

```typescript
const data = await res.json()

return { status: res.status, data }

 ```

There shouldn't be a error.

```
SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at parseJSONFromBytes (node:internal/deps/undici/undici:4306:19)
    at successSteps (node:internal/deps/undici/undici:4288:27)
    at consumeBody (node:internal/deps/undici/undici:4294:9)
    at _Response.json (node:internal/deps/undici/undici:4239:18)
```

## Any other comments?

I think when there is 204 in specs or there is no content in openapi specification the const data should not be generated and data should not be required in generated responses as there are none.

```
  System:
    OS: Linux 6.9 Pop!_OS 22.04 LTS
    CPU: (24) x64 AMD Ryzen 9 7900X 12-Core Processor
    Memory: 9.64 GB / 31.08 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  npmPackages:
    orval: ^7.1.1 => 7.1.1 
    react: rc => 19.0.0-rc-0751fac7-20241002 
    zod: 3.23.8 => 3.23.8 
```

### PR DESCRIPTION
**Pull request by soartec-lab**
## Status

**HOLD**

 I'm waiting for more options to gather in #1656 

## Description

fix #1656

This avoids an exception that occurs when executing `res.json()` if there is no content in the response body.

```ts
  params?: ListPetsParams,
  options?: RequestInit,
    ...options,
    method: 'GET',
  });

-  const data: Pets = await res.json();
+  const data: Pets =
    [204, 205, 304].includes(res.status) || !res.body ? {} : await res.json();

  return { status: res.status, data, headers: res.headers };
};
```

## Related PRs

none

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

you can check by sample apps

### DIFF
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -181,7 +181,7 @@ ${
   `
     : `const res = await fetch(${fetchFnOptions})
 
-  const data:${response.definition.success} = res.body ? await res.json() : {}
+  const data:${response.definition.success} = ([204, 205, 304].includes(res.status) || !res.body) ? {} : await res.json()
 
   ${override.fetch.includeHttpResponseReturnType ? 'return { status: res.status, data, headers: res.headers }' : `return data as ${responseTypeName}`}
 `;
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -181,7 +181,7 @@ ${
   `
     : `const res = await fetch(${fetchFnOptions})
 
-  const data:${response.definition.success}  = await res.json()
+  const data:${response.definition.success} = res.body ? await res.json() : {}
 
   ${override.fetch.includeHttpResponseReturnType ? 'return { status: res.status, data, headers: res.headers }' : `return data as ${responseTypeName}`}
 `;
--- a/docs/src/pages/guides/fetch-client.md
+++ b/docs/src/pages/guides/fetch-client.md
@@ -48,7 +48,7 @@ export const listPets = async (
     ...options,
     method: 'GET',
   });
-  const data = await res.json();
+  const data: Pets = res.body ? await res.json() : {};
 
   return { status: res.status, data };
 };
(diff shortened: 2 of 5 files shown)


packages/fetch/src/index.ts
```diff
@@ -181,7 +181,7 @@ ${
   `
     : `const res = await fetch(${fetchFnOptions})
 
-  const data:${response.definition.success}  = await res.json()
+  const data:${response.definition.success} = res.body ? await res.json() : {}
```
nimo23: Are you sure this will avoid the error? I  tried it with 

`const responseBody = res.body ? await res.json() : {};`

but it throws 

`Uncaught (in promise) SyntaxError: JSON.parse: unexpected end of data at line 1 column 1 of the JSON data`

if there is no response body. Using the following will not throw any errors:

`const responseBody = await res.json().catch(() => ({}));`

soartec-lab: As you pointed out, this code will cause an error if the body exists but is empty.
I think I need to check it by referencing the plain text once like `const text = res.text()`.

soartec-lab:
@zZHorizonZz @nimo23 @clemeth 

I have finalized the specifications based on discussions with everyone, so check them out.
Please note that the order of logic definitions has been changed for readability.

https://github.com/orval-labs/orval/issues/1656#issuecomment-2566450970

melloware (approved):

zZHorizonZz (approved):

clemeth (approved):

PR #1780, opened 2024-12-30
fix(fetch): pass headers to `fetch` when `headers` option is activated

### LINKED ISSUE #1779: Generated endpoint headers are not passed to `fetch`

1. Set `headers: true` and `client: "fetch"` in `orval.config.ts`.
2. Run `orval`.
3. Call a generated function with endpoint headers.

## What happens?

The headers are not passed to `fetch`.

This is an example of generated code:
```typescript
export const getUser = async (
  headers: GetUserHeaders,
  options?: RequestInit
): Promise<User> => {
  const res = await fetch(getGetUserUrl(), {
    ...options,
    method: "GET",
  });
  const data = await res.json();

  return data as User;
};
```

The headers should be passed to `fetch`:

```diff
export const getUser = async (
  headers: GetUserHeaders,
  options?: RequestInit
): Promise<User> => {
  const res = await fetch(getGetUserUrl(), {
    ...options,
+   headers: { ...headers, ...options?.headers },
    method: "GET",
  });
  const data = await res.json();

  return data as User;
};
```

With `axios` it works as expected:

```typescript
export const getUser = <TData = AxiosResponse<User>>(
  headers: GetUserHeaders,
  options?: AxiosRequestConfig
): Promise<TData> => {
  return axios.get(`http://127.0.0.1:3999/user`, {
    ...options,
    headers: { ...headers, ...options?.headers },
  });
};
```

```
orval: 7.3.0
react: 19.0.0
```

clemeth:
## Status

**READY**

## Description

Passes headers to `fetch` when `headers` option is activated. 

Fixes #1779.

I have not tested the patch since the build system doesn't support the system I am on.

### DIFF
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -22,6 +22,7 @@ import {
 export const generateRequestFunction = (
   {
     queryParams,
+    headers,
     operationName,
     response,
     mutator,
@@ -169,6 +170,7 @@ ${
   const fetchFnOptions = `${getUrlFnName}(${getUrlFnProperties}),
   {${globalFetchOptions ? '\n' : ''}      ${globalFetchOptions}
     ${isRequestOptions ? '...options,' : ''}
+    ${headers ? 'headers: { ...headers, ...options?.headers },' : ''}
     ${fetchMethodOption}${fetchHeadersOption ? ',' : ''}
     ${fetchHeadersOption}${fetchBodyOption ? ',' : ''}
     ${fetchBodyOption}
--- a/tests/configs/fetch.config.ts
+++ b/tests/configs/fetch.config.ts
@@ -105,6 +105,17 @@ export default defineConfig({
       target: '../specifications/petstore.yaml',
     },
   },
+  headers: {
+    output: {
+      target: '../generated/fetch/headers/endpoints.ts',
+      schemas: '../generated/fetch/headers/model',
+      client: 'fetch',
+      headers: true,
+    },
+    input: {
+      target: '../specifications/petstore.yaml',
+    },
+  },
   formData: {
     output: {
       target: '../generated/fetch/form-data-optional-request/endpoints.ts',


melloware (approved):
LGTM but i will let @soartec-lab review

soartec-lab:
Sure, please give me a few days to review it 👍

soartec-lab (commented):
@clemeth 
Thanks!
Since we are managing test cases, could you please add a test case that uses `headers: true`?

https://github.com/orval-labs/orval/blob/master/tests/configs/fetch.config.ts

melloware (approved):

soartec-lab (approved):
Thanks!

camillecroci:
Hi @clemeth 
This PR says "pass headers to fetch when headers option is activated "
but when i set the property `headers:false` to my orval config, I still get headers in the generated response. Am I misunderstanding that pr?
(A bit more of the config just in case: 
```
output: {
	mode: 'split',
	target: './generated/client.ts',
	client: 'fetch',
	override: {
		mutator: './lib/custom-fetch.ts',
		name: 'customFetch',
	},
	mock: true,
	headers: false,
},
```

clemeth:
Hello, @camillecroci. This PR actually introduced a bug which was fixed by #1770 before the next release, so please refer to that for current code. I can still try to clarify some things.

The `headers` option enables generated functions to take the endpoint headers directly as an argument, like this:

```typescript
  const result = await getFoo({ "User-ID": userId });
```

With the option disabled, headers have to be passed through the `options` argument:

```typescript
  const result = await getFoo({ headers: { "User-ID": userId } });
```

It's unclear to me what you're looking for, but this doesn't seem like it. Can you share some more details?

camillecroci:
aaah, yes it is very different indeed ! 
In my case, I am trying to understand what changes in 7.3.0 or 7.4.x have started to add headers to the reponse of my generated function. This is an example : 
Generated function: 
```
export const postSavedArticles = async (
	options?: RequestInit
): Promise<postSavedArticlesResponse> => {
	return postSavedArticlesMutator<postSavedArticlesResponse>(
		getPostSavedArticlesUrl(),
		{
			...options,
			method: 'POST',
		}
	)
}
```
The return type: 
```
export type postSavedArticlesResponse = {
	data: PostSavedArticles | ErrorResponse
	status: number
	headers: Headers
}
```

It used to have only `data` and `status` and my `headers` can be quite big so I would prefer not to have to return them

clemeth:
> ```
> mutator: './lib/custom-fetch.ts',
> name: 'customFetch',
> ```

Could it be because of your custom `fetch` function?

camillecroci:
the customFetch function has not changed since those `headers` appeared. I think I need to dig into other changes from other versions

clemeth:
Oh, of course, I was too quick there. That behavior was changed in #1699, I believe. There is, however, the `includeHttpResponseReturnType` option, which omits both `status` and `headers` from the return types. Would that fit your use case? Otherwise, it could perhaps be split into two options, one to omit `status` and one to omit `headers`, in case you need one but not the other.

We are getting off topic here, so perhaps you could open a feature request if that is something you need.

camillecroci:
@clemeth thank you very much for finding that out. I will open a feature request then !

PR #1794, opened 2025-01-07
Feature/generate on changes

anymaniax:
## Status

**WIP**

## Description

It's a new feature to be able to only regenerate when the specification change. It's check the git history to be able to do a diff and check if the specification file between to generation. So could be use with a git hook on after a pull for example.

I am not sure if it could be useful or not and about the implementation.

@soartec-lab  @melloware what do you think about this?


melloware (approved):
I love it!

soartec-lab (commented):
@anymaniax 
I thought this was a good option too. Thanks!

PR #1795, opened 2025-01-08
fix(fetch): take param serializer setting into account

### LINKED ISSUE #1751: paramsSerializerOptions: arrayFormat does not seems to work.
Hello, 

I'm trying to use `arrayFormat ` but it does not seems to work, as in the docs 

```ts
paramsSerializerOptions: {
    qs: {
        arrayFormat: 'brackets',
    },
},
```

I'm not sure it's a bug since there is no qs library installed anyway. does this option only works if axios is used?

**Update:** I just tried to provide a custom params serializer function, it's not used in the generated code and all the code are using  `URLSearchParams`

AllieJonsson:
## Status

**READY**

## Description

Fixes #1751 
This adds support for paramSerializer and paramSerializerOptions in `fetch` client.

I also improved the docs a bit

## Todos

- [x] Tests
- [x] Documentation


melloware (dismissed):

soartec-lab:
The behavior is different from `axios`, but I would like to use `explode` and `customFetch` supported below:

https://github.com/orval-labs/orval/pull/1604

I also added it to the original issue.

soartec-lab:
@AllieJonsson 
Thank you for this suggestion.
#1751 was solved by using `explode`, so can I close this PR as well?

AllieJonsson:
> @AllieJonsson Thank you for this suggestion. #1751 was solved by using `explode`, so can I close this PR as well?

Great, then maybe I can just update the docs a bit in this pr to specify that the paramSerializer is only for axios. Ill fix it soon!

soartec-lab:
@AllieJonsson 
Thanks!
That's already outside the scope of this original PR, would you mind creating a new PR?

Issue #419, opened 2022-06-09
Bug: Generating invalid TS code


dbachorz:
## What are the steps to reproduce this issue?

1. Prepare JSON input for orval which is a swagger output with following data:
```
{
  "/health/live": {
    "get": {
      "operationId": "HealthController_checkLiveness",
      "summary": "Live check",
      "parameters": [],
      "responses": {
        "200": {
          "description": "The Health Check is successful",
          "content": {
            "application/json": {
              "schema": {
                "type": "object",
                "properties": {
                  // other properties not significant for this use case
                  "details": {
                    "type": "object",
                    "example": {
                      "database": {
                        "status": "up"
                      }
                    },
                    "additionalProperties": {
                      "type": "object",
                      "properties": {
                        "status": {
                          "type": "string"
                        }
                      },
                      "additionalProperties": {
                        "type": "string"
                      }
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}

```

## What happens?
Generated code from orval looks like this:

```
export type HealthControllerCheckLiveness200Details = {
  [key: string]: {
    status?: string;
    [key: string]: string;
  };
};
```

Which is an invalid TypeScript code.

## What were you expecting to happen?
Status shouldn't be optional

## Any logs, error output, etc?
`Property 'status' of type 'string | undefined' is not assignable to 'string' index type 'string'.(2411)`

## Any other comments?
TypeScript version: 4.7.3

## What versions are you using?
6.8.1

**Operating System:** MacOS Monterey 12.4
**Package Version:** 6.8.1
**Browser Version:** Chrome 102.0.5005.61

melloware:
Can you try 6.24.0 and report if it is still an issue?

alexeyraspopov:
@melloware I'm having similar issue. Tried `6.31.0` and the issue still there

melloware:
How about 7.6.0?

AllieJonsson:
Kind of duplicate of #600, in this case we should probably mark the second additional properties as unknown, as that type does not match the types of the properties in the object.

To solve this specific case, status should be in the required array of the object:
```json diff
                  "details": {
                    "type": "object",
                    "example": {
                      "database": {
                        "status": "up"
                      }
                    },
                    "additionalProperties": {
                      "type": "object",
+                     "required": [
+                       "status"
+                     ],
                      "properties": {
                        "status": {
                          "type": "string"
                        }
                      },
                      "additionalProperties": {
                        "type": "string"
[...]

PR #1802, opened 2025-01-09
fix(core): use union type for additional properties

### LINKED ISSUE #600: Additional properties causing TS error. "Property is not assignable to index type."
### What are the steps to reproduce this issue?

Create schema with typed `additionalProperties`
```
        "additionalProperties": {
          "type": "string",
          "example": "something"
        },
```

### Discussed in https://github.com/anymaniax/orval/discussions/599

<div type='discussions-op-text'>

<sup>Originally posted by **Tymek** September 30, 2022</sup>
When I have additional properties (like in **[this source file](https://github.com/Unleash/unleash/blob/9ba441b9cf63dbca4b394dc95571eade623c0564/src/lib/openapi/spec/sdk-context-schema.ts))**, generated type has issues like for example `Property 'remoteAddress' of type 'string | undefined' is not assignable to 'string' index type 'string'.` because of index type `[key: string]: string;`.

Is there a way of making it work with TypeScript?

#### More details

From schema JSON:
```
...
      "sdkContextSchema": {
        "description": "The Unleash context as modeled in client SDKs",
        "type": "object",
        "additionalProperties": {
          "type": "string",
          "example": "top-level custom context value"
        },
        "required": [
          "appName"
        ],
        "properties": {
          "appName": {
            "type": "string",
            "minLength": 1,
            "example": "My cool application."
          },
          "currentTime": {
            "type": "string",
            "format": "date-time",
            "example": "2022-07-05T12:56:41+02:00"
          },
          "environment": {
            "type": "string",
            "deprecated": true
          },
          "properties": {
            "type": "object",
            "additionalProperties": {
              "type": "string"
            },
            "example": {
              "customContextField": "this is one!",
              "otherCustomField": "3"
            }
          },
...
        }
      },
...
```

I get output like so:
```
/**
 * Generated by orval v6.10.2 🍺
 * Do not edit manually.
 * Unleash API
 * OpenAPI spec version: 4.15.1
 */
import type { SdkContextSchemaProperties } from './sdkContextSchemaProperties';

/**
 * The Unleash context as modeled in client SDKs
 */
export interface SdkContextSchema {
  appName: string;
  currentTime?: string;
  /** @deprecated */
  environment?: string;
  properties?: SdkContextSchemaProperties;
  remoteAddress?: string;
  sessionId?: string;
  userId?: string;
  [key: string]: string;
}
```

</div>

### LINKED ISSUE #1516: The additionalProperties field in OpenAPI schema makes build fail

Example OpenAPI definition:
```json
"MyObject": {
  "type": "object",
  "properties": {
    "empty": {
      "type": "boolean"
    }
  },
  "additionalProperties": {
    "type": "object"
  }
}
```

## What happens?

Orval generates the following code:
```typescript
export interface MyObject {
  empty?: boolean;
  [key: string]: { [key: string]: any }
}
```

No types clash.

Typescript throws build error:
Property 'empty' of type "boolean | undefined" is not assignable to 'string' index type '{ [key: string]: any }'. ts(2411)

**Operating System:** Windows 10
**Package Version:** 6.29.1

AllieJonsson:
## Status

**HOLD**

## Description

I think this fixes #1516, #600 and #419
The solution is to use union type when additionalProperties is defined. The resulting type changes from
```ts
export type FooType = {
  foo?: string;
  bar?: boolean;
  [key: string]: string;
}
```
which would generate the following errors:
`Property 'foo' of type 'string | undefined' is not assignable to 'string' index type 'string'.`
`Property 'bar' of type 'boolean | undefined' is not assignable to 'string' index type 'string'.`

The change in this PR makes the resulting type be
```ts
export type FooType = {
  foo?: string;
  bar?: boolean;
} | ({
  foo?: string;
  bar?: boolean;
} & {
  [key: string]: string;
 });
```
This results in a (I think?) correct type:
image
image
image

Please check and see if there are any edge cases I've forgotten to check, otherwise I think this would be an improvement

### DIFF
--- a/packages/core/src/generators/schema-definition.ts
+++ b/packages/core/src/generators/schema-definition.ts
@@ -146,6 +146,7 @@ function shouldCreateInterface(schema: SchemaObject) {
     !schema.anyOf &&
     !isReference(schema) &&
     !schema.nullable &&
-    !schema.enum
+    !schema.enum &&
+    !schema.additionalProperties
   );
 }
--- a/packages/core/src/getters/object.ts
+++ b/packages/core/src/getters/object.ts
@@ -133,15 +133,16 @@ export const getObject = ({
 
         if (arr.length - 1 === index) {
           if (item.additionalProperties) {
+            acc.value += '\n}';
             if (isBoolean(item.additionalProperties)) {
-              acc.value += `\n  [key: string]: unknown;\n }`;
+              acc.value += ` | (${acc.value} & {\n  [key: string]: unknown;\n })`;
             } else {
               const resolvedValue = resolveValue({
                 schema: item.additionalProperties,
                 name,
                 context,
               });
-              acc.value += `\n  [key: string]: ${resolvedValue.value};\n}`;
+              acc.value += ` | (${acc.value} & {\n  [key: string]: ${resolvedValue.value};\n })`;
             }
           } else {
             acc.value += '\n}';


anymaniax:
It would create another issue if you want to use `notInOriginalType`.

```ts
// would be a typescript error no?
const x = a['notInOriginalType']
```

AllieJonsson:
> It would create another issue if you want to use `notInOriginalType`.
> 
> ```ts
> // would be a typescript error no?
> const x = a['notInOriginalType']
> ```

Since it's an indexed type that works;
image

anymaniax:
I was talking of this case

Screenshot 2025-01-09 at 12 24 54

AllieJonsson:
@anymaniax Do you think this would be an improvement? At least this doesn't generate invalid typescript:
```ts
type FooType = {
  foo?: string;
  bar?: boolean;
} | { [key: string]: string }
```
The downside is that all non-string properties gets unionized to string...
image
But at least typescript doesn't complain
image

AllieJonsson:
never mind, it has the same issues as above...

PR #1803, opened 2025-01-09
feat(msw): create an index file exporting msw for split tags mode

### LINKED ISSUE #512: Mocks: Is there a way to have a single export for mocks in tags-split mode?
Generating everything in one file with the `single` mode results in a 20k line file which is obviously not ideal, makes my IDE freeze and file operations take extremely long time to finish. So I naturally choose the `tags-split` mode in orval, but when I tried to set up mocks according to the [examples](https://github.com/vercel/next.js/blob/canary/examples/with-msw/mocks/handlers.js), I faced the following issue.
I couldn't find a way to have a single function that exports all the mock functions. I figured it should be something like `export const handlers = getMyApiMSW()` in `handlers.js` but currently I can only do that in 'split' mode as it generates a separate `.msw` file for mocks.

Question is, can we have an `index.mocks.ts` file or similar generated that exports _only_ the mock functions but all of them?
(Using react-query client if it makes a difference.)

Thank you, awesome tool btw!

Edit: related to https://github.com/anymaniax/orval/issues/347

AllieJonsson:
## Status

**READY/WIP**

## Description

Fixes #512 
Is `index.msw.ts` a good name? Or should it be something like `mocks.ts`?

## Steps to Test or Reproduce

1. yarn generate:mock
2. check generated/mock/petstore-tags-split and there is a `index.msw.ts` file which exports * from all .msw.ts files

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -280,6 +280,28 @@ module.exports = {
 };
 ```
 
+### indexMockFiles
+
+Type: `Boolean`
+
+Valid values: true or false. Defaults to false.
+
+When true, adds a index.msw.ts file which exports all mock functions.
+This is only valid when `mode` is `tags-split`.
+
+Example:
+
+```js
+module.exports = {
+  petstore: {
+    output: {
+      schemas: 'src/gen/model',
+      indexMockFiles: true,
+    },
+  },
+};
+```
+
 ### title
 
 Type: `String` or `Function`.
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -547,9 +547,9 @@ Give you the possibility to set the locale for the mock generation. It is used b
 
 Type: `Boolean`
 
-Valid values: true or false. Defaults to false.
+Default Value: `false`.
 
-When true, adds a index.msw.ts file which exports all mock functions.
+When `true`, adds a `index.msw.ts` file which exports all mock functions.
 This is only valid when `mode` is `tags-split`.
 
 Example:
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -280,28 +280,6 @@ module.exports = {
 };
 ```
 
-### indexMockFiles
-
-Type: `Boolean`
-
-Valid values: true or false. Defaults to false.
-
-When true, adds a index.msw.ts file which exports all mock functions.
-This is only valid when `mode` is `tags-split`.
-
-Example:
-
-```js
-module.exports = {
-  petstore: {
-    output: {
-      schemas: 'src/gen/model',
-      indexMockFiles: true,
-    },
-  },
-};
-```
-
 ### title
 
 Type: `String` or `Function`.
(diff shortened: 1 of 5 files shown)


anymaniax:
it's not a fix for me but more a feature no? And for the file name I would `.mock` instead to follow the logic of the packages no? @melloware @soartec-lab what do you think?

melloware:
Agreed i think `index.mock.ts` would be good.

soartec-lab:
I completely agree with `index.msw.ts`  👍 
This is a great enhancement!

soartec-lab (commented):
If we don't support everything at once and continue to use it as before, this aggregated index will be unnecessary and just noise. How about making it optional?

AllieJonsson:
> If we don't support everything at once and continue to use it as before, this aggregated index will be unnecessary and just noise. How about making it optional?

I added a check for `output.indexFiles`, so you can set indexFiles: false in config to not generate this. Or did you have something else in mind?

anymaniax:
`indexFiles` also handle the index.ts too so I would add another property like `indexMockFiles` (maybe there is something better). Like that we can have index.ts and index.(mocks|msw).ts handled separatly

anymaniax:
@AllieJonsson  also thanks for all you are doing!

AllieJonsson:
> `indexFiles` also handle the index.ts too so I would add another property like `indexMockFiles` (maybe there is something better). Like that we can have index.ts and index.(mocks|msw).ts handled separatly

True, I just thought that if you want to have index files for non-mock files, you probably want it for mock files as well. But maybe separating and allowing just having one of them is better. Should it be enabled or disabled by default? `indexFiles` is true by default

AllieJonsson:
This is also really only valid for tags-split mode, right? So adding a whole new setting that only applies to this mode is maybe a bit overkill

AllieJonsson:
Or maybe this should be added to mode `tags` as well?

packages/core/src/writers/split-tags-mode.ts
```diff
@@ -31,6 +31,14 @@ export const writeSplitTagsMode = async ({
     output.tsconfig,
   );
 
+  const indexFilePath =
+    output.mock && output.indexFiles
+      ? upath.join(dirname, 'index.mock' + extension)
```
soartec-lab: Currently, the mock is only msw, so I thought it should match the extension of other files, which are `msw.ts`. If I were to change it, I would need to change it to match other things as well.

```suggestion
      ? upath.join(dirname, 'index.msw' + extension)
```

melloware:
yeah we have too many settings already @AllieJonsson  😄

soartec-lab:
@AllieJonsson 

It's exactly as @anymaniax wrote.
When using `tags-split`, there are cases where the schema needs indexing and mocking is not necessary, but we don't want to end up with exclusive control.

Also, reusing one option for different purposes will complicate things and confuse users. I agree that there are already too many configuration items, but I didn't think it was overkill to create one option for a new use case.

soartec-lab (commented):
Another comment I made was about the file extension. The only file type available is msw, so I would like to align them.

AllieJonsson:
Wait, this setting should be in output.mock.indexMockFiles instead of output.indexMockFiles...

docs/src/pages/reference/configuration/output.md
```diff
@@ -543,6 +543,30 @@ Default Value: `en`.
...
 Give you the possibility to set the locale for the mock generation. It is used by faker, see the list of available options [here](https://fakerjs.dev/guide/localization.html#available-locales). It should also be strongly typed using `defineConfig`.
 
+### indexMockFiles
+
+Type: `Boolean`
+
+Valid values: true or false. Defaults to false.
+
+When true, adds a index.msw.ts file which exports all mock functions.
+This is only valid when `mode` is `tags-split`.
```
soartec-lab:
```suggestion
Type: `Boolean`

Default Value: `false`.

When `true`, adds a `index.msw.ts` file which exports all mock functions.
This is only valid when `mode` is `tags-split`.
```

soartec-lab (commented):
@AllieJonsson 
I made one comment to the docs. The rest are fine. Thanks for the great enhance!

soartec-lab (approved):
Great!

PR #1808, opened 2025-01-10
Fix chaotic shuffle of generated types

### LINKED ISSUE #1382: Types in generated react-query schema have unstable order, causing noisy diffs

I don't have a minimally reproducible example, but here's what happens in our project:

1. Generate the `*.schema.ts` output of type `react-query` from some OpenAPI schema.
2. Add a new component in the OpenAPI schema.
3. Re-generate `*.schemas.ts`.

## What happens?

The newly added component correctly appears as a new type in the generated schema. However, the order of the **other** types in the generated schema changes unnecessarily, and seemingly at random. As a result, the diff of the schema is noisy, and it significantly increases the size of the PRs with schema changes.

I expected the new type to appear in the generated `schemas.ts` file, with the order of the other types staying the same.

No specific log output, but here's a screenshot of it happening in our project: Types moved around, even though they had nothing to do with the change to the OpenAPI schema.
image

## Any other comments?

I don't know how orval is implemented. But I suspect that the instability could come from modelling the component names as a hash set, where the order is not stable.

**Operating System:** MacOS 14.4.1
**Package Version:** 6.26.0
**Browser Version:** Node.js 20

FabienCharlier:
## Status

Fix #1382

**WIP**

## Description

Sometimes, when generating Orval schema, some generated types are moved around the file. This PR fixes it.
Related to issue [Types in generated react-query schema have unstable order, causing noisy diffs #1382 ](https://github.com/orval-labs/orval/issues/1382)

## Related PRs

No related PRs

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout <branch>
> grunt jasmine
```

1.

### DIFF
--- a/packages/core/src/generators/models-inline.ts
+++ b/packages/core/src/generators/models-inline.ts
@@ -6,9 +6,7 @@ export const generateModelInline = (acc: string, model: string): string =>
 export const generateModelsInline = (
   obj: Record<string, GeneratorSchema[]>,
 ): string => {
-  const schemas = Object.values(obj)
-    .flatMap((it) => it)
-    .sort((a, b) => (a.imports.some((i) => i.name === b.name) ? 1 : -1));
+  const schemas = Object.values(obj).flatMap((it) => it);
 
   return schemas.reduce<string>(
     (acc, { model }) => generateModelInline(acc, model),


FabienCharlier:
Considering that the sort is not necessary (you can use types before defining them in Typescript), and that it has never properly worked, I think that we should get rid of it. This will lead to the items being sorted in the same order than in the OpenApi file, which looks like a good solution to me.

melloware (approved):

melloware:
LGTM cc @AllieJonsson ?

AllieJonsson:
> LGTM cc @AllieJonsson ?

Seems logical!

PR #1812, opened 2025-01-13
fix(core): only generate one operation if it has multiple tags

### LINKED ISSUE #1677: Output mode 'tags' issue when operation has multiple tags
Duplicate operation are generated when Output mode is 'tags'  and operation has multiples tags, which is allowed by OpenAPI.

1.  set multiple tags on a operation:
e.g. 
```yaml
paths:
  /pets:
    get:
      summary: List all pets
      operationId: listPets
      tags:
        - pets
        - listAll
```
2. set output mode to `'tags'`
3. run

## What happens?
the operation will be generated on both `./pets` and `./listAll`
which causes the `index.ts` file to throw an error: 
`Module './pets' has already exported a member named 'ListPetsParams'. Consider explicitly re-exporting to resolve the ambiguity.`

only the first tag should be used in this scenario to export types and functions thus avoiding duplication

## Any other comments?
there could be a config to set which tag to use in this scenario.( future enhancement)

orval: 7.2.0

  System:
    OS: macOS 14.6.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 137.88 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh

AllieJonsson:
## Status

**READY**

## Description

Fixes #1677 

## Todos

- [x] Tests
- [x] Documentation

### DIFF
--- a/tests/specifications/multiple-tags.yaml
+++ b/tests/specifications/multiple-tags.yaml
@@ -1,0 +1,84 @@
+openapi: '3.0.0'
+info:
+  version: 1.0.0
+  title: Swagger Petstore
+  license:
+    name: MIT
+servers:
+  - url: http://petstore.swagger.io/v1
+paths:
+  /pets:
+    get:
+      summary: List all pets
+      operationId: listPets
+      tags:
+        - dogs
+        - cats
+      responses:
+        '200':
+          description: A paged array of pets
+          content:
+            application/hal+json:
+              schema:
+                $ref: '#/components/schemas/Pets'
+  /pets/dog/{dogId}:
+    get:
+      summary: Info for a specific dog
+      operationId: showDogById
+      tags:
+        - dog
+      parameters:
+        - name: dogId
+          in: path
+          required: true
+          description: The id of the dog to retrieve
+          schema:
+            type: string
+      responses:
+        '200':
+          description: Expected response to a valid request
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Dog'
+  /pets/cat/{catId}:
+    get:
+      summary: Info for a specific cat
+      operationId: showCatById
+      tags:
+        - cat
+      parameters:
+        - name: catId
+          in: path
+          required: true
+          description: The id of the cat to retrieve
+          schema:
+            type: string
+      responses:
+        '200':
+          description: Expected response to a valid request
+          content:
+            application/json:
+              schema:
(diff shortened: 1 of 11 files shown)


tests/specifications/multiple-tags.yaml
```diff
@@ -0,0 +1,84 @@
...
+          description: The id of the cat to retrieve
+          schema:
+            type: string
+      responses:
+        '200':
+          description: Expected response to a valid request
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Cat'
```
soartec-lab: These definitions that are not related to multiple tags are unnecessary, right?

```suggestion
```

soartec-lab (commented):
@AllieJonsson 
Thanks for the handle on this.

As suggested in the original issue, how about just using the first tag?
That way, we can control the cases where multiple tags occur by inputting `OpenAPI`.

tests/specifications/multiple-tags.yaml
```diff
@@ -0,0 +1,44 @@
...
+            application/hal+json:
+              schema:
+                $ref: '#/components/schemas/Pets'
+components:
+  schemas:
+    Pet:
+      type: object
+      oneOf:
+        - $ref: '#/components/schemas/Dog'
+        - $ref: '#/components/schemas/Cat'
```
soartec-lab: This is also an unnecessary definition 👍

AllieJonsson:
> @AllieJonsson Thanks for the handle on this.
> 
> As suggested in the original issue, how about just using the first tag? That way, we can control the cases where multiple tags occur by inputting `OpenAPI`.

Yes, if the user does not set a function for this setting, then the first tag will be used. Do you think this setting is unnecessary? I think the issue stated it would be nice to have

soartec-lab:
@AllieJonsson 
Yes, I think this setting is unnecessary because you can change the tag to be output by changing the input of `OpenAPI`. If there is a case that cannot be avoided, I would like to deal with it as a new issue.

soartec-lab (dismissed):
Looks good. Thank you.

soartec-lab:
@AllieJonsson 
Could you fix conflict ?

AllieJonsson:
> @AllieJonsson Could you fix conflict ?

@soartec-lab Done!

soartec-lab (approved):
nice!

PR #1818, opened 2025-01-13
fix(core): corrctly combine schemas if type is array

### LINKED ISSUE #1817: Nullable properties have additional `| null` declaration in OpenAPI 3.1

1. Generate the type for a nullable property from an OpenAPI 3.1 schema:
```yaml
openapi: 3.1.0
info:
  title: Nullables
  description: 'OpenAPI 3.1 examples'
  version: 1.0.0
paths: {}
components:
  schemas:
    Test:
      properties:
        example_nullable:
          type:
            - string
            - 'null'
      title: Test
      type: object
```

## What happens?

The type is generated as follows:
```typescript
export interface Test {
  example_nullable?: TestExampleNullable;
}

export type TestExampleNullable = string | null | null;
```

Type type is generated as follows:
```typescript
export type TestExampleNullable = string | null;
```

N/A

## Any other comments?

This is due to [this PR](https://github.com/orval-labs/orval/pull/1739) that now also checks whether a `null` type is included in the array of types. However, with the OpenAPI 3.1 specification, nullable properties are declared by adding the `null` type to the type array in the property, rather than specifying `nullable: true`. As such, the `nullable` variable is now `true` for the property where it was `false` before. I'm not entirely sure where the second `| null` is coming from however.

```
 System:
    OS: macOS 15.2
    CPU: (8) arm64 Apple M3
    Memory: 136.42 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    @tanstack/react-query: ^5.63.0 => 5.63.0 
    @tanstack/vue-query: ^4.22.0 => 4.22.0 
    axios: ^0.26.1 => 0.26.1 
    msw: ^2.0.2 => 2.0.2 
    react: ^18.3.1 => 18.3.1 
    swr: ^2.2.4 => 2.2.4 
    vue: ^3.3.4 => 3.3.4 
    zod: ^3.23.8 => 3.23.8
```

I'm using Orval 7.4.1, but this issue was introduced in 7.4.0.

AllieJonsson:
## Status

**READY**

## Description

Fixes #1817 as it was not properly fixed in #1739

## Related PRs

| fix/null-type | #1739 |

### DIFF
--- a/packages/core/src/getters/scalar.ts
+++ b/packages/core/src/getters/scalar.ts
@@ -8,6 +8,7 @@ import {
 import { escape, isString } from '../utils';
 import { getArray } from './array';
 import { getObject } from './object';
+import { combineSchemas } from './combine';
 import { resolveExampleRefs } from '../resolvers';
 
 /**
@@ -27,10 +28,7 @@ export const getScalar = ({
 }): ScalarValue => {
   // NOTE: Angular client does not support nullable types
   const isAngularClient = context.output.client === OutputClient.ANGULAR;
-  const typeIncludesNull =
-    Array.isArray(item.type) && item.type.includes('null');
-  const nullable =
-    (typeIncludesNull || item.nullable) && !isAngularClient ? ' | null' : '';
+  const nullable = item.nullable && !isAngularClient ? ' | null' : '';
 
   const enumItems = item.enum?.filter((enumItem) => enumItem !== null);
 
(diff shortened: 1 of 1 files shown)


soartec-lab (commented):
@AllieJonsson 
Could you add a test?

AllieJonsson:
> @AllieJonsson Could you add a test?

Oops, seems like I forgot to do that. Ill fix!

AllieJonsson:
@soartec-lab I just checked and there exists a test already, 'null-type' which contains the 
```yaml
type:
  - 'string'
  - 'null'
```
which is what this code fixes

soartec-lab:
OK, I checked it with the following test:

https://github.com/orval-labs/orval/blob/master/tests/specifications/null-type.yaml

soartec-lab (approved):

PR #1824, opened 2025-01-15
feat: export types from @orval/core

### LINKED ISSUE #1590: Export VerbOptions and other important types

1. create an `orval.config.ts`
2. use `defineConfig` to create your config
3. specify input and output transformers
4. refactor your transformers and move them into their own functions / files

Because we are using typescript and want to type our function parameters, 
we need access to both `OpenAPIObject` which is what is passed to an `InputTransformer` and `VerbOptions` which is passed to an `OutputTransformer`. 

However, Orval does not expose these types. 
The `OpenAPIObject` type can be accessed by adding and importing the `openapi3-ts/oas30` package.
The `VerbOptions` can _only_ be accessed if we employ longwinded and arcane typescript unpacking magic.
This is the code I have written to gain access to `VerbOptions`:

```ts

type ConfigType = Parameters<typeof defineConfig>[0];

type ValueOf<T> = T[keyof T];

type UnwrappedConfigValueType = Unwrap<ValueOf<Unwrap<ConfigType>>>;

type Unwrap<T> = T extends Promise<infer U> ? Unwrap<U> : T extends () => infer U ? Unwrap<U> : T;

type OutputOptionsType = ExtractOutput<UnwrappedConfigValueType>;

type ExtractOutput<T> = T extends { output?: infer O }
  ? O extends string | undefined
    ? never
    : O
  : never;

type OverrideOutputType = ExtractOverride<OutputOptionsType>;

type ExtractOverride<T> = T extends { override?: infer O }
  ? O extends undefined
    ? never
    : O
  : never;

type ExtractTransformer<T> = T extends { transformer?: infer O }
  ? O extends (options: infer U) => unknown
    ? U
    : never
  : never;

export type VerbOptions = ExtractTransformer<OverrideOutputType>;
```

## What happens?

I have to add and import extra packages as well as write a vast amount of boiler plate code so that my transformer functions can be typed safely and correctly.

I expect the orval package to export these types so I can use them right away.

N/A

## Any other comments?

These types are inferred, but that only works as long as the transformer function is inline.
As soon as we refactor because we have a larger amount of configuration / extras, we run into this issue.

```
  System:
    OS: Windows 11 10.0.22631
    CPU: (12) x64 Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
    Memory: 2.71 GB / 15.78 GB
  npmPackages:
    orval: ^7.0.1 => 7.0.1
    react: ^18.3.1 => 18.3.1
```

AllieJonsson:
## Status

**READY**

## Description

Fixes #1590 
re-exports types from core


melloware:
Thanks @AllieJonsson !

Issue #1849, opened 2025-01-27
I want to override `export type` or `export interface` in schema generation to `const`


fjodor-rybakov:
Hi all! I want to override `export type` or `export interface` in schema generation to `const`. For example:

My schema:
```yml
components:
  schemas:
    InvalidRequestDataException:
      type: object
      properties:
        message:
          type: string
          const: 'Invalid data'
        code:
          type: integer
          const: 1
          description: 'Unique exception code'
      required: [ 'message', 'code' ]
```

What i get now:
```ts
export interface InvalidRequestDataException {
  message: 'Invalid data',
  code: 1
}
```

What i want:
```ts
export const InvalidRequestDataException = {
  message: 'Invalid data',
  code: 1
} as const;
export type InvalidRequestDataExceptionType = typeof InvalidRequestDataException;
```

How i can replace `interface` to `const`?

_Originally posted by @fjodor-rybakov in https://github.com/orval-labs/orval/discussions/1848_

PR #1850, opened 2025-01-27
feat: generate const object for const schema definition

fjodor-rybakov:
## Status

**READY**

## Description

https://github.com/orval-labs/orval/issues/1849

Add new `useConstForSchemaDefinition` for generate `const` instead `interface` for schema definition.
For example:
```yml
#swagger api
components:
  schemas:
    InvalidRequestDataException:
      type: object
      properties:
        message:
          type: string
          const: 'Invalid data'
        code:
          type: integer
          const: 1
          description: 'Unique exception code'
      required: [ 'message', 'code' ]
```
With enabled `useConstForSchemaDefinition` i will get follow generated code
```ts
export const InvalidRequestDataExceptionValue = {
  message: 'Invalid data';
  code: 1;
} as const;
export type InvalidRequestDataException = typeof InvalidRequestDataExceptionValue;
```

## Related PRs

## Todos

- [X] Tests
- [X] Documentation

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout <branch>
> grunt jasmine
```

1.


soartec-lab:
@fjodor-rybakov 
Could you please explain in more detail why you want to use this in const?

fjodor-rybakov:
> @fjodor-rybakov Could you please explain in more detail why you want to use this in const?

I have some api like this:
```yml
openapi: 3.1.0
paths:
  /get-bubble-gums/{id}:
    get:
      parameters:
        - in: path
          name: id
          schema:
            type: number
      responses:
        '200':
          description: Suceess response
          content:
            application/json:
              schema:
                #some success schems
        '404':
          description: Not found
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/BubbleGumIdNotFoundException'
                
components:
  schemas:
     BubbleGumIdNotFoundException:
        type: object
        properties:
          message:
            type: string
            const: 'Bubble gum not found'
          code:
            type: integer
            const: 1
            description: 'Unique exception code'
        required: [ 'message', 'code' ]
```
In specification I write `BubbleGumIdNotFoundException` constant object(all properties have const). In my code I want to get this constant object and use `code` property in code logic. Like:
```ts
if (response.body.code === BubbleGumIdNotFoundExceptionValue.code) {
  // do some
} else {
  // do else
```

packages/core/src/generators/interface.ts
```diff
@@ -45,11 +45,20 @@ export const generateInterface = ({
...
-      scalar.value === 'unknown' ? '{}' : scalar.value;
+    if (
+      context.output.override?.useConstForSchemaDefinition &&
+      scalar.type === 'object' &&
+      schema.properties &&
+      Object.values(schema.properties).every((item) => 'const' in item)
+    ) {
+      model += `export const ${name}Value = ${scalar.value} as const;\nexport type ${name} = typeof ${name}Value;\n`;
+    } else {
+      // If `scalar.value` is 'unknown', replace it with `{}` to avoid type error
```
soartec-lab: This is self-explanatory and no comment is needed 👍

packages/core/src/generators/interface.ts
```diff
@@ -45,11 +45,20 @@ export const generateInterface = ({
...
     !context?.output.override?.useTypeOverInterfaces
   ) {
-    // If `scalar.value` is 'unknown', replace it with `{}` to avoid type error
-    const blankInterfaceValue =
-      scalar.value === 'unknown' ? '{}' : scalar.value;
+    if (
+      context.output.override?.useConstForSchemaDefinition &&
+      scalar.type === 'object' &&
+      schema.properties &&
+      Object.values(schema.properties).every((item) => 'const' in item)
```
soartec-lab, line 51: 👍

packages/core/src/types.ts
```diff
@@ -114,6 +114,7 @@ export type NormalizedOverrideOutput = {
   useDates?: boolean;
   coerceTypes?: boolean; // deprecated
   useTypeOverInterfaces?: boolean;
+  useConstForSchemaDefinition?: boolean;
```
soartec-lab: Adding a prop is also required here.

https://github.com/orval-labs/orval/blob/master/packages/core/src/types.ts#L360

And please normalize `type` here:

https://github.com/orval-labs/orval/blob/master/packages/orval/src/utils/options.ts#L124

Then, `?` is not needed 👍 

```suggestion
  useConstForSchemaDefinition: boolean;
```

soartec-lab (commented):
@fjodor-rybakov

good. I double-checked the specifications and understood that const definitions are declarations for managing constants like enums.

- https://json-schema.org/draft/2020-12/json-schema-validation#name-const
- https://github.com/OAI/OpenAPI-Specification/issues/1313
- https://swagger.io/specification/

Please comment what you would like to change.
And please also add the following two files.

1. add test using `const` to  `default.config.ts`

https://github.com/orval-labs/orval/blob/master/tests/configs/default.config.ts

2. add `useConstForSchemaDefinition` property to guide

https://github.com/orval-labs/orval/blob/master/docs/src/pages/reference/configuration/output.md

packages/core/src/generators/interface.ts
fjodor-rybakov: Done

packages/core/src/types.ts
fjodor-rybakov: Done

fjodor-rybakov:
> @fjodor-rybakov
> 
> good. I double-checked the specifications and understood that const definitions are declarations for managing constants like enums.
> 
> * https://json-schema.org/draft/2020-12/json-schema-validation#name-const
> * [Defining constant value in response OAI/OpenAPI-Specification#1313](https://github.com/OAI/OpenAPI-Specification/issues/1313)
> * https://swagger.io/specification/
> 
> Please comment what you would like to change. And please also add the following two files.
> 
> 1. add test using `const` to  `default.config.ts`
> 
> https://github.com/orval-labs/orval/blob/master/tests/configs/default.config.ts
> 
> 2. add `useConstForSchemaDefinition` property to guide
> 
> https://github.com/orval-labs/orval/blob/master/docs/src/pages/reference/configuration/output.md

1. Done
2. Done

soartec-lab:
@fjodor-rybakov 
I was wondering if this should be an option.
If all objects include const, I think it would be okay to output const as you do now, without specifying it as an option.
Can you tell me why you made it an option in the first place?

fjodor-rybakov:
> @fjodor-rybakov I was wondering if this should be an option. If all objects include const, I think it would be okay to output const as you do now, without specifying it as an option. Can you tell me why you made it an option in the first place?

Indeed, you are right, this can be made the default behavior without set `useConstForSchemaDefinition`. I was thinging that if I make changes, i do it with a feature flag for enable or disable my feature. We make set `useConstForSchemaDefinition: true` by default or remove `useConstForSchemaDefinition` option altogether. What do you think?

fjodor-rybakov:
> > @fjodor-rybakov I was wondering if this should be an option. If all objects include const, I think it would be okay to output const as you do now, without specifying it as an option. Can you tell me why you made it an option in the first place?
> 
> Indeed, you are right, this can be made the default behavior without set `useConstForSchemaDefinition`. I was thinging that if I make changes, i do it with a feature flag for enable or disable my feature(Maybe this is redundant). We make set `useConstForSchemaDefinition: true` by default or remove `useConstForSchemaDefinition` option altogether. What do you think?

soartec-lab:
@fjodor-rybakov 
In my opinion, this option can be removed.
This is because this is in accordance with the OpenAPI specification, and because we want to reduce optional controls as much as possible.
I don't think it's too late to consider switching to an option again when a situation arises where you want to disable it.

fjodor-rybakov:
> @fjodor-rybakov In my opinion, this option can be removed. This is because this is in accordance with the OpenAPI specification, and because we want to reduce optional controls as much as possible. I don't think it's too late to consider switching to an option again when a situation arises where you want to disable it.

Okey, I am get rid of `useConstForSchemaDefinition` option

melloware:
Running build again!

fjodor-rybakov:
It turned out that not everything was so simple. `getScalar` can return `value` with some imported type instead const. Do you have some idea for quic fix it? @soartec-lab @melloware

soartec-lab:
@fjodor-rybakov 

Well, it seems like you need to change the output as below:

```diff
- export type StringConstValueNullable = 'string' | null | null;
+ export const StringConstValueNullable = 'string' as const;
+ export type StringConstValueNullableType = typeof StringConstValueNullable | null;

- import type { StringConstValueNullable } from './stringConstValueNullable';
+ import { StringConstValueNullable } from './stringConstValueNullable';

export const StringConstValue = {
  value: 'string',
  valueWithoutType: 'string',
  valueNullable: StringConstValueNullable,
} as const;
export type StringConst = typeof StringConstValue;
```

fjodor-rybakov:
> @fjodor-rybakov
> 
> Well, it seems like you need to change the output as below:
> 
> ```diff
> - export type StringConstValueNullable = 'string' | null | null;
> + export const StringConstValueNullable = 'string' as const;
> + export type StringConstValueNullableType = typeof StringConstValueNullable | null;
> 
> - import type { StringConstValueNullable } from './stringConstValueNullable';
> + import { StringConstValueNullable } from './stringConstValueNullable';
> 
> export const StringConstValue = {
>   value: 'string',
>   valueWithoutType: 'string',
>   valueNullable: StringConstValueNullable,
> } as const;
> export type StringConst = typeof StringConstValue;
> ```

Hmm... I think some wrong in test spec `const.yaml`.
In follow schema `valueNullable` always `const` and at the same time type can be **string** and **null**
```yaml
openapi: 3.1.0
info:
  title: Const
  version: 1.0.0
paths: {}
components:
  schemas:
    StringConst:
      type: object
      required:
        - value
        - valueWithoutType
        - valueNullable
      properties:
        value:
          type: string
          const: string
        valueWithoutType:
          const: string
        valueNullable:
          type:
            - string
            - 'null'
          const: string
```
In my opinion, adding nulls to types is confusing. As if constant could change...

fjodor-rybakov:
@soartec-lab can you review?

soartec-lab:
@fjodor-rybakov 
Of course.
Since you have made additional corrections, do you have any additional explanations or updates to this PR explanation?

tests/specifications/const-for-schema-definition.yaml
```diff
@@ -0,0 +1,40 @@
...
+          content:
+            application/json:
+              schema:
+                type: string
+        '404':
+          description: Not found
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/BubbleGumIdNotFoundException'
```
soartec-lab: Delete unnecessary definitions around here and keep the definitions to a minimum.

fjodor-rybakov:
> Since you have made additional corrections, do you have any additional explanations or updates to this PR explanation?

I am only fixed some bugs. I think there is nothing else to change

soartec-lab:
okey, please give me a moment to check later.

soartec-lab (commented):
I made a comment about the test.
Also, there are too many line breaks, so please check if you can reduce them.

```diff
import axios from 'axios'
import type {
  AxiosRequestConfig,
  AxiosResponse
} from 'axios'
+
export const BubbleGumIdNotFoundExceptionValue = {
  message: 'Bubble gum not found',
  /** Unique exception code */
  code: 1,
} as const;
export type BubbleGumIdNotFoundException = typeof BubbleGumIdNotFoundExceptionValue;

-
-
-
-
  export const getGetBubbleGumsId = <TData = AxiosResponse<string>>(
    id: number, options?: AxiosRequestConfig
 ): Promise<TData> => {
    return axios.get(
      `/get-bubble-gums/${id}`,options
    );
  }

export type GetGetBubbleGumsIdResult = AxiosResponse<string>
```

If you force a force push, you won't be able to see the difference and the review will be delayed, so it's okay to just commit. Because we squash merge.

soartec-lab:
@fjodor-rybakov 
Let's polish this up and merge it. Please check it when you have time 🙌

tests/specifications/const-for-schema-definition.yaml
fjodor-rybakov: Done

fjodor-rybakov:
> I made a comment about the test. Also, there are too many line breaks, so please check if you can reduce them.
> 
> ```diff
> import axios from 'axios'
> import type {
>   AxiosRequestConfig,
>   AxiosResponse
> } from 'axios'
> +
> export const BubbleGumIdNotFoundExceptionValue = {
>   message: 'Bubble gum not found',
>   /** Unique exception code */
>   code: 1,
> } as const;
> export type BubbleGumIdNotFoundException = typeof BubbleGumIdNotFoundExceptionValue;
> 
> -
> -
> -
> -
>   export const getGetBubbleGumsId = <TData = AxiosResponse<string>>(
>     id: number, options?: AxiosRequestConfig
>  ): Promise<TData> => {
>     return axios.get(
>       `/get-bubble-gums/${id}`,options
>     );
>   }
> 
> export type GetGetBubbleGumsIdResult = AxiosResponse<string>
> ```
> 
> If you force a force push, you won't be able to see the difference and the review will be delayed, so it's okay to just commit. Because we squash merge.

Problem with new line breaks is not related to my changes. I check in master

I have only been able to find out problem with imports. Perhaps this should be considered as a separate PR?

fjodor-rybakov:
@soartec-lab I find out bug with many line break before implementation, can you look?

soartec-lab (commented):
@fjodor-rybakov 

Well, thank you for your Boy Scouts.
There is no problem with the source code, but I noticed one thing. Since the const test already exists, could you please merge it into the following? This is my last request.

https://github.com/orval-labs/orval/blob/master/tests/specifications/const.yaml

fjodor-rybakov:
> @fjodor-rybakov
> 
> Well, thank you for your Boy Scouts. There is no problem with the source code, but I noticed one thing. Since the const test already exists, could you please merge it into the following? This is my last request.
> 
> https://github.com/orval-labs/orval/blob/master/tests/specifications/const.yaml

Okey, np :)

soartec-lab (approved):
It's perfect! thanks!

PR #1876, opened 2025-02-04
fix(core): Fixed workspace setting does not append a newline to the end of a line in an existing index file #1875

### LINKED ISSUE #1875: When I set workspace and generate code, the line breaks keep increasing.
I always use orval. Thanks.

This phenomenon occurs in the following steps.

1. setup project
```
$ node -v
v22.13.1
$ mkdir orval-sample
$ cd orval-sample
$ npm init --yes
$ npm install orval
```

2. create orval.config.js
```js
module.exports = {
    petstore: {
        input: 'https://petstore3.swagger.io/api/v3/openapi.json',
        output: {
            workspace: "generated",
            target: "./zod/sample.ts",
            schemas: "./types",
            client: "zod",
            mode: "tags-split",
            clean: true,
        },
    },
}
```
3. add orval command
  package.json
  ```diff
  {
    "scripts": {
        "test": "echo \"Error: no test specified\" && exit 1",
+     "orval": "orval --config ./orval.config.js"
    },
  }
  ```

4. execute orval command
```
$ npm run orval
```

5. The end of the line in `generated/index.ts` grows longer and longer each time the command is executed.
```diff
export * from './zod/pet/pet';
export * from './zod/store/store';
export * from './zod/user/user';
export * from './types';

+
+
```

maroKanatani:
## Status

**READY**

## Description

Fixed workspace setting does not append a newline to the end of a line in an existing index file.

I don't know the details of the code, so if you think another way is better or whatever, you can close this PR.

- Fix #1875

## Related PRs

## Todos

- [x] Tests
- [ ] Documentation

## Steps to Test or Reproduce
Please follow the steps in the following Issue
- https://github.com/orval-labs/orval/issues/1875

### DIFF
--- a/packages/orval/src/write-specs.ts
+++ b/packages/orval/src/write-specs.ts
@@ -132,8 +132,8 @@ export const writeSpecs = async (
         await fs.appendFile(
           indexFile,
           uniq(importsNotDeclared)
-            .map((imp) => `export * from '${imp}';`)
-            .join('\n') + '\n',
+            .map((imp) => `export * from '${imp}';\n`)
+            .join(),
         );
       } else {
         await fs.outputFile(
--- a/packages/orval/src/write-specs.ts
+++ b/packages/orval/src/write-specs.ts
@@ -133,7 +133,7 @@ export const writeSpecs = async (
           indexFile,
           uniq(importsNotDeclared)
             .map((imp) => `export * from '${imp}';\n`)
-            .join(),
+            .join(""),
         );
       } else {
         await fs.outputFile(
--- a/yarn.lock
+++ b/yarn.lock
@@ -1253,23 +1253,23 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@orval/angular@npm:7.4.1, @orval/angular@workspace:packages/angular":
+"@orval/angular@npm:7.5.0, @orval/angular@workspace:packages/angular":
   version: 0.0.0-use.local
   resolution: "@orval/angular@workspace:packages/angular"
   dependencies:
-    "@orval/core": "npm:7.4.1"
+    "@orval/core": "npm:7.5.0"
   languageName: unknown
   linkType: soft
 
-"@orval/axios@npm:7.4.1, @orval/axios@workspace:packages/axios":
+"@orval/axios@npm:7.5.0, @orval/axios@workspace:packages/axios":
   version: 0.0.0-use.local
   resolution: "@orval/axios@workspace:packages/axios"
   dependencies:
-    "@orval/core": "npm:7.4.1"
+    "@orval/core": "npm:7.5.0"
   languageName: unknown
   linkType: soft
 
-"@orval/core@npm:7.4.1, @orval/core@workspace:packages/core":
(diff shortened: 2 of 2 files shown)


packages/orval/src/write-specs.ts
```diff
@@ -132,8 +132,8 @@ export const writeSpecs = async (
         await fs.appendFile(
           indexFile,
           uniq(importsNotDeclared)
-            .map((imp) => `export * from '${imp}';`)
-            .join('\n') + '\n',
+            .map((imp) => `export * from '${imp}';\n`)
+            .join(),
```
soartec-lab: It should be aligned with the pattern written in else.

https://github.com/orval-labs/orval/blob/419fa4831163a04ee4f129a01d756842ed61139f/packages/orval/src/write-specs.ts#L139-L144

```suggestion
            .map((imp) => `export * from '${imp}';`)
            .join('\n'),
```

packages/orval/src/write-specs.ts
```diff
@@ -132,8 +132,8 @@ export const writeSpecs = async (
         await fs.appendFile(
           indexFile,
           uniq(importsNotDeclared)
-            .map((imp) => `export * from '${imp}';`)
-            .join('\n') + '\n',
+            .map((imp) => `export * from '${imp}';\n`)
```
soartec-lab, line 135: If there is already a newline at the end, will a new definition be added below that newline?
In that case, you may need to first remove the original trailing newline and then add another trailing newline.

soartec-lab (commented):
@maroKanatani 
Thank you for the nice update. I made one change request and a question

packages/orval/src/write-specs.ts
maroKanatani: I was a little worried that if I implemented it with `.join('\n')`, the last line would not have a newline.

```diff
"export * from './zod/pet/pet';\n" +
"export * from './zod/store/store';\n" +
"export * from './zod/user/user';\n" +
+"export * from './types';\n" // my pattern
-"export * from './types';"    // with `.join('\n')`
```

With `.join('\n') + '\n'`  before the modification, a newline is inserted even if the result of uniq(imports) is empty.

packages/orval/src/write-specs.ts
maroKanatani: > If there is already a newline at the end, will a new definition be added below that newline?

No, the behavior appears to be the same, especially before this modification.

No, the behavior does not seem to change, especially before this modification.
I have confirmed that it is appended by preparing and executing petstore index.ts with a new line at the end as shown below.

before 
```diff
export * from './zod/user/user';
export * from './types';
export * from './zod/store/store';
- export * from './zod/pet/pet'; // remove this line
```

after

```diff
export * from './zod/user/user';
export * from './types';
export * from './zod/store/store';
+ export * from './zod/pet/pet';
```

maroKanatani:
@soartec-lab Thanks for the review! I have returned your comment.

packages/orval/src/write-specs.ts
soartec-lab: I see. You're right.

soartec-lab (approved):
Thanks!

PR #1885, opened 2025-02-07
feat(fetch): split responses in separate types based on status

### LINKED ISSUE #1881: Response type is not matched with status when using Fetch client
Consider the following schema:
```json
"responses": {
  "200": {
    "description": "The evaluation of the position.",
    "content": {
      "application/json": {
         "schema": {
           "$ref": "#/components/schemas/CloudEval"
          }
        }
      }
    },
  "404": {
    "description": "The position was not found in the cloud evaluation database.",
    "content": {
      "application/json": {
        "schema": {
          "properties": {
            "error": {
              "type": "string",
              "example": "No cloud evaluation available for that position"
            }
          }
        }
      }
    }
  }
}
```
The generated code when using `fetch` is:
```ts
export type apiCloudEvalResponse = {
  data: CloudEval | ApiCloudEval404;
  status: number;
  headers: Headers;
};
```
but what I would like to have is instead
```ts
export type apiCloudEvalResponse = (
  | {
      data: CloudEval;
      status: 200;
    }
  | {
      data: ApiCloudEval404;
      status: 404;
    }
  | {
      data: CloudEval | ApiCloudEval404;
      status: Exclude<HTTPStatusCode, 200 | 404>;
    }
) & {
  headers: Headers;
};
```
so that I easily can typecheck my code by doing
```ts
const couldEval = apiGetCloudEval();
// cloudEval.data is typed to CloudEval | ApiCloudEval404 here
if (couldEval.status === 200) {
  // couldEval.data is typed to CloudEval here
}
```

@soartec-lab what are your thoughs?
HTTPStatusCode would be an enum we would have to add, but the list of http status codes are finite so that should be no issue (e.g. https://gist.github.com/sketchbuch/c256015f63199fc06b24bb0c20a5f3fd)

AllieJonsson:
## Status

**READY**

## Description

Fix #1881 
Splits response types into multiple types, to be able to check against status and get correctly typed data.

Test with specifications with multiple responses, as well as a default response

## Todos

- [x] Tests
- [ ] Documentation

### DIFF
--- a/packages/core/src/writers/single-mode.ts
+++ b/packages/core/src/writers/single-mode.ts
@@ -10,7 +10,7 @@ import {
 } from '../utils';
 import { generateImportsForBuilder } from './generate-imports-for-builder';
 import { generateTarget } from './target';
-import { getOrvalGeneratedTypes } from './types';
+import { getHTTPStatusCodes, getOrvalGeneratedTypes } from './types';
 
 export const writeSingleMode = async ({
   builder,
@@ -113,6 +113,10 @@ export const writeSingleMode = async ({
 
     if (implementation.includes('NonReadonly<')) {
       data += getOrvalGeneratedTypes();
+      data += '\n';
+    }
+    if (implementation.includes('<HTTPStatusCodes,')) {
+      data += getHTTPStatusCodes();
       data += '\n';
     }
 
--- a/packages/core/src/writers/single-mode.ts
+++ b/packages/core/src/writers/single-mode.ts
@@ -10,7 +10,7 @@ import {
 } from '../utils';
 import { generateImportsForBuilder } from './generate-imports-for-builder';
 import { generateTarget } from './target';
-import { getHTTPStatusCodes, getOrvalGeneratedTypes } from './types';
+import { getOrvalGeneratedTypes } from './types';
 
 export const writeSingleMode = async ({
   builder,
@@ -113,10 +113,6 @@ export const writeSingleMode = async ({
 
     if (implementation.includes('NonReadonly<')) {
       data += getOrvalGeneratedTypes();
-      data += '\n';
-    }
-    if (implementation.includes('<HTTPStatusCodes,')) {
-      data += getHTTPStatusCodes();
       data += '\n';
     }
 
--- a/packages/core/src/writers/types.ts
+++ b/packages/core/src/writers/types.ts
@@ -25,12 +25,3 @@ type NonReadonly<T> = [T] extends [UnionToIntersection<T>] ? {
     : T[P];
 } : DistributeReadOnlyOverUnions<T>;
 `;
-
-export const getHTTPStatusCodes = () => `
-export type HTTPStatusCode1xx = 100 | 101 | 102 | 103;
-export type HTTPStatusCode2xx = 200 | 201 | 202 | 203 | 204 | 205 | 206 | 207;
-export type HTTPStatusCode3xx = 300 | 301 | 302 | 303 | 304 | 305 | 307 | 308;
-export type HTTPStatusCode4xx = 400 | 401 | 402 | 403 | 404 | 405 | 406 | 407 | 408 | 409 | 410 | 411 | 412 | 413 | 414 | 415 | 416 | 417 | 418 | 419 | 420 | 421 | 422 | 423 | 424 | 426 | 428 | 429 | 431 | 451;
-export type HTTPStatusCode5xx = 500 | 501 | 502 | 503 | 504 | 505 | 507 | 511;
-export type HTTPStatusCodes = HTTPStatusCode1xx | HTTPStatusCode2xx | HTTPStatusCode3xx | HTTPStatusCode4xx | HTTPStatusCode5xx;
-`;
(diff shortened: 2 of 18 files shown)


melloware (approved):

soartec-lab:
Thank you for this improvement. I'm checking other PRs, so I'll check this one in turn.

packages/fetch/src/index.ts
```diff
@@ -105,36 +106,40 @@ ${
...
-      : 'unknown';
+  const allResponses = [...response.types.success, ...response.types.errors];
+  const nonDefaultStatuses = allResponses
+    .filter((r) => r.key !== 'default')
+    .map((r) => r.key);
+  const responseDataTypes = allResponses
+    .map(
+      (r) => `export type ${responseTypeName}${pascal(r.key)} = {
+  ${isContentTypeNdJson(r.contentType) ? 'stream: Response' : `data: ${r.value || 'unknown'}`}
+  status: ${r.key === 'default' ? `Exclude<HTTPStatusCodes, ${nonDefaultStatuses.join(' | ')}>` : r.key}
```
AllieJonsson: I just realized this probably fails if the spec contains only a default status and no other statuses. Will need to change this, so dont merge yet

melloware:
I converted it to a Draft for now @AllieJonsson so no one merges it.

packages/core/src/writers/types.ts
```diff
@@ -25,3 +25,12 @@ type NonReadonly<T> = [T] extends [UnionToIntersection<T>] ? {
     : T[P];
 } : DistributeReadOnlyOverUnions<T>;
 `;
+
+export const getHTTPStatusCodes = () => `
```
soartec-lab: I'm starting to notice that having this boiler template for each file causes noise.
I think this is a definition for the default, but how about making the default status definition a `number` instead of a strict enum?

soartec-lab (commented):
@AllieJonsson 
It's not important to me, but it makes the code redundant, so I made a suggestion.

packages/core/src/writers/types.ts
AllieJonsson: Unfortunately, that would not work. Image this scenario:
```ts
type response200 = {
  data: myData;
  status: 200;
}
type responseDefault = {
  data: myError;
  status: number;
}
type responseComposite = response200 | responseDefault;
type response = responseComposite & {
  headers: Headers;
}
```

Now checking if status is 200 is not enough:
```ts
if (resp.status === 200) {
  // resp.data is still typed to myData | myError
}
```

And you cannot exclude certain numbers from `number`:

`type Ex = Exclude<number, 200>` is resolved to `type Ex = number`

soartec-lab: I see, you are right. Please wait a moment as I confirm this.

soartec-lab (commented):
@AllieJonsson 
I just finished checking everything. Thank you for the great changes 🙌 
could you regenerate sample apps?

packages/core/src/writers/single-mode.ts
```diff
@@ -115,6 +115,10 @@ export const writeSingleMode = async ({
       data += getOrvalGeneratedTypes();
       data += '\n';
     }
+    if (implementation.includes('<HTTPStatusCodes,')) {
```
soartec-lab: I don't want to include the processing of `fetch`client only in the processing of each core writer. Could you move these operations into the `fetch` package to confine their scope of influence?

packages/swr/src/index.ts
```diff
@@ -589,23 +590,21 @@ export const ${swrMutationFetcherName} = (${swrProps} ${swrMutationFetcherOption
...
       : ''
   }
   ${
-    isRequestOptions && isMutator
+    params.isRequestOptions && params.isMutator
       ? `type SecondParameter<T extends (...args: any) => any> = Parameters<T>[1];\n\n`
       : ''
-  }`;
+  }
+  ${params.output.httpClient === OutputHttpClient.FETCH ? generateFetchHeader(params) : ''}
```
soartec-lab: The functions when processing is divided by `httpClient` are consolidated into `client.ts`, so please move them 👍

soartec-lab (commented):
@AllieJonsson 

thank you. Please check my comment about refactoring the `swr` and `query` packages

soartec-lab:
CI is failing 👀

soartec-lab:
Hi @AllieJonsson Could you check this?

AllieJonsson:
Yes, sorry, I've had the flu 🤒 I'll fix this later today

AllieJonsson:
@soartec-lab Issue fixed, but there seems to be some zod errors in the samples?

soartec-lab:
Hi @AllieJonsson, The issue has been fixed and merged. Could you please update the sample apps again?

AllieJonsson:
> Hi @AllieJonsson, The issue has been fixed and merged. Could you please update the sample apps again?

Thanks! It looks good now

melloware (approved):

soartec-lab (approved):
Thanks a bunch!

hypnoboutique:
Thank you so much for this fix! 🙏

Do you have any idea when it is likely to be released?

melloware:
hopefully today @hypnoboutique

PR #1890, opened 2025-02-10
fix: improve playground editor UX

crazy-poncho:
## Status

**READY**

## Description

There're some annoying UX issues in playground editor. For example, can't effectively type changes in editor, because there's no debounce on user changes, so each new character potentially sends request with invalid input and it triggers error response and editor component disappears. Also, even if you copy-paste valid inputs and you get success response, because component unmounts, cursor position is lost. 

That's why i have some playground editor UX improvements:
- show error message in output editor when you get error response from `POST api/generate` endpoint
- use `debounce` for generate query to improve performance
- use `keepPreviousData` query option to prevent editor unmount, so we keep cursor position 

## Steps to Test or Reproduce
Just go to home page and scroll to playground editor, or visit `https://orval.dev/playground` page and try to reproduce mentioned problems.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -51,6 +51,7 @@
     "remark-toc": "^7.0.0",
     "remark-unwrap-images": "^2.0.0",
     "unist-util-visit": "^2.0.3",
+    "use-debounce": "^10.0.4",
     "yaml": "^2.2.1"
   },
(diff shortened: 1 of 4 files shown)


melloware (approved):
I love this.  @anymaniax any objections?

PR #1919, opened 2025-02-19
feat(core): ability to provide custom validation ruleset

### LINKED ISSUE #1887: Feature Request: Ability to provide custom validation ruleset
Right now we default to whatever the `@ibm-cloud/openapi-ruleset` package gives us. 

https://github.com/orval-labs/orval/blob/3d122c3db1883ceeb192e75ac75472240e6f7537/packages/core/src/utils/validator.ts#L7

https://github.com/orval-labs/orval/blob/3d122c3db1883ceeb192e75ac75472240e6f7537/packages/core/src/utils/validator.ts#L17

Ideally we provide an alternative a way to supply our own custom ruleset if desired.

Here's the API I'm thinking of:

```ts
export default defineConfig({
  'api-client': {
    input: {
      target: 'swagger.json'

      // This can now be either a boolean (for the default ruleset), 
      // OR a path to your own custom ruleset
      validation: '../path-to-my-ruleset.json',
    },
    ....
   },
});
```

ezequiel:
Fixes #1887

## Status

**READY**

## Description

This feature will give devs the ability to provide their own custom rulesets. Right now we always default to `@ibm-cloud/openapi-ruleset` with no way of modifying it

#### ruleset.ts
```ts
import { oas } from '@stoplight/spectral-rulesets';

export default {
  extends: [oas],
  rules: {
    'operation-operationId': 'error',
  },
};

```

#### orval.config.ts
```ts
import validation from './validate';

export default defineConfig({
  api: {
    input: {
      target: '../docs/swagger.json',
      validation,
    },
    ...
});
```

### DIFF
--- a/packages/core/src/utils/validator.ts
+++ b/packages/core/src/utils/validator.ts
@@ -3,6 +3,7 @@ import {
   ibmOpenapiValidatorErrors,
   ibmOpenapiValidatorWarnings,
 } from './logger';
+import fs from 'node:fs/promises';
 
 const ibmOpenapiRuleset = require('@ibm-cloud/openapi-ruleset');
 const { Spectral } = require('@stoplight/spectral-core');
@@ -12,9 +13,14 @@ const { Spectral } = require('@stoplight/spectral-core');
  * More information: https://github.com/IBM/openapi-validator/#configuration
  * @param specs openAPI spec
  */
-export const ibmOpenapiValidator = async (specs: OpenAPIObject) => {
+export const ibmOpenapiValidator = async (
+  specs: OpenAPIObject,
+  validation: boolean | object,
+) => {
+  const ruleset =
+    typeof validation === 'boolean' ? ibmOpenapiRuleset : validation;
   const spectral = new Spectral();
-  spectral.setRuleset(ibmOpenapiRuleset);
+  spectral.setRuleset(ruleset);
   const results = await spectral.run(specs);
 
   const errors: { message: string; path: string[] }[] = [];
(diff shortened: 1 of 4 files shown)


packages/core/src/utils/validator.ts
```diff
@@ -12,9 +13,16 @@ const { Spectral } = require('@stoplight/spectral-core');
...
  * @param specs openAPI spec
  */
-export const ibmOpenapiValidator = async (specs: OpenAPIObject) => {
+export const ibmOpenapiValidator = async (
+  specs: OpenAPIObject,
+  validation: string | boolean,
+) => {
+  const ruleset =
+    typeof validation === 'string'
+      ? (await import(validation)).default
```
ezequiel: It would be nice to be able to import TypeScript files here though. Doesn't seem possible with the built-in `import()`

ezequiel: Actually, let me make this an `Object` instead, so the onus of importing is on the user instead, not orval

packages/orval/src/import-open-api.ts
```diff
@@ -79,7 +79,7 @@ const generateInputSpecs = async ({
       const transfomedSchema = transformerFn ? transformerFn(schema) : schema;
 
       if (input.validation) {
-        await ibmOpenapiValidator(transfomedSchema);
+        await ibmOpenapiValidator(transfomedSchema, input.validation);
```
melloware, line 82: What does passing this param here do?

melloware (changes requested):

packages/orval/src/import-open-api.ts
ezequiel: `input.validation` will either be a boolean or an object (a custom ruleset provided by the user)

When `ibmOpenapiValidator` detects a boolean, it'll opt to use the `@ibm-cloud/openapi-ruleset` ruleset by default (this ensures `input.validation` is backwards-compatible). And when it's an object, it'll just use that directly.

melloware (approved):

PR #1924, opened 2025-02-21
feat(zod): add default support

### LINKED ISSUE #1922: Ability to use OpenAPI's 'default' with generated Zod schema object
Feature request to add the ability to use the `default` property from an OpenAPI schema property, to apply that value to the Zod schema object. For example a property of:

```yaml
...
properties:
    exampleField:
      type: string
      maxLength: 255
      default: "Hello World"
...
```

Would generate a Zod object with the following property (assuming the above schema was used in a list endpoint):

```js
export const exampleListResponseResultsItemExampleResponseMax = 255;
export const exampleListResponseResultsItemExampleResponseDefault = "Hello World";

export const exampleListResponse = zod.object({
  "exampleField": zod.string().max(exampleListResponseResultsItemExampleResponseMax ).optional().default(exampleListResponseResultsItemExampleResponseDefault),
...
```

To maintain backward compatibility, it could be behind a config option, at the most basic level a bool:

```js
 module.exports = {
   petstore: {
     output: {
       ...
       override: {
         zod: {
           default: true, // Default: false
         },
       },
     },
     ...
   },
 };
```

Thank you for this great library!

ruangustavo:
Closes #1922

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -18,6 +18,7 @@ import {
   jsStringEscape,
   pascal,
   resolveRef,
+  stringify,
   ZodCoerceType,
 } from '@orval/core';
 import uniq from 'lodash.uniq';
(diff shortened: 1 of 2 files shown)


melloware:
nice PR!

packages/zod/src/index.ts
```diff
@@ -396,7 +437,9 @@ export const generateZodValidationSchemaDefinition = (
     ]);
   }
 
-  if (!required && nullable) {
+  if (!required && schema.default) {
+    functions.push(['default', defaultVarName]);
+  } else if (!required && nullable) {
```
AkulinskiLoftyworks, line 442: I am having a bug where i set a type of boolean and schema.default is false. 
`if (!required && schema.default)` ends up falsy thus default of false is never pushed

melloware: @AkulinskiLoftyworks  open a new bug with your example config and code.

PR #1933, opened 2025-02-23
fix(query): check for v5 flag

### LINKED ISSUE #1931: React Query v5 code generation fails with pnpm catalog protocol
When using pnpm's catalog protocol (`catalog:`) in package.json, orval fails to generate React Query v5 code even when explicitly setting version to 5 in the config.

### Reproduction

package.json:
```json
{
  "dependencies": {
    "@tanstack/react-query": "catalog:"
  }
}
```

orval.config.ts:
```ts
export default defineConfig({
  books: {
    output: {
      client: "react-query",
      override: {
        query: {
          version: 5
        }
      }
    }
  }
});
```

### Error
```
Invalid argument not valid semver ('catalog:' received)
```

## Analysis

The issue occurs in `isQueryV5WithDataTagError` function. Unlike `hasQueryV5`, it doesn't respect the config version override:

https://github.com/orval-labs/orval/blob/7741f077aa49e7fc9a15671c025fc5b5f4bbcf96/packages/query/src/index.ts#L1088-L1099

This issue was introduced in v7.4.1 with the addition of `isQueryV5WithDataTagError`. v7.4.0 works correctly as it doesn't have this check.

### Proposed Solution

Make `hasQueryV5WithDataTagError` respect the config version override similar to `hasQueryV5`:

```diff
const hasQueryV5WithDataTagError =
+  query.version === 5 ||
  isQueryV5WithDataTagError(
    context.output.packageJson,
    outputClient as 'react-query' | 'vue-query' | 'svelte-query',
  );
```

melloware:
Fix #1931

### DIFF
--- a/packages/query/src/index.ts
+++ b/packages/query/src/index.ts
@@ -1092,10 +1092,12 @@ const generateQueryHook = async (
       outputClient as 'react-query' | 'vue-query' | 'svelte-query',
     );
 
-  const hasQueryV5WithDataTagError = isQueryV5WithDataTagError(
-    context.output.packageJson,
-    outputClient as 'react-query' | 'vue-query' | 'svelte-query',
-  );
+  const hasQueryV5WithDataTagError =
+    query.version === 5 ||
+    isQueryV5WithDataTagError(
+      context.output.packageJson,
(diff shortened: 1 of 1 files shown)


soartec-lab (approved):
Sounds good!

Issue #1943, opened 2025-02-26
Hono: Preventing Duplicate Imports in the `index.context.ts`


manakuro:
## Summary
Duplicate type imports are being generated in the `index.context.ts`, leading to redundant definitions in the output.

## Current Behavior
When importing types in the `index.context.ts`, the same type may appear multiple times.  For example, in my project, the following duplication occurred:

```ts
import {
  ChannelRequest, 
  ChannelRequest, // Duplicated
  GetChannelsChannelIdMessagesParams,
  PostChannelsChannelIdMessagesBody,
  PutChannelsChannelIdMessagesBody,
  DeleteChannelsChannelIdMessagesBody,
  PostChannelsChannelIdAttachmentBodyOne,
  Attachment,
  User,
  PostUsersUserIdSessionTokenBody,
  User, // Duplicated
  UserNotificationSettings,
  ChannelNotificationSettings,
  PostUsersUserIdMarkAsReadBody,
  Reaction,
} from './index.schemas';
```

It is unclear whether this issue originates from the OpenAPI schema file. However, preventing duplicate imports in the generated code would improve maintainability and prevent potential conflicts.

## Expected Behavior
Duplicate type imports should be filtered out before being included in the final output. A deduplication step could likely be added to the following section of code to address this issue:

https://github.com/orval-labs/orval/blob/f8a0ab216ca8151f034826598fc582bb4e8487fd/packages/hono/src/index.ts#L679-L698

## Proposed Fix
A simple deduplication function could be used to ensure unique imports:

```ts
// packages/hono/src/index.ts
imps = uniqBy(imps, 'name');

const uniqBy = <T>(arr: T[], key: keyof T): T[] => {
  const set = new Set();
  return arr.filter((a) => !set.has(a[key]) && set.add(a[key]));
};
```

This ensures that each imports appears only once based on its name.

## Environment
orval version: 7.5.0
Node.js version: 22.11.0
OS: macOS 15.3.1

If this issue is assigned to me, I’d be happy to work on a fix.

manakuro:
@soartec-lab Could you please check this issue?

PR #1944, opened 2025-02-26
fix(hono): remove duplicate type imports in the `index.context.ts`

manakuro:
## Status

**READY**

## Description
Fixes: #1943 

This fixes an issue where duplicate type imports were generated in the `index.context.ts`. The solution introduces a deduplication step to ensure that each type is imported only once.

## Related PRs

None

## Todos

- [x] Tests
- [ ] Documentation

## Steps to Test or Reproduce

It is unclear which specific part of the OpenAPI schema file causes this issue. However, when tested with the OpenAPI schema used in my project, the fix produced the expected behavior.


packages/hono/src/index.ts
```diff
@@ -697,6 +700,8 @@ const generateContext = async (
     })
     .filter((imp) => contexts.includes(imp.name));
 
+  imps = uniqBy(imps, 'name');
```
manakuro: The `imps` array contained duplicate objects like the following:

```ts
[
  {
    name: 'ChannelRequest',
    specKey: undefined,
    schemaName: 'ChannelRequest'
  },
  {
    name: 'ChannelRequest',
    specKey: undefined,
    schemaName: 'ChannelRequest'
  },
  { name: 'GetChannelsChannelIdMessagesParams' },
  { name: 'PostChannelsChannelIdMessagesBody', isConstant: false },
  { name: 'PutChannelsChannelIdMessagesBody', isConstant: false },
  { name: 'DeleteChannelsChannelIdMessagesBody', isConstant: false },
  { name: 'PostChannelsChannelIdAttachmentBodyOne', isConstant: false },
  { name: 'Attachment', specKey: undefined, schemaName: 'Attachment' },
  { name: 'User', specKey: undefined, schemaName: 'User' },
  { name: 'PostUsersUserIdSessionTokenBody', isConstant: false },
  { name: 'User', specKey: undefined, schemaName: 'User' },
  {
    name: 'UserNotificationSettings',
    specKey: undefined,
    schemaName: 'UserNotificationSettings'
  },
  {
    name: 'ChannelNotificationSettings',
    specKey: undefined,
    schemaName: 'ChannelNotificationSettings'
  },
  { name: 'PostUsersUserIdMarkAsReadBody', isConstant: false },
  { name: 'Reaction', specKey: undefined, schemaName: 'Reaction' }
]
```

To address this, I added a deduplication process based on the `name` property.

packages/hono/src/utils/uniqBy.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    },
+    {
+      name: 'should do nothing if no duplicates exist',
+      param: [{ id: '1' }, { id: '2' }, { id: '3' }, { id: '4' }],
+      expected: [{ id: '1' }, { id: '2' }, { id: '3' }, { id: '4' }],
+    },
+  ])(`$name`, ({ param, expected }) => {
+    expect(uniqBy(param, 'id')).toStrictEqual(expected);
+  });
+});
```
manakuro: This test ensures that duplicate objects are removed from the array.

packages/hono/src/utils/uniqBy.ts
```diff
@@ -0,0 +1,4 @@
+export const uniqBy = <T>(arr: T[], key: keyof T): T[] => {
```
soartec-lab: For now, there is no need to make the `key` argument generic.

soartec-lab: We currently maintain the files together.
Although I understand the benefits of narrowing down the test target and extracting functions, I prioritize the benefits of placing the processing that will actually be used nearby.
You can write the implementation directly in the index.

soartec-lab (commented):
@manakuro 
This is a nice approach. I made a comment from a refactoring perspective.

packages/hono/src/utils/uniqBy.ts
manakuro: >For now, there is no need to make the key argument generic.
We currently maintain the files together.
Although I understand the benefits of narrowing down the test target and extracting functions, I prioritize the benefits of placing the processing that will actually be used nearby.
You can write the implementation directly in the index.

Thank you for the review. I appreciate your feedback and understand your point.

The reason I moved the `uniqBy` to a separate file was that I wanted to write tests for it. However, if there's no need to separate it, how about simplifying the deduplication logic into a single line like this?

```ts
const imps = Object.values(verbOptions)
    .flatMap((verb) => {
      const imports: GeneratorImport[] = [];
      if (verb.params.length) {
        imports.push(...verb.params.map((param) => param.imports).flat());
      }

      if (verb.queryParams) {
        imports.push({
          name: verb.queryParams.schema.name,
        });
      }

      if (verb.body.definition) {
        imports.push(...verb.body.imports);
      }

      return imports;
    })
    .filter((imp) => contexts.includes(imp.name))
    .filter((imp, i, arr) => arr.findIndex((v) => v.name === imp.name) === i); // added
```

While this approach is less performant than using the `uniqBy`, I don’t think the impact would be significant. In terms of time complexity, it would be O(n²).

soartec-lab: @manakuro 
It looks make sense 🙌 
Thank you also for the form of performance. It's not a big concern, so I prefer an implementation that allows for more maintainable functionality. I also understand the background behind adding the tests you did 👍

manakuro: Got it. 👍 
I've applied the above fix and confirmed that it produces the expected result. Additionally, updating the `samples/hono` was also expected.

melloware (approved):

soartec-lab (approved):
Thanks!

Issue #1715, opened 2024-11-26
Add description property to Zod generated schema


karlismelderis-mckinsey:
How hard would it be for you to add related schema name as `description` to generated Zod schema?

It would allow us to log Schema name in case zod validation fails by pulling data from `<zodSchema>.description` instead of passing around arbitrary string argument

here is an example:
```ts
const getCatsParams = zod.object({
  name: zod.string(),
}, {description: 'GetCatsParams'});

console.log(getCatsParams.description)
```

if you agree to this change I can try to create a PR

melloware:
i would say submit the PR for review!

karlismelderis-mckinsey:
> i would say submit the PR for review!

🫡
Will try to help out here

🙏
We have some other work ahead of us so I don't expect any movement until Q1 next year

PR #1962, opened 2025-03-09
feat: Add descriptions for zod

TheJokersThief:
## Status

**READY**

Fix https://github.com/orval-labs/orval/issues/1715

## Description

Adds parameter descriptions to generated zod definitions, using the `zod.describe("<string>")` function.

Example output:

```ts
export const addListQueryLimitRegExp = new RegExp('^\\+\\d{10, 15}');


export const addListQueryParams = zod.object({
  "limit": zod.string().regex(addListQueryLimitRegExp).optional().describe('How many items to return at one time (max 100)'),
  "birthdate": zod.string().date().optional().describe('birth date')
})
```

Reference docs: https://zod.dev/?id=describe

## Todos

- [x] Tests
- [ ] Documentation

## Related Issues

Found this before doing the work:

- https://github.com/orval-labs/orval/issues/1715

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -446,6 +446,10 @@ export const generateZodValidationSchemaDefinition = (
     functions.push(['optional', undefined]);
   }
 
+  if (schema.description) {
+    functions.push(['describe', `'${jsStringEscape(schema.description)}'`]);
+  }
+
   return { functions, consts: uniq(consts) };
 };
 
@@ -797,6 +801,7 @@ const parseParameters = ({
       }
 
       const schema = deference(parameter.schema, context);
+      schema.description = parameter.description;
 
       const mapStrict = {
         path: strict.param,
--- a/packages/zod/src/zod.test.ts
+++ b/packages/zod/src/zod.test.ts
@@ -343,6 +343,46 @@ describe('generateZodValidationSchemaDefinition`', () => {
     });
   });
 
+  describe('description handling', () => {
+    const context: ContextSpecs = {
+      output: {
+        override: {
+          useDates: false,
+        },
+      },
+    } as ContextSpecs;
+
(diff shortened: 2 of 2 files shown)


melloware (approved):

marty0678:
I think (I may be wrong here, still digging into it) this change resulted in 100x CPU usage now with Zod + Vite 6 in some situations. I upgraded from Orval 7.6 to 7.21 and the generated Zod schema file more than doubled in size for this one large OpenAPI spec (from ~4MB to ~9MB on the generated output). Digging into it, it appears to be because of the describe() additions with all the documentation in this one particular OpenAPI schema.

The result is the node process going from idling at 0.0% CPU to idling at ~100% CPU. I only noticed because this fanless Macbook Air started to throttle after I upgraded Orval so I went digging what caused this.

My monkey patch solution for now, is to just open the file and remove all the describe() calls in a post script, which gets it back to the 0% idling and ~4MB file. But I wanted to flag it for others who may hit this and also ask if describe could/should be a configuration option in future versions as in this case it's a nice to have but not at the performance cost personally.

If so, I can open an issue for this, just didn't want to do that if that's not desired. Thank you!

melloware:
@marty0678 is the CPU at runtime using the zod schema or the Orval geenration time taking the CPU?

marty0678:
> @marty0678 is the CPU at runtime using the zod schema or the Orval geenration time taking the CPU?

It's the Zod schema at runtime, so technically this is a performance issue with Zod I would imagine, just the new generated schema from Orval triggers it. 

I'll do some digging on the Zod side, I'm still on Zod 3 and Orval 7 and I understand there was significant performance improvements in Zod 4 so perhaps this isn't an issue there.

melloware:
OK let us know its possibler we could make this a Zod switch to turn it off but this feels like a Zod bug?

PR #2018, opened 2025-04-09
feat(core): add union enum generation

### LINKED ISSUE #2010: Add option to generate union instead of enum
## Related

- https://github.com/orval-labs/orval/issues/907

## Problem

I've noticed I don't use the generated `enum`s in `schemas`.
I use only the generated `type`.

I assume there are other users-devs who also need ability to disable enum/const generation.

So I propose to add a flag `override.disableEnumGeneration`, related to `override.useNativeEnums`

AllieJonsson:
## Status

**READY**

## Description

fixes #2010

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -2240,7 +2240,7 @@ module.exports = {
 };
 ```
 
-#### useNativeEnums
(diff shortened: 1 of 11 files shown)


melloware (approved):

soartec-lab (approved):
Thanks!!

PR #2026, opened 2025-04-14
feat(core): add better support for form-data arrays

### LINKED ISSUE #1417: form-data array submission support

1. Have a doc with body that includes  form-data with array type field

E.g. : 

```
 post:
      summary: Create
      tags:
        - Docs
      requestBody:
        content:
          multipart/form-data:
            schema:
              type: object
              properties:
                files:
                  type: array
                  items:
                    type: string
                    format: binary
``` 
2. generate orval client (client: axios | axios-functions)

## What happens?
[According to PHP documentation ](https://www.php.net/manual/en/features.file-upload.multiple.php)we need to user `files[]` field name to show it's an array of files. And when we are using orval to generate an endpoint for 'mutlipart/form-data' it creates form with field name just like `files` neither `files[]`. As a result have an error from server that `files` is not an array

Expecting to have correct generated form-data

```
const postDocs = (
    postDocsBody: PostDocsBody,
    options?: SecondParameter<typeof customInstance>,
  ) => {
    postDocsBody.files.forEach(value => formData.append("files[]", value)); 

    return ...
  };
```

 - Output: 

```
const postDocs = (
    postDocsBody: PostDocsBody,
    options?: SecondParameter<typeof customInstance>,
  ) => {
    postDocsBody.files.forEach(value => formData.append("files", value));

   return ...
  };
```

**Operating System:** macOS
**Package Version:** 6.29.1

AllieJonsson:
## Status

**READY**

## Description

Fixes #1417 (appending [] to formData key name) and adds support for exploding the formData properties as well.
I need this in my current project at work, so I added it here as well.

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -992,13 +992,17 @@ module.exports = {
 
 Type: `Boolean`.
 
-Use to generate a <a href="https://tanstack.com/query/latest/docs/react/reference/useQuery" target="_blank">useQuery</a> custom hook. If the query key isn't provided that's the default hook generated.
+Use to generate a <a href="https://tanstack.com/query/latest/docs/react/reference/useQuery" target="_blank">useQuery</a> custom hook.
+If the query key isn't provided that's the default hook generated.
 
 ##### useMutation
 
 Type: `Boolean`.
 
 Use to generate a <a href="https://tanstack.com/query/latest/docs/react/reference/useMutation" target="_blank">useMutation</a> custom hook.
+The hook will only be generated if the operation is not a `GET` operation, and not configured to generate a [query](#useQuery).
+
+The [operations override](#operations) will take precedence if both are configured.
 
 ##### useInfinite
 
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -1898,7 +1898,8 @@ Use this property to provide a config to your http client or completely remove t
 
 Type: `Boolean` or `String` or `Object`.
 
-Valid values: path of the formData function or object with a path and name.
+Valid values: path of the formData function or object with a path and name. You can also define how
+the names of form entries are handled regarding arrays.
 
 Use this property to disable the auto generation of form data if you use multipart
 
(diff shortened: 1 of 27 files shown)


melloware (approved):

melloware (approved):

soartec-lab:
I'll check it later!

anymaniax:
@AllieJonsson not related to the pull request. Could you check your discord messages when you have the time

docs/src/pages/reference/configuration/output.md
```diff
@@ -1968,6 +1968,101 @@ export const customFormUrlEncodedFn = <Body>(body: Body): URLSearchParams => {
...
+  formData.append(`age`, pet.age.toString());
+}
+if (pet.relatives !== undefined) {
+  pet.relatives.forEach((value) =>
+    formData.append(`relatives[]`, JSON.stringify(value)),
+  );
+}
+```
+
+Type `explode` setting results in the following generated code:
```
soartec-lab, line 2051: This may not be necessary.
If necessary, why not consider specifying `explode` separately?

https://swagger.io/docs/specification/v3_0/serialization/

soartec-lab: And if this is removed, it can be made into a simple option like `override.formData.serializeWithBrackets: true` 👍

soartec-lab (commented):
@AllieJonsson 
I added comments 👍 
could you check it ?

docs/src/pages/reference/configuration/output.md
AllieJonsson: But that doesn't allow you to specify how `body` data in `multipart/form-data` requests are serialized, right?

soartec-lab: I thought that if `override.formData.serializeWithBrackets` was set to `true`, brackets would be added, and if it was set to `false`, it would behave as before.

AllieJonsson: Yes, I was thinking about the `explode`, how would you configure it to explode form data?
What I need for it to work with my ASP.Net backend is instead of 
```ts
data.forEach(
  (value, index) => {
    formData.append(
      `receptionFacilitiesPerFaction`,
/* 
 * value is of type
 * {
 *   fractionId: number
 *   receptionFacilityIds: number[]
 * }
*/
      JSON.serialize(value),
    );
  },
);
```
I need to set
```ts
data.forEach(
  (value, index) => {
    formData.append(
      `receptionFacilitiesPerFaction[${index}].fractionId`,
      value.fractionId.toString(),
    );
    value.receptionFacilityIds.forEach((v, rIndex) => {
      formData.append(
        `receptionFacilitiesPerFaction[${index}].receptionFacilityIds[${rIndex}]`,
        v.toString(),
      );
    });
  },
);
```

soartec-lab: Wow, so `ASP.net` can't accept it unless it's in a format like `receptionFacilitiesPerFaction[0]&receptionFacilitiesPerFaction[1]`?

AllieJonsson: I was surprised as well, I couldn't get it to work with sending json objects in a multipart/form-data, so I asked ChatGPT, which gave me this solution, and it worked!
https://chatgpt.com/share/67ff8e02-1fb0-8000-901b-61e336ddad0e

soartec-lab: I see, I didn't know that.

soartec-lab: @AllieJonsson 
Okey, then we already have `override.formData` so let's just move up the options level like `override.formData.arrayHandling`.

soartec-lab (approved):
It's good!

soartec-lab (approved):
Thanks!

PR #2027, opened 2025-04-14
fix(query): prioritize query over mutation for non-GET operations in generateQueryHook

### LINKED ISSUE #2025: Duplicate non-infinite hooks generated when infinite query enabled for POST endpoints
I have produced an example project/repo showcasing this bug [here](https://github.com/Edward-Upton/orval-duplicate-hook-bug), where there are duplicate hooks being generated with the same signature in [`src/index.ts`](https://github.com/Edward-Upton/orval-duplicate-hook-bug/blob/main/src/index.ts#L263).

The specific issue seems to appear when you have a POST operation, e.g.:
```
/items:
    post:
      operationId: searchItems
      summary: Search for items with pagination
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/SearchRequest"
      responses:
        "200":
          description: Successful response
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/PaginatedItemsResponse"
        "400":
          description: Bad request
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/ErrorResponse"
```

and then enable infinite query generation on it:

```
      override: {
        operations: {
          searchItems: {
            query: {
              useQuery: true,
              useInfinite: true,
              useInfiniteQueryParam: "cursor",
            },
          },
        },
      },
```

which then results in multiple of the same hooks for the normal (non-infinite-query) hook being generated

Some other things I've tried are:
- Removing the `useQuery: true` option --> doesn't change the generated code (i.e. still fails).
- Using a top-level `useInfinite: true` (not operation specific) --> this doesn't generate an infinite query hook for POSTs (potentially another bug).

Note that this does not fail if the endpoint is a GET endpoint, only for POSTs (my example repo includes a GET endpoint to showcase this).

Edward-Upton:
## Status

**READY**

Fix #2025

## Description

Ensures that only one of the queries or mutations are generated, giving priority to queries.

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

Run on the test repo that shows the bug in action

https://github.com/Edward-Upton/orval-duplicate-hook-bug

### DIFF
--- a/packages/query/src/index.ts
+++ b/packages/query/src/index.ts
@@ -1129,6 +1129,15 @@ const generateQueryHook = async (
     isQuery = operationQueryOptions.useSuspenseQuery;
   }
 
+  // For non-GET operations, only register query OR mutation hooks, not both
+  let isMutation =
+    operationQueryOptions?.useMutation || override.query.useMutation;
+
+  // If both query and mutation are true for a non-GET operation, prioritize query
+  if (verb !== Verbs.GET && isQuery) {
+    isMutation = false;
+  }
+
   if (isQuery) {
     const queryKeyMutator = query.queryKey
       ? await generateMutator({
@@ -1286,11 +1295,6 @@ const generateQueryHook = async (
         : undefined;
   }
 
-  let isMutation = verb !== Verbs.GET && override.query.useMutation;
-  if (operationQueryOptions?.useMutation !== undefined) {
-    isMutation = operationQueryOptions.useMutation;
-  }
-
   if (isMutation) {
     const mutationOptionsMutator = query.mutationOptions
       ? await generateMutator({
@@ -1365,7 +1369,7 @@ const generateQueryHook = async (
 
     const mutationOptionsFn = `export const ${mutationOptionsFnName} = <TError = ${errorType},
     TContext = unknown>(${mutationArguments}): ${mutationOptionFnReturnType} => {
-    
+
 ${hooksOptionImplementation}
 
       ${
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -992,13 +992,17 @@ module.exports = {
 
 Type: `Boolean`.
 
-Use to generate a <a href="https://tanstack.com/query/latest/docs/react/reference/useQuery" target="_blank">useQuery</a> custom hook. If the query key isn't provided that's the default hook generated.
+Use to generate a <a href="https://tanstack.com/query/latest/docs/react/reference/useQuery" target="_blank">useQuery</a> custom hook.
+If the query key isn't provided that's the default hook generated.
 
 ##### useMutation
 
 Type: `Boolean`.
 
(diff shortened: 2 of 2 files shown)


melloware:
Running the tests now!

packages/query/src/index.ts
```diff
@@ -1129,6 +1129,17 @@ const generateQueryHook = async (
...
   }
 
+  // For non-GET operations, only register query OR mutation hooks, not both
+  let isMutation = verb !== Verbs.GET && override.query.useMutation;
+  if (operationQueryOptions?.useMutation !== undefined) {
+    isMutation = operationQueryOptions.useMutation;
+  }
+
+  // If both query and mutation are true for a non-GET operation, prioritize query
+  if (verb !== Verbs.GET && isQuery && isMutation) {
```
soartec-lab: If `isQuery` is true, we don't need to consider `isMutation`, right?

```suggestion
  if (verb !== Verbs.GET && isQuery) {
```

packages/query/src/index.ts
```diff
@@ -1129,6 +1129,17 @@ const generateQueryHook = async (
     isQuery = operationQueryOptions.useSuspenseQuery;
   }
 
+  // For non-GET operations, only register query OR mutation hooks, not both
+  let isMutation = verb !== Verbs.GET && override.query.useMutation;
```
soartec-lab: The verb is checked last, after all options have been checked, so it is best to leave this step for last.

```suggestion
  let isMutation = operationQueryOptions.useMutation || override.query.useMutation;
  
  if (verb !== Verbs.GET && isQuery) {
      isMutation = false;
  }
```

soartec-lab (commented):
@Edward-Upton 

Thank you for the nice correction. I made a comment.
I think it needs to be reflected in the documentation, but I'll think about it for a bit.

Edward-Upton:
I have resolved the comments, thanks for the review 😄

soartec-lab (commented):
@Edward-Upton 
Thanks!
I thought about documentation. Please clarify the specifications in the guide for the following options.

1. Only valid when the verbs of the two mutations are not get and not query
2. Prioritize `output.mutation` over `output.orverride.mutation`

- https://orval.dev/reference/configuration/output#usemutation
- https://orval.dev/reference/configuration/output#usemutation

Could you update docs ?

soartec-lab (approved):
It's good !

AllieJonsson:
It seems like this change broke samples, where we now get multiple implementation of the hooks

soartec-lab:
@AllieJonsson 
Oh, will that be fixed by just re-running orval in samples?

AllieJonsson:
@soartec-lab No, I tried generating samples in #2035 but the issue persists

AllieJonsson:
@soartec-lab isn't it weird that we don't see if samples break in the PR, but in the following PRs? That must mean that we don't correctly build/test samples?

PR #2037, opened 2025-04-17
fix(ci): run update-samples in workflow

AllieJonsson:
## Status

**READY**

## Description

Build samples before testing them, to make sure we don't break samples in a PR

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -33,6 +33,8 @@ jobs:
         run: yarn build
       - name: formatting
         run: yarn format:check
+      - name: update samples
+        run: yarn update-samples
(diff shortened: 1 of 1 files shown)


AllieJonsson:
I'm not sure, what do you think?
Having that be an error means you always have to generate samples when creating a pull request, isn't it enough that the build will fail if the test in the following step fails?

soartec-lab:
@AllieJonsson 
Ah, I've missed updating samples a few times in the past, so I thought that if I could check them all at once, I would be able to notice unintended changes.

soartec-lab:
Hi @AllieJonsson , are there any plans to address this in the near future? If you don't, I'll close this for now.

PR #2040, opened 2025-04-17
fix(core): no not sanitize operationName when overridden

### LINKED ISSUE #552: Sanitization of custom operation names
The [sanitization step](https://github.com/anymaniax/orval/blob/f3928e184e9e44e686162e2eda194092e8a564ac/src/core/generators/verbsOptions.ts#L81
) shouldn't be done for overriden operation names. Once they've requested control, it would make more sense to leave this job up to the user. As it stands now, it's not possible to use valid characters such as `_` or `$` as part of the method's name.

1. Define a custom `operationName` function in the orval.config's options.
2. Attempt to name operations using valid variable symbols such as `_` or `$`

## What happens?

The resulting string `my_very_long_httpRequest_GET` will be "sanitized" to `myverylonghttpRequestGET`

Keep the defined string as the operation name.

## Any other comments?

Once they assume control, any sanitization should be left up to the user. As it stands, the api is too restrictive for no good reason.

**Package Version: 6.9.6**

ajubin:
## Status

**READY**

Fix #552 

## Description

The current behaviour is really strange when overriding `operationName` in `output`, see this related issue: https://github.com/orval-labs/orval/issues/552

Maybe I should add in documentation that  when overriding, you need to sanitize on your side

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout <branch>
> grunt jasmine
```

1.

### DIFF
--- a/packages/core/src/generators/verbs-options.ts
+++ b/packages/core/src/generators/verbs-options.ts
@@ -80,10 +80,9 @@ const generateVerbOptions = async ({
 
   const overrideOperationName =
     overrideOperation?.operationName || output.override?.operationName;
-  const overriddenOperationName = overrideOperationName
+  const operationName = overrideOperationName
     ? overrideOperationName(operation, route, verb)
-    : camel(operationId);
-  const operationName = sanitize(overriddenOperationName, { es5keyword: true });
+    : sanitize(camel(operationId), { es5keyword: true });
 
   const response = getResponse({
     responses,


soartec-lab (approved):
It's good!

soartec-lab:
@ajubin 
Thanks for fixed. Could you check format ?

ajubin:
I fixed the formatting error

PR #2058, opened 2025-04-22
feat: allow index files customization

### LINKED ISSUE #2044: Allow filtering and customizing exports in generated index.ts
## Description
Currently, when Orval generates code, it automatically exports everything from the generated index.ts file. There's no way to:
1. Filter out specific exports (e.g., schemas)
2. Append custom exports from other files
3. Control what gets exported from the generated index.ts

## Use Case
In our project, we need to:
- Exclude schema exports from the generated index.ts file to avoid polluting the public API
- Append exports from our custom files to the generated index.ts
- Have more granular control over what gets exported from the generated code

## Current Behavior
The generated index.ts automatically exports everything:
```typescript
export * from './schemas';
export * from './endpoints';
// ... other exports
```

## Desired Behavior
Ability to configure exports in orval.config.js:
```javascript
module.exports = {
  myApi: {
    output: {
      // ... other config
      indexFiles(statements) {
        const exports = statementes.filter((name) => name !== "./schemas")
        exports.push("./custom-files")
        return exports;
      }
    }
  }
}
```

## Benefits
1. Better control over the public API surface
2. Ability to keep schemas internal when needed
3. Easier integration with custom code
4. More flexible code organization

## Additional Context
This would be particularly useful for larger projects where we need more control over what gets exposed through the generated API client.

## Possible Implementation Approaches
1. Add an `exports` configuration option in the output config
2. Support glob patterns for include/exclude
3. Allow specifying additional files to export from

Would love to hear thoughts on the best approach to implement this!

maastrich:
## Status

**READY**

## Related

Fix https://github.com/orval-labs/orval/issues/2044

## Description

add a way to customize index files:

```ts
export default defineConfig({
  output: {
    // ....rest of the config
    indexFiles: {
      schema: (schema) => schemaPaths.filter(schema => schema.name === "Dog"),
      workspace: (implementationPaths) => implementationPaths.filter(path => path.includes('do-not-expose-this-file')),
    },
  },
});
```

add a way to expose custom files:

```ts
export default defineConfig({
  output: {
    // ....rest of the config
    client() {
        return {
            // ... client definition
            extraFiles() {
                return [
                    {
                        path: 'src/file-to-expose.ts',
                        content: 'export default function () {}',
                        exposeIndexFile: true,
                    },
                    {
                        path: 'src/file-to-not-expose.ts',
                        content: 'export default function () {}',
                        exposeIndexFile: false,
                    },
                    {
                        path: 'src/file-to-not-expose-2.ts',
                        content: 'export default function () {}',
                    }
                ]
            }
        }
    }
});
```

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout feat/customize-index-files
> grunt jasmine
> yarn
> yarn build
> cd tests
> yarn
> yarn generate:multi-client
```

1. check generated clients in `generated/multi-client/*`
2. check `index.ts` in each client, it should only include files that were exposed (no models)
3. check zod client and see exports for `./extrafiles/index`

### DIFF
--- a/packages/orval/src/utils/options.ts
+++ b/packages/orval/src/utils/options.ts
@@ -3,6 +3,7 @@ import {
   ConfigExternal,
   createLogger,
   FormDataArrayHandling,
+  GeneratorSchema,
   GlobalMockOptions,
   GlobalOptions,
   HonoOptions,
@@ -24,14 +25,15 @@ import {
   NormalizedMutator,
   NormalizedOperationOptions,
   NormalizedOptions,
+  NormalizedOutputIndexFiles,
   NormalizedOverrideOutput,
   NormalizedQueryOptions,
   OperationOptions,
   OptionsExport,
   OutputClient,
   OutputHttpClient,
+  OutputIndexFiles,
   OutputMode,
-  OutputOptions,
   OverrideOutput,
   PropertySortOrder,
   QueryOptions,
@@ -196,7 +198,7 @@ export const normalizeOptions = async (
       tsconfig,
       packageJson,
       headers: outputOptions.headers ?? false,
-      indexFiles: outputOptions.indexFiles ?? true,
+      indexFiles: normalizeIndexFiles(outputOptions.indexFiles),
       baseUrl: outputOptions.baseUrl,
       unionAddMissingProperties:
         outputOptions.unionAddMissingProperties ?? false,
@@ -428,6 +430,24 @@ export const normalizePath = <T>(path: T, workspace: string) => {
     return path;
   }
   return upath.resolve(workspace, path);
+};
+
+const normalizeIndexFiles = (
+  indexFiles: OutputIndexFiles | boolean | undefined = true,
+): NormalizedOutputIndexFiles => {
+  if (isBoolean(indexFiles)) {
+    return indexFiles
+      ? {
+          workspace: (implementations: string[]) => implementations,
+          schemas: (schemas: GeneratorSchema[]) => schemas,
+        }
+      : false;
+  }
+  return {
+    workspace: (implementations: string[]) => implementations,
+    schemas: (schemas: GeneratorSchema[]) => schemas,
+    ...indexFiles,
+  };
 };
 
 const normalizeOperationsAndTags = (
--- a/tests/configs/multi-client.config.ts
+++ b/tests/configs/multi-client.config.ts
@@ -1,0 +1,70 @@
+import { defineConfig } from 'orval';
+
+export default defineConfig({
+  reactQuery: {
+    input: '../specifications/petstore.yaml',
+    output: {
+      client: 'react-query',
+      mode: 'tags-split',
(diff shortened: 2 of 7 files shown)


maastrich:
I've an update for the readme if you want me to add it in the pull request 👀

melloware:
Sure

maastrich:
Hey @melloware, I've pushed it with my last commit
https://github.com/orval-labs/orval/pull/2058/commits/b1b956fe66cc3680227d985b52eb5a84152126e3

melloware:
OK i have added some others for review

packages/orval/src/utils/options.ts
```diff
@@ -430,6 +432,24 @@ export const normalizePath = <T>(path: T, workspace: string) => {
   return upath.resolve(workspace, path);
 };
 
+const normalizeIndexFiles = (
+  indexFiles: OutputIndexFiles | boolean | undefined = true,
```
AllieJonsson, line 436: We don't need to have the default value of the param here

packages/orval/src/utils/options.ts
```diff
@@ -430,6 +432,24 @@ export const normalizePath = <T>(path: T, workspace: string) => {
...
+): NormalizedOutputIndexFiles => {
+  if (isBoolean(indexFiles)) {
+    return indexFiles
```
AllieJonsson, line 439: Add a const `defaultIndexFiles` or similar and use it here and in the case where `indexFiles` is not a boolean

tests/configs/multi-client.config.ts
AllieJonsson, line 1: I'm not sure we should have a file like this, why not just add these cases to their respective config file `react-query.config.ts` and `zod.config.ts`?

AllieJonsson (changes requested):
Thanks for your contribution, I added a few comments!

packages/orval/src/utils/options.ts
```diff
@@ -430,6 +432,24 @@ export const normalizePath = <T>(path: T, workspace: string) => {
...
+      ? {
+          workspace: (implementations: string[]) => implementations,
+          schemas: (schemas: GeneratorSchema[]) => schemas,
+        }
+      : false;
+  }
+  return {
+    workspace: (implementations: string[]) => implementations,
+    schemas: (schemas: GeneratorSchema[]) => schemas,
+    ...indexFiles,
```
AllieJonsson, line 449: If a user puts 
```ts
{
  schemas: undefined
}
```
in the config, we will get undefined here. I'd suggest doing something like
```ts
const defaultIndex = {
  workspace: (implementations: string[]) => implementations,
  schemas: (schemas: GeneratorSchema[]) => schemas,
};

return {
  workspace: indexFiles.workspace ?? defaultIndex.workspace,
  schemas: indexFiles.schemas ?? defaultIndex.schemas,
};
```

melloware:
No response from OP in 7 months closing

PR #2079, opened 2025-05-02
fix(query): override operation mutation and query conflict

anymaniax:
## Status

**READY**

## Description

Solve conflict when bot useMutation and useQuery are set when override by operation


AllieJonsson (approved):

melloware (approved):

maapteh:
@anymaniax or @melloware we had it with only (at least the description mentions when doing both??):

```
override: {
        operations: {
          reprint: {
            query: {
              useMutation: true,
            },
          },
```

Nothing else, maybe we can add a unit test for it?

Issue #2085, opened 2025-05-13
zod.discriminatedUnion does not work in combination with inheritance


timmisset:
When an openapi spec contains a oneOf with a discriminator, this will generate a zod.discriminatedUnion.

return `zod.discriminatedUnion('${propertyName}', [${typeSchemas.join(
        ',',
      )}])`;

However, the typeSchemas are not necessarly ZodObjects, which is the requirement for a DiscriminatedUnion. 
For example:

zod.object({ propA: string }).and({ propB: string}) will become a ZodIntersect which is not a valid option for a DiscriminatedUnion.

The previous implementation would concat the different options of the union with .or() which worked fine (at least for our use-case)

Hope you guys can either revert this change or provide a config option to disable it

melloware:
PR is welcome

timmisset:
It is caused by this PR: https://github.com/orval-labs/orval/pull/1907 from @Georgegriff.

Georgegriff:
Can you please provide an example openapi spec with the broken behaviour so I can understand the issue a little more

timmisset:
Hi @Georgegriff 
This is an example openapi:

`openapi: 3.0.3
info:
  title: Test OpenApi``
  version: 1.0.0
paths:
  /test:
    post:
      operationId: getTest
      requestBody:
        content:
          application/json:
            schema:
              oneOf:
                - $ref: "#/components/schemas/Horse"
                - $ref: "#/components/schemas/Cow"
              discriminator:
                propertyName: animalType
                mapping:
                  Horse: "#/components/schemas/Horse"
                  Cow: "#/components/schemas/Cow"
      responses:
        "200":
          description: OK

components:
  schemas:
    Animal:
      type: object
      required: [animalType]
      properties:
        animalType:
          type: string
    Horse:
      allOf:
        - $ref: "#/components/schemas/Animal"
        - type: object
          properties:
            hasShoes:
              type: boolean
    Cow:
      allOf:
        - $ref: "#/components/schemas/Animal"
        - type: object
          properties:
            brand:
              type: string
`
which now results into

`/**
 * Generated by orval v7.9.0 🍺
 * Do not edit manually.
 * Test OpenApi``
 * OpenAPI spec version: 1.0.0
 */
import {
  z as zod
} from 'zod';


export const getTestBody = zod.discriminatedUnion('animalType', [zod.object({
  "animalType": zod.string()
}).and(zod.object({
  "hasShoes": zod.boolean().optional()
[...]

PR #1907, opened 2025-02-18
fix: implement discriminations correctly in zod

Georgegriff:
## Status

**READY/WIP/HOLD**

## Description

A few sentences describing the overall goals of the pull request's commits.

## Related PRs

List related PRs against other branches:

| branch              | PR       |
| ------------------- | -------- |
| other_pr_production | [link]() |
| other_pr_master     | [link]() |

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Changelog Entry (unreleased)

## Steps to Test or Reproduce

Outline the steps to test or reproduce the PR here.

```bash
> git pull --prune
> git checkout <branch>
> grunt jasmine
```

1.

PR #2099, opened 2025-05-19
flatten zod objects in a discriminatedUnion

timmisset:
## Status

READY

## Description

The PR flattens allOf objects when they are used within the scope of a discriminatedUnion. A discriminatedUnion only accepts ZodObjects.
Fix https://github.com/orval-labs/orval/issues/2085

**IMPORTANT:** 

I created a PR for this issue, but it will not be a solution for all future scenarios. Since there can be multiple levels of nesting there may be a need for a more structured solution, perhaps always flattening the allOf properties regardless of the situation?

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -530,6 +530,24 @@ export const parseZodValidationSchemaDefinition = (
   const parseProperty = (property: [string, any]): string => {
     const [fn, args = ''] = property;
 
+    if (fn === 'flattenAllOf') {
+        const flattenedProperties: Record<string, any> = args.reduce(
+          (
+            acc: Record<string, any>,
+            {
+              functions
+            }: { functions: [string, any][];}
+          ) => {
+            functions.forEach(([_fn, schema]) => {
+              acc = { ...acc, ...schema}
+            })
+            return acc;
+          },
+          {}
+        )
+        const functionName = getObjectFunctionName(isZodV4, strict);
+        return `${parseProperty([functionName, flattenedProperties])}${strict ? parseProperty(['strict', undefined]) : ''}`
+    }
     if (fn === 'allOf') {
       return args.reduce(
         (
@@ -562,7 +580,7 @@ export const parseZodValidationSchemaDefinition = (
       const [, propertyName] = fn.split('discriminator__');
       const typeSchemas = args.map(
         ({ functions }: { functions: [string, any][]; consts: string[] }) => {
-          const schemaFunctions = functions.map(parseProperty).join('');
+          const schemaFunctions = functions.map(([fn, args]) => fn === 'allOf' ? parseProperty(['flattenAllOf', args]) : parseProperty([fn, args])).join('');
           return schemaFunctions;
         },
       );
@@ -656,7 +674,6 @@ ${Object.entries(args)
     ) {
       return `.coerce.${fn}(${args})`;
     }
-
     return `.${fn}(${args})`;
   };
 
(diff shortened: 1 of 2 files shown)


melloware:
Assigned to @soartec-lab  for review

soartec-lab:
@timmisset 

Oh, that means the sample apps haven't been updated since before, so I'll look for another way.
I wanted to run a test to check this change, but I'll think of another way to do it.

packages/zod/src/index.ts
```diff
@@ -562,7 +578,13 @@ export const parseZodValidationSchemaDefinition = (
       const [, propertyName] = fn.split('discriminator__');
       const typeSchemas = args.map(
         ({ functions }: { functions: [string, any][]; consts: string[] }) => {
-          const schemaFunctions = functions.map(parseProperty).join('');
+          const schemaFunctions = functions
+            .map(([fn, args]) =>
+              fn === 'allOf'
+                ? parseProperty(['flattenAllOf', args])
```
soartec-lab, line 584: Is it possible to specify the identifier of `function` when selecting `fn`, rather than determining it at runtime?

https://github.com/orval-labs/orval/blob/ea74a17e22bd8ef089dd2035ae372160077c1081/packages/zod/src/index.ts#L372

soartec-lab (commented):
@timmisset 
I have confirmed the output results. I have consulted with you about the implementation, so please check it.

packages/zod/src/index.ts
```diff
@@ -530,6 +530,22 @@ export const parseZodValidationSchemaDefinition = (
   const parseProperty = (property: [string, any]): string => {
     const [fn, args = ''] = property;
 
+    if (fn === 'flattenAllOf') {
+      const flattenedProperties: Record<string, any> = args.reduce(
+        (
+          acc: Record<string, any>,
+          { functions }: { functions: [string, any][] },
+        ) => {
+          functions.forEach(([_fn, schema]) => {
```
timmisset, line 539: @soartec-lab, the issue that's now coming up is that the 'schema' parameter in this line is not necessarily an object which causing problems with the flattening. I think that workaround is making the code far too complex for the purpose of the PR that causes this: https://github.com/orval-labs/orval/pull/1907. 
To correctly flatten the allOf you will need to know quite a few things that happen downstream (nested allOfs, object types etc).
Both the PR from @Georgegriff  and this one will adding complexity for the minimal gain of rewriting discriminatedUnions. 

My suggestion would be to simply rollback the PR from @Georgegriff  as he also suggested: https://github.com/orval-labs/orval/issues/2085#issuecomment-2889973437.

soartec-lab: @timmisset 
That certainly seems reasonable.

> My suggestion would be to simply rollback the PR from @Georgegriff as he also suggested: https://github.com/orval-labs/orval/issues/2085#issuecomment-2889973437.

So does that mean I can choose to merge this PR or revert #1907?

timmisset: @soartec-lab please revert #1907, thanks!

soartec-lab: Hi, @Georgegriff , thanks for your response to #1907.
But it seems to have introduced some issues as explained in this PR.
We are trying to revert this as the drawbacks of the bugs it created are greater than the improvements in #1907, but what are your thoughts on this?

Georgegriff: I can't see a good way of supporting both so I think reverting makes sense

Georgegriff: We might need to be careful when reverting to not accidentally roll back subsequent changes, a manual revert, looking at the changes and tests  to remove, might make sense

soartec-lab: Okey

@timmisset 
If you like, could you close this PR and open a new one that reverts the changes above?

timmisset: @soartec-lab PR is ready: https://github.com/orval-labs/orval/pull/2118

PR #2103, opened 2025-05-20
fix: Angular client props nullability

### LINKED ISSUE #2092: Angular schema nullability definitions incorrectly done
I'm using orval 7.9.0 and I have the following component schema:
```json
"components": {
	"schemas": {
	   "TestDto":{
		  "type":"object",
		  "required":[
			 "Foo",
			 "Bar"
		  ],
		  "properties":{
			 "Foo":{
				"type":"string"
			 },
			 "Bar":{
				"nullable":true,
				"type":"string"
			 },
			 "Foobar":{
				"nullable":true,
				"type":"string"
			 }
		  }
	   }
	}
}
```

and the generated model with `angular` client output is as such:
```typescript
export interface TestDto {
  Foo: string;
  /** @nullable */
  Bar: string;
  /** @nullable */
  Foobar?: string;
}
```
As you can see the Bar and Foobar properties both have the "nullable" comment (but Bar surprisingly isn't nullable nor optional at all) and Foobar is just an optional property. 

If I generate with `zod` client output I get what I expected:
```typescript
export interface TestDto {
  Foo: string;
  /** @nullable */
  Bar: string | null;
  /** @nullable */
  Foobar?: string | null;
}
```

Is there any specific reason for why angular generation works this way or is it just a bug?

jonimake:
## Status

**READY**

## Description

Handle angular http client parameter nullability in a more robust way that doesn't force all schema properties to be non nullable. Adds a new utility type `DeepNonNullable` that is used to wrap all angular client query param props instead of making all scalars non-nullable in angular configuration.

Fix #2092 for reference

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

You can either run the tests for the project (there is a new test case for the happy path for angular in `propts.test.ts`) or you can generate the sample angular client. The http client query params should be wrapped with `DeepNonNullable` type to enforce no nulls can be set there.


melloware:
Running the build now!

melloware:
@jonimake looks like tests are failing?

niklissad:
any updates? it's look like a critical bug with strict types enabled in project

melloware:
this has been moved here can you guys respond in that PR please: https://github.com/orval-labs/orval/pull/2244

PR #2118, opened 2025-05-28
feat(zod): revert #1907 to not use discriminatedUnion

### LINKED ISSUE #2085: zod.discriminatedUnion does not work in combination with inheritance

timmisset:
## Status

Ready

## Description

Fix #2085 and https://github.com/orval-labs/orval/pull/2099
revert #1907


melloware:
cc @Georgegriff

soartec-lab (approved):
Thanks！

Georgegriff:
Cross posting from issue: https://github.com/orval-labs/orval/issues/2085#issuecomment-3077567650

PR #2130, opened 2025-05-31
fix: use minLength and maxLength when mocking strings if defined

### LINKED ISSUE #2124: Mocks don't use minLength and maxLength when mocking string values

I have following yaml.

```
openapi: '3.0.2'
info:
  title: API
  version: '1.0'
 
components:
  schemas:
    value:
      type: object
      required:
       - string
       - integer
      properties:
        string:
          type: string
          minLength: 5
          maxLength: 20
        integer:
          type: integer
          minimum: 0
          maximum: 10

paths:
  /path1:
    get:
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:

```

This is my Orval config

```

export default defineConfig({
  test: {
    output: {
      target: './api.ts',
      mock: true,
    },
    input: {
    },
  },
});
```

## What happens?

I get following mock

```
    string: faker.string.alpha(20),
    integer: faker.number.int({min: 0, max: 10}),
...overrideResponse})
```

Should result to following utilizing the minLength and maxLength values.

```
    string: faker.string.alpha({ length: { min: 5, max: 20 }}),
    integer: faker.number.int({min: 0, max: 10}),
...overrideResponse})
```

Orval 7.9.0

SampsaKaskela:
## Status

**WIP**

## Description

Fixes #2124

Uses minLength and maxLength properties for mocking strings if defined in yaml. Wanted to keep current behavior same so if neither is defined then uses default length of 20 as currently, although in my opinion this could also be changed to range so user could test different string lengths. I decided that if only minLength is defined then code uses minLength + 20 as max so user of the library would be able to test some range different of values at least. In case of only maxLength being defined, minLength is set to 0. My reasoning with this behavior was that if user had bothered to define either value then they are likely aware of existence of these properties and so we can make assumption that other one was left blank on purpose. Open to different ideas if this is not desirable.

## Related PRs

--

## Todos

Which file should the tests be and does this need documentation?

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

Build the repo as normally and test changes by defining minLength and maxLength for string value in yaml.

### DIFF
--- a/samples/basic/petstore.yaml
+++ b/samples/basic/petstore.yaml
@@ -143,8 +143,8 @@ components:
         name:
           type: string
           description: 'Name of pet'
-          maxLength: 0
-          minLength: 40
+          minLength: 0
+          maxLength: 40
         age:
           type: integer
           format: int32
--- a/samples/react-query/basic/petstore.yaml
+++ b/samples/react-query/basic/petstore.yaml
@@ -143,8 +143,8 @@ components:
         name:
           type: string
           description: 'Name of pet'
-          maxLength: 0
-          minLength: 40
+          minLength: 0
+          maxLength: 40
         age:
           type: integer
           format: int32
--- a/samples/react-query/custom-client/petstore.yaml
+++ b/samples/react-query/custom-client/petstore.yaml
@@ -143,8 +143,8 @@ components:
         name:
           type: string
           description: 'Name of pet'
-          maxLength: 0
-          minLength: 40
+          minLength: 0
+          maxLength: 40
         age:
           type: integer
           format: int32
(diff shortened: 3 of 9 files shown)


melloware:
hmmm looks like tests are failing?

SampsaKaskela:
Yeah miss counted the }. Didn't realize yarn test command didn't actually run tests within /tests directory. Should pass now.

SampsaKaskela:
I noticed the samples that have minLength and maxLength defined are actually backwards. For example
```
        name:
          type: string
          description: 'Name of pet'
          maxLength: 0
          minLength: 40
```

Should I fix those in this PR?

melloware (approved):

PR #2131, opened 2025-05-31
fix: use minItems and maxItems when mocking arrays

### LINKED ISSUE #2129: Mocks don't use minItems and maxItems when mocking arrays

I have following yaml.

```
openapi: '3.0.2'
info:
  title: API
  version: '1.0'
 
components:
  schemas:
    value:
      type: object
      required:
       - array
      properties:
        array:
          type: array
          items:
              type: string
          minItems: 1
          maxItems: 20

paths:
  /path1:
    get:
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
```

This is my Orval config

```

export default defineConfig({
  test: {
    output: {
      target: './api.ts',
      mock: true,
    },
    input: {
    },
  },
});

```
## What happens?

I get following mock. I guess it gets the 1 and 10 from arrayMin and arrayMax default values in Orval config.

```
  array: Array.from({ length: faker.number.int({ min: 1, max: 10 }) }, (_, i) => i + 1).map(() => (faker.string.alpha(20))), 
...overrideResponse})
```

It should set min and max properties according to minItems and maxItems.

```
  array: Array.from({ length: faker.number.int({ min: 1, max: 20 }) }, (_, i) => i + 1).map(() => (faker.string.alpha(20))), 
...overrideResponse})
```

No errors

## Any other comments?

In config there is arrayMin and arrayMax but to my knowledge those apply to all arrays. I'm guessing it would be desirable that arrayMin and arrayMax override the regular array sizes but if they are not given then minItems and maxItems should be used if they exist and as a 3rd fallback default to 1 and 10 as currently.

Orval 7.9.0

SampsaKaskela:
## Status

**READY**

## Description

Fixes #2129

Use minItems and maxItems properties when mocking arrays. arrayMin and arrayMax options can still be used to override these. Removed default arrayMin and arrayMax values from normalizedOptions so we can detect when they are not set. Fallback values are handled when values are actually used further down the code.

## Related PRs

--

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

Build repo and use minItems and maxItems in yaml for array types.

### DIFF
--- a/packages/mock/src/faker/getters/scalar.ts
+++ b/packages/mock/src/faker/getters/scalar.ts
@@ -225,8 +225,8 @@ export const getMockScalar = ({
       return {
         value:
           `Array.from({ length: faker.number.int({ ` +
-          `min: ${mockOptions?.arrayMin}, ` +
-          `max: ${mockOptions?.arrayMax} }) ` +
+          `min: ${mockOptions?.arrayMin ?? item.minItems ?? 1}, ` +
+          `max: ${mockOptions?.arrayMax ?? item.maxItems ?? 10} }) ` +
           `}, (_, i) => i + 1).map(() => (${mapValue}))`,
         imports: resolvedImports,
         name: item.name,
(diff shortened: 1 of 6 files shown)


SampsaKaskela:
As an additional note I noticed that when updating samples some files changed before I committed any changes so I excluded those files from this commit. I can commit those files as well if desired even tho they are not related to changes in this PR but past PRs.

melloware (approved):

PR #2132, opened 2025-05-31
feat: Added class client builder

### LINKED ISSUE #2127: Generate a client class, that takes http client as a constructor argument
My use case is probably pretty niche - I have multiple backends, that I need to connect to, they are dynamic and can change at runtime, thus I need multiple instances of clients - mutators do not suffice here. I also think it is a bit simpler to work with classes instead of functions + mutators in general, so it might be a nice to have for more people.

I have already implemented this usinga a lightly modified version of angular client and can make a PR - somethng like 'axios-class' or 'class' client. Would you be interested in this contribution?

blazmrak:
## Status

**WIP**

## Description

Added a class client generator for axios.

Closes #2127

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

Added the generated file in the basic sample.


melloware:
Hmm so this is just a raw Axios builder? I wonder if `class` is the right name?

blazmrak:
I'm open to anything. I chose class, because it could be implemented for any http client in the future.

melloware:
@soartec-lab thoughts?

soartec-lab:
Hi, @blazmrak .

Thank you for your interesting suggestions. Let me ask some questions to deepen my understanding.

1. This doesn't mean that new clients will be added, but rather that the implementation format of the client is class-based or function-based, right?
2. In that case, I will think about the name again, but would it be okay to use a different option, for example, `implementationFormat: "functionBase" | "classBase"`?
3. Also, could you please explain in more detail what advantages you have of dynamically changing the URL using a class compared to the current method of passing the URL as an argument to fetch?

blazmrak:
Yeah, introducing another option is probably the way to go, although it explodes the combinations with `client` + `mode` + `httpClient`... I tried to avoid that, but this makes it so that even other clients can be generated as classes with less hacks.

Regarding the 3rd point, the main reason for this feature is that I need to connect to multiple servers exposing the same API (docker daemons for my use case), that are not known at build time and classes over functions are the way to go for this. It's just easier to have a long lived instance of a client instead of having to pass the same parameters over and over again and juggle the connections, practically having to hand roll a wrapper around the client generated from open api spec.

This is very niche, as I said in the issue, I'm probably the first one where this requirement is pretty much a necessity and will probably be the only one for a long while.

Another benefit, in my eyes at least, is that classes with custom instance injected are easier to reason about, test and are more flexible - you just instantiate new instance in code instead of having to configure it via environment variables with the added benefit of autocompletion and namespacing in the code, also making it clearer to the reader about what is happening (`await listPets()` vs `await this.client.listPets()`). This is very much my personal preference and a side product, not the reason for the PR, I mention it, just because someone else was also interested in this feature for this reason.

soartec-lab:
@blazmrak 

I understand. Thank you for the clear explanation.
If you have multiple servers, you can separate them, for example by tags, define different custom instances, and output clients for multiple servers, allowing you to reuse them without having to specify the URL each time.

https://orval.dev/reference/configuration/inputhttps://orval.dev/reference/configuration/input

Does this method solve your use-case?

blazmrak:
No, like I said, the servers are dynamic and only known at runtime. The usual use case, where you are generating the client for a specific API provider ahead of time, doesn't work for me.

Concretely: I'm building a self hosting platform, which will manage multiple servers - think poor man's kubernetes/digital ocean. When the user adds/removes the servers from the "cluster" dynamically I'll have to make the connection to the docker daemon that runs on that server.

I'm generating docker client from the open api spec, because existing clients are either outdated, incomplete or don't offer the option to make the connection over SSH.

soartec-lab:
@blazmrak 
I now understand edge cases well.
Since the `class-base` implementation raises concerns about `tree-shaking` and other issues, I think it should be defined as a separate option like the `implementationFormat` option mentioned above, rather than simply adding client.

blazmrak:
@soartec-lab I already changed it to use `implementationFormat` option.

soartec-lab:
@blazmrak 
Thanks. This is an edge case, so I'll wait to see what other users think about it.

melloware:
I agree this is more an edge case and not a core use case. Thanks though @blazmrak !

PR #2134, opened 2025-06-01
fix: use multipleOf when mocking numbers

### LINKED ISSUE #2133: Mocks don't use multipleOf when mocking numbers

I have following yaml

```
openapi: '3.0.2'
info:
  title: API
  version: '1.0'
 
components:
  schemas:
    value:
      type: object
      required:
       - integer
      properties:
        integer:
          type: integer
          minimum: 0
          maximum: 100
          multipleOf: 10

paths:
  /path1:
    get:
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
```

This is my Orval config

```

export default defineConfig({
  test: {
    output: {
      target: './api.ts',
      mock: true,
    },
    input: {
    },
  },
});
```

## What happens?

I get following mock

```
    integer: faker.number.int({min: 0, max: 100}),
...overrideResponse})
```

I should get following which utilizes multipleOf

```
    integer: faker.number.int({min: 0, max: 100, multipleOf: 10}),
...overrideResponse})
```

No errors

Orval 7.9.0

SampsaKaskela:
## Status

**WIP**

## Description

Fixes #2133

Use multipleOf when mocking numbers if defined.

## Related PRs

--

## Todos

Tests are currently failing due to TS errors because Faker 8 that is currently used doesn't support multipleOf that was added in version 9. This change doesn't break current mocks but Faker 8 doesn't utilize it either. Upgrading to version 9 will cause tests to pass but not sure if this is breaking change that should be considered separately.

## Steps to Test or Reproduce

Build repo and use multipleOf in yaml file.

### DIFF
--- a/packages/mock/src/faker/getters/scalar.ts
+++ b/packages/mock/src/faker/getters/scalar.ts
@@ -137,7 +137,7 @@ export const getMockScalar = ({
     case 'number':
     case 'integer': {
       let value = getNullable(
-        `faker.number.int({min: ${item.minimum}, max: ${item.maximum}})`,
+        `faker.number.int({min: ${item.minimum}, max: ${item.maximum}, multipleOf: ${item.multipleOf}})`,
         item.nullable,
       );
       const numberImports: GeneratorImport[] = [];
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "rimraf": "^5.0.5",
     "tsup": "^8.4.0",
     "turbo": "^1.13.3",
-    "typescript": "^4.7.4",
+    "typescript": "^5.2.2",
     "typescript-eslint": "^8.26.0",
     "vitest": "^0.6.3",
     "zx": "^7.2.3"
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "@commitlint/config-conventional": "^19.8.0",
     "@eslint/eslintrc": "^3.2.0",
     "@eslint/js": "^9.25.1",
-    "@faker-js/faker": "^8.4.0",
+    "@faker-js/faker": "^9.8.0",
     "@release-it-plugins/workspaces": "^4.2.0",
     "@release-it/conventional-changelog": "^10.0.0",
     "@types/node": "^20.13.0",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -14,7 +14,7 @@
     "test": "vitest run test.ts --global"
   },
   "devDependencies": {
-    "@faker-js/faker": "^8.4.0",
+    "@faker-js/faker": "^9.8.0",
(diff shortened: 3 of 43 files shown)


melloware (changes requested):
looks like tests failing?

SampsaKaskela:
As mentioned, for tests to pass upgrade to Faker v9 would be needed but I'm not sure if upgrading to Faker v9 would be safe to do. It didn't cause issues for me and tests did pass so is it okay to upgrade or would that be considered breaking change although it is only devDependency?

melloware:
Personally i have no problem with Faker 9  I think we should try and stay on later libraries. We are also upgrading to Zod4 so I think its OK

SampsaKaskela:
Seems like Faker 9 requires Typescript 5 to work so I upgraded to TS v5 for samples that had the issues. Some samples were already using TS 5 but not all. While this fixes the problems this makes me bit concerned that would putting this to patch release mean that it will potentially break checks if someone is using TS v4 and tsc command to type check their generated code like these checks do.

melloware:
I can talk to the team about making this 8.0 as we are also updating to Zod 4. Probably not a bad idea to bump major versions @anymaniax @soartec-lab

soartec-lab:
@SampsaKaskela @melloware 
To maintain compatibility, you should check the version definition in package.json and change the behavior.
This is an implementation that maintains compatibility with zod v4. Please refer to this.

https://github.com/orval-labs/orval/blob/master/packages/zod/src/compatibleV4.ts#L11

packages/mock/src/faker/getters/scalar.ts
```diff
@@ -137,7 +137,7 @@ export const getMockScalar = ({
     case 'number':
     case 'integer': {
       let value = getNullable(
-        `faker.number.int({min: ${item.minimum}, max: ${item.maximum}})`,
+        `faker.number.int({min: ${item.minimum}, max: ${item.maximum}, multipleOf: ${item.multipleOf}})`,
```
soartec-lab, line 140: How about making `multipleOf` available only for faker v9?

SampsaKaskela: Sounds good to me. Since this is merged and closed, is it okay if I do it in #2141 ?

soartec-lab: @SampsaKaskela 
Thank you. This is a different issue from #2142, so please create a new one.

soartec-lab:
@melloware 

Regarding the version, I don't think there is a need to make a major version upgrade right away.
It will take a little more time to stabilize `zod v4`, and there are still some other breaking changes remaining.
Therefore, we would like to release the next release while maintaining compatibility, and then make a major update after a certain period of time has passed.

PR #2141, opened 2025-06-02
fix: use faker.number.float when mocking numbers and int when mocking integers

### LINKED ISSUE #2140: Use faker float when mocking numbers and int when mocking integers.
At the moment when mocking any numbers they will always be integers since they are mocked with faker.number.int function. Only integers should be mocked with int and regular numbers should be mocked with float function. You can currently specify float or double as a format for number but it doesn't take into account for example minimum and maximum attributes. We could settle with some kind of default precision for floating point numbers since I don't think people want numbers with tons of decimals as default so maybe defaulting to say 2 would be reasonable. Mock options could allow setting fractionDigits option (same name as in Fakerjs 9) that allows user to set what they want as precision when mocking floats. This behavior would be overridden if user specifies multipleOf in source file since that will then determine the precision.

SampsaKaskela:
## Status

**READY**

## Description

Fixes #2140

Uses faker.number.float instead of int when mocking number types. Still uses int for integer types. Add fractionDigits to mock options to specify how many decimals to show with default value of 2. Added fractionDigits: 1 to one of the tests to verify it doesn't cause type errors and that override works. Float function uses 999999999 for max value as default since default value otherwise is 1.0 so we get wider range of possible values. Removed double and float formats since they would hinder with new functionality.

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

Build repo and use type: number in yaml and use fractionDigits in override.mock.fractionDigits in config.


soartec-lab:
Hi, @SampsaKaskela did you regenerate the samples and there was no change?

SampsaKaskela:
Yes I did. None of the samples did use type number. They all use integer type for numbers. Tests do use type number but generated test results are not commited to repo so looks like nothing has changed.

packages/mock/src/faker/getters/scalar.ts
```diff
@@ -144,6 +144,12 @@ export const getMockScalar = ({
         `faker.number.int({min: ${item.minimum}, max: ${item.maximum}${isFakerV9 ? `, multipleOf: ${item.multipleOf}` : ''}})`,
         item.nullable,
       );
+      if (type === 'number') {
+        value = getNullable(
+          `faker.number.float({min: ${item.minimum}, max: ${item.maximum ?? 999999999}, ${item.multipleOf ? `multipleOf: ${item.multipleOf}` : `fractionDigits: ${mockOptions?.fractionDigits ?? 2}`}})`,
```
soartec-lab: > Only one of multipleOf or fractionDigits should be passed.

It seems that `multipleOf` and `fractionDigits` can only accept one or the other.
In this implementation, if `multipleOf` is specified, both will be defined, which violates the interface, right?

Ref: https://fakerjs.dev/api/number

soartec-lab: The default for `fractionDigits` is determined when normalizing.

https://github.com/orval-labs/orval/blame/master/packages/orval/src/utils/options.ts

The default was removed by this PR, but it needs to be reverted.

https://github.com/orval-labs/orval/commit/55e36b66276eb5f5fdc1ce4e573c7ab57169643b

```suggestion
          `faker.number.float({min: ${item.minimum}, max: ${item.maximum ?? 999999999}, ${item.multipleOf ? `multipleOf: ${item.multipleOf}` : `fractionDigits: ${mockOptions?.fractionDigits}`}})`,
```

SampsaKaskela: The code currently does specify only one. If you have specified multipleOf in yaml it will write multipleOf and if not it will write fractionDigits.

SampsaKaskela: fractionDigits is new option introduced in this PR. In PR you mention I moved the default values of arrayMin and arrayMax out of normalized options so we could determine if user actually is overriding the minItems and maxItems used in yaml. If default values are in options, we would have no way of determining if we should use the mockOptions or minItems and maxItems in yaml. That was at least the reasoning behind the change although I'm sure there could be cleaner solution than what I used. Of course in this case fractionDigits default value can be in normalized options because we don't care if user has defined it or not unlike with arrayMin and arrayMax so I can move it there.

soartec-lab: Ah, I totally understand 👍

soartec-lab: @SampsaKaskela 
How about leaving the default specification, and then prioritizing it if it is specified on an item-by-item basis, and using the default value if it is not specified?

SampsaKaskela: So in the PR it has this logic for example in scalar.ts for the arrayMin usage

`min: ${mockOptions?.arrayMin ?? item.minItems ?? 1}`

Would you want me to change it to

`min: ${item.minItems ?? mockOptions?.arrayMin}`

So basically use mockOptions value only if the value is not defined in yaml? I can definitely do that just why I made mockOptions priority if defined was that if user defines their overrides then they might get confused if it does nothing. But I can change it to this behavior if you think it is better.

soartec-lab: @SampsaKaskela 
That's right, I'll accept your suggestion.
Then, since `fractionDigits` is the only one that doesn't have a definition in the item, let's move the default definition there 👍

SampsaKaskela: Okay I leave the array behavior as is and move fractionDigits default to normalized options.

SampsaKaskela: @soartec-lab Now that I started thinking about the arrayMin and arrayMax I think your suggestion might actually be better when thinking about docs. Like docs at the moment say default values are 1 and 10 so if we go with my implementation then behavior is bit different depending if you defined those or not. If we went with your suggestion then we just would have to update the docs and mention that it is the fallback value in case minItems and maxItems don't exist.

soartec-lab: @SampsaKaskela 

Yes, that's right. I respect your opinion, but I was thinking the same thing you're thinking now.
orval outputs based on the specifications defined by openapi, so if you want to define min and max, you should define them in openapi as actual specifications.
And since that should be different for each item, I think it's better to respect individual definitions rather than overwriting everything with the overridden value.

SampsaKaskela: Do you want me to make it so in this PR or do you want me to make separate PR for it?

soartec-lab: @SampsaKaskela 
Since I will be merging this PR, I would like to create a new one.

soartec-lab (approved):
Thanks!

PR #2142, opened 2025-06-03
feat: use multipleOf only if faker v9 is installed

SampsaKaskela:
## Status

**READY**

## Description

Check if Faker v9 is installed and use multipleOf if it is. If not, omit multipleOf to maintain compatibility.

## Related PRs

Discussed in #2134 about maintaining compatibility.

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

Install Faker v8 and perform tests.


SampsaKaskela:
@soartec-lab This should now work with Faker 8 and 9. Is this what you were looking for?

soartec-lab:
That's absolutely true. I'll check out this PR later.

packages/mock/src/faker/getters/scalar.ts
```diff
@@ -132,12 +133,15 @@ export const getMockScalar = ({
...
+  const isFakerV9 =
+    !!context.output.packageJson &&
+    isFakerVersionV9(context.output.packageJson);
 
   switch (type) {
     case 'number':
     case 'integer': {
       let value = getNullable(
-        `faker.number.int({min: ${item.minimum}, max: ${item.maximum}, multipleOf: ${item.multipleOf}})`,
+        `faker.number.int({min: ${item.minimum}, max: ${item.maximum}${isFakerV9 ? `, multipleOf: ${item.multipleOf}` : ''}})`,
```
soartec-lab, line 144: Should it be limited to only when `item.multipleOf` is present?

```suggestion
        `faker.number.int({min: ${item.minimum}, max: ${item.maximum}${isFakerV9 && item.multipleOf ? `, multipleOf: ${item.multipleOf}` : ''}})`,
```

SampsaKaskela: Sure I can do that. Was following the practice that was used with min and max. Should I make it so that those are omited too if not defined?

soartec-lab: I understand your intention. It's OK as is 👍

soartec-lab (approved):
Thanks!

PR #2143, opened 2025-06-03
fix(fetch): respect OpenAPI 3.0 explode default value for query params

yamish123:
## Status

**READY**

## Description

Fix explode parameter default value handling for query parameters in fetch client. According to [OpenAPI 3.0 specification](https://swagger.io/docs/specification/v3_0/serialization/#query), the default value of `explode` for query parameters should be `true` when the style is `form` (which is the default). The previous implementation only considered parameters with explicitly set `explode: true`, incorrectly excluding parameters where `explode` was undefined (should default to `true`).

## Related PRs

## Todos

- [ ] Tests (Auto-generated tests updated via `yarn generate`)
- [ ] Documentation

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout <branch>
> cd tests
> yarn generate
```

Check the generated code in `tests/generated/fetch/parameters/endpoints.ts` to verify that query parameters without explicit `explode: false` are included in the `explodeParameters` array

### DIFF
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -60,7 +60,7 @@ export const generateRequestFunction = (
   const explodeParameters = parameters.filter((parameter) => {
     const { schema } = resolveRef<ParameterObject>(parameter, context);
 
-    return schema.in === 'query' && schema.explode;
+    return schema.in === 'query' && schema.explode !== false;
   });
 
   const explodeParametersNames = explodeParameters.map((parameter) => {


soartec-lab:
@yamish123 

Oops, up until now, `explode: false` behaved the same as `explode: true`, right?

soartec-lab:
Correction. What you are saying is that the parameters should be in the format when `explode` is set to `true` even if the `explode` option is not specified, right?

yamish123:
> Correction. What you are saying is that the parameters should be in the format when explode is set to true even if the explode option is not specified, right?

That is correct. There is only one problem, and that is when the `explode` is `undefined`.

| Query Parameter Definition | Current Behavior | Expected Behavior | Status |
|---------------------------|---------------------------|------------------|-------------------|
| `explode: true` | ✅ Included in explodeParameters | ✅ Included in explodeParameters | Correct |
| `explode: false` | ✅ Excluded from explodeParameters | ✅ Excluded from explodeParameters | Correct |
| `explode` not specified | ❌ Excluded from explodeParameters | ✅ Should be included in explodeParameters | **Fixed by this PR** |


According to the [OpenAPI 3.0 specification](https://spec.openapis.org/oas/v3.0.2#fixed-fields-9), the default value of `explode` is `true` when `style` is `form`, so if `explode` is `undefined`, it should be treated as `true`.

soartec-lab:
@yamish123 
That's correct.
What's bothering me now is that it's a breaking change.
I think you need to add `override.fetch.explode: true` to maintain compatibility. So If you like, would you please add an option to maintain compatibility?

yamish123:
@soartec-lab 
I have added the override.fetch.explode option for backward compatibility.
[b15116c](https://github.com/orval-labs/orval/pull/2143/commits/b15116ce685c40cd1bcaedd220af02bbab954789)

Also, changed override.fetch.ignoreHttpResponseReturnType to optional for users who specify only explode
[b3ef9f9](https://github.com/orval-labs/orval/pull/2143/commits/b3ef9f9a2b290be094f6ca832ad03fc250e837ed)

docs/src/pages/guides/fetch-client.md
```diff
@@ -163,3 +163,64 @@ module.exports = {
   },
 };
 ```
+
+#### query parameter explode behavior
```
soartec-lab: Okay, so this should go in as an option to "references".

https://github.com/orval-labs/orval/blob/master/docs/src/pages/reference/configuration/output.md?plain=1#L962

soartec-lab (commented):
@yamish123 
Thanks! I will ask you to move the document.

docs/src/pages/guides/fetch-client.md
yamish123: @soartec-lab 
I have moved the documentation on `explode` to output.md :+1:.
[4f38a65](https://github.com/orval-labs/orval/pull/2143/commits/4f38a651f44d7264504fc33998a8398a9a995479)

yamish123: @soartec-lab 
I have moved the documentation on `explode` to output.md :+1:.
[1cc5957](https://github.com/orval-labs/orval/pull/2143/commits/1cc5957a109c297928c35d8928e5ef0ff9b942cc)

soartec-lab (approved):
Thanks for the great improvement 🙌

PR #2150, opened 2025-06-06
feat: allow defining min and max ranges for mocked numbers and strings

### LINKED ISSUE #2148: Allow defining default min and max ranges for numbers and strings when mocking
Currently you can defined arrayMin and arrayMax in mock options to default to if they are not defined in input file. Same could be allowed for strings and numbers. In many cases not every value can be given min and max lengths in input file. Values might be too arbitrary or specification might be incomplete. At the moment only workaround is to override all numbers manually but for large apis this is tedious and defeats the purpose of using generator.

Currently for numbers the default range is 0 to Number.MAX_SAFE_INTEGER which is the default in Faker but might not be desired always because not everyone wants their mock to return numbers that are in trillions or more. For string we are currently defaulting to 20 which is fair default but user might want something different and even some cases want to test with different string lengths.

Proposal would be following which follows same naming convention as arrayMin and arrayMax currently.

```
{
    output: {
      mock: true,
      override: {
        mock: {
          stringMin: 10,
          stringMax: 30,
          numberMin: 0,
          numberMax: 1000
        },
      },
    },
  }
```

SampsaKaskela:
## Status

**READY**

## Description

Fix #2148

Adds stringMin, stringMax, numberMin and numberMax options that can be used to define default ranges for generated values if range is not defined in specification.

## Todos

- [ ] Tests
- [x] Documentation

## Steps to Test or Reproduce

Build repo and use any of the above options in mock options. Values should be reflected in generated mocks.


melloware (dismissed):

soartec-lab (dismissed):
Thanks for your all improvements.

soartec-lab:
@SampsaKaskela Could you fix a conflict?

SampsaKaskela:
Noticed that bigInts didn't get these defaults if set since they were implemented in separate if statement. I moved the logic inside the switch case so it uses same implementation as the regular int and just switches out to bigInt if conditions are met.

I also added integer type to tests since there actually seemed to have been minor bug that if you specified int64 as format then regardless of type it would get converted to number. Not that it is common mistake anyways.

Issue #1839, opened 2025-01-21
Add support for error handling in Orval built in fetch


oferitz:
I understand that this issue has been discussed here: [#1387](https://github.com/orval-labs/orval/issues/1387#issuecomment-2244738742), and the suggested solution is to create a custom fetch. 

However, creating a custom fetch can sometimes be an overhead and introduce unwanted side effects, as highlighted in this issue: [#1838](https://github.com/orval-labs/orval/issues/1838).

Given these challenges, I believe it would be reasonable to consider adding support for error handling directly in Orval's built-in fetch. This enhancement would simplify the developer experience and avoid the need for custom implementations in many cases.

Your consideration of this feature would be much appreciated!

soartec-lab:
@oferitz 
I understand that you want to add the following to the source code automatically generated by orval. Is my understanding correct?

```
if (!response.ok) {
  throw new Error('Network response was not ok')
}
```

Ref: https://tanstack.com/query/latest/docs/framework/react/guides/query-functions#usage-with-fetch-and-other-clients-that-do-not-throw-by-default

oferitz:
I want to preserve the original error so something like:


```javascript
fetch('https://api.example.com/data')
  .then(async response => {
    if (!response.ok) {
      const errorData = await response.json();
      // Preserve the original status and error details
      const error = new Error(errorData.message || response.statusText);
      error.status = response.status;
      error.data = errorData;
      throw error;
    }
    return response.json();
  })
  .catch(error => {
    // Handle both network errors and response errors
    console.error('Status:', error.status);
    console.error('Message:', error.message);
    console.error('Data:', error.data);
  });
```

soartec-lab:
@oferitz 
Thanks. The implementation of error handling varies depending on the project. Therefore, we are not thinking about supporting one case with orval.

oferitz:
@soartec-lab Thanks for your feedback. I understand your point about implementation variations. Let me propose a more standardized approach that strictly follows the Response API specification.
The following example only uses standard Response API properties (status and statusText), making it more universal and aligned with web standards. This approach:

- Preserves the essential error information from the Response object
- Follows standard practices for HTTP error handling
- Doesn't make assumptions about error response body structure

This minimal implementation could serve as a foundation that users can extend based on their specific needs. Would you consider supporting this basic level of error handling in Orval?

```ts
// Basic error handling aligned with Response API spec
fetch('https://api.example.com/data')
  .then(async response => {
    if (!response.ok) {
[...]

PR #2168, opened 2025-06-17
feat(fetch): add ability to throw on error

### LINKED ISSUE #2047: Allow `fetch`client's `data` to be set to success type only
# Background

In #1885, we defined the type of `response`'s `data` and `status` as a combination.
When used with the `fetch` client, this is a more accurate type definition.

However, when used as an HTTP client such as `swr` or `react-query`, it becomes a union type of 200 and 400, and the type cannot be specified. And in those libraries, it is common to use exceptions for cases other than 200 and refer to `data` only when it is 200.

For example, `swr`:

https://swr.vercel.app/docs/error-handling

# Proposal

There are cases where you want to define only 200 data as before and make only the status a union type, so by making these options selectable, we would like to maintain the format before #1885.

AllieJonsson:
## Status

**READY**

## Description

Fixes #2047 

When using a package like `swr` or `query`, they use exceptions to know if the response is an error or not. 
To accomplish this, I added the `shouldThrowOnError` setting. When set to true, the function is typed to return only the success type (or success types, if there are multiple), when set to false, the behavior is unchanged.

When the endpoint does not have an error specified:
```ts
export type listPetsResponse200 = {
  data: Pets;
  status: 200;
};

export type listPetsResponseSuccess = listPetsResponse200 & {
  headers: Headers;
};

  params: ListPetsParams,
  options?: RequestInit,
): Promise<listPetsResponseSuccess> => {
    ...options,
    method: 'GET',
  });

  if (!res.ok) {
    const err: globalThis.Error & { info?: any; status?: number } =
      new globalThis.Error();
    const data = body ? JSON.parse(body) : {};
    err.info = data;
    err.status = res.status;
    throw err;
  }

  const data: listPetsResponseSuccess['data'] = body ? JSON.parse(body) : {};
  return {
    data,
    status: res.status,
    headers: res.headers,
  } as listPetsResponseSuccess;
};
```

When the endpoint does have an error specified:
```ts
export type createPetsResponse200 = {
  data: Pet;
  status: 200;
};

export type createPetsResponseDefault = {
  data: Error;
  status: Exclude<HTTPStatusCodes, 200>;
};

export type createPetsResponseSuccess = createPetsResponse200 & {
  headers: Headers;
};
export type createPetsResponseError = createPetsResponseDefault & {
  headers: Headers;
};

export const createPets = async (
  createPetsBody: CreatePetsBody,
  params: CreatePetsParams,
  options?: RequestInit,
): Promise<createPetsResponseSuccess> => {
  const res = await fetch(getCreatePetsUrl(params), {
    ...options,
    method: 'POST',
    body: JSON.stringify(createPetsBody),
  });

  if (!res.ok) {
    const err: globalThis.Error & {
      info?: createPetsResponseError['data'];
      status?: number;
    } = new globalThis.Error();
    const data: createPetsResponseError['data'] = body ? JSON.parse(body) : {};
    err.info = data;
    err.status = res.status;
    throw err;
  }

  const data: createPetsResponseSuccess['data'] = body ? JSON.parse(body) : {};
  return {
    data,
    status: res.status,
    headers: res.headers,
  } as createPetsResponseSuccess;
};
```

### DIFF
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -254,27 +254,23 @@ ${override.fetch.shouldThrowOnError ? '' : `export type ${responseTypeName} = ($
   }
 `;
   const throwOnErrorImplementation = `if (!res.ok) {
-    const err: globalThis.Error & {${hasError ? `info?: ${errorName}${override.fetch.includeHttpResponseReturnType ? "['data']" : ''}, ` : ''}status?: number} = new globalThis.Error();
-    ${
-      hasError
-        ? `const body = [204, 205, 304].includes(${isNdJson ? 'stream' : 'res'}.status) ? null : await ${isNdJson ? 'stream' : 'res'}.text()
-    const data: ${errorName}${override.fetch.includeHttpResponseReturnType ? `['data']` : ''} = body ? JSON.parse(body) : {}
+    const err: globalThis.Error & {info?: ${hasError ? `${errorName}${override.fetch.includeHttpResponseReturnType ? "['data']" : ''}` : 'any'}, status?: number} = new globalThis.Error();
+    const data ${hasError ? `: ${errorName}${override.fetch.includeHttpResponseReturnType ? `['data']` : ''}` : ''} = body ? JSON.parse(body) : {}
     err.info = data;
-    `
-        : ''
-    }err.status = ${isNdJson ? 'stream' : 'res'}.status;
+    err.status = ${isNdJson ? 'stream' : 'res'}.status;
     throw err;
   }`;
   const fetchResponseImplementation = isNdJson
-    ? `  const stream = await fetch(${fetchFnOptions})
+    ? `  const stream = await fetch(${fetchFnOptions});
+  const body = [204, 205, 304].includes(res.status) ? null : await stream.text();
   ${override.fetch.shouldThrowOnError ? throwOnErrorImplementation : ''}
   ${override.fetch.includeHttpResponseReturnType ? `return { status: stream.status, stream, headers: stream.headers } as ${override.fetch.shouldThrowOnError ? successName : responseTypeName}` : `return stream`}
   `
     : `const res = await fetch(${fetchFnOptions})
 
-    ${override.fetch.shouldThrowOnError ? throwOnErrorImplementation : ''}
+  const body = [204, 205, 304].includes(res.status) ? null : await res.text();
+  ${override.fetch.shouldThrowOnError ? throwOnErrorImplementation : ''}
 
-  const body = [204, 205, 304].includes(res.status) ? null : await res.text()
   const data: ${override.fetch.shouldThrowOnError ? successName : responseTypeName}${override.fetch.includeHttpResponseReturnType ? `['data']` : ''} = body ? JSON.parse(body) : {}
   ${override.fetch.includeHttpResponseReturnType ? `return { data, status: res.status, headers: res.headers } as ${override.fetch.shouldThrowOnError ? successName : responseTypeName}` : 'return data'}
 `;
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -146,6 +146,7 @@ ${
       const name = `${responseTypeName}${pascal(r.key)}${'suffix' in r ? r.suffix : ''}`;
       return {
         name,
+        success: response.types.success.some((s) => s.key === r.key),
         value: `export type ${name} = {
   ${isContentTypeNdJson(r.contentType) ? `stream: TypedResponse<${r.value}>` : `data: ${r.value || 'unknown'}`}
   status: ${
@@ -159,18 +160,32 @@ ${
       };
     });
 
-  const compositeName = `${responseTypeName}Composite`;
-  const compositeResponse = `${compositeName} = ${responseDataTypes.map((r) => r.name).join(' | ')}`;
+  const successName = `${responseTypeName}Success`;
+  const errorName = `${responseTypeName}Error`;
+  const hasError = responseDataTypes.some((r) => !r.success);
 
   const responseTypeImplementation = override.fetch
     .includeHttpResponseReturnType
     ? `${responseDataTypes.map((r) => r.value).join('\n\n')}
     
-export type ${compositeResponse};
-    
-export type ${responseTypeName} = ${compositeName} & {
+${`export type ${successName} = (${responseDataTypes
+  .filter((r) => r.success)
+  .map((r) => r.name)
+  .join(' | ')}) & {
   headers: Headers;
-}\n\n`
+}`};
+${
+  hasError
+    ? `export type ${errorName} = (${responseDataTypes
+        .filter((r) => !r.success)
+        .map((r) => r.name)
+        .join(' | ')}) & {
+  headers: Headers;
+}`
+    : ''
+};
+
+${override.fetch.shouldThrowOnError ? '' : `export type ${responseTypeName} = (${hasError ? `${successName} | ${errorName}` : successName})\n\n`}`
     : '';
 
   const getUrlFnProperties = props
@@ -190,7 +205,9 @@ export type ${responseTypeName} = ${compositeName} & {
     .join(',');
 
   const args = `${toObjectString(props, 'implementation')} ${isRequestOptions ? `options?: RequestInit` : ''}`;
-  const returnType = `Promise<${responseTypeName}>`;
+  const returnType = override.fetch.shouldThrowOnError
+    ? `Promise<${successName}>`
+    : `Promise<${responseTypeName}>`;
 
   const globalFetchOptions = isObject(override?.requestOptions)
     ? `${stringify(override?.requestOptions)?.slice(1, -1)?.trim()}`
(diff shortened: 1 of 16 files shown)


soartec-lab:
@AllieJonsson 
Thanks for the improvement.

In addition to this case, there is a case where error handling is done in `custom-fetch`.
There are many ways to handle error handling, such as controlling specific responses individually so that exceptions are not thrown.

https://github.com/orval-labs/orval/blob/master/samples/next-app-with-fetch/custom-fetch.ts#L52-L53

The issue I originally reported was simply to limit the type definition of response data to 200.

---

Although the scope is different from the first issue, there was also a request to add default error handling.

https://github.com/orval-labs/orval/issues/1839

I support this response because I think it will benefit them.

melloware:
Look like merge conflicts?

AllieJonsson:
> @AllieJonsson Thanks for the improvement.
> 
> In addition to this case, there is a case where error handling is done in `custom-fetch`. There are many ways to handle error handling, such as controlling specific responses individually so that exceptions are not thrown.
> 
> https://github.com/orval-labs/orval/blob/master/samples/next-app-with-fetch/custom-fetch.ts#L52-L53
> 
> The issue I originally reported was simply to limit the type definition of response data to 200.
> 
> Although the scope is different from the first issue, there was also a request to add default error handling.
> 
> #1839
> 
> I support this response because I think it will benefit them.

I completely forgot about custom fetch. 
I assume when `shouldThrowOnError` is true, we should just type it as the success response, and let the customFetch handle the error? Then maybe the name of the setting is wrong. Maybe `forceSuccessResponse` is better?
Is there a need for the `customFetch` method to know about the error type in this case?

So many questions... 😄 

```diff
export type createPetsResponse200 = {
  data: Pet;
  status: 200;
};

export type createPetsResponseDefault = {
  data: Error;
  status: Exclude<HTTPStatusCodes, 200>;
};

- export type createPetsResponseComposite =
-   | createPetsResponse200
-   | createPetsResponseDefault;

- export type createPetsResponse = createPetsResponseComposite & {
-   headers: Headers;
- };

+ export type createPetsSuccessResponse = createPetsResponse200 & {
+   headers: Headers;
+ };
+ 
+ export type createPetsErrorResponse = createPetsResponseDefault & {
+   headers: Headers;
+ };

export const createPets = async (
  createPetsBodyItem: CreatePetsBodyItem[],
  options?: RequestInit,
- ): Promise<createPetsResponse> => {
-   return customFetch<createPetsResponse>(getCreatePetsUrl(), {
+ ): Promise<createPetsSuccessResponse> => {
+   return customFetch<createPetsSuccessResponse>(getCreatePetsUrl(), {
    ...options,
    method: 'POST',
    headers: { 'Content-Type': 'application/json', ...options?.headers },
    body: JSON.stringify(createPetsBodyItem),
  });
};
```

soartec-lab:
@AllieJonsson 

This is great. Thank you for these questions as they are very valuable to me 🙌 
So I agree with the sample code you provided.

I also think the option name `forceSuccessResponse` is more appropriate than `shouldThrowOnError`.
`shouldThrowOnError` would be a good name to add an option to define the default exception throw handling as an alternative.

packages/fetch/src/index.ts
```diff
@@ -236,17 +259,26 @@ export type ${responseTypeName} = ${compositeName} & {
     ${fetchBodyOption}
   }
 `;
+  const throwOnErrorImplementation = `if (!${isNdJson ? 'stream' : 'res'}.ok) {
+    const err: globalThis.Error & {info?: ${hasError ? `${errorName}${override.fetch.includeHttpResponseReturnType ? "['data']" : ''}` : 'any'}, status?: number} = new globalThis.Error();
```
MattFromGer, line 281: Does it make sense to derive this from a generic error type? So the library consumer can for example write a wrapper that handles all API errors?

melloware:
Looks like merge conflicts need to be resolved.

pedrosimao:
Is this PR stuck? Does the author plans to solve the conflicts?

I am also facing a similar problem, where I have a `customFetch` function that throw errors.
After the 7.4+ Orval version, all the returned types of our app were broken. As I am trying to solve the issue I found this PR, but I tend to believe there could be a better solution.

I think while we are using external clients or `customFetch` we should allow this `customFetch` function itself to be the source of truth of which type is being returned, instead of forcing a type.

For instance, the code generated should be like this:
```
export const getAllPosts = async (params?: GetAllPostsParams, options?: RequestInit) => {
  
  return customFetch<getAllPostsResponse>(getGetAllPostsUrl(params),
  {      
    ...options,
    method: 'GET'
    
    
  }
);}
```

Instead of this:
```
export const getAllPosts = async (params?: GetAllPostsParams, options?: RequestInit): Promise<getAllPostsResponse> => {
  
  return customFetch<getAllPostsResponse>(getGetAllPostsUrl(params),
  {      
    ...options,
    method: 'GET'
    
    
  }
);}
```

Removing `: Promise<getAllPostsResponse>` would allow `customFetch` to return any type it wants, and not only  throwing or not throwing errors.

Would anyone agree or disagree with my view?
If there is some disagreement on this being the better design choice, we could add an option named something like `allowCustomReturnType` on the mutator options. Not sure if the same would be applicable to custom clients like swr and so, since we don't use it.

AllieJonsson:
> Is this PR stuck? Does the author plans to solve the conflicts?
> 
> I am also facing a similar problem, where I have a `customFetch` function that throw errors. After the 7.4+ Orval version, all the returned types of our app were broken. As I am trying to solve the issue I found this PR, but I tend to believe there could be a better solution.
> 
> I think while we are using external clients or `customFetch` we should allow this `customFetch` function itself to be the source of truth of which type is being returned, instead of forcing a type.
> 
> For instance, the code generated should be like this:
> 
> ```
> export const getAllPosts = async (params?: GetAllPostsParams, options?: RequestInit) => {
>   
>   return customFetch<getAllPostsResponse>(getGetAllPostsUrl(params),
>   {      
>     ...options,
>     method: 'GET'
>     
>     
>   }
> );}
> ```
> 
> Instead of this:
> 
> ```
> export const getAllPosts = async (params?: GetAllPostsParams, options?: RequestInit): Promise<getAllPostsResponse> => {
>   
>   return customFetch<getAllPostsResponse>(getGetAllPostsUrl(params),
>   {      
>     ...options,
>     method: 'GET'
>     
>     
>   }
> );}
> ```
> 
> Removing `: Promise<getAllPostsResponse>` would allow `customFetch` to return any type it wants, and not only throwing or not throwing errors.
> 
> Would anyone agree or disagree with my view? If there is some disagreement on this being the better design choice, we could add an option named something like `allowCustomReturnType` on the mutator options. Not sure if the same would be applicable to custom clients like swr and so, since we don't use it.

Sorry, I've been on vacation and in parental leave and haven't been in the office for several months, I'm looking to pick up on this pr in a couple of weeks (hopefully before the next release 😄).

Yes, it seems like removing the response type in the function should work, but I think we only would want to do this for customFetch, maybe possibly we should do it as a setting

Odonno:
I agree that if the `customFetch` handles error, we should be able to export a `Promise` of the type of the `200` data type, otherwise it should naturally default to the basic response type.

pedrosimao:
> Yes, it seems like removing the response type in the function should work, but I think we only would want to do this for customFetch, maybe possibly we should do it as a setting

Hi @AllieJonsson , thanks for replying, even during your parental leave. Really amazing that you look into this.

I am not sure about all libraries, because I haven't tested them, but if we are adding this `allowCustomReturnType` as an option, then I feel like it would probably be useful for any external client library, and not only customFetch. Most of these libraries probably do expect a generic type and return a customized version of this generic data type anyway... not forcing a type into something that will probably evolve, and which is out of our control, seems like a better design choice overall to me.

ptlthg:
I just started evaluating using Orval in my own projects, and the ability to customize the return type here would be a great help, so +1 to the conversation going on here.

I did notice that setting `output.override.fetch.includeHttpResponseReturnType` to `false` removes all the custom return type generation and just adds `: Promise<SuccessDataType>`, but I still would rather have it use the return type of my custom fetch function.

soartec-lab:
@MattFromGer @Odonno @pedrosimao @ptlthg @AllieJonsson 

Thanks everyone for the discussion.
Let me clarify the discussion. There are three confusing points here.

**1. The `shouldThrowOnError` option for allowing exception throwing to be customized within fetch.**

As the PR title suggests, this is the first thing they're trying to address.

**2. The `forceSuccessResponse` option limits the return type to success only.**

This is what I suggested in the original issue, and @AllieJonsson is also working on it since the problem still persists.

**3. Allowing generic types to be injected into the response type.**

This is what they're proposing in the latest discussion.

---

Do these three provide enough proposals for discussion?
If there are no issues, we'll focus on 1 or 2 here and move any new proposals to a separate issue.

soartec-lab (commented):
Hi @AllieJonsson If it's completed please let me know 👍

AllieJonsson:
> Hi @AllieJonsson If it's completed please let me know 👍

I think its completed. I recall there was something I wanted to change, but I cant remember what it was 🤔

soartec-lab:
And if we run into any issues, you will probably remember what you forgot, so we can add it then 👍

AllieJonsson:
> Hi, @AllieJonsson Thanks, this is a great change. Could you please add an explanation of the new options to the documentation?

This is the thing I was thinking of adding! 😅 I'll try to get it done this weekend

soartec-lab:
@AllieJonsson 
Oh year, I'm glad you remembered!

AllieJonsson:
> @AllieJonsson Oh year, I'm glad you remembered!

@soartec-lab I think I'm done now, please check if you think there are anything more to change/add. :)

PR #2173, opened 2025-06-19
fix(query): make queryKey function parameters optional

### LINKED ISSUE #1840: Generated queryKey parameters should be optional since undefined is already handled
## Problem
The generated queryKey function requires parameters to be mandatory when the type has required fields, even though:
1. The function already handles undefined cases internally with `...(params ? [params] : [])`
2. We need to pass undefined when invalidating TanStack Query cache

## Current Behavior
```typescript
  limit: string; // required
};

// Generated code - params is required despite handling undefined
export const getListPetsQueryKey = (params: ListPetsParams) => {
  return [`http://localhost:8000/pets`, ...(params ? [params] : [])] as const;
};

// Cannot invalidate cache like this
queryClient.invalidateQueries(getListPetsQueryKey());
```

## Current Workaround
We are currently using type assertion as a workaround:

```typescript
queryClient.invalidateQueries(getListPetsQueryKey({} as ListPetsParams));
```
However, this is not ideal as it bypasses TypeScript's type checking and could lead to runtime errors.

## Expected Behavior
Parameters should be optional since undefined is handled:
```typescript
export const getListPetsQueryKey = (params?: ListPetsParams) => {
  return [`http://localhost:8000/pets`, ...(params ? [params] : [])] as const;
};

// Should be able to invalidate cache
queryClient.invalidateQueries(getListPetsQueryKey());
```

orval version:  "7.4.1"
Node.js version:  "v22.13.0"
pnpm version: "9.12.2"

him0:
## Summary
- Make all query and body parameters optional in generated queryKey functions to enable cache invalidation without type assertion
- Since queryKey functions already handle undefined parameters with `...(params ? [params] : [])`, the parameters should be optional
- Path parameters remain required as they are essential for route identification

## Fixes
Closes #1840

## Problem
Currently, generated queryKey functions require parameters to be mandatory when the type has required fields, even though the function handles undefined cases internally. This prevents using TanStack Query's prefix matching for cache invalidation without type assertion workarounds.

Users had to resort to type assertion workarounds like `getListPetsQueryKey({} as ListPetsParams)` to invalidate cache, which bypasses TypeScript's type checking and could lead to runtime errors.

## Solution
Added a helper function `makeParamsOptional` that converts all parameter types to optional by replacing `param: Type` with `param?: Type` using regex replacement in the queryKey function generation logic.

## Examples

**Before:**
```typescript
export const getListPetsQueryKey = (params: ListPetsParams) => {
  return [`/pets`, ...(params ? [params] : [])] as const;
};

// Cannot invalidate cache like this - TypeScript error
queryClient.invalidateQueries(getListPetsQueryKey());

// Required workaround
queryClient.invalidateQueries(getListPetsQueryKey({} as ListPetsParams));
```

**After:**
```typescript
export const getListPetsQueryKey = (params?: ListPetsParams) => {
  return [`/pets`, ...(params ? [params] : [])] as const;
};

// Can invalidate all variants using TanStack Query's prefix matching
queryClient.invalidateQueries(getListPetsQueryKey());

// Can still invalidate specific variants
queryClient.invalidateQueries(getListPetsQueryKey({ limit: "10" }));
```

## Benefits
- Enables type-safe cache invalidation without workarounds
- Supports TanStack Query's prefix matching pattern for invalidating all related queries
- Maintains backward compatibility for existing code
- Path parameters remain required where they should be

## Test plan
- [x] Build passes
- [x] Generated queryKey functions have optional parameters for query/body params
- [x] Path parameters remain required
- [x] TanStack Query prefix matching works as expected
- [x] Existing functionality remains unchanged

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -77,7 +77,7 @@ export const generateRequestFunction = (
     explodeParameters.length > 0
       ? `const explodeParameters = ${JSON.stringify(explodeParametersNames)};
 
-    if (value instanceof Array && explodeParameters.includes(key)) {
+    if (Array.isArray(value) && explodeParameters.includes(key)) {
       value.forEach((v) => normalizedParams.append(key, v === null ? 'null' : v.toString()));
       return;
     }
--- a/packages/query/src/index.ts
+++ b/packages/query/src/index.ts
@@ -1279,6 +1279,8 @@ const generateQueryHook = async (
     ];
 
     const queryKeyFnName = camel(`get-${operationName}-queryKey`);
+    // Convert "param: Type" to "param?: Type" for queryKey functions
+    // to enable cache invalidation without type assertion
     const makeParamsOptional = (params: string) => {
       if (!params) return '';
       return params.replace(/(\w+):\s*([^,}]+)/g, '$1?: $2');
--- a/packages/query/src/index.ts
+++ b/packages/query/src/index.ts
@@ -1283,7 +1283,19 @@ const generateQueryHook = async (
     // to enable cache invalidation without type assertion
     const makeParamsOptional = (params: string) => {
       if (!params) return '';
-      return params.replace(/(\w+):\s*([^,}]+)/g, '$1?: $2');
+      // Handle parameters with default values: "param?: Type = value" -> "param: Type = value" (remove optional marker)
+      // Handle regular parameters: "param: Type" -> "param?: Type"
+      return params.replace(
+        /(\w+)(\?)?:\s*([^=,}]*?)\s*(=\s*[^,}]*)?([,}]|$)/g,
+        (match, paramName, optionalMarker, type, defaultValue, suffix) => {
+          // If parameter has a default value, don't add '?' (it's already effectively optional)
+          if (defaultValue) {
+            return `${paramName}: ${type.trim()}${defaultValue}${suffix}`;
+          }
+          // Otherwise, make it optional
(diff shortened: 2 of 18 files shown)


him0:
It looks like the regenerated files that change because of this should be committed as well, but I’m not sure which command to use.
If another commit is required, could you let me know how to proceed?

melloware (approved):

melloware:
you can just push another commit to this PR

him0:
I’ve committed the generated changes and pushed them—please take a look.

melloware:
LGTM i assigned some others to review

melloware (approved):

soartec-lab (dismissed):
LGTM!

soartec-lab:
@him0 Could you check failed test?

packages/fetch/src/index.ts
```diff
@@ -77,7 +77,7 @@ export const generateRequestFunction = (
     explodeParameters.length > 0
       ? `const explodeParameters = ${JSON.stringify(explodeParametersNames)};
 
-    if (value instanceof Array && explodeParameters.includes(key)) {
+    if (Array.isArray(value) && explodeParameters.includes(key)) {
```
him0, line 80: It started throwing errors after the files were generated, so I fixed it.

melloware (approved):

soartec-lab (approved):
Thanks！

PR #2178, opened 2025-06-23
fix(core): prevent unwanted response merging on same schema value

### LINKED ISSUE #2043: Axios: generator does not seem to support multiple response types #1588
## Description
If you use an openAPI schema with multiple responses like this:

```json
  "/api/user": {
    "get": {
      "operationId": "get_user",
      "summary": "Get User",
      "parameters": [],
      "responses": {
        "200": {
          "description": "OK",
          "content": {
            "application/json": {
              "schema": { "$ref": "#/components/schemas/UserResponse" }
            }
          }
        },
        "400": {
          "description": "Bad Request",
          "content": {
            "application/json": {
              "schema": { "$ref": "#/components/schemas/ApiErrorResponse" }
            }
          }
        }
```

The generated fetch will only have the 200 response type (in this example UserResponse). It does not seem to generate a union type for the method return type.

## Expected Solution

There was a similar issue with fetch that was solved
- [the issue](https://github.com/orval-labs/orval/issues/1588)
- [the fix](https://github.com/orval-labs/orval/pull/1776)

## Current Workaround

I provide my own response type, but I manually typed it (instead of using openapi spec)

```ts
export const customClient = <T>(
  config: AxiosRequestConfig,
  maybeOptions?: AxiosRequestConfig,
): Promise<ReponseType<T>> => {

          // ... implementation
}

type ReponseType<TRes> = AxiosResponse<
  | TRes
  | {
      status: 400;
      statusText: string;
      data: {
        message: {
          name: string;
          message: string;
          statusCode: 400;
          errorLevel: string;
          context: object;
        };
      };
    }
>;


```

### LINKED ISSUE #2177: Multiple response status codes with same schema only generate one type in composite
## Bug description

  When an OpenAPI operation defines multiple response status codes that return the **same schema**, Orval only generates TypeScript types for one of them instead of creating a proper union type that includes all status codes.

  ## Expected behavior

  For an endpoint `POST /subscriptions` that can return:
  - **200** with `CreateSubscriptionResponseDto` (existing subscription)
  - **201** with `CreateSubscriptionResponseDto` (new subscription created)
  - **403** with `BaseError`
  - **404** with `BaseError`

  Orval should generate:

  ```typescript
  export type insuranceConfigurationControllerCreateSubscriptionResponseComposite =
    | insuranceConfigurationControllerCreateSubscriptionResponse200
    | insuranceConfigurationControllerCreateSubscriptionResponse201
    | insuranceConfigurationControllerCreateSubscriptionResponse403
    | insuranceConfigurationControllerCreateSubscriptionResponse404;
```

  ## Actual behavior

  With the fetch client, Orval only generates types for one status code per schema, missing others that share the same schema:

  ```typescript
  export type insuranceConfigurationControllerCreateSubscriptionResponseComposite =
    | insuranceConfigurationControllerCreateSubscriptionResponse201  // Missing 200
    | insuranceConfigurationControllerCreateSubscriptionResponse404; // Missing 403
```

  ## Steps to reproduce

  1. Create an OpenAPI spec with multiple status codes sharing the same schema:

  ```yaml
  /subscriptions:
    post:
      responses:
        '200':
          description: 'Existing subscription returned'
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/CreateSubscriptionResponseDto'
        '201':
          description: 'New subscription created'
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/CreateSubscriptionResponseDto'  # Same schema as 200
        '403':
          description: 'Forbidden'
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/BaseError'
        '404':
          description: 'Not found'
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/BaseError'  # Same schema as 403
```

  2. Generate types with Orval using fetch client
  3. Observe missing status codes in the composite type

  ## Environment

  - Orval version: 7.9.0
  - Client: fetch
  - Mode: tags-split

 ## Configuration

  ```typescript
  const config: ConfigExternal = {
    output: {
      client: "fetch", 
      mode: "tags-split",
      target: "./gen/client.ts",
      override: {
        mutator: {
          path: "./fetch-client.ts",
          name: "fetchWrapper",
        },
      },
    },
    input: {
      target: "./openapi.yaml",
    },
  };
```

##  Impact

  This breaks TypeScript type safety and business logic handling when using the fetch client:
  - Cannot distinguish between 200 vs 201: Unable to handle "existing subscription" vs "new subscription" scenarios differently
  - Cannot distinguish between error types: Unable to handle 403 vs 404 errors differently in the application
  - Runtime vs compile-time mismatch: Generated types don't reflect actual API behavior

  ## Workaround

  Currently considering duplicating schema definitions to force Orval to generate separate types, but this creates OpenAPI spec bloat.

  Note: This issue may be specific to the fetch client implementation, as it might handle response type generation differently than axios.

Ben1306:
## Status

WIP

### Problem

Fixes #2177
Fix #2043
When an endpoint returns multiple status codes with the same schema (e.g., 200 and 201 both returning `CreateSubscriptionResponseDto`), Orval was merging them into a single type, causing the loss of status code-specific type information.

### Solution
Updated the `getResponse` function in `packages/core/src/getters/response.ts` to use a custom `uniqueKey` function that includes both the status code and content type: `${type.key}-${type.value}`.

This ensures that each status code gets its own distinct type, even when they share the same schema.

### Changes
- Modified `getResponse` function to pass custom uniqueKey function to `getResReqTypes`
- Added comprehensive unit tests to verify the behavior
- Tests cover the specific scenario described in the issue

### Testing
Added unit tests that verify:
- Multiple status codes with same schema generate separate types
- Each status code (200, 201, 403, 404) gets its own type entry
- Both success and error types are properly separated

### Example
Before: Only one type generated for 2XX/2YY or 4XX/4YY responses with same schema
After: Separate types generated for each status code, preserving all response variations

### DIFF
--- a/packages/core/src/getters/response.test.ts
+++ b/packages/core/src/getters/response.test.ts
@@ -1,0 +1,118 @@
+import { describe, it, expect } from 'vitest';
+import { ResponsesObject } from 'openapi3-ts/oas30';
+import { getResponse } from './response';
+
+const context = {
+  output: {
+    override: {
+      formData: { arrayHandling: 'serialize', disabled: false },
+      enumGenerationType: 'const',
+    },
+  },
+  specKey: 'spec',
+  target: 'spec',
+  workspace: '',
+  specs: {
+    spec: {
+      components: { schemas: {} },
+    },
+  },
+};
+
+describe('getResponse', () => {
+  describe('multiple status codes with same schema', () => {
+    it('should generate separate types for each status code when using custom uniqueKey function', () => {
+      const responses: ResponsesObject = {
+        '200': {
+          description: 'Existing subscription',
+          content: {
+            'application/json': {
+              schema: {
+                $ref: '#/components/schemas/CreateSubscriptionResponseDto',
+              },
+            },
+          },
+        },
+        '201': {
+          description: 'New subscription created',
+          content: {
+            'application/json': {
+              schema: {
+                $ref: '#/components/schemas/CreateSubscriptionResponseDto',
+              },
+            },
+          },
+        },
+        '403': {
+          description: 'Forbidden',
+          content: {
+            'application/json': {
+              schema: {
+                $ref: '#/components/schemas/BaseError',
+              },
+            },
+          },
+        },
+        '404': {
+          description: 'Not found',
+          content: {
+            'application/json': {
+              schema: {
+                $ref: '#/components/schemas/BaseError',
+              },
+            },
+          },
+        },
+      };
+
+      const result = getResponse({
+        responses,
+        operationName: 'configurationControllerCreateSubscription',
+        context: context as any,
+      });
+
+      expect(result.types.success).toHaveLength(2);
+      expect(result.types.errors).toHaveLength(2);
+
+      const successKeys = result.types.success.map((type) => type.key);
+      const errorKeys = result.types.errors.map((type) => type.key);
+
+      expect(successKeys).toContain('200');
+      expect(successKeys).toContain('201');
+      expect(errorKeys).toContain('403');
+      expect(errorKeys).toContain('404');
+
+      expect(result.definition.success).toContain(
+        'CreateSubscriptionResponseDto',
+      );
+
+      expect(result.definition.errors).toContain('BaseError');
+    });
+
+    it('should handle empty responses object', () => {
+      const result = getResponse({
+        responses: {},
+        operationName: 'testOperation',
+        context: context as any,
+      });
+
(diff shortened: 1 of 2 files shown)


melloware:
looks like tests are failing

melloware (approved):

melloware:
I think this might also fix #2043 and #2177

PR #2199, opened 2025-07-08
feat: generate `zod.literal` when schema has `const` value

lukemorales:
## Status

**READY**

## Description

OpenAPI Schema 3.1 supports literal values when a `const` property exists in a string schema. This is important when we have discriminated unions based on a literal value. This PR adds support for generating zod schemas with literal values using `z.literal()` when a string schema has `const` property.

## Todos

- [x] Tests
- [ ] Documentation

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout feat/add-support-for-literal-values-in-zod
> grunt jasmine
> yarn
> yarn build
> cd tests
> yarn
> yarn generate:zod
```

1. check generated `typed-arrays-tuples-v3-1.ts` in generated/zod/*
2. check that a new property `example_string_const` has a value of `zod.literal("this_is_a_string_const")` instead of `zod.string()`

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -68,7 +68,7 @@ const possibleSchemaTypes = [
   'array',
 ];
 
-const resolveZodType = (schema: SchemaObject) => {
+const resolveZodType = (schema: SchemaObject | SchemaObject31) => {
   const schemaTypeValue = schema.type;
   const type = Array.isArray(schemaTypeValue)
     ? schemaTypeValue.find((t) => possibleSchemaTypes.includes(t))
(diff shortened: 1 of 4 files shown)


packages/zod/src/index.ts
```diff
@@ -156,7 +156,7 @@ export const generateZodValidationSchemaDefinition = (
   const type = resolveZodType(schema);
   const required = rules?.required ?? false;
   const nullable =
-    schema.nullable ??
+    ('nullable' in schema && schema.nullable) ||
```
lukemorales, line 159: this change was needed because `schema.nullable` does not exist in `SchemaObject31`

Issue #2204, opened 2025-07-11
Passing a reviver-function to JSON.parse


chcederquist:
I would like to add a reviver-function to the fetch client for dates. Currently I have to opt out of using the default fetch implementation, and pass my own mutator.
To avoid writing my own mutator(basically copy-pasting the generated one), it would be useful to just be able to pass a reviver-function to [JSON.parse](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#reviver) in the [fetch client code](https://github.com/orval-labs/orval/blob/981fdf6de15725c4f6cc5312c7a98fcb1f09f187/packages/fetch/src/index.ts#L251).

I'd be happy to a create a PR for this, if you agree with the suggestion.

melloware:
@chcederquist please create a PR and @soartec-lab can review as he is the author of the Fetch Client.

PR #2207, opened 2025-07-12
feat: added jsonReviver param to fetch config

chcederquist:
## Status

READY

## Description

Implements #2204

## Related PRs
None

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout <branch>
> cd tests
> yarn generate:fetch
```

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -1013,6 +1013,39 @@ Type: `String` or `Object`
 
 Allows you to provide a <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#reviver" target="_blank">reviver</a> function to the fetch client when it parses JSON. It is recommended to use this to revive dates when setting <a href="#usedates">useDates</a> to `true`
 
+Example:
+
+```js
+module.exports = {
+  petstore: {
+    output: {
+      override: {
+        fetch: {
+          jsonReviver: {
+            path: './api/mutator/custom-reviver.ts',
+            name: 'customReviver',
+            // default: true
+          },
+        },
+      },
+    },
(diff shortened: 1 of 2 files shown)


chcederquist:
Should I squash these commits into a single one?

melloware:
you don't have to we usually squash and merge.

melloware:
also i am wondering if in the docs you can add a section in Fetch Client about this feature with example code.  i think a lot of poeple will want to use this.

On this page: https://orval.dev/guides/fetch-client

chcederquist:
> also i am wondering if in the docs you can add a section in Fetch Client about this feature with example code. i think a lot of poeple will want to use this.
> 
> On this page: https://orval.dev/guides/fetch-client

Done. I've updated the date regex to be the same as isoDateFormat-regex in the useDates example for consistency.

soartec-lab:
@chcederquist @melloware 
Thanks for the great suggestions.
I would like to see other users' reactions to this improvement, because this approach can also be solved with the existing `custom-fetch`, and this is just an added convenience. I would like more information on whether this is an edge case or a desired solution.

melloware:
@soartec-lab you are right this could be solved with a Custom Fetch but 99% of people like things that work out of the box and converting Dates is the most common use case over the years people ask about.

chcederquist:
I agree that it is totally possible with custom fetch, and I actually have done that for now in my project using this library. But my worry as an implementer of it was that I would have to re-visit the custom fetch implementation, and have to keep an eye on the changelog if something changed, that would necessitate a change to my custom fetch.

soartec-lab:
I see, please give us some time to review this for merging 👍

soartec-lab (approved):
Thanks for the great improvement.

Issue #2203, opened 2025-07-11
Custom parameter serialization for fetch client


chcederquist:
The current documentation says, that parameter serialization is only available for Axios-clients: https://orval.dev/reference/configuration/output#paramsserializer

However, this is also relevant for non-Axios clients. Currently I'm using [useDates: true](https://orval.dev/reference/configuration/output#usedates), and reviving dates using the [mutator](https://orval.dev/reference/configuration/output#mutator) option.

The only missing part is serialising dates. Currently, the implementation for fetch clients [stringifies all parameters using toString()](https://github.com/orval-labs/orval/blob/981fdf6de15725c4f6cc5312c7a98fcb1f09f187/packages/fetch/src/index.ts#L81), but for dates, they should be serialized according to the ISO 8601-standard, meaning it should use [toISOString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) instead.

Currently I've written my own post-generation hook that manually replaces it, but I think it would be better to handle it natively, or at least provide a way for us to write our own parameter serialization for query params.

I'd be happy to make a PR once a good solution has agreed upon :-)

melloware:
it seems like param serializer is supported by Fetch client see this ticket: https://github.com/orval-labs/orval/issues/1751   A lot of people use the `qs` library with this feature.

melloware:
But feel free to submit a PR.

chcederquist:
> it seems like param serializer is supported by Fetch client see this ticket: [#1751](https://github.com/orval-labs/orval/issues/1751) A lot of people use the `qs` library with this feature.

I believe the paramsSerializer PR for that issue was abandoned, and the issue was instead solved by using explode. However, exploding parameters is only relevant for general objects and arrays. From what I can tell, there's currently no way to modify serialization of dates.

chcederquist:
> But feel free to submit a PR.

Do you think it's relevant to do a specific config property in overrides: fetch, setting whether dates should be converted to ISO strings or just stringified, or should there be a more general parameter serialization?

melloware:
@soartec-lab can you weigh in on this?

soartec-lab:
Sure!

@chcederquist 
Ah, I see. So when you want to parse a Date, you need a solution that has nothing to do with `explode`.
Could #2207 be one solution to this?

chcederquist:
[...]

PR #2213, opened 2025-07-18
feat: params of format date-time are now stringified using toISOString when useDates === true

chcederquist:
## Status

READY

## Description

Implements #2203 

## Related PRs

None

## Todos

- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout master
> yarn build
> cd test
> yarn generate:fetch
```

### DIFF
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -73,13 +73,22 @@ export const generateRequestFunction = (
 
     return schema.name;
   });
+  const hasDateParams =
+    context.output.override.useDates &&
+    parameters.some(
+      (p) =>
+        'schema' in p &&
+        p.schema &&
+        'format' in p.schema &&
+        p.schema.format === 'date-time',
+    );
 
   const explodeArrayImplementation =
     explodeParameters.length > 0
       ? `const explodeParameters = ${JSON.stringify(explodeParametersNames)};
 
     if (Array.isArray(value) && explodeParameters.includes(key)) {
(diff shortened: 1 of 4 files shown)


packages/fetch/src/index.ts
```diff
@@ -73,13 +73,22 @@ export const generateRequestFunction = (
...
     return schema.name;
   });
+  const hasDateParams =
+    context.output.override.useDates &&
+    parameters.some(
+      (p) =>
+        'schema' in p &&
+        p.schema &&
+        'format' in p.schema &&
+        p.schema.format === 'date-time',
```
soartec-lab, line 83: Is it possible to avoid writing it so conservatively and just access the object?

```suggestion
    parameters.some((p) => p.schema && p.schema.format === 'date-time')
```

chcederquist: Not without TS-warnings. Both p(ParameterObject | ReferenceObject) and schema(ReferenceObject | SchemaObject) are union types, so I need to ensure that 'schema' is in p (infers that p is a ParameterObject) and that 'format' is in p.schema (infers that p.schema is SchemaObject). And since p.schema is possibly undefined, I must check that as well.

chcederquist:
@soartec-lab 
I've added a docs note and addressed the code comment

packages/fetch/src/index.ts
soartec-lab: Okay, that sounds reasonable 👍

soartec-lab (approved):
LGTM!

PR #2219, opened 2025-07-20
fix: use zod.union for oneOf and anyOf

dalen:
## Status

**READY**

## Description

Instead of chaining `.or` use `zod.union`.

This creates much more performant types, and avoids TypeScript error TS2589: Type instantiation is excessively deep and possibly infinite.

The previous code generated types like

```
zod.literal(0).or(zod.literal(1)).or(zod.literal(2))
```

Which translates into a typescript type:

```
zod.ZodUnion<[zod.ZodUnion<[zod.ZodLiteral<0>, zod.ZodLiteral<1>]>, zod.ZodLiteral<2>]>
```

With this change we instead generate this:

```
zod.union([zod.literal(0),zod.literal(1),zod.literal(2)])
```

Which produces a simpler type:

```
zod.ZodUnion<readonly [zod.ZodLiteral<0>, zod.ZodLiteral<1>, zod.ZodLiteral<2>]>
```


melloware:
This is great but looks like test failure?

dalen:
hmm, `yarn test:ci`  was working for me locally. But I'll take a look

dalen:
I think I see the issue.

Can't use `zod.union` on enums with just one item. I'll make a fix for that.

dalen:
I made a fix for that, and added a test for that case.

Issue #1723, opened 2024-11-29
Global useQuery Configuration does not apply to POST requests


vitminakov:
## What are the steps to reproduce this issue?

1. In the orval.config.js, configure the operations as follows:
```
operations: {
  query: {
    useQuery: true,
  },
};
```
2. Use this configuration on a POST request in the OpenAPI schema.
3. Generate the API using orval

## What happens?

Even though `useQuery: true` is configured globally, the `POST` request generates a `useMutation` instead of a `useQuery`.

## What were you expecting to happen?

I expected the `POST` request to generate `useQuery` based on the global configuration.

## Any logs, error output, etc?

…

## Any other comments?

When specifying the `operationId` explicitly in the `orval.config.js`:
```
operations: {
  getSideMenu: {
    query: {
      useQuery: true,
    },
  },
};
```

I have found this https://github.com/orval-labs/orval/issues/515
Is it related?

## What versions are you using?

```
  System:
    OS: macOS 12.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 107.91 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  npmPackages:
    @tanstack/react-query: 5.61.5 => 5.61.5 
    axios: 1.7.8 => 1.7.8 
    orval: 7.3.0 => 7.3.0 
    react: 18.3.1 => 18.3.1 
```

melloware:
See this ticket: https://github.com/orval-labs/orval/issues/1162

did you try..

```js
query: {
        useQuery: true,
        useMutation: false,
      }
```

vitminakov:
there is my config
```
/* eslint-disable no-undef */
function header(info) {
  return [
    'eslint-disable',
    'Generated by orval 🍺',
    'Do not edit manually.',
    ...(info.title ? [info.title] : []),
    ...(info.description ? [info.description] : []),
    ...(info.version ? [`OpenAPI spec version: ${info.version}`] : [])
  ];
}

module.exports = {
  bff: {
    output: {
      mode: 'single',
      target: './src/api/bff/index.ts',
      schemas: './src/api/bff/model',
      client: 'react-query',
      clean: true,
      prettier: true,
      headers: true,
      override: {
        header,
        mutator: {
          path: './src/shared/config/axios/instance.ts',
          name: 'customInstance'
        },
        query: {
          useQuery: true,
          useMutation: false,
          usePrefetch: true
        }
      }
    },
    input: {
      target: './merged.yaml'
    }
[...]

PR #2230, opened 2025-07-25
Fix for #2227 - correctly using the operations override

### LINKED ISSUE #2227: MAJOR operations option regression after version 7.5 even in latest (7.10)
We tried upgrading from 7.5 into 7.10 but experience major regression in the operations override option. 

- 7.10 not acknowledging the operation override
- 7.9 not acknowledging the operation override
- 7.8 acknowledges the operation override, BUT is generating during compilation **duplicate identifiers** for the operations in override
- 7.7 acknowledges the operation override, BUT is generating during compilation **duplicate identifiers** for the operations in override
- 7.6 acknowledges the operation override, BUT is generating during compilation **duplicate identifiers** for the operations in override
- 7.5 the override feature works perfectly, and no duplicate identifiers!

Since version 7.9 and 7.10 we experience a bug where the operations override itself is not working.

We have in orval config (only two operations):

```
      override: {
         ...
        operations: {
          reprint: {
            query: {
              useMutation: true,
            },
          },
          getAddressSuggestions: {
            query: {
              useQuery: true,
            },
          },
        },
```

But this operation id is not changed in a query but stays a mutation. 
The `getAddressSuggestions` is only one time in our open api spec.

```
export const useReprintMutationOptions = <TError = ErrorType<ErrorBody | void>, TContext = unknown>(options?: {
  mutation?: UseMutationOptions<
```

```
export const useGetAddressSuggestions = <TError = ErrorType<ErrorBody | void>, TContext = unknown>(
  options?: {
    mutation?: UseMutationOptions<
```

It would be nice when https://github.com/orval-labs/orval/pull/2079 had a test to validate this regression/solution? Then we had a real solution instead :)

## 7.8 the override works ...
After this version it stays the original operation instead of the wanted change
So in 7.8 i can see clearly the POST is changed in hook with QUERY and not MUTATION. But with another side effect namely duplicate identifiers.

Image

**but we are getting for stage `Transpiling & type-checking TypeScript` now (also in 7.7 version):**

```
generated/orval/index.ts(1986,17): error TS2300: Duplicate identifier 'useGetAddressSuggestions'.
generated/orval/index.ts(1996,17): error TS2300: Duplicate identifier 'useGetAddressSuggestions'.
generated/orval/index.ts(2006,17): error TS2300: Duplicate identifier 'useGetAddressSuggestions'.
generated/orval/index.ts(2011,17): error TS2300: Duplicate identifier 'useGetAddressSuggestions'.
generated/orval/index.ts(2057,18): error TS2300: Duplicate identifier 'useGetAddressSuggestions'.

generated/orval/index.ts(2434,17): error TS2300: Duplicate identifier 'useReprint'.
generated/orval/index.ts(2446,17): error TS2300: Duplicate identifier 'useReprint'.
generated/orval/index.ts(2458,17): error TS2300: Duplicate identifier 'useReprint'.
generated/orval/index.ts(2465,17): error TS2300: Duplicate identifier 'useReprint'.
generated/orval/index.ts(2513,18): error TS2300: Duplicate identifier 'useReprint'.
```

It all seems to be related to the override option.

maapteh:
Fix #2227 

This is a fix for #2227 where compilation changes the intentional code. By adding the correct code it should not be mangled anymore and stay exactly as intended.

I noticed i had to add to main package.json a resolution that ansi-styles stays on v4 since v5 is ESM. Not sure if thats only for me?

Also the naming `override` in the Orval config, and the naming override in `GeneratorVerbOptions` in the @orval/query library is **very confusing** since its not that override :) that actual override is in the second argument, which are the `GeneratorOptions`, which is the actual operational override you set in your config :exploding_head:.

_Will try this version in our own codebase first (_pnpm way it is correct for my apps now_), maybe we can also get an alpha since not easy to rewire with Bazel in our CI and see your compiled package?_

Thanks for this wonderfull library btw, i will be glad to think/work how we can improve the testability of at least the query package :)

### DIFF
--- a/packages/query/src/index.ts
+++ b/packages/query/src/index.ts
@@ -1187,7 +1187,6 @@ const generateQueryHook = async (
     hasOperationQueryOption;
 
   let isMutation = override.query.useMutation && verb !== Verbs.GET;
-  
 
   if (operationQueryOptions?.useMutation !== undefined) {
     isMutation = operationQueryOptions.useMutation;
--- a/packages/query/src/index.ts
+++ b/packages/query/src/index.ts
@@ -1168,23 +1168,26 @@ const generateQueryHook = async (
   let implementation = '';
   let mutators = undefined;
 
+  // Allows operationQueryOptions (which is the Orval config override for the operationId)
+  // to override non-GET verbs
+  const hasOperationQueryOption = !!(
+    operationQueryOptions &&
+    (operationQueryOptions?.useQuery ||
+      operationQueryOptions?.useSuspenseQuery ||
+      operationQueryOptions?.useInfinite ||
+      operationQueryOptions?.useSuspenseInfiniteQuery)
+  );
+
   let isQuery =
-    Verbs.GET === verb &&
-    (override.query.useQuery ||
-      override.query.useSuspenseQuery ||
-      override.query.useInfinite ||
-      override.query.useSuspenseInfiniteQuery);
-
-  // Allows operationQueryOptions to override non-GET verbs
-  const hasOperationQueryOption =
-    operationQueryOptions?.useQuery ||
-    operationQueryOptions?.useSuspenseQuery ||
-    operationQueryOptions?.useInfinite ||
-    operationQueryOptions?.useSuspenseInfiniteQuery;
-
-  isQuery = isQuery || hasOperationQueryOption;
+    (Verbs.GET === verb &&
+      (override.query.useQuery ||
+        override.query.useSuspenseQuery ||
+        override.query.useInfinite ||
+        override.query.useSuspenseInfiniteQuery)) ||
+    hasOperationQueryOption;
 
   let isMutation = override.query.useMutation && verb !== Verbs.GET;
+  
 
   if (operationQueryOptions?.useMutation !== undefined) {
     isMutation = operationQueryOptions.useMutation;


packages/query/src/index.ts
```diff
@@ -1168,23 +1168,26 @@ const generateQueryHook = async (
...
+      operationQueryOptions?.useSuspenseInfiniteQuery)
+  );
 
-  isQuery = isQuery || hasOperationQueryOption;
+  let isQuery =
+    (Verbs.GET === verb &&
+      (override.query.useQuery ||
+        override.query.useSuspenseQuery ||
+        override.query.useInfinite ||
+        override.query.useSuspenseInfiniteQuery)) ||
```
maapteh, line 1186: ps im not even sure about this, the override.query.useQuery in all my cases even for gets and nothing/something in config overrides is always true :) So i thought it would just be `Verbs.GET === verb || hasOperationQueryOption` but im afraid to leave this part out since its introduced ...

But i know for 100% my config override is in `operationQueryOptions` :)

melloware:
Build running now

melloware:
Does this also fix https://github.com/orval-labs/orval/issues/1723 ?

maapteh:
> Does this also fix #1723 ?

Yes, that was the exact same bug we had. A POST which we wanted the useQuery for :) 
I looked before adressing my bug, but thats the exact same behaviour indeed and we needed the hook, so a useMutation: false was also like mentioned there not working :( 
I do not know about the other added lines in 7.10 but this fix is for that bug, which came because of compilation since its intention was ok.

maapteh:
```
 output: {
      client: 'react-query',
      ...
      override: {
        ...
        operations: {
          ...
          getAddressSuggestions: {
            query: {
              useQuery: true,
            },
          },
        },
```

Now our POST from open api spec for operationId `getAddressSuggestions` is a useQuery instead of useMutation.

image

melloware (approved):

maapteh:
Thanks for releasing, its fixed!

PR #2244, opened 2025-07-27
fix: Angular client props nullability

### LINKED ISSUE #2092: Angular schema nullability definitions incorrectly done

snebjorn:
## Status

**READY**

## Description

Fix #2092 

I'm not exactly sure what the intended goal is here.

Given
```yml
parameters:
  - name: requirednullableString
    in: query
    required: true
    schema:
      type: string
      nullable: true
```
What is the expected outcome?
With this implementation the parameter becomes a required string. Is that expected?

## Related PRs

Taking over the work done in #2103


snebjorn:
seems all that was missing was formatting

jonimake:
Hey, thank you for taking care of the final issues. I was really busy with work and on my summer holiday for the past one and half months or so so I couldn't really look into this at all.

Anyway, really glad to see it got in. Really helps us at work.

PR #2269, opened 2025-08-02
docs: needs globally installed prettier

snebjorn:
## Status

**READY**

## Description

As discussed in #2246 I've updated the docs to say that a global prettier is needed


soartec-lab:
Thank you. There may be a v7.11.2 release, but this PR will not be included in that release, so I will hold off on merging it.

PR #2272, opened 2025-08-02
feat(fetch): changed default value of `override.fetch.explode` from `false` to `true` to clarify its meaning

soartec-lab:
## Status

**READY**

## Description

Starting with #2143, the fetch client has changed so that array query parameters without the explode option behave the same as when explode is true. 
The previous behavior was the same as when explode is false, so we've added the `override.fetch.explode` option to maintain compatibility.

The default for `override.fetch.explode` is `false`.
When `override.fetch.explode` is `false`, array query parameters without explode specified behave the same as when explode is `true`.
It would be confusing to have the same behavior as if explode was `true` even though it was set to `false`, so I changed the default value from `false` to `true` to change the meaning.

### Before

- `override.fetch.explode` defaults to `false`
- When `override.fetch.explode` is `false`, array query parameters without the explode setting behave the same as when explode is set to `true`.

### After

- `override.fetch.explode` defaults to `true`
- When `override.fetch.explode` is `true`, array query parameters without the explode setting behave the same as when explode is set to `true`.

## Related PRs

- #2143

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

none

### DIFF
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -1001,11 +1001,11 @@ If you want to return a defined return type instead of an automatically generate
 ##### explode
 
 Type: `Boolean`.
-Default: `false`
+Default: `true`
 
 By default, the `fetch` client follows the OpenAPI specification for query parameter explode behavior. This means that query parameters will be exploded unless explicitly set to `false` in the OpenAPI schema.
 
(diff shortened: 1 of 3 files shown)


melloware (approved):

soartec-lab:
@yamish123
I want to share this with you so I mention it.

soartec-lab:
@melloware 
I've rebased it so I'll request a review again.

melloware (approved):

Issue #347, opened 2022-04-02
Export MSW controllers to a separate index file.


stijnvanhulle:
Hi, At this moment we are using Orval 🍻 to generate our own API client which is already working really great. We can import our Axios controllers and React-query controllers from the index.ts(api-client/lib/gen/clients/axios/index.ts) file. But we also need the MSW controllers to mock our API calls.

At this moment we have a couple of them but those can become really fast 100 imports that we need to specify with the correct path. 

```
import { petControllerMSW } from 'api-client/lib/gen/clients/axios/pet-controller/pet-controller.msw'
import { jupilerControllerMSW } from 'api-client/lib/gen/clients/axios/jupiler-controller/jupiler-controller.msw'
```

It would be nice to have something like this: just one import that contains all the different MSW controllers.
```
import { petControllerMSW, jupilerControllerMSW } from 'api-client/lib/gen/clients/msw'
```

I have tried playing around with **workspace** and the different **modes** but all modes have the same thing: MSW is excluded in the exports. 

### Idea

**src/core/writers/specs.ts**
I tried something out by adding a new file **msw.ts** that contains all the controllers but not sure if this would be the best way to solve this issue, any clues on what can be possible here?  Maybe a new client just for the MSW?

```
  if (output.workspace) {
    const workspacePath = output.workspace;
    let imports = implementationPaths
      .filter((path) => !path.endsWith('.msw.ts'))
      .map((path) =>
        relativeSafe(workspacePath, getFileInfo(path).pathWithoutExtension),
      );

    if (output.schemas) {
      imports.push(
        relativeSafe(workspacePath, getFileInfo(output.schemas).dirname),
      );
    }

    const indexFile = join(workspacePath, '/index.ts');

    if (await pathExists(indexFile)) {
      const data = await readFile(indexFile, 'utf8');
      const importsNotDeclared = imports.filter((imp) => !data.includes(imp));
      await appendFile(
        indexFile,
        uniq(importsNotDeclared)
          .map((imp) => `export * from '${imp}';`)
          .join('\n') + '\n',
      );
    } else {
      await outputFile(
        indexFile,
        uniq(imports)
          .map((imp) => `export * from '${imp}';`)
          .join('\n') + '\n',
      );
    }

    // indexFile for MSW => this creates a msw file with all exports for the MSW controllers
    const indexMSWFile = join(workspacePath, '/msw.ts');

    let importsMSW = implementationPaths
[...]

PR #2274, opened 2025-08-02
feat: indexMockFiles now only exports mock handlers

### LINKED ISSUE #2247: barrel (indexFiles) support for target and mock, not just schemas
Today there's barrel support for the generated schemas (aka models)
https://orval.dev/reference/configuration/output#indexfiles

It would be handy if this support could include targets (aka endpoints) and mocks.
It's just nicer to import endpoints from a barrel :)

The need for a mock barrel is so we easily can import the mock handlers needed to setup [msw](https://mswjs.io/).

```ts
// src/mocks/node.ts
import { setupServer } from 'msw/node'
import { handlers } from './handlers.js'
 
export const server = setupServer(...handlers)
```
https://mswjs.io/docs/quick-start#3-process-level-integration

snebjorn:
## Status

**READY**

## Description

Closes #2247, #347

This is a breaking change!

`indexMockFiles` now only exports the mock handlers to be used by msw's `setupWorker`/`setupWorker`

example:
```ts
import { setupServer } from 'msw/node';
import * as mocks from './endpoints/index.msw';

const handlers = Object.entries(mocks).flatMap(([, getMock]) => getMock());
const server = setupServer(...handlers);

export { server };
```

The angular-app sample have been setup to utilize this.

### DIFF
--- a/packages/core/src/writers/split-tags-mode.ts
+++ b/packages/core/src/writers/split-tags-mode.ts
@@ -3,6 +3,7 @@ import { generateModelsInline, generateMutatorImports } from '../generators';
 import { OutputClient, WriteModeProps } from '../types';
 import {
   camel,
+  pascal,
   getFileInfo,
   isFunction,
   isSyntheticDefaultImportsAllow,
@@ -202,7 +203,10 @@ export const writeSplitTagsMode = async ({
               tag,
               tag + '.' + getMockFileExtensionByTypeName(output.mock!),
             );
-            fs.appendFile(indexFilePath, `export * from '${localMockPath}'\n`);
+            fs.appendFile(
+              indexFilePath,
+              `export { get${pascal(tag)}Mock } from '${localMockPath}'\n`,
+            );
           }
         }
 
--- a/docs/src/pages/guides/msw.md
+++ b/docs/src/pages/guides/msw.md
@@ -156,3 +156,16 @@ import { setupServer } from 'msw/node';
 const server = setupServer();
 server.use(...getPetsMock());
 ```
+
+You can also turn on [`indexMockFiles`](../reference/configuration/output#indexmockfiles) which will allow you dynamically import all mock handlers.
+
+```ts
+// node.ts
+import * as mocks from './endpoints/index.msw';
+import { setupServer } from 'msw/node';
+
+const handlers = Object.entries(mocks).flatMap(([, getMock]) => getMock());
+const server = setupServer(...handlers);
+
+export { server };
+```
--- a/docs/src/pages/reference/configuration/output.md
+++ b/docs/src/pages/reference/configuration/output.md
@@ -584,7 +584,7 @@ Type: `Boolean`
 
 Default Value: `false`.
 
-When `true`, adds a `index.msw.ts` file which exports all mock functions.
+When `true`, adds a `index.msw.ts` file which exports arrays with all mock functions.
 This is only valid when `mode` is `tags-split`.
 
 Example:
@@ -600,6 +600,13 @@ module.exports = {
     },
   },
 };
+```
+
+```ts
+// index.msw.ts
+export { getPetsMock } from './pets/pets.msw';
+export { getStoresMock } from './stores/stores.msw';
+// etc...
 ```
 
 ## docs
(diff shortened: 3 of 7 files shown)


packages/core/src/writers/split-tags-mode.ts
```diff
@@ -202,7 +203,10 @@ export const writeSplitTagsMode = async ({
               tag,
               tag + '.' + getMockFileExtensionByTypeName(output.mock!),
             );
-            fs.appendFile(indexFilePath, `export * from '${localMockPath}'\n`);
+            fs.appendFile(
+              indexFilePath,
+              `export { get${pascal(tag)}Mock } from '${localMockPath}'\n`,
```
snebjorn, line 208: this is rather fragile. Open to suggestions :)

soartec-lab:
@snebjorn 
Perhaps this solution will also solve the #347 ?

packages/core/src/writers/split-tags-mode.ts
soartec-lab, line 208: @snebjorn 
So by exporting only one function that aggregates the handler functions that generate mocks, rather than exporting all the functions as we have done up until now, you can easily create all the mock definitions iteratively on the consuming application side?

snebjorn: Short answer: yes :)

The idea is to only export handler functions (much like #347), and I found the aggregate handler function (`get{tag}Mock`) which made it a lot easier to do.
The goal is to be able to iterate over mock handlers without having to change that part of the code when the OpenApi spec changes.

Example:
Currently you have to do this, and update it everytime the OpenApi spec changes
```ts
import { getPetsMock } from './endpoints/pets/pets.msw';
import { getFooMock } from './endpoints/foo/foo.msw';
import { getBarMock } from './endpoints/bar/bar.msw';
// ...etc

const handlers = [...getPetsMock(), ...getFooMock(), ...getBarMock(), /* ...etc */];
const server = setupServer(...handlers);
```

In this PR you can do this and never touch it again 🎸 
```ts
import * as mocks from './endpoints/index.msw';

const handlers = Object.entries(mocks).flatMap(([, getMock]) => getMock());
const server = setupServer(...handlers);
```

I didn't investigate why there were other exported functions. I assumed only the handlers would be useful, and that has been my experience so far :)

soartec-lab: Thanks! I think that makes sense. 
Hi, @AllieJonsson what do you thinks about this change.

AllieJonsson: LGTM!

melloware (dismissed):

soartec-lab:
@snebjorn 
The use case of running all handlers and viewing all API calls in an application would be helpful to users.
Could you add this to the documentation after the section "A function that returns an Array that aggregates all handlers in the file"?

https://orval.dev/guides/msw#a-function-that-returns-an-array-that-aggregates-all-handlers-in-the-file

Spence1115:
I know this has been closed but this breaking change is a blocker for us upgrading - we don't always want to import _every_ mock, we might want to override some of them on our side, and not having a single file to export them all means we've got to refactor a bunch of our libraries in order to upgrade. Real shame this functionality wasn't kept as an option.

melloware:
@snebjorn thoughts?

snebjorn:
having `index.msw.ts` be a normal barrel was my initial solution for this, but I couldn't get it to work so I ended with this. I can take a look again now that I'm more familiar with the codebase.

@Spence1115 will this solve your problem?
```ts
// current index.msw.ts
export { getPetsMock } from './pets/pets.msw';
```
becomes
```ts
// index.msw.ts normal barrel
export { getSearchPetsMockHandler } from './pets/pets.msw';
export { getListPetsMockHandler } from './pets/pets.msw';
export { getCreatePetsMockHandler } from './pets/pets.msw';
export { getShowPetByIdMockHandler } from './pets/pets.msw';
export { getShowPetTextMockHandler } from './pets/pets.msw';
export { getUploadFileMockHandler } from './pets/pets.msw';
export { getDownloadFileMockHandler } from './pets/pets.msw';
```

Spence1115:
@snebjorn we've written our own handler for the index file which more or less does what the _original_ here did, because there are some instances we want to import the `mockHandler`, sometimes the entire `mock` array, sometimes we want the response itself etc.

As we're working in a monorepo, we've got our setup where a service has a library within the monorepo, so we import from the library itself, hence why we need everything exported.

PR #2280, opened 2025-08-04
fix(core): export DeepNonNullable for use in clients, fixes #2279

### LINKED ISSUE #2279: DeepNonNullable isn't exported @orval/core - regression
#2244 added this to the angular clients

```ts
import type { DeepNonNullable } from '@orval/core/src/utils/deep-non-nullable';
```

2 issues
- `@orval/core` doesn't export `DeepNonNullable`
- `@orval/core/src/utils/deep-non-nullable` isn't a valid import path. It should probably be `@orval/core`

snebjorn:
## Status

**READY**

## Description

fixes #2279

This should be released ASAP as angular client generation is broken since v7.11.0

### DIFF
--- a/packages/angular/src/index.ts
+++ b/packages/angular/src/index.ts
@@ -44,7 +44,7 @@ const ANGULAR_DEPENDENCIES: GeneratorDependency[] = [
   },
   {
     exports: [{ name: 'DeepNonNullable' }],
-    dependency: '@orval/core/src/utils/deep-non-nullable',
+    dependency: '@orval/core',
   },
 ];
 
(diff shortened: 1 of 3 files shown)


ygorazambuja:
while the release is not out, how can i use this branch ?

snebjorn:
I did this:
c/p this into your project and search/replace the imports to point to it

https://github.com/orval-labs/orval/blob/ee67eeaf48d619eb62935016b033559d593b6f9d/packages/core/src/utils/deep-non-nullable.ts#L1-L9

PR #2292, opened 2025-08-07
feat(fetch): added support for urlEncodeParameters

### LINKED ISSUE #2172: `urlEncodeParameters` config ignored when using `fetch`
The output config [`urlEncodeParameters`](https://orval.dev/reference/configuration/output#urlencodeparameters) says it is only supported in Tanstack Query clients and that is true but it additionally only works if your `httpClient` is set to `axios` (the default). Changing `httpClient` to `fetch` makes the setting no longer do anything.

Near as I can tell, this problem was introduced in the v7.x release when `fetch` support was added for Tanstack Query clients. The code was split, and the original `axios` implementation retained the fix from #679, but the `fetch` implementation does not use it.

You can easily reproduce this in [the playground](https://orval.dev/playground).

1. Choose the Basic example
2. Find the `showPetById` function in `endpoints.ts` and note that is not encoding the `petId` param in the route
3. Add the line `"urlEncodeParameters": true` to `orval.config.ts` under the `"output"` section
4. Find the `showPetById` function in `endpoints.ts` again and note that it now properly wraps `petId` with `encodeURIComponent`, awesome!
5. Add the line `"httpClient": "fetch",` to `orval.config.ts` under the `"output"` section
6. Find the `getShowPetByIdUrl` function in `endpoints.ts` and note that it does not encode the `petId` param in the route
7. You can change or remove `urlEncodeParameters` but the output won't differ as the code that generates for `fetch` never uses it

chcederquist:
## Status

READY

## Description

Fixes #2172 

## Related PRs

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

Outline the steps to test or reproduce the PR here:
Follow the instructions in the bug report (#2172)


melloware (approved):

melloware:
Thanks!  Will let @soartec-lab review since this is fetch related

soartec-lab:
Of course, I'll check this out when I get back from my short week-long vacation 👍

packages/fetch/src/utils.ts
```diff
@@ -0,0 +1,10 @@
+import { TEMPLATE_TAG_REGEX } from '@orval/core';
+
+export const wrapRouteParameters = (
+  route: string,
+  prepend: string,
+  append: string,
+): string => route.replaceAll(TEMPLATE_TAG_REGEX, `\${${prepend}$1${append}}`);
+
+export const makeRouteSafe = (route: string): string =>
+  wrapRouteParameters(route, 'encodeURIComponent(String(', '))');
```
soartec-lab: Please combine these functions into one, not separate them. Also, do not create files like util, but move them to index.

packages/fetch/src/index.ts
```diff
@@ -38,6 +39,11 @@ export const generateRequestFunction = (
   }: GeneratorVerbOptions,
   { route, context, pathRoute }: GeneratorOptions,
 ) => {
+  let _route = route;
```
soartec-lab: You can make the variable immutable and remove the conventional underscore:

```
const implmentionRoute = context.output.urlEncodeParameters ? makeRouteSafe(_route) : route

`return stringifiedParams.length > 0 ? \`${implmentionRoute}${'?${stringifiedParams}'}\` : \`${implmentionRoute}\``
      : `return \`${implmentionRoute}\``
```

melloware (changes requested):
marking unnapproved waiting for changes

packages/fetch/src/utils.ts
chcederquist: Moved into index.ts and made into one function

packages/fetch/src/index.ts
chcederquist: Made it immutable

packages/fetch/src/index.ts
```diff
@@ -38,6 +39,10 @@ export const generateRequestFunction = (
   }: GeneratorVerbOptions,
   { route, context, pathRoute }: GeneratorOptions,
 ) => {
+  const _route = context.output.urlEncodeParameters
```
soartec-lab: @chcederquist 

Please give your variables meaningful names. 
Underscores in variable names are sometimes used to give special meanings in other projects or cultures, so they should not be misleading when reading the source code 👍

chcederquist: I've updated the name

jaydeep987:
@soartec-lab seems this change was removed somehow in next versions ? Even in latest `v8.2.0` does not seem to have url path/query param encoding logic which was fixed in this PR.

melloware:
@jaydeep987 looking at the code it looks like this URL encoding was moved to the TanStack Query code outside of Fetch Or Axios?  Are you not using TanStack Query?

jaydeep987:
@melloware Thank you for quick reply.
We are using plain fetch implementation for now due to several reasons. Not yet using TanStack query.

Our sdk's orval version was not maintained since last few months, (it is at 7.11.2) and now I am upgrading it slowly as SDK generated by orval is core part of app.

melloware:
ahhh that makes sens it looks like someone refactored it to only be in the React Query code.  You might want to open a new ticket

PR #2300, opened 2025-08-09
feat(msw): exposed RequestHandlerOptions on mock handlers

Georgegriff:
## Status

**READY

## Description

Mock service worker handlers expose a third arg, very similaer to an event listener that lets you do stuff like {once: true} allowing you to run a handle only one time, this is extremely helpful for mocking different api responses in a single test

## Todos

- [ ] Tests
- [ ] Documentation


packages/mock/src/msw/index.ts
Georgegriff, line 1: the only change, rest is generated sample files

PR #2301, opened 2025-08-09
feat: support uint64 type in `output.override.useBigInt`

ptlthg:
## Status

**READY**

## Description

Adds support for treating the `uint64` data type as a bigint in the same way that `int64` is handled. I'm using orval on an api that has `uint64`s as well as `int64`, and the discrepancy on how they were handled bothered me, so here we are!

Not sure if I need to provide any steps or anything here, it's a pretty basic change, but let me know.


melloware (approved):
Nice improvement

Issue #2182, opened 2025-06-26
Zod: Support JSON-Schema minimum/exclusiveMinimum → Zod .min()/.gt()


danielArchlet:
### 🚀 Feature Request: Support JSON-Schema `minimum`/`exclusiveMinimum` → Zod `.min()`/`.gt()` refinements

**Motivation**  
Currently Orval’s Zod client generates only the bare primitives (`z.number()`, `z.string()`, etc.) and does not translate JSON-Schema numeric keywords (`minimum`, `exclusiveMinimum`, `maximum`, `exclusiveMaximum`) into Zod refinements (`.min()`, `.gt()`, `.max()`, `.lt()`). This means that any numeric constraints declared in our OpenAPI definitions are lost in the generated schemas, leading to gaps between backend validation (e.g. Pydantic’s `gt=0`) and frontend validation in React-Hook-Form.

**Current Behavior**  
Given an OpenAPI schema like:

```yaml
components:
  schemas:
    BaseBid:
      type: object
      properties:
        price:
          type: number
          exclusiveMinimum: 0
```

Orval emits:

```ts
// generated by Orval Zod client
export const BaseBidSchema = z.object({
  price: z.number(),
});
```

—missing the .gt(0) that enforces exclusiveMinimum: 0.

**Expected Behavior**
Orval should detect numeric keywords and emit the corresponding Zod refinements. For the example above, the generated schema would be:

```t
export const BaseBidSchema = z.object({
  price: z.number().gt(0),
});
```

**Proposal**
Extend the Zod client to map:
- minimum: n → .min(n)
- exclusiveMinimum: n → .gt(n)
- maximum: n → .max(n)
- exclusiveMaximum: n → .lt(n)

Thanks for the work on proper Zod refinements so far! Is support for these numeric refinements under consideration? Happy to help with test cases if that would be useful.

Zerohidz:
+1

melloware:
PR from anyone would be great!

harish1094:
+1

PR #2308, opened 2025-08-14
Missing consts  and some export includes `,`;

harish1094:
## Status

**READY**

## Description

Not generating `export const` for `maximum` value for following data :
```
"max_val_steps": {
    "title": "Max Val Steps",
    "anyOf": [
      {
        "maximum": 8000.0,
        "minimum": 1.0,
        "type": "integer"
      },
      {
        "$ref": "#/components/schemas/MissingType"
      }
    ],
    "default": "???",
    "metadata": {
    }
  }
```

Also some generated export includes `,` like the following:

```
export const getResponseAvgRelTimeMinOne = 2;
,export const getResponseAvgRelTimeMaxOne = 2;
export const getResponseMostRecentRelTimeMinOne = 2;
,export const getResponseMostRecentRelTimeMaxOne = 2;

```

- [x] If the `anyOf` function contains value like `maximum`, `consts` are not being exported(added) inside the  `parseProperty`
- [x] Some` export consts` includes `,` and gives error in reactjs projects. added replaceAll to remove those `,`

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -582,11 +582,17 @@ export const parseZodValidationSchemaDefinition = (
       }
 
       const union = args.map(
-        ({ functions, consts:consts2 }: { functions: [string, any][], consts:string[] }) => {
+        ({
+          functions,
+          consts: argConsts,
+        }: {
+          functions: [string, any][];
+          consts: string[];
+        }) => {
           const value = functions.map(parseProperty).join('');
           const valueWithZod = `${value.startsWith('.') ? 'zod' : ''}${value}`;
           // consts are missing here
-          consts += consts2?.join("\n");
+          consts += argConsts?.join('\n');
           return valueWithZod;
         },
       );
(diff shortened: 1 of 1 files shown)


harish1094:
@melloware , @anymaniax 

Looks like similar to this https://github.com/orval-labs/orval/issues/2182,
Could you please take a look?

packages/zod/src/index.ts
```diff
@@ -582,9 +582,11 @@ export const parseZodValidationSchemaDefinition = (
       }
 
       const union = args.map(
-        ({ functions }: { functions: [string, any][] }) => {
+        ({ functions, consts:consts2 }: { functions: [string, any][], consts:string[] }) => {
```
melloware: consts2 seems weird?

melloware:
@harish1094 have you confirmed this fixes #2182 ?

harish1094:
> @harish1094 have you confirmed this fixes #2182 ?

No, but I tested this changes with the specific case I mentioned in the description, and this change resolved the issue.

PR #1730, opened 2024-12-02
feat(query): Add default mutation key

b-nijhuis-innovadis:
## Status

**READY**

## Description

Add a default mutation key for React Query mutation generation if none was provided.

Fix https://github.com/orval-labs/orval/issues/1249

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Changelog Entry (unreleased)

## Steps to Test or Reproduce

Generate a React Query mutation without supplied mutation key to check if default key is used. Generate with a supplied mutation key to check if that one is used instead.

Issue #243, opened 2021-11-01
Customize default query keys


mpetito-envative:
Is it possible to configure the convention used for query key generation?

I see that you can override the query key when using the generated hook, but this requires remembering to do so each time.  Instead, I'd like to configure how the query key is constructed project-wide because our APIs already have consistent hierarchical routes.

For ex., instead of this generated query key:

```
export const getGetResourceByIdQueryKey = (resourceId: number,) => [`/resource/${resourceId}`];
```

It would make a lot more sense in my project for the query key to be split by path segments, e.g.:

```
['resource', resourceId]
```

I don't see any obvious way to configure or override the convention used for query keys.

My motivation is to simplify query invalidation by having predictable query keys and support prefix invalidation.

anymaniax:
Hello @mpetito-envative, not really the only thing that you can do at the moment is to override the options in the orval config. Could be a new option but how would you do it?

mpetito-envative:
It seems like this would be a runtime behavior because the queryKey depends on properties passed into the query hook.

It is already possible to do something similar with the mutator override for the purpose of replacing the http request.  Perhaps a queryKey override would allow for replacement of the queryKeyFn at runtime?

Alternatively (or in addition), it might make sense to change the default convention used for queryKey generation.  I know this would be a breaking change, but I can't think of any disadvantage to splitting the route on `/` and it would offer an easy path for prefix invalidation if routes already represent a hierarchy.

anymaniax:
It's also what I thought but if we do a sort of mutator I will need to pass the route and all params to the function. For the other one, could make sense but since it's a breaking change I will add it for the next major release

knyzorg:
@anymaniax Any updates on this? This issue just bit me and I'm not sure how to move forward without hardcoding the path in some form and matching with a predicate.

missing1984:
+1. This is an awesome tool but the queryKey is blocking me from using it.

anymaniax:
@missing1984 it's on my todo list for the next version. Sorry I have a lot of stuff to do lately

OliverDudgeon:
Another use case for query-key customisation is when there are more than one orval-generated api clients. If both that a `/version` endpoint for example, then the queries keys clash.

la55u:
Did you guys manage to find a workaround?

missing1984:
I wrote a custom script base on "ts-morph" to process the generated file. It will be great to have native support though

knyzorg:
[...]

PR #2314, opened 2025-08-18
feat: Add useOperationIdAsQueryKey option for query key generation

### LINKED ISSUE #2096: Feature Request: Add operationId to queryKey for better React Query hook identification
### Is your feature request related to a problem? Please describe.

Currently, when generating React Query hooks using Orval, the `queryKey` returned from the `...QueryKey` functions only includes the API path, for example:

```ts
return [`/some/path`] as const;
```

This makes it difficult to uniquely identify or cache queries in larger projects where many paths may be dynamic or similar.

---
### Describe the solution you’d like

I’d like to request an option in Orval’s config to include the operationId alongside the API path in the queryKey, like:

```ts
return ['getSomeData', `/some/path`] as const;
```

---
### Describe alternatives you’ve considered

We are currently using a custom post-generation script to parse the generated api-docs.ts and inject the operationId manually, but this approach is brittle and could break with future Orval updates.

---
### Additional context

```yaml
reactQuery:
  includeOperationIdInQueryKey: true
```

Thanks for the great tool and all your hard work!

mortik:
## Summary
Fix #2096 

This PR adds a new configuration option `useOperationIdAsQueryKey` that allows users to generate query keys using the OpenAPI operation ID instead of the route string for TanStack Query implementations.

## Changes

- Added `useOperationIdAsQueryKey?: boolean` to `QueryOptions` and `NormalizedQueryOptions` types
- Modified query key generation logic in `packages/query/src/index.ts` to use operation ID when enabled
- Updated `normalizeQueryOptions` function to handle the new option
- Added comprehensive tests in `packages/query/src/utils.test.ts`
- Added documentation in `/docs/src/pages/reference/configuration/output.md`

## Configuration Example

```javascript
module.exports = {
  petstore: {
    output: {
      client: 'react-query',
      override: {
        query: {
          useOperationIdAsQueryKey: true,
        },
      },
    },
    input: './petstore.yaml',
  },
};

### DIFF
--- a/packages/query/src/index.ts
+++ b/packages/query/src/index.ts
@@ -1328,13 +1328,6 @@ const generateQueryHook = async (
         ? getRouteAsArray(route) // Note: this is required for reactivity to work, we will lose it if route params are converted into string, only as array they will be tracked // TODO: add tests for this
         : `\`${route}\``;
 
-    const queryKeyPathParams = props
-      .filter(
-        (prop) =>
-          prop.type === GetterPropType.NAMED_PATH_PARAMS ||
-          prop.type === GetterPropType.PARAM,
-      )
-      .map((prop) => prop.name);
 
     // Use operation ID as query key if enabled, otherwise use route string
     const queryKeyIdentifier = override.query.useOperationIdAsQueryKey

### CODE REVIEW

**Comment by mortik** [comment]
This should resolve #2096

**Review by AllieJonsson — DISMISSED** [review]
LGTM!

**Comment by mortik** [comment]
I realized that this approach still needs an update to include the path params like id etc. i will update the pr to include that in the querykey function

**Comment by soartec-lab** [comment]
Please give me a little time to salvage past discussions about cases where you might want to use arbitrary strings for query keys. And once a policy has been decided, it will need to be documented in the documentation to avoid any confusion.

**Comment by soartec-lab** [comment]
Hi, @mortik  Sorry for the long wait, I've retraced the steps so far.

First, if you want to customize the query key, you can currently do so by using `override.query.querykey`.

https://orval.dev/reference/configuration/output#querykey

For `mutator`, starting with #1730, the `operationId` defined in the `OpenAPI` is now used as the query key by default.

- https://github.com/orval-labs/orval/pull/1730

Other discussions:

- https://github.com/orval-labs/orval/issues/243

---

With this change, when you add an option, the `OpenAPI` `operationId` will be used as the query key by default. This makes sense to me, as it's the same behavior as the default for `mutator` 👍

File: packages/query/src/index.ts [hunk 1]
```diff
@@ -1324,11 +1324,24 @@ const generateQueryHook = async (
...
+        (prop) =>
+          prop.type === GetterPropType.NAMED_PATH_PARAMS ||
+          prop.type === GetterPropType.PARAM,
+      )
+      .map((prop) => prop.name);
+
+    // Use operation ID as query key if enabled, otherwise use route string
+    const queryKeyIdentifier = override.query.useOperationIdAsQueryKey
+      ? [`'${operationName}'`, ...queryKeyPathParams]
+      : routeString;
```
soartec-lab: Should this be common and no separation is necessary?

```suggestion
      ? operationName : routeString;
```

soartec-lab (commented):
@mortik  I added a question.

melloware:
@mortik looks like conflicts too

mortik:
> Hi, @mortik Sorry for the long wait, I've retraced the steps so far.
> 
> First, if you want to customize the query key, you can currently do so by using `override.query.querykey`.
> 
> https://orval.dev/reference/configuration/output#querykey
> 
> For `mutator`, starting with #1730, the `operationId` defined in the `OpenAPI` is now used as the query key by default.
> 
> * [feat(query): Add default mutation key #1730](https://github.com/orval-labs/orval/pull/1730)
> 
> Other discussions:
> 
> * [Customize default query keys #243](https://github.com/orval-labs/orval/issues/243)
> 
> With this change, when you add an option, the `OpenAPI` `operationId` will be used as the query key by default. This makes sense to me, as it's the same behavior as the default for `mutator` 👍

i tried to use the `override.query.querykey` option but sadly the function does not get the nescessary information from the schema to use, for example, the operationId as a querykey. Would also be an interessting change if the custom query key function would get more context information to change the default querykey behavior.

mortik:
> @mortik looks like conflicts too

i resolved the conflict :)

packages/query/src/index.ts
soartec-lab: @mortik 
I still don't understand why this change is necessary so could you explain for me?
Or, incorporate this change.

https://github.com/orval-labs/orval/pull/2314#discussion_r2365680855

packages/query/src/index.ts
```diff
@@ -1328,11 +1328,23 @@ const generateQueryHook = async (
         ? getRouteAsArray(route) // Note: this is required for reactivity to work, we will lose it if route params are converted into string, only as array they will be tracked // TODO: add tests for this
         : `\`${route}\``;
 
+    const queryKeyPathParams = props
+      .filter(
+        (prop) =>
+          prop.type === GetterPropType.NAMED_PATH_PARAMS ||
+          prop.type === GetterPropType.PARAM,
+      )
+      .map((prop) => prop.name);
```
soartec-lab: I think this process is no longer necessary.

```suggestion
```

PR #2333, opened 2025-08-25
fix: fixed broken void to null unification (revert #2229)

### LINKED ISSUE #2291: `null` vs `undefined` breaking change `7.10.0` <=> `7.11.0`
## Related contributors

- @snebjorn @melloware @soartec-lab 

## Problem

#2226 #2229 change shouldn't have been accepted.
The documentation example doesn't say the endpoint returns null, it says there's no response.body.

`null` is not `undefined`.

If the fetch client returns null, when the documentation says it should return nothing; the problem is in the client, not the typings.

---

Now that the generated typings are invalid, some might check `response === null`, introducing runtime issues.

soartec-lab:
This reverts commit b015a1e1867b019923408a2477f767e86f0c1525.

## Status

**READY**

## Description

fix #2291 

#2229 contains bugs and is still controversial, so we are reverting it.

## Related PRs

#2229

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

Compare with #2229


melloware:
@snebjorn you would have to submit a new PR that only fixes the Angular client.

PR #2352, opened 2025-09-02
fix(axios): provide type parameter when inferring mutator `options` type

### LINKED ISSUE #2351: Type parameter should be provided
Currently, we are omitting the mutator function type parameter from the `option` parameter.

```ts
export const getFoo = (options?: SecondParameter<typeof getFooMutator>) => { // <--- Missing type parameter "Bar"
	return getFooMutator<Bar>({url: `/api/foo`, method: 'GET'}, options);
};
```

In doing so, we are preventing users from generating strongly typed functions such as this:

```ts
getFoo({
  onMessage(message) {
    setBar(message); // <--- ❌ Error: `message` is typeof `unknown`
  }
});
```

- Generic options are useful to create functions for Server Sent Events which are representable in OpenAPI
- Generic options are useful to support filtering or sorting functionalities in the type system

But the crucial point is: There is really no need (that I can see) for us to omit the type parameter. We are already providing the type parameter in the actual call to `getFooMutator()`.

```ts
export const getFoo = (options?: SecondParameter<typeof getFooMutator<Bar>>) => {
	return getFooMutator<Bar>({url: `/api/foo`, method: 'GET'}, options);
};

getFoo({
  onMessage(message) {
    setBar(message); // <--- ✅ `message` is typeof `Bar`
  }
});
```

I would be happy to take a stab at this problem.

hahanein:
## Status

**READY**

## Description

When a mutator has a second parameter, the generated `options` was typed as `SecondParameter<typeof mutator>`, which drops the mutator's generic and widens the options type. That prevents strongly-typed per-endpoint config.

Instantiate the mutator with the endpoint success type when extracting the second parameter:

```ts
  SecondParameter<typeof mutator<SuccessType>>
```

This preserves the correct, strongly-typed options.

Closes #2351.

## Todos

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/axios/src/index.ts
+++ b/packages/axios/src/index.ts
@@ -137,7 +137,7 @@ const generateAxiosImplementation = (
 
     return `const ${operationName} = (\n    ${propsImplementation}\n ${
       isRequestOptions && mutator.hasSecondArg
-        ? `options${context.output.optionsParamRequired ? '' : '?'}: SecondParameter<typeof ${mutator.name}>,`
+        ? `options${context.output.optionsParamRequired ? '' : '?'}: SecondParameter<typeof ${mutator.name}<${response.definition.success || 'unknown'}>>,`
         : ''
     }) => {${bodyForm}
       return ${mutator.name}<${response.definition.success || 'unknown'}>(


hahanein:
@melloware Can you maybe take a look at this? I don't know how to take care of those todos or if that is necessary in this case.

PR #2358, opened 2025-09-09
Upgrade faker to v10

Julien-Molina:
## Status

**READY*

## Description

Upgrade faker dependency to version 10


snebjorn:
What was this intended to solve? I don't believe updating faker in the root package does much.

melloware:
We should update it everywhere

PR #2334, opened 2025-08-25
Enables goto source/reference/implementation

snebjorn:
## Status

**READY**

## Description

Enables goto source/reference/implementation

In VS Code ctrl clicking this
image

Jumps here
image

## Related PRs

List related PRs against other branches:

- continuation of #2325 merge this first
- #2303

PR #2378, opened 2025-09-17
chore: change to tsdown as tsup is unmaintained

snebjorn:
## Status

**READY**

## Description

Change bundler to tsdown because tsup is [now unmaintained](https://github.com/egoist/tsup/commit/8b6907df5d944e5e4a639a297875a35818ba9fa5)

image

Luckily tsdown is much nicher!

It comes with "GOTO reference" implemented in #2334 out of the box.
#2334 also broke setting breakpoints in the TS files, which I found out yesterday, and tsdown fixes that issue 🚀

tsup is dead, long live tsdown


melloware:
build failed

PR #2412, opened 2025-09-30
fix: use dynamicImport function instead of native import for package.json loading

MtBlue81:
## Summary

This PR fixes an issue with package.json loading by using the `dynamicImport` function instead of native `import` statement.

## Changes

- **packages/orval/src/utils/package-json.ts**: Replace native `import()` with `dynamicImport<any>()` function call for loading package.json files

## Motivation
Starting with Node.js v22, dynamic imports require `{ with: { type: ‘json’ }`.

For example, specifying `packageJson` in `samples/react-query/basic/orval.config.ts` to generate code results in the following error:

```
react-query-basic:generate-api: node:internal/modules/esm/assert:88
react-query-basic:generate-api:         throw new ERR_IMPORT_ATTRIBUTE_MISSING(url, 'type', validType);
react-query-basic:generate-api:               ^
react-query-basic:generate-api:
react-query-basic:generate-api: TypeError [ERR_IMPORT_ATTRIBUTE_MISSING]: Module "file:///<local-path>/orval/samples/react-query/basic/package.json" needs an import attribute of "type: json"
```

## Testing

- Tested with the react-query basic sample in my local machine
- Verified that package.json loading works correctly with the new implementation


melloware:
@MtBlue81 does this fix any open tickets?   @snebjorn  can you review?

melloware (changes requested):
Build is failing

lukemorales:
@melloware I didn't find any open issues but I'm facing this problem after orval@7.12 was released

melloware:
thanks @lukemorales !

packages/orval/src/utils/package-json.ts
```diff
@@ -21,7 +21,7 @@ export const loadPackageJson = async (
 
   const normalizedPath = normalizePath(packageJson, workspace);
   if (fs.existsSync(normalizedPath)) {
-    const pkg = await import(normalizedPath);
+    const pkg = await dynamicImport<any>(normalizedPath);
```
snebjorn: please use unknown over any

melloware:
```suggestion
    const pkg = await dynamicImport<unknown>(normalizedPath);
```

melloware (approved):

packages/orval/src/utils/package-json.ts
```diff
@@ -21,7 +21,7 @@ export const loadPackageJson = async (
 
   const normalizedPath = normalizePath(packageJson, workspace);
   if (fs.existsSync(normalizedPath)) {
-    const pkg = await import(normalizedPath);
+    const pkg = await dynamicImport<unknown>(normalizedPath);
```
melloware, line 29:
```suggestion
    const pkg = await dynamicImport<any>(normalizedPath);
```

packages/orval/src/utils/package-json.ts
melloware: It has to be `any` as it fails with `PackageJson cannot be unknown` TypeScript failure

MtBlue81: I think it might be better to set generics to unknown and use type guard for pkg. 
Should I create a PR when I have time?

melloware: Sure

MtBlue81: @melloware I added the type guard.

lukemorales:
@melloware is this going to be released any time soon? Since it's a bug preventing my team from updating orval, having patch release with it instead of waiting for everything planned for the next minor version would be much appreciated

melloware:
yes i will try and release soon!

melloware:
@lukemorales 7.13.1 is out if you want to test it please?

lukemorales:
@melloware I just updated our dependencies and we have no errors! thanks for the patch release!

melloware:
Excellent!

PR #2423, opened 2025-10-02
feat: add a way to provide an axios instance

jdrouet:
When using the `axios` client, I'd like to be able to provide an axios instance. The way to do in the doc is to define a global constant which is then used globally. This is a problem because it allows only a single instance of the client in the application and it's potentially prone to security issues.
With this, we can create an instance of axios an pass it as an argument like the following

```
const instance = axios.create({ headers: { Authentication: 'Bearer foo' } });
const api = createClient();

return api(instance);
```


melloware:
this looks pretty good to me build is running now

melloware:
sorry do a git pull from main please

jdrouet:
I rebased 😉

melloware:
running build again

melloware:
ok it looks like it is failing the samples check you might want to run `yarn update-samples` then `yarn test:samples`

jdrouet:
@melloware should be fixed now 🤔

melloware:
`Parameter 'axios' cannot reference itself.`

jdrouet:
I will come with a proper fix today or tomorrow 😉

melloware:
Any update on this?

PR #2435, opened 2025-10-12
chore: removes the last require() usages

snebjorn:
This removes the last usages of legacy `require` paving the way for ESM

I had to bump the minimum required node version to 22.18.0. This is a **breaking change**.

Even though tests pass and no meaningful changes to samples were made, I doubt this will work for all users.

### DIFF
--- a/.github/workflows/deploy-doc.yaml
+++ b/.github/workflows/deploy-doc.yaml
@@ -13,10 +13,10 @@ jobs:
       - uses: actions/checkout@v5
         with:
           fetch-depth: 0
-      - name: Use Node.js 20
+      - name: Use Node.js 22
         uses: actions/setup-node@v5
         with:
-          node-version: 20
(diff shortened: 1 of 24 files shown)


melloware:
Ok I marked this for 7.14.0

melloware:
Actually if this is that major should it be 8.0.0?

snebjorn:
I added this to the orval package to make it clear that node 22.18.0 is the minimum.
```
"engines": {
  "node": ">=22.18.0"
},
```

I don't know what'll happen to bun and deno users. But we don't officially support those, right?

melloware:
No i think all other package managers are "great if it works" and if people submit fixes we will accept them.

Noyabronok:
> Ok I marked this for 7.14.0

Per semver, breaking changes must get a major version bump

Issue #1557, opened 2024-08-01
Can't fetch refs from url


iArranzm:
Hi! hope you can help me with this.

## What happens?
When I'm trying to generate my services, I'm getting an error when trying to fetch an schema from an URL like this (changed Ref for privacy)
```yaml
properties:
        pet:
          allOf:
            - $ref: "https://raw.githubusercontent.com/orval-labs/orval/master/samples/angular-app/petstore.yaml#/components/schemas/Pet"
            - description: pet info
```
The error is 

> orval --config orval.config.ts
🍻 Start orval v6.31.0 - A swagger client generator for typescript
fakeService: Cleaning output folder
🛑 fakeService - Oops... 🍻. Ref not found: https://raw.githubusercontent.com/orval-labs/orval/master/samples/angular-app/petstore.yaml#/components/schemas/Pet

…

## What were you expecting to happen?
I was expecting that Refs could be retrieved from and URL. At first I thought it was because the URL I was trying to access required me to be connected to the VPN, but when I switch to your test YAML, I get the same error.

## Any other comments?

I'm don't think it's a bug, probably it's a feature not supported or I am doing something wrong.  If you need further info I'll be glad to provide it.

## What versions are you using?

System:
    OS: macOS 14.5
    CPU: (12) arm64 Apple M2 Pro
    Memory: 61.92 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    orval: ^6.31.0 => 6.31.0

melloware:
Duplicate of https://github.com/orval-labs/orval/issues/863

melloware:
Yeah I am sure its a bug or unimplemented feature.

atsoy:
I fear it happens in also in `v7.0.1`.
Wanted to upgrade the version from `6.15.0` (where is was fine), but it can't find `$ref`.

```yaml
openapi: 3.1.0
info:
  description: Default description
  version: 1.0.0
  title: Default Title
paths:
  foo:
    get:
      operationId: myOperationId
      summary: Just a summary
      parameters:
        - $ref: '../parameters/path/path-bar.yaml#/components/parameters/path-bar'
      responses:
        '200':
          description: Default response
          content:
            application/json:
              schema:
                $ref: '../schemas/index.yaml#/components/schemas/DefaultDTO' // can't find this one, and it's not a single case
```

P.S: @melloware since it works in `6.15.0`, I would tend to assume it's a bug.

melloware:
[...]

PR #2439, opened 2025-10-15
fix: don't mangle external URL $ref

xandris:
It should be valid for a local file `target.yaml` to reference an external schema like GeoJSON. Preserve these URLs as-is and don't turn them into fake file paths.

Fix https://github.com/orval-labs/orval/issues/1557

### DIFF
--- a/packages/core/src/getters/ref.ts
+++ b/packages/core/src/getters/ref.ts
@@ -87,9 +87,15 @@ export const getRefInfo = (
     };
   }
 
-  const path = isUrl(context.specKey)
-    ? resolveUrl(context.specKey, pathname)
-    : upath.resolve(getFileInfo(context.specKey).dirname, pathname);
+  let path: string;
+
+  if (isUrl($ref)) {
+    path = $ref;
+  } else if (isUrl(context.specKey)) {
+    path = resolveUrl(context.specKey, pathname);
+  } else {
+    path = upath.resolve(getFileInfo(context.specKey).dirname, pathname);
+  }
 
   return {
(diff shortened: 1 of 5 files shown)


packages/core/src/getters/ref.ts
```diff
@@ -87,9 +87,11 @@ export const getRefInfo = (
...
   }
 
-  const path = isUrl(context.specKey)
-    ? resolveUrl(context.specKey, pathname)
-    : upath.resolve(getFileInfo(context.specKey).dirname, pathname);
+  const path = isUrl($ref)
+    ? $ref
+    : isUrl(context.specKey)
+      ? resolveUrl(context.specKey, pathname)
+      : upath.resolve(getFileInfo(context.specKey).dirname, pathname);
```
melloware: This code is a bit hard to follow with nested ternary operators. Maybe someting like this...

```js
let path;

if (isUrl($ref)) {
  path = $ref;
} else if (isUrl(context.specKey)) {
  path = resolveUrl(context.specKey, pathname);
} else {
  const { dirname } = getFileInfo(context.specKey);
  path = upath.resolve(dirname, pathname);
}
```

melloware (changes requested):

packages/core/src/getters/ref.ts
xandris: sure thing :) some days i wish js had switch expressions

melloware:
Thanks !  building and running tests now.   great improvement!

xandris:
glad to be useful :) this is an awesome project tyvm

PR #2444, opened 2025-10-17
fix(hono): zValidator not imported consistently

xandris:
The branches in `generateHandlers` independently compute whether the generated code requires the zValidator import. This code hasn't been kept in sync, however, notably for the `tags` and `tags-split` modes. In this case, `generateHandlers` assesses that `zValidator` isn't needed, but `getHonoHandlers` produces code that validates the response body anyway, resulting in a Typescript compilation error for first-time users of orval.

I don't prefer code generation with string manipulation but don't want to make any drastic changes to someone else's codebase. Instead, have `getHonoHandlers` itself report whether `zValidator` is required by bundling this flag alongside the generated code in a tuple. Also, handle the one verb and many verbs per file cases uniformly by making `getHonoHandlers` variadic. It performs the same kind of reduction that was used before, but also reduces over the zValidator flag.

Add tests for this condition to the hono configs. Tag the verbs in `examples.yaml` to trigger the behavior.

### DIFF
--- a/packages/hono/src/index.ts
+++ b/packages/hono/src/index.ts
@@ -131,47 +131,64 @@ export const generateHono: ClientBuilder = async (verbOptions, options) => {
   };
 };
 
-const getHonoHandlers = ({
-  handlerName,
-  contextTypeName,
-  verbOption,
-  validator,
-}: {
-  handlerName: string;
-  contextTypeName: string;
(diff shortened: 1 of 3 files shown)


xandris:
It looks like I introduced a whitespace change. I'll remediate that.

PR #2446, opened 2025-10-18
fix(hono): some configs generate wrong imports

xandris:
Apologies for the larger pull request. Hopefully the negative diffstat makes up for it.

`single` mode generates these files:

```text
.
├── endpoints.context.ts
├── endpoints.handlers.ts
├── endpoints.ts
├── endpoints.validator.ts
└── endpoints.zod.ts
```

But the contexts file has this import:

```typescript
import { ListPetsParams,
CreatePetsParams,
CreatePetsBody } from './endpoints.schemas';
```

When in reality the TypeScript types live in `endpoints.ts`.

`tags` mode generates these files:

```text
.
├── endpoints.schemas.ts
├── endpoints.validator.ts
├── health.context.ts
├── health.handlers.ts
├── health.ts
├── pets.context.ts
├── pets.handlers.ts
├── pets.ts
└── pets.zod.ts
```

But strangely, an absolute import:

```typescript
import {
  ListPetsParams,
  CreatePetsParams,
  CreatePetsBody
} from 'endpoints.schemas';
```

Refactor the hono client to prevent these sorts of issues.

First, introduce a helper that generates import paths. Prefer computing the relative path between absolute paths. Callers must know the importer and importee paths but otherwise shouldn't perform path manipulation even if the relationship between the two is supposedly known.

With path handling abstracted, this lets us remove some duplicated code. Much of the duplication relates to mode handling, with the difference being the import paths in the code templates. Split `generateHandlers` into `generateHandlerFiles` and `generateHandlerFile`. `generateHandlerFiles` decides how to group verbs into files and defers code generation to `generateHandlerFile`. split `generateContext` and `generateHandlers` similarly.

Remove some duplicate path computation. `generateZvalidator` reports its output path, feed that to `generateHandlerFiles`. Compute the path to the schemas in `generateExtraFiles` because its location doesn't vary per tag split.

Externalize the zValidator implemention. It's static and outlining it makes the code easier to read.

Use type imports in the orval code to satisfy the TypeScript compiler.

Generate type imports where we can; it's a good practice.

Use a Set to avoid a quadratic unique algorithm. As a side effect, generate sorted imports.

I notice there are a lot of uses `.reduce` that mutate the accumulator and `.flatMap` that `.push` into an array; these are usually clearer as a traditional for-loop.


melloware:
looks like you might need to run `yarn update-samples`

melloware (changes requested):
see comments

xandris:
apologies, i will update the samples when i get home.

did you have any code review comments or was it just the samples?

melloware:
So far it looks pretty good to me

xandris:
uh it seems to only change the samples on Windows. so either a case-insensitivity issue, line ending issue, or path separators. i don't have windows, is there a way to get a diff

xandris:
oh nevermind it's probably the import paths. those should always be forward slash regardless of the underlying os

melloware:
Did you mean to close this?

xandris:
ah dang i didn't. i must have pushed the 'close with comment' button again i stg

PR #2456, opened 2025-10-20
fix: generate prefetch hook for hook based mutator

### LINKED ISSUE #2441: useRefresh for react-query with hook mutator is wrong.
When I use `useRefresh` with a hook mutator, the output is incorrect.

```ts
export const prefetchFindPetsByStatusQuery = async <TData = Awaited<ReturnType<ReturnType<typeof useFindPetsByStatusHook>>>, TError = void>(
queryClient: QueryClient, params?: FindPetsByStatusParams, options?: { query?:UseQueryOptions<Awaited<ReturnType<ReturnType<typeof useFindPetsByStatusHook>>>, TError, TData>,}
): Promise<QueryClient> => {
    // >>> this is wrong
    const queryOptions = useFindPetsByStatusQueryOptions(params,options)
    await queryClient.prefetchQuery(queryOptions);
    return queryClient;
}
```

I would expect something like this

```ts
export const usePrefetchFindPetsByStatus = (params, options) => {
      const findPetsByStatus = useFindPetsByStatusHook(params, options)
      return useCallback(() => queryClient.prefetchQuery(queryOptions), [options, queryClient, params]);
};
```

If it looks okay, I can create PR with the fix.

aizerin:
Generate a different implementation for refetch function when `isHook` is true.

fixes #2441

### DIFF
--- a/packages/query/src/index.ts
+++ b/packages/query/src/index.ts
@@ -167,6 +167,7 @@ const REACT_QUERY_DEPENDENCIES_V3: GeneratorDependency[] = [
       { name: 'useQuery', values: true },
       { name: 'useInfiniteQuery', values: true },
       { name: 'useMutation', values: true },
+      { name: 'useQueryClient', values: true },
       { name: 'UseQueryOptions' },
       { name: 'UseInfiniteQueryOptions' },
       { name: 'UseMutationOptions' },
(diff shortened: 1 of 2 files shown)


melloware:
Do any docs need to be updated as well for config?

aizerin:
I don’t think so. This just fixes existing functionality by generating a new method for hook-based mutators.

melloware (approved):

Issue #434, opened 2022-06-18
v8 ideas 💡 


anymaniax:
I am already thinking about version 7 for a moment and wanted a place to talk about it.

1. Split the lib into multiple packages to have a core and a package by template/client. Like that we can have multiple advantages. Versioning for the template and be able to use an older version of a template if your project is on an older version of react-query for example. Also having a core will help people that want to create their own template and then use it with other features of orval.

2. Rethink the config file since I started this lib a lot of stuff has been added and I think we can improve some points. For example, the generation could be more custom. I think for example that having a mix of [this](https://redux-toolkit.js.org/rtk-query/usage/code-generation#multiple-output-files) and some of Orval features could be great

3. move to esm? 

if you have any other ideas throw them here so that we can discuss it

melloware:
I definitely like the idea of breaking the lib up so its easier to find and make changes to the different libs.  Changing the config file though might be a big and painful change.  However it may be worth it but I am not sure.

anymaniax:
I was also thinking about having a migrate command to do it easily for users

stijnvanhulle:
https://github.com/anymaniax/orval/issues/347 This one should be great to have :)

anymaniax:
What do you think of [this](https://github.com/lukemorales/query-key-factory) for query keys

melloware:
> What do you think of [this](https://github.com/lukemorales/query-key-factory) for query keys

I LOVE this idea. I find myself doing this a lot in my code...

```ts
const QUERY_LIST = "list-cars";
 const queryList = useGetEntityCars(
        { request: JSON.stringify(lazyParams) },
        {
            query: {
                queryKey: [QUERY_LIST, lazyParams],
                refetchOnWindowFocus: false,
                retry: false,
                cacheTime: 0
            }
        }
    );
```

So then later I can do this...
`queryClient.invalidateQueries([QUERY_LIST]);`

So that library looks like what I wanted to do all along!

melloware:
Although I do hate tying myself to another runtime library written by just one guy.  it would be cool if you could just cope his code and include it in the generated code?

DominicBach:
I've been playing with the codebase for a bit and have been experimenting with using the Typescript AST factory to generate code as opposed to plain string manipulation. I've used it to encapsulate mock generation logic into separate type classes, which I could then compose together to build more complex representations.

It could also solve your idea for additional customization by allowing users to hook into the AST generation process to change the code at generation time.

[...]

PR #2460, opened 2025-10-21
Convert to ESM

snebjorn:
⚠️ This is a breaking change ⚠️ 

Packages now compiles to ESM. No more CommonJS. Users needing CommonJS should stay on the previous version.

Updated a bunch of packages that was stuck on old versions as they moved to pure ESM and we couldn't use them. But now we can.

Related: 
- #434

Breaking changes:
- Orval will no longer look for orval.config.cjs
- orval.config.(js|mjs|ts) must be ESM
```js
// orval.config
// before
module.exports = {...

// after
export default {...
```
- user provided hooks and mutators must be ESM
```js
// my-mutator
// before
module.exports = ...

// after
export default ...
```

For TypeScript users there's no changes

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -18624,7 +18624,7 @@ __metadata:
     "@types/react": "npm:18.3.24"
     axios: "npm:^1.12.2"
     hono: "npm:^4.10.2"
-    msw: "npm:^2.11.2"
+    msw: "npm:^2.11.6"
     npm-run-all: "npm:^4.1.5"
     openapi3-ts: "npm:^4.5.0"
     orval: "workspace:*"


melloware:
Yeah this should be an 8.0 feature I think?

melloware:
I just read through that post and it seems a lot of users have concerns moving to ESM. 

I know in my life it's been a nightmare working on a few different open source react projects.

snebjorn:
> I just read through that post and it seems a lot of users have concerns moving to ESM.
> 
> I know in my life it's been a nightmare working on a few different open source react projects.

I could be wrong. But Orval is a CLI tool, it doesn't integrate directly with the project it's being used in. There are no changes to the output. So I don't understand what the concern is 😃 But I might have missed something.

The only breaking change, I'm aware of, are the user provided mutators and hooks must be ESM now.

melloware:
yeah the mutators etc being ESM to me is a bigger change than one might think but I could be wrong.

People were mad about uprgading to Node 22 and that should have been harmless!

snebjorn:
> People were mad about uprgading to Node 22 and that should have been harmless!

to be fair that breaking change wasn't released as a major. I'd probably also get slightly upset if that happened to me 😄

melloware:
Yep that was my fault!  Also i think i just caused you a merge conflict

melloware:
I will release 7.15.0 tomorrow and then merge this and the docs one as 8.0.0!

melloware (approved):

PR #2463, opened 2025-10-22
Updated docs project

snebjorn:
I had some fun updating the docs project.

I updated the various packages. Switched some packages that was abandoned for maintained ones.
I changed the MDX setup so it's closer to docs https://nextjs.org/docs/pages/guides/mdx 
I changed react files from .js to .tsx
I changed MDX files from .md to .mdx

Unfortunately we can't use the turbopack feature of next@16 because our mdx plugins does not have serializable options.

> **Good to know:**
> remark and rehype plugins without serializable options cannot be used yet with [Turbopack](https://nextjs.org/docs/app/api-reference/turbopack), because JavaScript functions can't be passed to Rust.

source: https://nextjs.org/docs/pages/guides/mdx#using-plugins-with-turbopack

### DIFF
--- a/docs/next.config.ts
+++ b/docs/next.config.ts
@@ -24,6 +24,20 @@ const nextConfig: NextConfig = {
       destination: '/docs/overview',
     },
   ],
+  webpack: (config, { dev, isServer }) => {
+    if (!dev && isServer) {
+      // we're in build mode so enable shared caching for the GitHub API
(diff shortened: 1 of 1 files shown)


melloware:
This is awesome!

snebjorn:
I didn't click all the buttons, but it appears to work

PR #2468, opened 2025-10-23
fix(hono): zValidator wrapper doesn't typecheck with 0.7.4

xandris:
`validator.ts` doesn't type check in client code if you update to `@hono/zod-validator` 0.7.4.

Update the zValidator wrapper with the latest polymorphic/backwards compatible types from 0.7.4. Try hard to not rely on `any`. Type check the code with `tsc` during linting to catch errors.

I'm not personally a fan of how `@hono/zod-validator` is typed. This may be a situation, however, where TypeScript's verifier can't exhaustively check things because of the lack of higher-kinded types. I think the solution is for them to provide overloads for v3 and v4 zod schemata and make the innards of the function very vague about types.


melloware:
might have to run `yarn update-samples` ?

xandris:
i did

```
git clean -dffx
yarn
yarn update-samples
```

and can't get any diffs to happen :/

xandris:
@melloware I dug into this. The content of the files triggering the diff check is JSON stored as Base64-encoded DEFLATE streams:

* `samples/react-app/docs-html-plugin/assets/hierarchy.js`
* `samples/react-app/docs-html-plugin/assets/navigation.js`
* `samples/react-app/docs-html-plugin/assets/search.js`
* `samples/react-app/docs-html/assets/hierarchy.js`
* `samples/react-app/docs-html/assets/navigation.js`
* `samples/react-app/docs-html/assets/search.js`

For instance:

```javascript
window.hierarchyData = 'eJyrVirKzy8pVrKKjtVRKkpNy0lNLsnMzwMKVNfWAgCbHgqm';
```

The DEFLATE streams can be subtly different from platform to platform so this isn't surprising. Typedoc generates these files, so I don't think Orval codegen has any influence whether they change or not

melloware:
hmm so you think its a flaw in our check ?

xandris:
yes since the check is enforcing the output of zlib to be consistent across all machines

melloware:
@snebjorn i am thinking of pulling this check out of the build?

xandris:
I think it's still a useful check, but files that don't pertain to Orval shouldn't be enforced

melloware:
I opened this ticket: https://github.com/nickcharlton/diff-check/issues/12

melloware:
@xandris do you mind reviewing these two hono tickets? https://github.com/orval-labs/orval/issues?q=is%3Aissue%20state%3Aopen%20label%3Ahono

PR #2470, opened 2025-10-24
fix(zod): update import for tree shaking

### LINKED ISSUE #2450: Zod: Zod v4 is recommended to use wildcard import for tree shaking
https://github.com/orval-labs/orval/issues/2278
The fix for this issue behaved differently than I expected.

The current output is as follows.

```ts
import zod from 'zod';
```

However, to fully benefit from Bundler's tree-shaking, it should be a wildcard import like this.

```ts
import * as zod from 'zod';
```

I apologize for requesting this a second time, but I would appreciate your consideration.

melloware:
Fix #2450

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -44,9 +44,9 @@ const ZOD_DEPENDENCIES: GeneratorDependency[] = [
   {
     exports: [
       {
-        default: true,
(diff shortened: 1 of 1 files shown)


melloware:
@chcederquist having trouble making it the way he expects?

PR #2497, opened 2025-10-28
ci: re-enabled outdated sample check

snebjorn:
removed and gitignored the troublesome js files mentioned in https://github.com/orval-labs/orval/pull/2468#issuecomment-3439030674

I intentionally made the gitignore wider than it needed to be. The reasoning is that I don't know what future js files with base64 issues in them might show up.

If someone have some insight into when these show up we can tighten it


melloware:
Weird tests are failing?

melloware (approved):

PR #2499, opened 2025-10-28
chore: replaces lodash with remeda

snebjorn:
Replaces lodash with remeda which have better typescript support

Migration was pretty straight forward https://remedajs.com/migrate/lodash/


melloware (approved):

melloware:
Love this change it simplifies the dependencies

PR #2508, opened 2025-10-29
chore(revert): drop `Tanstack Query` v3 compatibility

melloware:


melloware:
@snebjorn i don't think its this commit as the tests are still failing after the revert

snebjorn:
this one definitely causes the tests to fail. 

image

But more stuff was merged after this one. For instance #2503 is suspiciously not changing sample output. I'm not saying it should or there's an error. But without the PR check we have no idea :) 

So from e00076fd4d67 forward every commit could introduce new errors

melloware:
I think we can just do a git checkout from that commit.  And see if everything passes.  Then apply that commit and fix tests etc etc

PR #2536, opened 2025-11-03
Bugs/2520 zod strict with all of

### LINKED ISSUE #2520: Zod: Strict with allOf OpenAPI is not producing usable zod schemas
Hello,
Using orval to generate zod schemas, using the option 'strict' for complex bodies is applying strict to individual part of the schema but not on the final one.

# Reproduction
orval: `7.15.0`
node version: `22.21.0`

https://github.com/IP-In/orval-zod-bug-report

In `openapi.yaml`, the request POST - /fish has it's body which is the merge of 2 items
A. object with 1 mandatory property 'name'
B. object with 1 mandatory property 'swimming'

```
openapi: 3.1.0
info:
  title: Random API
  version: 1.0.0
paths:
  /fish:
    post:
      requestBody:
        content:
          application/json:
            schema:
              allOf:
                - type: object
                  required:
                    - name
                  properties:
                    name:
                      type: string
                - type: object
                  required:
                    - swimming
                  properties:
                    swimming:
                      type: boolean
...
```

`npm run orval`
_Generate the zod schema_
orval.config.js
```
...
output: {
			client: 'zod',
			target: './generated',
			override: {
				zod: {
					strict: {
						response: true,
						query: true,
						param: true,
						header: false,
						body: true
					}
				}
			}
		}
...
```

`npm run test`
_Try to parse `{name: "Nemo", swimming: true}` which gives the following_

Image

# Expected
A single object is generated with the combination objects with the strict option.

```
export const postFishBody = zod.object({
    "name": zod.string(),
    "swimming": zod.boolean()
}).strict();
```
_I couldn't find an alternative and Zod has already it's opinion on this use case it seems.
https://github.com/colinhacks/zod/issues/1284_

# Actual
```
export const postFishBody = zod.object({
  "name": zod.string()
}).strict().and(zod.object({
  "swimming": zod.boolean()
}).strict())
```

sentik:
# Fix: Correct handling of `allOf` with strict mode in Zod schemas

## Summary

This MR fixes an issue where Zod schema generation from OpenAPI specifications with `allOf` containing multiple object schemas produced incorrect validation code when strict mode was enabled. Previously, `.strict()` was applied to each object separately and then combined using `.and()`, which caused validation issues. Now, all object properties are merged into a single object, and strict validation is applied only once to the merged schema.

**Fixes #2520**

## Problem

When generating Zod schemas from OpenAPI specifications with `allOf` containing multiple object schemas and strict mode enabled, the generator produced invalid Zod code:

```typescript
// ❌ Before (incorrect):
zod.object({ "name": zod.string() }).strict().and(zod.object({ "swimming": zod.boolean() }).strict())
```

This caused validation issues because:
- Strict mode was applied to intermediate schemas instead of the final merged schema
- The `.and()` method creates an intersection that doesn't properly validate unknown keys in strict mode
- Each object schema was validated separately, allowing unknown keys that should be rejected

### Example OpenAPI Schema

```yaml
schema:
  allOf:
    - type: object
      required: ["name"]
      properties:
        name:
          type: string
    - type: object
      required: ["swimming"]
      properties:
        swimming:
          type: boolean
```

## Solution

The fix detects when all parts of an `allOf` schema are objects and strict mode is enabled. In such cases, it merges all object properties into a single object and applies `.strict()` (or `strictObject` for Zod v4) only once to the final merged schema.

```typescript
// ✅ After (correct):
zod.object({
  "name": zod.string(),
  "swimming": zod.boolean()
}).strict()
```

### Core Logic Changes

The fix modifies the `parseZodValidationSchemaDefinition` function in `packages/zod/src/index.ts`:

1. **Detection**: Checks if all parts of an `allOf` schema are objects when strict mode is enabled
2. **Property Merging**: Combines properties from all constituent objects into a single `mergedProperties` object
3. **Single Strict Application**: Applies `.strict()` once to the merged object (or uses `strictObject` for Zod v4)

### Design Decisions

1. **Why merge instead of keeping `.and()`?**
   - Strict mode must be applied to the final merged schema, not intermediate ones
   - Merging produces a single object schema with all properties, which strict mode can validate correctly
   - The `.and()` method creates intersections that don't properly handle unknown keys in strict mode

2. **Why conditional merging?**
   - Only merges when all parts are objects AND strict mode is enabled
   - Falls back to `.and()` for non-object schemas or when strict mode is disabled
   - Maintains backward compatibility with existing behavior

3. **Why support both Zod v3 and v4?**
   - Zod v3 uses `zod.object().strict()` pattern
   - Zod v4 introduces `zod.strictObject()` as a single method
   - Supporting both ensures compatibility with existing projects and future versions

### 1. Core Implementation (`packages/zod/src/index.ts`)

- Modified `parseZodValidationSchemaDefinition` to detect and handle object-only `allOf` schemas
- Added property merging logic for strict mode scenarios
- Added support for both Zod v3 and v4 object creation methods

### 2. Unit Tests (`packages/zod/src/zod.test.ts`)

Added comprehensive test cases:
- `handles allOf with strict mode for objects (issue #2520)` - tests Zod v3 behavior
- `handles allOf with strict mode for objects in Zod v4` - tests Zod v4 behavior

Both tests verify:
- Properties are merged correctly
- Strict is applied only once
- No `.and()` is used in the output

### 3. Integration Test

- Created OpenAPI specification: `tests/specifications/all-of-strict.yaml`
- Added configuration **all-of-strict** in `tests/configs/default.config.ts`:

## Related Issues

Fixes #2520

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -709,32 +709,95 @@ export const parseZodValidationSchemaDefinition = (
     const [fn, args = ''] = property;
 
     if (fn === 'allOf') {
-      return args.reduce(
-        (
-          acc: string,
-          {
-            functions,
-            consts: argConsts,
-          }: { functions: [string, any][]; consts: string[] },
-        ) => {
-          const value = functions.map((prop) => parseProperty(prop)).join('');
-          const valueWithZod = `${value.startsWith('.') ? 'zod' : ''}${value}`;
+      const allOfArgs = args as ZodValidationSchemaDefinition[];
+      // Check if all parts are objects and we need to merge them for strict mode
+      const allAreObjects =
+        strict &&
+        allOfArgs.length > 0 &&
+        allOfArgs.every((partSchema) => {
+          if (partSchema.functions.length === 0) return false;
+          const firstFn = partSchema.functions[0][0];
+          // Check if first function is object or strictObject
+          // For Zod v3 with strict, it will be object followed by strict
+          return firstFn === 'object' || firstFn === 'strictObject';
+        });
 
-          if (argConsts.length > 0) {
-            consts += argConsts.join('\n');
+      if (allAreObjects) {
+        // Merge all object properties into a single object
+        const mergedProperties: Record<string, ZodValidationSchemaDefinition> =
+          {};
+        let allConsts = '';
+
+        for (const partSchema of allOfArgs) {
+          if (partSchema.consts.length > 0) {
+            allConsts += partSchema.consts.join('\n');
           }
 
-          if (!acc) {
-            acc += valueWithZod;
-            return acc;
-          }
+          // Find the object function (might be first or second after strict)
+          const objectFunctionIndex = partSchema.functions.findIndex(
+            ([fnName]) => fnName === 'object' || fnName === 'strictObject',
+          );
 
+          if (objectFunctionIndex !== -1) {
+            const objectArgs = partSchema.functions[objectFunctionIndex][1];
+            if (
+              objectArgs &&
+              typeof objectArgs === 'object' &&
+              !Array.isArray(objectArgs)
+            ) {
+              // Merge properties (later schemas override earlier ones)
+              Object.assign(
+                mergedProperties,
+                objectArgs as Record<string, ZodValidationSchemaDefinition>,
+              );
+            }
+          }
+        }
+
+        if (allConsts.length > 0) {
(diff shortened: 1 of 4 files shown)


sentik:
endpoints.ts for all-of-strict.yaml (example from https://github.com/orval-labs/orval/issues/2520)

[endpoints.txt](https://github.com/user-attachments/files/23294652/endpoints.txt)

PR #2542, opened 2025-11-06
fix(cli): --watch accepts multiple paths again

snebjorn:
- Added conflict error messages to cli arguments that can't be used together.
- `--watch` accepts multiple paths again

Also cleaned up the logic that calls `generateSpec`

I noticed that jiti was recently added to load ts files. I don't think we need a package for that as node 22.18 should support it


snebjorn:
there's similar logic in orval.ts and generate.ts. That's a job for another day :)

PR #2544, opened 2025-11-10
chore: better debugging and editor support

snebjorn:
Enabled https://tsdown.dev/options/package-exports#dev-exports it should:
> During development, you may want your exports to point directly to your source files for better debugging and editor support

Unfortunately nodejs is real picky about imports... it requires the file extension and doesn't support barrels (index).

If anyone got a better solution - let me know


ScriptType:
Normally you can do tsx
```json
"dev": "tsx --watch src/index.ts"
```
but i think it wont work for monorepo and i think adding the file extension is the only real solution possible.

snebjorn:
> Normally you can do tsx
> 
> ```json
> "dev": "tsx --watch src/index.ts"
> ```
> 
> but i think it wont work for monorepo and i think adding the file extension is the only real solution possible.

The problem I ran into was samples/ and tests/ would use the src ts files from packages instead of the built output, which caused a bunch of other issues.

The goal:
References should work across packages inside packages/
samples/ and tests/ should use the built output, but still trigger breakpoints set in packages/

anymaniax:
Pretty weird that we need to do that no? Other open source libraries follow that pattern and have the same issue?

melloware:
closing in favor of https://github.com/orval-labs/orval/pull/3503

PR #2547, opened 2025-11-13
chore: update tsdown to 0.16.4

snebjorn:


snebjorn:
strange... updating tsdown breaks types on linux

snebjorn:
apparently tsdown doesn't use semver and also don't put breaking changes under breaking changes :(

well I figured it out

melloware (approved):

PR #2517, opened 2025-10-30
chore: make changes to satisfy eslint rules

snebjorn:
Loosened eslint ruleset to make it feasible to fix linting errors

Fixed linting in:
- @orval/angular
- @orval/axios
- @orval/fetch
- @orval/hono
- @orval/mcp
- @orval/swr

Missing:
- @orval/core
- @orval/mock
- @orval/orval
- @orval/query
- @orval/zod

Help wanted 😄

PR #2576, opened 2025-11-18
fix: msw lazy delay

Noyabronok:
https://github.com/orval-labs/orval/pull/2517 broke support for msw lazy delay execution.  
This PR fixes the issue and adds unit tests to avoid similar mistakes in the future


snebjorn:
Absolutely smashing with the tests, nicely done!

However the types could use some love.

image

`mock` doesn't support the inputs used
image

what indeed will `${delay}` print?
image

PR #2594, opened 2025-11-28
feat: custom mutator on mcp

Totorro35:
**Zod:** 
- Fix string escape in Zod
- Inject default enum in place to avoid error

**Mcp:**
- Add option to enable transport config on MCP
- Use client name instead hand-crafted
- Add missing required args on server tool declaration
- Add missing import for custom mutator


melloware:
Assigned to @soartec-lab to review as this is MCP related.

melloware:
@Totorro35 it might be better to submit the ZOD fix separately.

soartec-lab (commented):
@Totorro35 
This is a great improvement. Please separate the PR for the zod changes so we can move forward with the merge process 👍

PR #2606, opened 2025-12-01
feat: mcp options

Totorro35:
- Add option to enable transport config on MCP
- Use client name instead hand-crafted
- Add missing required args on server tool declaration
- Add missing import for custom mutator

Fix https://github.com/orval-labs/orval/pull/2594


melloware:
@Totorro35 where are we with this?

PR #2608, opened 2025-12-01
fix(core): topologically sort schema declarations by dependencies

### LINKED ISSUE #1511: Orval generates some enums that are used before they are declared

1. Open codesandbox link here: https://codesandbox.io/p/devbox/slgr4l
2. run `pnpm run orval` in terminal or start development from terminal GUI in 

I already run that command so you can also look at the generated TS file. I added both "stripped" version of schema yaml and full one. Currently stripped one is used for generation and should be enough but if needed orval.config can be switched to use full version.

## What happens?

Backend is written in Django and Django rest framework so we use drf spectacular for generating open api schema. What happens is, Orval correctly generates all the types but for some reason few enums are used before they are generated. For example on line 1540: `export const SubjectRequestIcon = { ...SubjectIconEnum, ...BlankEnum, ...NullEnum } as const;` , both `BlankEnum` and `NullEnum` are used but they are declared much further in the file (line ~8449 and ~5053). These two are most prevalent since they are used on every model that uses Django's choice filed and can be null in database. There are two more enums that are used before they are declared if it helps: `ProductIconEnum` and `LevelEnum`.

That enumas that are used to construct other enums are generated before them so we don't get TS errors.

There is no logs, errors or anything in CLI - everything runs fine but the generated file has typescript errors:
- `Block-scoped variable 'BlankEnum' used before its declaration.ts(2448)`
- `Variable 'BlankEnum' is used before being assigned.ts(2454)`

## Any other comments?
I attached schema.yaml if it helps and here is orval output config:
```
output: {
    target: './src/features/api/orval/',
    indexFiles: true,
    client: 'fetch',
    mode: 'tags-split',
    clean: true,
    prettier: true,
    baseUrl: 'http://0.0.0.0:8081',
    urlEncodeParameters: true,
    headers: true,
    override: {
        useNamedParameters: true,
        suppressReadonlyModifier: true,
        useTypeOverInterfaces: true,
        contentType: {
            include: [
                'application/json',
                'text/plain',
                'application/x-www-form-urlencoded',
                'multipart/form-data',
            ],
        },
    },
},
```

**Operating System:** MacOS
**Package Version:** 6.31.0
**Browser Version:** Newest Chrome

webjunkie:
Fixes #1511

I ran into this when using orval with a Django REST Framework OpenAPI spec. The generated TypeScript wouldn't compile because enums were declared in the wrong order:

```typescript
export const Combined = { ...BlankEnum } // TS error: used before declaration
// ... hundreds of lines later ...
export const BlankEnum = { '': '' }
```

The issue is that schemas get output in alphabetical order, but combined enums using spread syntax need their dependencies declared first.

I added dependency tracking through the resolvers and a topological sort before output. Added a test that reproduces the original issue.

### DIFF
--- a/packages/core/src/generators/schema-definition.test.ts
+++ b/packages/core/src/generators/schema-definition.test.ts
@@ -200,7 +200,11 @@ describe('generateSchemasDefinition', () => {
 
     // Verify dependencies come before the schema that spreads them
     const order = result.map((schema) => schema.name);
-    expect(order.indexOf('ZzzFirst')).toBeLessThan(order.indexOf('AaaCombined'));
-    expect(order.indexOf('ZzzSecond')).toBeLessThan(order.indexOf('AaaCombined'));
+    expect(order.indexOf('ZzzFirst')).toBeLessThan(
+      order.indexOf('AaaCombined'),
+    );
+    expect(order.indexOf('ZzzSecond')).toBeLessThan(
+      order.indexOf('AaaCombined'),
+    );
   });
 });
--- a/packages/core/src/generators/schema-definition.ts
+++ b/packages/core/src/generators/schema-definition.ts
@@ -168,7 +168,9 @@ export const generateSchemasDefinition = (
   return sortSchemasByDependencies(deduplicatedModels);
 };
 
-const sortSchemasByDependencies = (schemas: GeneratorSchema[]): GeneratorSchema[] => {
+const sortSchemasByDependencies = (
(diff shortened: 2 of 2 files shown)


melloware:
you might have to run `yarn update-samples` script.

webjunkie:
Ah, thanks! It said unchanged previously, but that is nonsense. Had to clean caches to get it updated.

melloware:
ok now `yarn format` :)

PR #2614, opened 2025-12-01
Upgrade openapi to 3.1

snebjorn:
**Major refactor changes**

OpenAPI parsers, validators, loaders, upgraders, etc have all been switched to https://github.com/scalar/scalar
Input specs are upgraded to OpenAPI 3.1 and all functions get served OpenAPI 3.1 so no longer a mix of 3.0 and 3.1

All specs and schemas are no longer passed to functions along a spec key so they needed to pick the correct schema. Instead only the relevant spec and schema is passed to functions. Hopefully this makes things easier.
Example: 
```ts
// before
specs.reduce(unreadable code...
context.specs[specKey]
// now
spec
context.spec
```

> [!NOTE]
> `SchemaObject` is union-typed with a boolean, this is because the OpenAPI spec allows for this
> > The empty schema (which allows any instance to validate) MAY be represented by the boolean value true and a schema which allows no instance to validate MAY be represented by the boolean value false.
https://swagger.io/specification/
>
> However this screws up the type. So in order to fix this a type-guard excluding the boolean must be used, and also handle the boolean case. Pretty annoying - but that's in spec.
> 
> A lot of types can be either "the thing" or a reference to a thing. We have `isReference` (and still do) but now we also have `{ isDereferenced } from '@scalar/openapi-types/helpers'` which is the reverse, but gives better types then `!isReference`
>
> Nullable types are now harder to "detected" thanks to OpenAPI 3.1 
> ```
> # OpenAPI v3.0
> type: string
> nullable: true
> 
> # OpenAPI v3.1
> type:
> - "string"
> - "null" 
> ```
> https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0
>
> We need to do something better then what I did, because null is now in the type prop and it's an array and... and... 
> `const isNullable = isArray(schema.type) && schema.type.includes('null')`

> [!CAUTION]
> **BREAKING CHANGES**
> - input.validation is gone. input spec is now always validated, as well as transformer output.
> - input.parserOptions is gone. New parser doesn't take swagger options.
> - input.converterOptions is gone.
> - `InputFiltersOption` renamed `InputFiltersOptions`
> - generated filenames and models may change name from prior versions
> 
> Plus I expect some regressions

**FEATURES**
- cli argument `--project` now accepts a list of projects. Example: `--project dogstore catstore`

### DIFF
--- a/docs/src/pages/reference/cli.mdx
+++ b/docs/src/pages/reference/cli.mdx
@@ -43,10 +43,11 @@ $ orval --config ./api/orval.config.js
 
 ### Project
 
-The `--project` option, shorthand `-p`, can be used to focus on one project of your Orval config.
+The `--project` option, shorthand `-p`, can be used to focus on projects of your Orval config.
 
 ```bash
 $ orval --project petstore
+$ orval --project dogstore catstore
 ```
 
 ### Watch


melloware:
Nice looks like all tests are passing!

snebjorn:
Yep, hopefully this doesn't break stuff I'm not aware of :)

melloware (approved):

melloware:
Well we will find out I will do another RC release so we can test.

PR #2617, opened 2025-12-02
#2616 fix(fetch): Fix client generation when params contain dates but exploded params do not.

### LINKED ISSUE #2616: Fetch client generates invalid typescript when params contain dates, but exploded params don't
When the fetch client is generated, if the `params` object contains parameters with a `date-time` format, but the params that are in exploded params do not, then the value of `hasDateParams` is `true`, but there are no date params in the explodeParams.  This means that the generated code looks like this:
```
        if (Array.isArray(value) && explodeParameters.includes(key)) {
            value.forEach((v) => {
                normalizedParams.append(key, v === null ? "null" : v instanceof Date ? v.toISOString() : v.toString());
            });
            return;
        }
```
Which results in a typescript error:  `The left-hand side of an 'instanceof' expression must be of type 'any', an object type or a type parameter.`

jlillywhite:
Fix #2616.  
- Adds a separate check to see if the explodedParams contains date-time params. 
- Also fixes a typo in the name of the `normalParamsImplementation`

### DIFF
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -74,9 +74,9 @@ export const generateRequestFunction = (
 
     return schema.name;
   });
-  const hasDateParams =
+  const hasExplodedDateParams =
     context.output.override.useDates &&
-    parameters.some(
+    explodeParameters.some(
       (p) =>
         'schema' in p &&
         p.schema &&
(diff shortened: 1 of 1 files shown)


melloware:
@soartec-lab one for you to review

Issue #1644, opened 2024-09-27
Orval overuses `AnyOf` types delimiter


jaxxreal:
## What are the steps to reproduce this issue?

1. Generate orval client with the following config:
```
export default defineConfig({
  apiClient: {
    input: {
      target: 'http://localhost:3001/swagger-json',
    },
    output: {
      target: './src/client/orval.ts',
      schemas: './src/orval-types',
      mode: 'single',
      client: 'swr',
      override: {
        mutator: {
          path: './src/client/httpClient.ts',
          name: 'makeRequest',
        },
        useNamedParameters: true
      },
    },
    hooks: {
      afterAllFilesWrite: 'eslint --fix',
    },
  },
});
```

## What happens?

Look at the types generated. Names like `ListOfActivitiesResponseAnyOfDataAnyOfItemAnyOfAttributesAnyOfDataAnyOfItemAnyOfTargetAnyOfId` are pretty common. OpenAPI schema for the given example has NO `anyOf` conditions.

## What were you expecting to happen?

If an OpenAPI spec has no `anyOf` condition I don't expect to see them on my types. So the expected type name would be:
`ListOfActivitiesResponseDataItemAttributesDataItemTargetId`

## Any logs, error output, etc?

-

## Any other comments?

It's an issue because nearly every single generated file has a decent amount of `AnyOf`.

I know that the project is looking for contributors, so I would appreciate minimal guidance on where to start my search for a solution. Maybe it's a configuration issue? Please let me know.

## What versions are you using?

`orval@6.28.2`

Please execute `npx envinfo --system --npmPackages orval,zod,axios,msw,swr,@tanstack/react-query,@tanstack/vue-query,react,vue` and paste the results here.

```
System:
    OS: macOS 14.6.1
    CPU: (12) arm64 Apple M3 Pro
    Memory: 94.64 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
```

PR #2654, opened 2025-12-16
inline combined types by default, add `useCombinedTypeAliases` opt-out

superxiao:
**Breaking change**: Inlines combined types by default, cleaning up generated types by removing unnecessary intermediate type aliases.

Helps with #1644 - eliminates verbose names like `ListOfActivitiesResponseAnyOfDataAnyOfItem...`.

```typescript
// orval.config.ts - to restore old behavior
output: {
  override: {
    useCombinedTypeAliases: true,
  },
},
```

Estimated impact on a real project:

| Metric | Before | After | Change |
|--------|--------|-------|--------|
| Types | 1,549 | ~252 | -84% |
| Lines | 7,156 | ~5,859 | -18% |

Fewer intermediate types means better readability and potentially faster type inference.

The change handles two cases:

**1. Intermediate object types** - things like `ResponseAnyOf`, `ResponseOneOfTwo` go away:

```typescript
// before
export type Response401OneOf = { defined: true; ... };
export type Response401OneOfTwo = { defined: false; ... };
export type Response401 = Response401OneOf | Response401OneOfTwo;

// after
export type Response401 = { defined: true; ... } | { defined: false; ... };
```

**2. Primitive union properties** - explicit `anyOf`/`oneOf` with primitives get inlined:

```typescript
// before (anyOf: [{type: string}, {type: number}])
export type MyObjectField = string | number;
export interface MyObject {
  field?: MyObjectField;
}

// after
export interface MyObject {
  field?: string | number;
}
```

Set `useCombinedTypeAliases: true` to restore the old behavior of creating intermediate type aliases.

Property-level inline objects still get named types (complex types benefit from a name anyway):

```typescript
// anyOf with inline objects at property level - still creates named type
export type MyObjectResult = { data?: string } | { error?: string };
export interface MyObject {
  result?: MyObjectResult;
}
```

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -630,6 +630,7 @@ export type NormalizedFetchOptions = {
   forceSuccessResponse: boolean;
   jsonReviver?: Mutator;
   useZodSchemaResponse: boolean;
+  runtimeValidation: boolean;
 };
 
 export type FetchOptions = {
@@ -637,6 +638,7 @@ export type FetchOptions = {
   forceSuccessResponse?: boolean;
   jsonReviver?: Mutator;
   useZodSchemaResponse?: boolean;
+  runtimeValidation?: boolean;
 };
 
 export type InputTransformerFn = (spec: OpenApiDocument) => OpenApiDocument;
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -20,6 +20,16 @@ import {
 } from '@orval/core';
 import { isDereferenced } from '@scalar/openapi-types/helpers';
 
+const WILDCARD_STATUS_CODE_REGEX = /^[1-5]XX$/i;
+
+const getStatusCodeType = (key: string): string => {
+  if (WILDCARD_STATUS_CODE_REGEX.test(key)) {
+    const prefix = key[0];
+    return `HTTPStatusCode${prefix}xx`;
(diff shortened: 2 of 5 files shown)


superxiao:
As the real world example in the PR description shows, such an option could cut down some excessive single-use types for a project I'm working on (where the generated file is still growing fast). Let me know if you think this is a good idea or not.

melloware:
@anymaniax @soartec-lab what do you think?  I am all for less generated code...

soartec-lab:
@superxiao @melloware 
I agree with this. I don't find many cases where I want to refer to polymorphic types, so it might be a good idea to make this behavior the default and separate types optional. If that's the case, V8 is the right time.

melloware:
agreed this should be the default.

melloware:
@superxiao can you make this the default `true` and make them opt out of it with `false`

superxiao:
PR has been updated to make the inlining default behavior. New option is renamed from `inlineCombinedTypes` to `useCombinedTypeAliases` - `false` and `undefined` is the default.

melloware:
This is fantastic!

soartec-lab:
@superxiao 
Thanks for great improvement. Could you also make a PR to document it?

superxiao:
https://github.com/orval-labs/orval/pull/2657

PR #2662, opened 2025-12-21
test(core): add wildcard content type tests and response coverage

superxiao:
- Test wildcard */* media type generates Blob for request/response
- Test binary media types produce Blob, JSON ignores contentMediaType
- Add wildcard to isBinaryContentType test
- Add wildcardFile field to comprehensive FormData test
- Test both request and response for core type generation rules


superxiao:
This is a test only change. Some OpenAPI generators (e.g. orpc) generates wildcard media key and `contentMediaType` from `z.file()` with no mimetype  declaration - so I think having it explicitly tested is useful. File/Blob in response is also added in tests.

Issue #2243, opened 2025-07-27
Angular: Different response types causes issues


snebjorn:
Given this openapi spec

```yml
paths:
  /my-endpoint:
    get:
      responses:
        "200":
          description: Expected response to a valid request
          content:
            text/plain:
              schema:
                type: string
            application/xml:
              schema:
                $ref: "#/components/schemas/Pet"
            application/json:
              schema:
                $ref: "#/components/schemas/Pet"

```

Generates for Angular (not sure about the others)
```ts
getMyEndpoint<TData = string | Pet>(...) {
  return this.http.get<TData>(`/my-endpoint`, options);
}
```
This is incorrect. The "string" return will be treated as a json-string which is not what the openapi defined.
At least for Angular's HttpClient "text/plain" and "application/json" cannot be mixed in the same `http.get` call as it needs either a `responseType: "json"` (default) or `responseType: "text"` option.

Furthermore the mocks generated also break. I'm getting the below. Which stops importing `Pet` even though it's still used in the file. Not sure what is going on here.

```diff
diff --git a/samples/angular-app/src/api/endpoints/pets/pets.msw.ts b/samples/angular-app/src/api/endpoints/pets/pets.msw.ts
index 97626236..c702c8dd 100644
--- a/samples/angular-app/src/api/endpoints/pets/pets.msw.ts
+++ b/samples/angular-app/src/api/endpoints/pets/pets.msw.ts
@@ -8,7 +8,7 @@ import { faker } from '@faker-js/faker';
 
 import { HttpResponse, delay, http } from 'msw';
 
-import type { Pet, Pets } from '../../model';
+import type { Pets } from '../../model';
 
 export const getListPetsResponseMock = (): Pets =>
   Array.from(
@@ -77,23 +77,22 @@ export const getCreatePetsMockHandler = (
 
 export const getShowPetByIdMockHandler = (
   overrideResponse?:
+    | string
     | Pet
     | ((
         info: Parameters<Parameters<typeof http.get>[1]>[0],
-      ) => Promise<Pet> | Pet),
+      ) => Promise<string | Pet> | string | Pet),
 ) => {
   return http.get('*/v:version/pets/:petId', async (info) => {
     await delay(1000);
 
     return new HttpResponse(
-      JSON.stringify(
-        overrideResponse !== undefined
-          ? typeof overrideResponse === 'function'
-            ? await overrideResponse(info)
-            : overrideResponse
-          : getShowPetByIdResponseMock(),
-      ),
[...]

PR #2674, opened 2025-12-23
fix(mock,angular): Different response types causes issues

ScriptType:
## Summary
DRAFT

Fixes_partially #2243 - Angular client now properly handles OpenAPI endpoints with multiple response content types using function overloading.

## Problem

When an OpenAPI endpoint defines multiple response content types (e.g., `application/json`, `text/plain`, `application/xml`), the Angular client generated a single method with a union return type but no way to specify which content type to request. Angular's `HttpClient` requires `responseType` to be set correctly at request time, causing runtime failures when non-JSON responses were returned.

Additionally, MSW mock imports were incorrectly dropped when `text/plain` was present among content types.

### Angular Generator (`@orval/angular`)
- Generates **function overloads** for endpoints with multiple content types
- Each overload accepts an `accept` parameter typed to a specific content type literal
- Runtime `responseType` is dynamically set based on the `Accept` header:
  - `json` for JSON content types
  - `text` for text/* and XML content types
  - `blob` for binary content types
- Default content type prioritizes `application/json` > any `+json` type > first in list

### MSW Mock Generator (`@orval/mock`)
- Fixed import filtering to preserve model type imports when `text/plain` is among content types

### Core Utilities (`@orval/core`)
- Added `getDefaultContentType()` utility for consistent default selection across generators

## Generated Code Example

```typescript
// Before: No way to request specific content type
showPetById(petId: string): Observable<string | Pet>

// After: Type-safe overloads with content negotiation
showPetById(petId: string, accept: 'text/plain'): Observable<string>;
showPetById(petId: string, accept: 'application/json'): Observable<Pet>;
showPetById(petId: string, accept?: string): Observable<Pet>; // defaults to JSON
```

## Breaking Changes

None. This is additive - existing code using singular content type endpoints works unchanged.

## Future Work
( I can try also to do this if you want )
- [ ] Extend function overloading to Axios generator (`@orval/axios`)
- [ ] Extend function overloading to Fetch generator (`@orval/fetch`)
- [ ] Update MSW mock generator to dynamically respond based on `Accept` header

## Testing

- Angular package builds successfully
- Sample `angular-app` regenerated and builds
- Manual testing confirms correct `Accept` headers and response parsing


melloware:
PRs for fixing Axios and fetch and others would be awesome!

ScriptType:
The thing is i don't know if there are scenarios where it is breaking for fetch/axios.
Angular Http didn't worked with union types so it is pretty safe to do this, but if fetch/axios works but maybe not 100% correct it could break code when it now defaults to json instead for example text like it did before.
Someone needs to look at it and decides how to handle it there.

melloware:
No worries this fix is great for Angular.

ScriptType:
@melloware Than i would say i make it open for Review (atleast when this is the route we want to go).

We still need to create Issues for the other generators (if you think that we need it there). :)

melloware (approved):

melloware:
@snebjorn if you want to review

melloware:
I am going to merge this so it can be in RC5

PR #2687, opened 2025-12-25
React Query: Introduce `react-compiler` support

theisgroenbech:
First of all thank you for a super nice library! 🥳 

The current output for `React Query` is does not adhere to the rules of React and thereby breaks `react-compiler` optimisations.

```
Mutating a value returned from a function whose return value should not be mutated (react-compiler/react-compiler)
```

The issue is setting the `queryKey` of `query` here:

```
  const query = useQuery(queryOptions, queryClient) as UseQueryResult<TData, TError> & {
    queryKey: DataTag<QueryKey, TData, TError>
  }

  query.queryKey = queryOptions.queryKey

  return query
```

The fix is using spreading instead, which does not break the rules of `React` and thereby introduces support for React Compiler.

The fixed output looks like this:

```
  const query = useQuery(queryOptions, queryClient)

  return { ...query, queryKey: queryOptions.queryKey } as UseQueryResult<
    TData,
    TError
  > & { queryKey: DataTag<QueryKey, TData, TError> }
}
```

### DIFF
--- a/packages/query/src/index.ts
+++ b/packages/query/src/index.ts
@@ -960,6 +960,45 @@ const generatePrefetch = ({
   }
 };
 
+const getQueryReturnStatement = ({
+  outputClient,
+  hasSvelteQueryV4,
+  hasSvelteQueryV6,
+  hasQueryV5,
+  hasQueryV5WithDataTagError,
+  queryResultVarName,
+  queryOptionsVarName,
+}: {
+  outputClient: OutputClient | OutputClientFunc;
+  hasSvelteQueryV4: boolean;
+  hasSvelteQueryV6: boolean;
+  hasQueryV5: boolean;
+  hasQueryV5WithDataTagError: boolean;
+  queryResultVarName: string;
+  queryOptionsVarName: string;
+}): string => {
+  if (isAngular(outputClient)) {
+    return `return ${queryResultVarName};`;
+  }
+
+  if (isVue(outputClient)) {
+    const queryKeyType = hasQueryV5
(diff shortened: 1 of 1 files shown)


melloware:
Looks like you have to run yarn update samples

theisgroenbech:
> Looks like you have to run yarn update samples

Thanks @melloware! I have updated the samples 🥳

packages/query/src/index.ts
```diff
@@ -1363,15 +1363,17 @@ export function ${queryHookName}<TData = ${TData}, TError = ${errorType}>(\n ${q
...
+        ? `${queryResultVarName}.queryKey = unref(${queryOptionsVarName}).queryKey as ${hasQueryV5 ? `DataTag<QueryKey, TData${hasQueryV5WithDataTagError ? ', TError' : ''}>` : 'QueryKey'};
+
+  return ${queryResultVarName};`
+        : hasSvelteQueryV4 || hasSvelteQueryV6
+          ? `${queryResultVarName}.queryKey = ${queryOptionsVarName}.queryKey;
 
-  return ${queryResultVarName};
+  return ${queryResultVarName};`
+          : `return { ...${queryResultVarName}, queryKey: ${queryOptionsVarName}.queryKey };`
+  }
```
theisgroenbech: I can also move this out into a helper function if you prefer that?

melloware:
Running tests  now.

melloware:
Maybe a helper function would be better

theisgroenbech:
@melloware moved into helper!

Issue #2556, opened 2025-11-14
improve `@orval/query` code quality


snebjorn:
1. Make `@orval/query` pass linting.
2. Add `@orval/query` to the linting CI check

melloware:
this should be a little easier after @ScriptType PR to refactor this as each file can be tackled individually.

PR #2413, opened 2025-09-30
feat: add support for Svelte Query v6 compatibility 

jacokok:
Add compatibility to @tanstack/svelte-query v6.

https://tanstack.com/query/latest/docs/framework/svelte/migrate-from-v5-to-v6

> @tanstack/svelte-query v6 adapter fully migrates to the runes syntax.
> Like the Angular and Solid adapters, most functions for the Svelte adapter now require options to be provided as a "thunk" (() => options) to provide reactivity.

This change will check if you have the newer v6 package and use the thunk options. See examples below of output of v5 vs v6.

### v5
```ts
// Mutation
return createMutation(mutationOptions, queryClient);

// Query
const query = createQuery(queryOptions, queryClient) as CreateQueryResult<TData, TError> 
& { queryKey: DataTag<QueryKey, TData, TError>; };
```

### v6
```ts
// Mutation
return createMutation(() => ({ ...mutationOptions, queryClient }));

// Query
const query = createQuery(() => ({ ...queryOptions, queryClient })) as CreateQueryResult<TData, TError> 
& { queryKey: DataTag<QueryKey, TData, TError> };
```

Issue #1187, opened 2024-02-01
Support for Svelte Query v5


ericpgreen2:
Are there any plans to support Svelte Query v5?

One change is that Svelte Query v5 has a [new model for reactivity](https://tanstack.com/query/v5/docs/framework/svelte/reactivity).

Orval has been very valuable for my team, so thank you for your work!

melloware:
PRs are welcome!

melloware:
This was implemented here: https://github.com/orval-labs/orval/pull/2413

PR #2691, opened 2025-12-30
refactor(query): split index.ts into multiple Files

ScriptType:
This PR refactors the `@orval/query` package by splitting the large `index.ts` file (2042 lines) into smaller, focused files for better maintainability.

### New Files Created

| File | Lines | Purpose |
|------|-------|---------|
| `dependencies.ts` | 475 | Framework dependencies (React/Vue/Svelte/Angular) + version detection utilities |
| `query-options.ts` | 215 | `QueryType`, `generateQueryOptions`, `generateQueryArguments`, `getQueryOptionsDefinition` |
| `return-types.ts` | 172 | `generateQueryReturnType`, `generateMutatorReturnType`, `getQueryFnArguments`, `getQueryReturnStatement` |
| `mutation-generator.ts` | 247 | `generateMutationHook` (extracted from `generateQueryHook`) |
| `query-generator.ts` | 971 | `getFrameworkPrefix`, `generatePrefetch`, `generateQueryImplementation`, `generateQueryHook` |

### Updated Files

- `index.ts`: Reduced from 2042 to 122 lines (re-exports + builder)

## Benefits

- **Better maintainability**: Each File has a single responsibility
- **Easier navigation**: Developers can quickly find relevant code
- **Clean separation**: Query and mutation generation are now in separate files

## Breaking
For PR´s which are not merged right now this change will be Breaking !

### DIFF
--- a/packages/query/src/query-options.ts
+++ b/packages/query/src/query-options.ts
@@ -48,8 +48,8 @@ export const generateQueryOptions = ({
         omitBy(
           options,
           (_, key) =>
-            (type !== QueryType.INFINITE ||
-              type !== QueryType.SUSPENSE_INFINITE) &&
+            type !== QueryType.INFINITE &&
+            type !== QueryType.SUSPENSE_INFINITE &&
             INFINITE_QUERY_PROPERTIES.has(key),
         ),
       )?.slice(1, -1)}`


melloware:
I actually think this is pretty clean @anymaniax @soartec-lab

melloware:
@ScriptType want to tackle this one as part of this PR?  https://github.com/orval-labs/orval/issues/2556

ScriptType:
sure

packages/query/src/return-types.ts
```diff
@@ -0,0 +1,172 @@
...
+    return '{ pageParam }';
+  }
+
+  return '{ signal }';
+};
+
+export const getQueryReturnStatement = ({
+  outputClient,
+  hasSvelteQueryV4,
+  hasSvelteQueryV6,
```
AllieJonsson, line 137: V5?

ScriptType: I have also no clue, what is up with the V6. (i think tanstack released v6 for svelte which is runebased)
But it is like this in master.

melloware: V5 and V6 were never added see this ticket: https://github.com/orval-labs/orval/issues/1187

melloware: So that v6 should be removed

ScriptType: https://github.com/orval-labs/orval/pull/2413
It got introduced here?

melloware: Weird that v5 was skipped?

AllieJonsson: Yeah, feels a bit weird we have v4 and v6 support only, but it seems like v6 is the one to use when you use svelte5, so i guess its fine

ScriptType:
@melloware done

Sneaked in also a fix

query-option.ts line 46
This is always true as QueryType never can be Infinite or Suspense Infinite, which resulted in getNextPageParam and getPreviousPageParam allways being omitted
```typescript
  const queryConfig = isObject(options)
    ? ` ${stringify(
        omitBy(
          options,
          (_, key) =>
            (type !== QueryType.INFINITE ||
              type !== QueryType.SUSPENSE_INFINITE) &&
            INFINITE_QUERY_PROPERTIES.has(key),
        ),
      )?.slice(1, -1)}`
    : '';
```

melloware:
Fantastic. I will link the other ticket

melloware:
@AllieJonsson @ScriptType technically is https://github.com/orval-labs/orval/issues/1187 implemented then?  He talks about some new reactivity model in Svelte 5?

melloware:
Nevermind it absolutely was implemented so I closed that ticket.

ScriptType:
@melloware 
Yes it is implemented to my understanding.
Just read a bit about the v6 thing, it seems v5 was pretty buggy on tanstack thats why they switched to "v6" fully utilizing the svelte runes (svelte signals basically).

Is there anything still needed in my PR ?

PR #2703, opened 2026-01-06
fix(core): mark const enums as value imports

### LINKED ISSUE #2624: Incorrect `import type` and value space reference in const+typeof generation for schemas with   both enum and const
When a schema property has both `enum` and `const` fields, and the schema contains only properties
  with `const` values, orval generates incorrect TypeScript code with two issues:

  1. **Wrong import statement**: Uses `import type` instead of `import` for enum constants used as
  values
  2. **Type/value space confusion**: References enum type name as a value in const object

```typescript
import type { AType } from './aType';

export const AValue = {
  type: AType, // 'AType' cannot be used as a value because it was imported using 'import type'.
} as const;
export type A = typeof AValue;
```

  ## Reproduction

  See minimal reproduction example: **https://github.com/ioslh/temp-orval-generate**

- [openapi.json](https://github.com/ioslh/temp-orval-generate/blob/master/openapi.json)
- [orval config](https://github.com/ioslh/temp-orval-generate/blob/master/orval.config.js)
- the generated type with import error: https://github.com/ioslh/temp-orval-generate/blob/master/src/model/a.ts

You can clone the repo and run `npm run generate` to see the generated result.

froggy1014:
## Summary
  - Use value imports for const + enum object properties to avoid TS “import type” value-space errors.
  - Ensure const literals are escaped and reuse the enum value directly instead of aliasing a type.

  ## Changes

 - packages/core/src/getters/object.ts
     - Detect const properties that reference enums
     - Mark imports as value imports (isConstant)
     - Skip alias generation when a const literal is present
     - Escape string const values

  Fixes #2624

  ## Before
  
  ```ts
  import type { AType } from './aType';

  export const AValue = {
    type: AType, // 'AType' cannot be used as a value because it was imported using 'import type'.
  } as const;
  ```

  ## After

  ```ts
  import { AType } from './aType';

  export const AValue = {
    type: AType,
  } as const;

### DIFF
--- a/packages/core/src/generators/interface.test.ts
+++ b/packages/core/src/generators/interface.test.ts
@@ -87,6 +87,7 @@ export type TestSchema = typeof TestSchemaValue;
     expect(got).toEqual(want);
   });
 
+  // With enumGenerationType: const - mimic default enum output
   it('should inline const literal when enum + const are both present', () => {
     const schema: OpenApiSchemaObject = {
       type: 'object',
@@ -129,6 +130,7 @@ export type ConstEnum = typeof ConstEnumValue;
     expect(got).toEqual(want);
   });
 
+  // With enumGenerationType: const - keep referenced enums type-only
   it('should use type-only imports for referenced schemas in interfaces', () => {
     const schema: OpenApiSchemaObject = {
       type: 'object',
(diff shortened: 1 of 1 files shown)


melloware:
looks like you might haev to run `yarn update-samples`

AllieJonsson (approved):

froggy1014:
> looks like you might haev to run `yarn update-samples`

@melloware 

it runs fine locally. if you don't mind could we re-run test job ?

melloware:
hmmm this is failing `yarn update-samples` ?

melloware:
it shows files have changed so the build fails

froggy1014:
@melloware 

I've add some test code for cases and code changes to meet the other test - !

melloware:
runnin build now

Issue #2171, opened 2025-06-18
Nullable Enum Fields on v3.1 OpenAPI Spec Generated as Union Types Instead of Enum Objects


lukemorales:
# The Problem
When generating TypeScript types from v3.1 OpenAPI schemas that contain nullable enum fields, orval currently generates them as simple union types instead of proper enum objects with their corresponding const declarations.

## Current behavior:
```ts
export type GetAppSettings200Currency = "USD" | "CAD" | null;
```

## Expected behavior:
The generator should create both the type and const object declaration:
```ts
 export type GetAppSettings200Currency =
  | (typeof GetAppSettings200Currency)[keyof typeof GetAppSettings200Currency]
  | null;

// eslint-disable-next-line @typescript-eslint/no-redeclare
export const GetAppSettings200Currency = {
  USD: "USD",
  CAD: "CAD",
} as const;
```

## Reproduction
Use the following yaml in the orval playground:
```yaml
openapi: 3.1.0
info:
  title: Reproduction
  description: Testing
  version: 0.0.1
servers:
- url: http://localhost:3002
  description: Local
paths:
  "/app-settings":
    get:
      responses:
        '200':
          description: The App Settings
          content:
            application/json:
              schema:
                type: object
                properties:
                  currency:
                    type:
                    - string
                    - 'null'
                    enum:
                    - USD
                    - CAD
                    -
                required:
                - currency
      operationId: getAppSettings
      tags:
      - app settings
      - public
      parameters: []
      description: Get App Settings
      summary: getAppSettings
components:
  schemas: {}
```

## Possible fix?

I did a patch in `@orval/core` in our repo with the following changes, but I'm not super confident this is an acceptable fix since it worked well for our use case, but I lack knowledge of the specs you should follow and not sure it supports other use-cases:

```diff
diff --git a/dist/index.js b/dist/index.js
index 916f026ba4ea25c094bddbf3bf96000e831148b6..5bbe42764e4ae5532e83d67e5ffe239149488b58 100644
--- a/dist/index.js
+++ b/dist/index.js
@@ -50569,7 +50569,7 @@ var getScalar = ({
 }) => {
   const isAngularClient = context.output.client === OutputClient.ANGULAR;
   const nullable = item.nullable && !isAngularClient ? " | null" : "";
-  const enumItems = item.enum?.filter((enumItem) => enumItem !== null);
+  const enumItems = item.enum;
   if (!item.type && item.items) {
     item.type = "array";
   }
[...]

PR #2714, opened 2026-01-09
chore: remove eslint comment for enum generation

### LINKED ISSUE #2459: Same Naming & Eslint comment
```ts
export type InitialStepRequestStepKind =
  (typeof InitialStepRequestStepKind)[keyof typeof InitialStepRequestStepKind];

// eslint-disable-next-line @typescript-eslint/no-redeclare
export const InitialStepRequestStepKind = {
  FORM: "FORM",
  CORE_ACTION: "CORE_ACTION",
  MANUAL: "MANUAL",
  OFFER: "OFFER",
  ONBOARDING: "ONBOARDING",
} as const;
```

I’m using Biome instead of ESLint, so the generated code shouldn’t include any ESLint-specific comments.
Also, having both the type and const share the same name causes import conflicts — they should be named differently or made configurable.

```js
  coreApi: {
    input: "...",
    output: {
      mode: "tags-split",
      target: "./clients",
      schemas: "./schemas",
      client: "axios",
      workspace: "./core",
      indexFiles: true,
      biome: true,
      override: {
        header: false,
        useTypeOverInterfaces: true,
        suppressReadonlyModifier: true,
        mutator: {
          path: "./custom-instance.ts",
          name: "customInstanceCore",
        },
      },
      clean: true,
    },
    hooks: {
      afterAllFilesWrite: ["bun run lint:unsafe"],
    },
  },
```

froggy1014:
Fixes #2459

### DIFF
--- a/packages/core/src/generators/interface.test.ts
+++ b/packages/core/src/generators/interface.test.ts
@@ -110,7 +110,7 @@ export type TestSchema = typeof TestSchemaValue;
       {
         name: 'ConstEnumKind',
         model:
-          "export type ConstEnumKind = typeof ConstEnumKind[keyof typeof ConstEnumKind];\n\n\n// eslint-disable-next-line @typescript-eslint/no-redeclare\nexport const ConstEnumKind = {\n  A: 'A',\n} as const;\n",
+          "export type ConstEnumKind = typeof ConstEnumKind[keyof typeof ConstEnumKind];\n\n\nexport const ConstEnumKind = {\n  A: 'A',\n} as const;\n",
         imports: [],
         dependencies: [],
       },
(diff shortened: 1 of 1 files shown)


melloware:
@froggy1014 can you check this one i want to see what happens 

https://github.com/orval-labs/orval/issues/2171

melloware:
what does the no-redeclare prevent ESLINT from not liking the fact that PetCountry is a const and a type?

```
export type PetCountry = (typeof PetCountry)[keyof typeof PetCountry];

// eslint-disable-next-line @typescript-eslint/no-redeclare
export const PetCountry = {
```

froggy1014:
@melloware 

I realized later that these rules aren't applied in the project. so I didn;t notice it until now.. 



```js
'no-redeclare': 'off', // eslint-rule ( js )
'@typescript-eslint/no-redeclare': 'error', // eslint-rule ( ts )
```

ig previous PR did not cover issue this one, my bad.

#2459

but this code is fine in TS world, but JS 

```ts
export type GetAppSettings200Currency = typeof GetAppSettings200Currency[keyof typeof GetAppSettings200Currency] | null;

export const GetAppSettings200Currency = {
  USD: 'USD',
  CAD: 'CAD',
} as const;

```

melloware:
so what should we do?   should i merge this PR?

froggy1014:
wait then it should be fine then. if there is any problem with this one.

user should 'off' the re-declare rule on their project

melloware (approved):

froggy1014:
@melloware

Just to clarify — `@typescript-eslint/no-redeclare` is off by default.
However, it is included in the `typescript-eslint/all` preset, which enables all @typescript-eslint rules as errors.

In that setup, users might run into lint errors with the generated files.
In that case, they would need to either add those paths to eslintignore or disable the @typescript-eslint/no-redeclare rule.

As far as I know, most projects tend to use the recommended preset rather than all.

That said, it might also be worth considering using a different Prefix or Suffix in the next major version to avoid this kind of confusion.

melloware:
I agree with the Biome user that this shouldn't have been in the generated code either way.  Let people configure their ESLint or Biome the way they want.

PR #2715, opened 2026-01-10
fix(core): add type guards for 'in' operator on schema objects (#2540)

### LINKED ISSUE #2540: Incorrect import path when referencing external YAML file schemas
### Environment
- **Orval Version**: 7.16.0
- **Node.js Version**: 22.18.0
- **Package Manager**: yarn
- **Client**: react-query

### Description
When using `$ref` to reference a schema defined in an external YAML file, Orval generates an incorrect import path in the TypeScript output. The import statement references the external YAML filename instead of the actual generated schema file.

### Configuration
```typescript

export default defineConfig({
  api: {
    input: {
      target: "./openapi/main.yaml",
    },
    output: {
      target: "./src/apis",
      schemas: "./src/schemas",
      client: "react-query",
      mode: "tags-split",
      clean: true,
      prettier: true,
      mock: false,
      override: {
        mutator: {
          path: "./src/utils/CustomAxiosInstance.ts",
          name: "customInstance",
        },
        query: {
          signal: false,
        },
      },
    },
  },
});
```

### Reproduction Steps

**1. Create external schema file (`common_schemas.yaml`):**
```yaml
components:
  schemas:
    InternalServerError_500:
      type: object
      properties:
        code:
          type: integer
          example: 500
        message:
          type: string
          example: Internal server error
```

**2. Reference the external schema in endpoint definition (`users/endpoints.yaml`):**
```yaml
post:
  operationId: RegisterUser
  summary: Register a new user
  tags:
    - user
  requestBody:
    required: true
    content:
      application/json:
        schema:
          type: object
          properties:
            name:
              type: string
  responses:
    200:
      description: OK
      content:
        application/json:
          schema:
            type: object
            properties:
              message:
                type: string
    500:
      description: Internal Server Error
      content:
        application/json:
          schema:
            $ref: "../common_schemas.yaml#/components/schemas/InternalServerError_500"
```

**3. Run Orval:**
```bash
yarn orval
```

### Expected Behavior

Generated import statement should reference the actual generated schema file:
```typescript
import type { InternalServerError500 } from "../../schemas/internalServerError500";
```

### Actual Behavior

Generated import statement references the external YAML filename:
```typescript
import type { InternalServerError500 } from "../../schemas/common_schemas";
```

However, the actual generated file is:
```
src/schemas/internalServerError500.ts
```

This causes a TypeScript compilation error:
```
Cannot find module '../../schemas/common_schemas' or its corresponding type declarations.
```

### Generated Files

**Generated API file (`src/apis/user/user.ts`):**
```typescript
/**
 * Generated by orval v7.16.0 🍺
 * Do not edit manually.
 */
import type { InternalServerError500 } from "../../schemas/common_schemas"; // ❌ Wrong path
// ... rest of the file
```

**Generated schema file (`src/schemas/internalServerError500.ts`):**
```typescript
/**
 * Generated by orval v7.16.0 🍺
 * Do not edit manually.
 */

export interface InternalServerError500 {
  code?: number;
  message?: string;
}
```

### Additional Context

- The schema file is correctly generated with the proper naming convention (camelCase)
- The issue only occurs when using external file references via `$ref: "../file.yaml#/..."`
- Internal references within the same file (e.g., `$ref: "#/components/schemas/..."`) work correctly
- This appears to be a regression or edge case in how Orval resolves external file references

### Workaround

Currently, the only workaround is to:
1. Manually update the import paths in generated files, or
2. Move all schemas to the same file as the endpoint definitions (not ideal for large APIs)

### Related Issues

Similar issues with import paths have been reported but with different root causes:
- #2382 (indexFiles=false with custom schemas path)
- #1547 (MSW import paths with indexFiles=false)

However, this issue is specifically about external YAML file references, not the indexFiles configuration.

ryo-kozin:
## Summary
Fixes #2540 by adding type guards before using the \`in\` operator on schema objects.

## Problem
When external YAML references are not properly resolved, primitive values (like strings) can be passed where schema objects are expected. This causes \`Cannot use 'in' operator to search for 'const' in <value>\` errors.

For example:
```
🛑 uzone_user - Cannot use 'in' operator to search for 'const' in リクエストは正常に完了しました。
```

## Changes
- Add type guards (\`typeof X === 'object' && X !== null\`) before using \`in\` operator in:
  - \`packages/core/src/generators/interface.ts:47\`
  - \`packages/core/src/getters/object.ts:170\`
  - \`packages/core/src/resolvers/object.ts:42,44\`
- Update \`packages/core/src/writers/schemas.ts\` to always use schema name for canonical path lookup instead of potentially incorrect importPath from external file references

## Testing
Tested with a real-world OpenAPI spec containing external YAML references and Japanese characters in example values. Code generation now succeeds without errors.

## Related
- Fixes #2540

### DIFF
--- a/packages/core/src/generators/interface.ts
+++ b/packages/core/src/generators/interface.ts
@@ -44,7 +44,7 @@ export function generateInterface({
       scalar.type === 'object' &&
       schema.properties &&
       Object.values(schema.properties).length > 0 &&
-      Object.values(schema.properties).every((item) => 'const' in item)
+      Object.values(schema.properties).every((item) => typeof item === 'object' && item !== null && 'const' in item)
     ) {
       const mappedScalarValue = scalar.value
         .replaceAll(';', ',')
--- a/packages/core/src/getters/object.ts
+++ b/packages/core/src/getters/object.ts
@@ -167,7 +167,7 @@ export function getObject({
 
         acc.hasReadonlyProps ||= isReadOnly || false;
 
-        const constValue = 'const' in schema ? schema.const : undefined;
+        const constValue = typeof schema === 'object' && schema !== null && 'const' in schema ? schema.const : undefined;
         const hasConst = constValue !== undefined;
         let constLiteral: string | undefined;
 
--- a/packages/core/src/resolvers/object.ts
+++ b/packages/core/src/resolvers/object.ts
@@ -39,9 +39,9 @@ function resolveObjectOriginal({
     ).test(resolvedValue.value)
   ) {
     let model = '';
-    const isConstant = 'const' in schema;
+    const isConstant = typeof schema === 'object' && schema !== null && 'const' in schema;
     const constantIsString =
-      'type' in schema &&
+      typeof schema === 'object' && schema !== null && 'type' in schema &&
       (schema.type === 'string' ||
         (Array.isArray(schema.type) && schema.type.includes('string')));
 
--- a/packages/core/src/writers/schemas.ts
+++ b/packages/core/src/writers/schemas.ts
@@ -68,9 +68,11 @@ function normalizeCanonicalImportPaths(
 ) {
   for (const schema of schemas) {
     schema.imports = schema.imports.map((imp) => {
+      // Always use the schema name (imp.name) to look up the canonical path,
+      // not imp.importPath which might be incorrect for external file references
       const resolvedImportKey = resolveImportKey(
         schemaPath,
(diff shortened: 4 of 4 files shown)


melloware:
Why did you close the PR

ryo-kozin:
@melloware
Sorry for the late reply. When I tried the latest version, it seems that the way errors are handled has changed, so this fix is no longer necessary.

PR #2711, opened 2026-01-09
Add Solid Query and SolidStart

davedbase:
fix #1486 

Hello from the SolidJS Core Team. A friend recommended Orval which prompted me to investigate using it for an upcoming project. I was sad to learn Solid Query was left out of support. SolidStart was also not supported. This PR brings both of them to Orval.

Tests for this PR currently pass correctly and I've gone ahead and added the documentation as well as SolidStart to the homepage list. The output looks accurate but I only briefly practically tested it in a SolidStart application with the Pet example. Since our project will depend on the integration, I'll be sure to fully vet it and contribute adjustments in the future.

Thanks for building and maintaining Orval! Open-source rules ❤️💙

PR #2720, opened 2026-01-11
fix(core): add `@orval/solid-start` to `orval` deps

soartec-lab:
This is a follow-up to #2711
Add missing `@orval/solid-start` dependency to `orval` package.


soartec-lab:
Hi, @davedbase Here's a share. I made a small fix as a follow-up to your PR.

PR #2726, opened 2026-01-11
chore: angular-catalog-sync 

froggy1014:
- https://github.com/orval-labs/orval/pull/2724 broke tests due to Angular version drift; this should prevent a repeat.
 
 - Runtime packages are now grouped and kept in sync with @angular/core (see https://github.com/angular/angular/blob/main/tools/bazel/npm_packages.bzl), and dependabot is updated accordingly.

Previously only `@angular/core` and `@angular/common` were in the catalog—was there a reason to limit it to those two?

- https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,12 +10,24 @@ updates:
     schedule:
       interval: 'monthly'
     groups:
-      angular:
+      angular-runtime:
         patterns:
-          - '@angular/*'
+          - '@angular/animations'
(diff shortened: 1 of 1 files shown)


melloware (approved):
This is fantastic. We probably need to do this for svelte and solid and a few others where we want all the libs at once!

PR #2733, opened 2026-01-12
[docs] Improve Algolia DocSearch

### LINKED ISSUE #2716: [docs] Algolia Search: 404 Error When Clicking Parent Category Items
## Problem

When users search for "Configuration" in Algolia DocSearch and click on the result, they are redirected to a 404 page. This happens because parent categories in `manifest.json` don't have a direct `path` or `href` - they only contain child routes. However, the Algolia index includes these parent categories with URLs, but no corresponding page exists at that URL.

Image

https://orval.dev/reference/configuration

Currently, this issue only affects the "Configuration" category under "API Reference", so the impact is relatively minor. However, as the documentation structure grows deeper with more nested routes, this problem could affect more parent categories. Addressing this proactively would be beneficial.

## Proposed Solutions

I have implemented and tested both solutions locally. I would like to get feedback on which approach would be preferred:

### Solution 1: Filter out parent categories from search results
- Use `transformItems` to exclude parent category items from search results entirely
- **Pros**: 
  - Cleaner search results (users only see actual pages)
  - Prevents confusion about non-clickable parent categories
- **Cons**: 
  - Parent category names won't appear in search results (though child pages will still be searchable)

### Solution 2: Redirect to first child page when clicking parent category
- Use `navigator` to detect parent category clicks and redirect to the first child page
- **Pros**: 
  - Parent categories remain visible in search results (useful for discoverability)
  - Users can still find parent categories by name
- **Cons**: 
  - Parent categories appear in results but redirect to a different page (might be slightly confusing)

Both solutions are implemented recursively, so they will automatically handle any nested routes added to `manifest.json` in the future. Both have been tested locally and work correctly.

## Additional Note

While working on this issue, I noticed that converting `docs/src/manifests/manifest.json` to Algolia's index format (`docs/algolia-index.json`) was challenging. I created a script to automate this conversion. While such scripts can become outdated when the structure changes, having one available might be helpful for future maintenance. I'm happy to contribute this script if it would be useful.

Also, I noticed that the text in Algolia search results appears white, making it difficult to read. I've fixed this by adding CSS styles to set the text color for better readability. This fix has been tested locally and works correctly.

## Questions

1. Which solution do you prefer (Solution 1 or Solution 2)?
2. Should I proceed with creating a pull request for the chosen solution?
3. Would the Algolia index generation script be useful to include in the repository?
4. Should I include the search result text color fix in the same pull request, or would you prefer it as a separate issue?

Thank you for your consideration.

IZUMIRU:
## Summary

Fixes #2716

This PR fixes the 404 error when users click on parent category items (e.g., "Configuration") in Algolia DocSearch results. It also adds support for Enter key navigation and preserves Algolia's default styling.

## Problems

- Parent categories (items in `manifest.json` that have `routes` but no `path`/`href`) appear in search results and cause 404 errors when clicked
- Enter key does not work to navigate to search results
- Search results had poor readability due to styling issues

### 1. Filter out parent categories from search results

Use `transformItems` to exclude parent category URLs from search results.

- `collectParentCategoryUrls` function recursively collects parent category URLs
- `transformItems` filters out items that match parent category URLs

### 2. Preserve Algolia's default styling

Remove `hitComponent` to keep Algolia DocSearch's default styling (including blue background highlight).

### 3. Add Enter key navigation support

Use `itemUrl` parameter in `navigator.navigate` function to support Enter key navigation.

Before|After
-|-
image|image


melloware:
awesome running the build now

melloware:
ahh please run `yarn format`

melloware:
All good!

melloware:
its deployed to Vercel but if I search for "Input" and click on INput Configuration I get the 404 page.

melloware:
I see Output goes to

`/reference/configuration/output`

but Input goes to 

`/reference/configuration#input`

IZUMIRU:
@melloware 
It works correctly locally, so this might be an issue with differences in the Algolia index. Sorry about that. I'll check and fix it.

PR #2734, opened 2026-01-12
[doc] Restore URL normalization for Algolia search results

IZUMIRU:
## Summary

This PR restores the URL normalization logic that was previously removed in PR #2733. In PR #2733, the implementation worked correctly in the local environment. However, this was because the local Algolia index did not include URLs with hash fragments (e.g., `/reference/configuration#input`), while the production Algolia index does. As a result, the URL normalization logic was mistakenly removed, thinking it was unnecessary. This PR restores the logic to handle production environments where the Algolia index contains URLs with hash fragments.

## Problem and Solution

The production Algolia index contains URLs with hash fragments (e.g., `/reference/configuration#input`), while the local Algolia index does not. Without URL normalization, search results with hash fragments cause navigation issues in production.

This PR restores the URL normalization logic in `transformItems` to:
- Remove hash fragments from URLs (e.g., `/reference/configuration#input` → `/reference/configuration/input`)

---

image

http://localhost:3000/reference/configuration/input

### DIFF
--- a/docs/src/components/useSearch.tsx
+++ b/docs/src/components/useSearch.tsx
@@ -118,7 +118,15 @@ export function SearchProvider({
                     return null;
                   }
 
-                  return item;
+                  const url = new URL(item.url);
+                  return {
+                    ...item,
+                    url: item.url
+                      .replace(url.origin, '')
+                      .replace('#__next', '')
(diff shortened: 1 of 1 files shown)


IZUMIRU:
@melloware 
I've made the changes. Please take a look when you have a chance.

PR #2750, opened 2026-01-14
Switched SolidStart to Fetch API signature

### LINKED ISSUE #2746: Solid Start not using fetch client request signature
When passing a mutator override with Solid Start, the request signature is not correct. It is using function signature and params that align with Axios and not Fetch. See the following on how the signature should differ:

https://orval.dev/guides/custom-client
https://orval.dev/guides/custom-axios

This can technically be worked around by writing a custom instance signature, but it really should align with with fetch by default: `
 fetch(input: string | Request | URL, init?: RequestInit | undefined)
`

Image

Image

Solid start should use a similar function to this to generate the correct signature:
https://github.com/orval-labs/orval/blob/d9858d4e9e12abd0d033034c612cfbeba58642a5/packages/query/src/client.ts#L70

The documentation here should also mention add solid-start to the list as fetch client defaults:
https://orval.dev/guides/fetch-client

davedbase:
Fix #2746 

Patching an issue reported regarding the signature which was using Axios reference instead of fetch. I considered support Axios as an option in the build but didn't watch to make the implementation way more complex. A user could just use the mutator override. Frankly we prefer to promote fetch instead of Axios.

### DIFF
--- a/packages/solid-start/src/index.ts
+++ b/packages/solid-start/src/index.ts
@@ -43,9 +43,7 @@ export const generateSolidStartTitle: ClientTitleBuilder = (title) => {
   return `${pascal(sanTitle)}`;
 };
 
-export const generateSolidStartHeader: ClientHeaderBuilder = ({
-  title,
-}) => `
+export const generateSolidStartHeader: ClientHeaderBuilder = ({ title }) => `
 /**
  * Cache Invalidation:
  *


davedbase:
@JavanPoirier should review this before merging to confirm it's aligned with his expectations.

JavanPoirier:
Tested and approve!

PR #2758, opened 2026-01-15
Fix (query) angular custom instance error

### LINKED ISSUE #2757: angular-query 8.0.0 error return type and/or custom instance problems
The result from a mutation/query error response is typed as the openapi defined response, instead of angulars HttpErrorResponse, and when logged to the console the data matches HttpErrorResponse

I tried adding a custom instance to change the error type
```
export interface ExtendedServerErrorResponse<Error> extends HttpErrorResponse {
  error: Error | any | null;
}
export type ErrorType<Error> = ExtendedServerErrorResponse<Error>;
```

However using the custom response-type.ts file from the (non-query) angular examples I get these newline characters messing things up in generated orval:

Image

ScriptType:
# Fix Angular Query Custom Mutator Generation

## Summary

Fixes #2757

When using a custom mutator with Angular Query, the generated code had multiple issues:

1. **Literal `\n` in output** - `String.raw` template literal caused `\n` to appear as literal text instead of newlines
2. **Incorrect `HttpClient` injection** - `inject(HttpClient)` was generated even though the mutator handles HTTP internally
3. **Wrong function call arguments** - `http` was passed as first argument but the function didn't expect it
4. **Signal wrapped incorrectly** - Signal was wrapped in object `{ signal, ...requestOptions }` instead of passed as separate argument

### Core Fix (`packages/query/src/`)

- **client.ts**: 
  - **Fixed `String.raw` issue** - Removed `String.raw` template literal in `generateAngularHttpRequestFunction` so `\n` is interpreted as newline
  - Added `hasMutator` parameter to `getHttpFunctionQueryProps()` to skip `http` prefix when custom mutator is used
  - Updated `getQueryOptions()` to return `requestOptions, signal` as separate arguments for mutator case (matching axios pattern)

- **query-generator.ts**:
  - Pass `!!mutator` to `getHttpFunctionQueryProps()`
  - Check `!mutator` before adding `http` prefix for infinite queries
  - Check `!mutator` before generating `inject(HttpClient)`

- **mutation-generator.ts**:
  - Check `!mutator` before adding `http` prefix in mutation function calls
  - Check `!mutator` before generating `inject(HttpClient)`

### Sample (`samples/angular-query/`)

Added a working example of Angular Query with custom mutator:
- `orval.config.ts`: Added `petstoreCustomInstance` config
- `src/api/mutator/custom-client-with-error.ts`: Custom mutator demonstrating:
  - `ErrorType<Error>` interface extending `HttpErrorResponse`
  - Internal `inject(HttpClient)` handling
  - Proper signal and responseType support

### Unit Tests (`packages/query/src/client.test.ts`)
- 17 new tests for `getHttpFunctionQueryProps` and `getQueryOptions`
- Verify `http` prefix is skipped when mutator is used
- Verify signal is passed as separate argument for mutator case

### Integration Tests (`samples/angular-query/src/app/angular-query-generation.spec.ts`)
- 16 tests covering native and custom mutator generation
- Verify exported functions work correctly
- Test `queryFn` execution with signal handling
- Validate `queryKey` parity between native and custom instance
- Verify `ErrorType` compatibility with `HttpErrorResponse`
- Confirm custom mutator path works without direct `HttpClient` injection

### Before (with bug)
```typescript
// In getListPetsQueryOptions - incorrectly injected HttpClient
const http = inject(HttpClient);  // ERROR: HttpClient not imported when using mutator

// queryFn incorrectly passed http as first arg and signal wrapped in object
const queryFn = ({ signal }) => listPets(http, params, { signal, ...requestOptions });
//                                        ^^^^         ^^^^^^^^^^^^^^^^^^^^^^^^^
//                                        Wrong!       Wrong - signal should be separate arg

// But the operation function (listPets) was generated correctly WITHOUT http param:
export const listPets = (params?: ListPetsParams, signal?: AbortSignal) => { ... }

// This caused TS2345: Argument of type 'unknown' is not assignable to parameter
```

### After (fixed)
```typescript
// No HttpClient injection when using custom mutator (mutator handles it internally)

// queryFn correctly passes only params and signal as separate arguments
const queryFn = ({ signal }) => listPets(params, signal);

// Matches the operation function signature
export const listPets = (params?: ListPetsParams, signal?: AbortSignal) => { ... }
```

### DIFF
--- a/packages/query/src/client.test.ts
+++ b/packages/query/src/client.test.ts
@@ -194,7 +194,7 @@ describe('getQueryOptions', () => {
   });
 
   describe('signal handling with mutator vs without', () => {
-    it('should return signal as separate arg for mutator case', () => {
+    it('should return signal as separate arg for Angular mutator case', () => {
       const mutatorNoSecondArg = { ...mockMutator, hasSecondArg: false };
       const result = getQueryOptions({
         isRequestOptions: false,
@@ -215,6 +215,29 @@ describe('getQueryOptions', () => {
         httpClient: OutputHttpClient.ANGULAR,
       });
       expect(result).toBe('{ signal }');
+    });
+
+    it('should return signal wrapped in object for fetch mutator case', () => {
+      const mutatorNoSecondArg = { ...mockMutator, hasSecondArg: false };
+      const result = getQueryOptions({
+        isRequestOptions: false,
+        mutator: mutatorNoSecondArg,
+        isExactOptionalPropertyTypes: false,
+        hasSignal: true,
+        httpClient: OutputHttpClient.FETCH,
+      });
+      expect(result).toBe('{ signal }');
+    });
+
+    it('should wrap signal in options for fetch mutator with hasSecondArg', () => {
+      const result = getQueryOptions({
+        isRequestOptions: true,
+        mutator: mockMutator,
+        isExactOptionalPropertyTypes: false,
+        hasSignal: true,
+        httpClient: OutputHttpClient.FETCH,
+      });
+      expect(result).toBe('{ signal, ...requestOptions }');
     });
   });
 });
--- a/packages/query/src/client.ts
+++ b/packages/query/src/client.ts
@@ -477,17 +477,26 @@ export const getQueryOptions = ({
       return 'requestOptions';
     }
 
-    // For mutator case, signal is always a separate argument (axios pattern)
-    // not wrapped in options object (fetch pattern)
-    return 'requestOptions, signal';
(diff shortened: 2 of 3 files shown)


ScriptType:
Ok, i did an error when updating getQueryOptions() to fix the issue, and i accidently updated all mutators and not only than when we want it.
Reopen now.

cyb456:
NG0203: The `_HttpClient` token injection failed. `inject()` function must be called from an injection context such as a constructor, a factory function, a field initializer, or a function used with `runInInjectionContext`. Find more at https://v21.angular.dev/errors/NG0203

Issue #2755, opened 2026-01-15
Orval 8 has broken support for Open API spec v3 nullable:true in typescript type generation - not mentioned in release notes


Georgegriff:
Sorry if i get my Open API spec version muddled up but this ticket is about how `nullable: true` used t work

Is this intentional, but i had some fields that are using:

```
field: {
  "type": "string",
  "nullable": true
}
```

the type generated used to be

field: string | null

now its just field: string

if i use 3.1.0 syntax it works, but obviously that would require a change to the API itself

```
"hint": {
  "type": ["string", "null"]
}
```

Georgegriff:
Here is the PR that dropped 3.0 support https://github.com/anymaniax/orval/issues/2614 it's not mentioned in the v8 release notes so that's why i opened to understand if it was intentional

melloware:
thanks for digging in @Georgegriff

melloware:
cou update the migration guide with this info?

Georgegriff:
https://github.com/orval-labs/orval/pull/2756

melloware:
Reopening and assigning to @snebjorn he thinks this is a bug.

snebjorn:
@Georgegriff seems you found bug with the nullable check. It should give you `string | null` for both 3.0 and 3.1.

Can you please provide a minimal spec that have this issue and I'll look into it, thanks

Georgegriff:
> @Georgegriff seems you found bug with the nullable check. It should give you `string | null` for both 3.0 and 3.1.
> 
> Can you please provide a minimal spec that have this issue and I'll look into it, thanks

string | null works it's nullable: true that does not anymore

snebjorn:
I need a specific example where it doesn't work cause I have many examples where it does work :)
You found an edge case it seems

Georgegriff:
> I need a specific example where it doesn't work cause I have many examples where it does work :) You found an edge case it seems

@snebjorn here you go, i've run this against orval 8 in my repo and trimmed it down 

```ts
import { defineConfig } from "orval";

export default defineConfig({
  internal_api: {
    input: {
      target: "./schema.json",
    },
    output: {
      target: "generated/internal.ts"
    },
  },
});

```

```json
{
  "openapi": "3.1.0",
  "info": {
    "title": "API V2",
    "version": "v2"
  },
  "paths": {
    "/questions/{id}": {
      "parameters": [
        {
          "name": "id",
          "in": "path",
          "description": "The ID for the question",
[...]

PR #2766, opened 2026-01-16
docs: openapi 3.0 was not dropped

snebjorn:
Internally we're upgrading input specs to 3.1, users shouldn't be impacted


snebjorn:
waiting for #2755 resolution

PR #2767, opened 2026-01-16
Feat (query) make inject query reactive

ScriptType:
## Summary

Enables Angular Query `inject*` functions to accept **both params AND options** as getter functions, allowing queries to automatically re-execute when Angular signal values change.

## Usage

```typescript
// Reactive params - query refetches when rows() changes
rows = signal(10);
petsQuery = injectListPets(() => ({ limit: this.rows() }));

// Reactive options - query enables/disables when isReady() changes
isReady = signal(false);
petsQuery = injectListPets({ limit: 10 }, undefined, () => ({
  query: { enabled: this.isReady() },
}));
```

## What Changed

- **Getter signature for params**: `T | (() => T)` in `inject*` hooks
- **Getter signature for options**: `options?: T | (() => T)` for reactive `enabled`, `staleTime`, etc.
- **Deferred resolution**: Both params and options resolved inside `injectQuery` lambda
- **HttpClient positioning**: `inject(HttpClient)` in `inject*` functions; `http` is first param in `getQueryOptions`/`getMutationOptions`

## Breaking Change

`getQueryOptions`/`getMutationOptions` now require `http: HttpClient` as **first** parameter:

```typescript
// Before
getListPetsQueryOptions(params?, options?)

// After
getListPetsQueryOptions(http: HttpClient, params?, options?)
```

I hope the breaking change is ok, as everything is still new :/ .

### DIFF
--- a/packages/query/src/query-generator.ts
+++ b/packages/query/src/query-generator.ts
@@ -255,8 +255,13 @@ const generateQueryImplementation = ({
     props.map((prop) => {
       // Add getter alternative: T becomes T | (() => T)
       const getterType = prop.definition.replace(
-        /^(\w+)(\??):\s*(.+)$/,
-        (_, name, optional, type) =>
+        /^(\w+)(\??): (.+)$/,
+        (
+          _match: string,
+          name: string,
+          optional: string,
+          type: string,
+        ): string =>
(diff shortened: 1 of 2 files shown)


melloware:
Looks like a LINT error

PR #2789, opened 2026-01-16
Chore (lint) switch to oxlint

ScriptType:
I tried oxlint in some of my private projects and have good experience with it. (don't forget to also use it in your IDE)
Just thought you are also maybe interested, just checkout the branch and compare results to master.

### Summary
Replaces ESLint with [oxlint](https://oxc.rs/docs/guide/usage/linter.html) for significantly faster linting.
### Benchmark Results
| Linter | Cold Cache | Warm Cache | Speedup |
|--------|------------|------------|---------|
| ESLint (old) | 12.4s | 7.7s | - |
| oxlint (new) | 0.5s | 0.5s | **~15-25x faster** |
*oxlint has no caching overhead since it doesn't need caching — it's already instant.*
### Changes
- Added `.oxlintrc.json` with migrated rules (generated via `npx @oxlint/migrate`)
- Updated root `package.json` lint script to use oxlint directly
- Added `scripts/lint-summary.js` for per-package summary output
- Added `scripts` to ignore patterns in both ESLint and oxlint configs
### New Commands
```bash
yarn lint          # Fast stylish output (0.5s)
yarn lint:summary  # Detailed output + per-package summary (0.7s)
```
### Why Not Use Turbo for Linting?
We intentionally run oxlint at the root level instead of per-package via Turbo because:
1. **oxlint is already instant** (~20ms per package). Turbo's process spawning overhead (~200ms per package) dominates the total time
2. **No caching benefit**: Linting is so fast that caching provides no meaningful speedup
3. **Single process is faster**: Running oxlint once on the entire repo (0.5s) is 5x faster than Turbo spawning 12 separate oxlint processes (2.7s):
   | Approach | Time |
   |----------|------|
   | oxlint at root | 0.5s |
   | oxlint via Turbo | 2.7s |
   | ESLint via Turbo | 12.4s |
Turbo excels when tasks take seconds or minutes — for sub-100ms tasks like oxlint, running a single process directly is objectively faster.
### Missing ESLint Rules
The following ESLint rules are not yet supported by oxlint:
**Import Sorting** (handled separately by Prettier):
- `simple-import-sort/imports`
- `simple-import-sort/exports`
**Type-Aware Rules** (style preferences, not correctness):
- `@typescript-eslint/no-unnecessary-condition`
- `@typescript-eslint/no-invalid-void-type`
- `@typescript-eslint/consistent-type-assertions`
- `@typescript-eslint/dot-notation`
- `@typescript-eslint/unified-signatures`
**Other**:
- `prettier/prettier` → Already handled by `yarn format`
- `turbo/no-undeclared-env-vars` → Turbo-specific
### Future Cleanup (Optional)
After validation, we can remove:
- `eslint.config.mjs`
- ESLint dependencies from root `package.json` (`eslint`, `@eslint/js`, `eslint-*`, `typescript-eslint`)

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -1,6 +1,7 @@
 {
   "$schema": "./node_modules/oxlint/configuration_schema.json",
   "plugins": ["typescript", "unicorn"],
+  "jsPlugins": ["eslint-plugin-simple-import-sort"],
   "categories": {
     "correctness": "off"
   },
@@ -20,6 +21,8 @@
     "**/dist"
   ],
   "rules": {
+    "simple-import-sort/imports": "error",
+    "simple-import-sort/exports": "error",
     "constructor-super": "error",
     "for-direction": "error",
     "no-async-promise-executor": "error",


melloware:
@snebjorn thoughts?

snebjorn:
interesting I didn't know about oxlint

> Import Sorting (handled separately by Prettier)

I don't believe prettier sorts, it just formats imports.
I feel strongly about sorting imports :)

> prettier/prettier → Already handled by yarn format

`yarn format` runs formatting
eslint plugin prettier/prettier makes eslint fail when code isn't formatted
It also turns off rules that are unnecessary or might conflict with Prettier

I don't think I care too much about the rest

melloware:
Prettier can definitely handle the import sorting I am pretty sure. If it's not now we can set it to.

ScriptType:
Yea they have also an Prettier alternative, oxfmt but when i tried it the angular support wasn't really given , so i think it wouldn't be a good choice especially when "niche" frameworks like solid are used or some come up in the future.
In oxlint you can also do the --type-check flag which is basically adding tsc --noEmit and report these errors.

It uses typescript-go for all the type stuff (maybe you have an issue with it because it`s "experimental") but the type-checking is "done" according to the typescript team.
https://github.com/microsoft/typescript-go

snebjorn:
It doesn't sort imports by default, and I don't see an option to turn it on

https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLAtgBwgJxgAmACMAafAQwF98AzHCdfAciKYB0oNs9CATfEmGp0GzckxAkQETDFTQAzslDkc9AO4AFVQiUpyAG3XkAnkqlEc5MAGs4MAMrl0cADKoocZDUMK4UiCIAKzgwGAB1K0xkEEwcOD8cADcvCytbewdMaw8Ac2QYHABXfxA-dFQC4tK4AA9MOBwMBBhDAHkGqxhcTQgFVDloGIReSRA6zubYQwAVRqhVVATvX1L+qFyDOABFIoh4FYM-KSCFWoc8rd39ryQfI9KARz34TXpMPRByBQBaTzheACxoVyKgDHkAMIMdDkGKGAxjdabOAAQRghVQRCKr0a7k8h2OIAAFjB0AZwkSBglsmA4A5dANUEkBiYYmAFOYQEkSgBJKCA2AOMBNWQo-kOGAmLYE0pxPpwSLkaIoOIJRopMYjNo0PG3EAGGhjDyJGBvci5GEyqTZHCJGJEchEOAI61NWDhVC8GBE5AADgADFJ4s9UPEzRbYXdVlJWkQPV6fUgAExSIp+GaOvT3QlwdBO3iA3iucgbIrmuAAMVwMPReTh2IgIEolCAA

ScriptType:
@snebjorn you need a plugin
https://www.npmjs.com/package/prettier-plugin-organize-imports

EDIT: i just saw oxlint has now support for js-plugins. (didn't had it when i used it)
Tested the simple-sort plugin now and it also seems to work.

snebjorn:
> you need a plugin https://www.npmjs.com/package/prettier-plugin-organize-imports

says it the same as using the "Organize Imports" action in VS Code which leaves a lot to be desired :)

> using the organizeImports feature of the TypeScript language service API. This is the same as using the "Organize Imports" action in VS Code.

> EDIT: i just saw oxlint has now support for js-plugins. (didn't had it when i used it)
> Tested the simple-sort plugin now and it also seems to work.

very nice ✔️ 

I looked at the docs and it says type-aware rules is an alpha feature https://oxc.rs/docs/guide/usage/linter/type-aware.html
type-aware rules are so much better, so we need to make sure it's fit for purpose.

hmm, is this just too early?

using typescript-go - experimental
type-aware rules - alpha
js-plugins - technical preview

I don't mind trying new tools. But it should be reliable. The above doesn't give me that feeling.

I found this while looking into oxlint. So at some point typescript-eslint will get the speedup from typescript-go as well
- https://github.com/typescript-eslint/typescript-eslint/issues/10940

melloware:
agree with  @snebjorn  is this too early?

piradata:
i think it is top early, and would change it to biome instead of eslint, as i never heard about this oxlint..

https://biomejs.dev/assist/actions/organize-imports/

it also supports organize imports, and can substitute pretty much eslint and prettier, as it does both formatting and linting

ScriptType:
biomejs is definetly possible, and 
The difference why i chose oxlint is because it aims to be a drop-in replacement of eslint, same rule-names etc.

But i think for now we can say it is maybe to early so i will close it here, and just look at it again in some months or a year from now .

PR #2796, opened 2026-01-19
fix(enum): CVE-2026-23947 x-enum values must be JS escaped

melloware:
fix(enum): x-enum values must be JS escaped


AllieJonsson (approved):

melloware:
cc @k14uz

PR #2799, opened 2026-01-19
docs: rehype-pretty-code

snebjorn:
new syntax highlighting https://rehype-pretty.pages.dev/ it supports diff

I'm still looking for a way to turn on line numbers globally so we don't have to add `showLineNumbers` everywhere
https://github.com/rehype-pretty/rehype-pretty-code/discussions/265

It works, but it needs to be styled
image

Current
image


snebjorn:
@anymaniax I'm told you're working on a major doc refactor. Figured you could use this. It still needs work

PR #2813, opened 2026-01-21
fix: clean workspace directory when clean option is enabled

takahiro-1227:
## Problem
When `clean: true` is set, the `workspace` directory was not being cleaned up, which caused `index.ts` files in the workspace directory to persist after regeneration.

## Solution
Added cleanup for the `workspace` directory similar to how `target` and `schemas` directories are cleaned. This ensures that all generated files, including `index.ts` in the workspace directory, are properly removed before regeneration.

## Changes
- Added cleanup logic for `options.output.workspace` in `generate-spec.ts`
- Uses the same pattern as `target` and `schemas` cleanup (`**/*` with `!**/*.d.ts` exclusion)

### DIFF
--- a/packages/orval/src/generate-spec.ts
+++ b/packages/orval/src/generate-spec.ts
@@ -41,6 +41,12 @@ export async function generateSpec(
         getFileInfo(options.output.schemas).dirname,
       );
     }
+    if (options.output.workspace) {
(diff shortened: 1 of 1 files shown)


takahiro-1227:
Maybe I should rethink. Workspace may be positioned at the same level to files that should not be removed.

PR #2816, opened 2026-01-21
feat(svelte-query): mutation reactivity, enable invalidates on svelte-query

AllieJonsson:
Reactivity was not implemented correctly for mutation, which is now fixed.
Enables mutationInavalidations in svelte-query


melloware:
Looks like a lint error with import sorts

PR #2820, opened 2026-01-21
[WIP] fix(core): use parameter name in formData for $ref schemas (#2814)

### LINKED ISSUE #2814: Bug: formData generation uses wrong variable name in v8.0.x
## Bug Description

In orval v8.0.x, when generating code for multipart/form-data endpoints, the generated `formData.append()` calls use the wrong variable name. It uses the camelCase version of the OpenAPI schema name instead of the actual function parameter name.

## Steps to Reproduce

1. Create an OpenAPI spec with a multipart/form-data endpoint:
```yaml
/api/maps/{id}/pgm:
  put:
    operationId: MapsController_savePgm
    parameters:
      - name: id
        required: true
        in: path
        schema:
          type: number
    requestBody:
      required: true
      content:
        multipart/form-data:
          schema:
            $ref: "#/components/schemas/SavePgmRequest"
    responses:
      "200":
        description: ""
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/SavePgmResponse"

components:
  schemas:
    SavePgmRequest:
      type: object
      properties:
        file:
          type: string
          format: binary
      required:
        - file
```

2. Generate code using orval v8.0.x with `client: "axios-functions"` and `httpClient: "axios"`

## Expected Behavior

The generated code should use the function parameter name:
```typescript
export const mapsControllerSavePgm = (
  id: number,
  mapsControllerSavePgmBody: MapsControllerSavePgmBody,
  options?: ...
) => {
  formData.append(\`file\`, mapsControllerSavePgmBody.file);  // Correct
  ...
}
```

## Actual Behavior

The generated code uses the schema name (camelCase) which is not defined:
```typescript
export const mapsControllerSavePgm = (
  id: number,
  mapsControllerSavePgmBody: MapsControllerSavePgmBody,  // Parameter name
  options?: ...
) => {
  formData.append(\`file\`, savePgmRequest.file);  // Wrong! Uses schema name
  ...
}
```

This causes TypeScript error: `TS2304: Cannot find name 'savePgmRequest'`

## Environment

- orval version: 8.0.0, 8.0.1, 8.0.2 (all affected)
- orval v7.18.0 works correctly (parameter is named `savePgmRequest: SavePgmRequest` and formData uses `savePgmRequest.file`)

## Workaround

Using a custom formData function via the `output.override.formData` configuration option.

## Additional Context

This appears to be a regression introduced in v8.0.0. In v7.18.0, the parameter name matched the schema name, so the formData code was correct. In v8.0.x, the parameter name was changed to a different format (e.g., `mapsControllerSavePgmBody`) but the formData generation still references the old schema-based name.

superxiao:
Pass `isRef: true` when calling `getSchemaFormDataAndUrlEncoded` from inline `requestBody` path, ensuring `formData` variable uses parameter name instead of schema name.

Fix #2814

### DIFF
--- a/packages/core/src/getters/res-req-types.test.ts
+++ b/packages/core/src/getters/res-req-types.test.ts
@@ -282,3 +282,82 @@ formData.append(\`wildcardFile\`, bodyRequestBody.wildcardFile);
     });
   });
 });
+
+// When requestBody uses $ref, the generated function parameter name varies:
+// - If schema properties have file types (contentMediaType), the type is rewritten
+//   as inline, so param becomes operation-based: uploadFile() → uploadFileBody
+// - If schema has no file types, the $ref type is preserved,
+//   so param uses schema name: uploadFile() → mySchemaName
+// formData variable must match whichever naming is used.
+describe('getResReqTypes (formData $ref variable name)', () => {
+  it('uses operation-based name when schema has file properties', () => {
+    // contentMediaType: 'application/octet-stream' marks this as a file field
+    const ctx: ContextSpec = {
+      ...context,
+      spec: {
+        components: {
+          schemas: {
+            FileUpload: {
+              type: 'object',
+              properties: {
+                file: {
+                  type: 'string',
+                  contentMediaType: 'application/octet-stream',
+                },
+              },
+            },
+          },
+        },
+      },
+    };
+    const reqBody: [string, OpenApiRequestBodyObject][] = [
+      [
+        'requestBody',
+        {
+          content: {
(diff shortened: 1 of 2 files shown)


superxiao:
@melloware I think this fix is not quite right. I will have a PR soon. We are not rushing for a new minior version right?

melloware:
No but i wouldn't mind releasing one in the next day or two it has some fixes

superxiao:
Update PR https://github.com/orval-labs/orval/pull/2825

PR #2828, opened 2026-01-22
fix(mock): escape mocked values

melloware:
cc @k14uz

### DIFF
--- a/packages/mock/src/faker/getters/scalar.ts
+++ b/packages/mock/src/faker/getters/scalar.ts
@@ -156,11 +156,11 @@ export function getMockScalar({
           'number',
         );
       } else if ('const' in item && typeof item.const === 'string') {
-        value = item.const;
+        value = jsStringEscape(item.const);
       }
 
       return {
-        value: jsStringEscape(value),
+        value,
         enums: item.enum,
         imports: numberImports,
         name: item.name,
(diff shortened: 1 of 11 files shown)


k14uz:
FYI: I reviewed the current fix and I think there’s still a gap around how `const` values are turned into code.

#### Observation
`jsStringEscape()` only escapes quotes, backslashes, and line terminators (see `jsStringEscape()` in `packages/core/src/utils/string.ts`). It does not serialize values into safe JS literals, and it does not address characters like `)`, `;`, `(` that can break surrounding expression contexts. [utils/string.ts#L227-L253](https://github.com/orval-labs/orval/blob/master/packages/core/src/utils/string.ts#L227-L253)

Because the mock generator is emitting **code expressions** (e.g. inside `faker.helpers.arrayElement([ ... ])`), escaping quotes alone isn’t sufficient if an untrusted value is ever emitted unquoted in an expression context.

#### Suggested approach
Instead of “escape”, serialize `const` into a JavaScript literal at the point it’s incorporated into generated code, e.g.:

```ts
// for const handling in mock scalar generation
if ('const' in item) {
  value = JSON.stringify(item.const);
}
```

This way you can also remove the `jsStringEscape()` wrapping at return, since `JSON.stringify()` already produces valid JavaScript literals.

Let me know if you'd like me to test a revised version.

PR #2830, opened 2026-01-22
fix(mock): CVE-2026-24132 escape mocked values

melloware:
cc @k14uz

### DIFF
--- a/packages/mock/src/faker/getters/scalar.ts
+++ b/packages/mock/src/faker/getters/scalar.ts
@@ -154,7 +154,7 @@ export function getMockScalar({
           existingReferencedProperties,
           'number',
         );
-      } else if ('const' in item && typeof item.const === 'string') {
+      } else if ('const' in item) {
         value = JSON.stringify(item.const);
       }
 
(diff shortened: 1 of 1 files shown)


k14uz:
Looks much better now! Two minor follow-ups I noticed:
1. In the `string` branch, this currently double-wraps ([scalar.ts#L260-L262](https://github.com/orval-labs/orval/blob/13ff3f73568a2829a66e550dd1f8b823c658b4a6/packages/mock/src/faker/getters/scalar.ts#L260-L262)):
```ts
value = `'${JSON.stringify((item as OpenApiSchemaObject).const)}'`;
```
Since `JSON.stringify("abc")` already returns `"abc"` (including the quotes), wrapping in `'...'` yields `'"abc"'`. I think this should just be:
```ts
value = JSON.stringify((item as OpenApiSchemaObject).const);
```

2. For number/boolean you currently only apply the const handling when `typeof item.const === 'string'`. It might be cleaner to handle any const type:
```ts
if ('const' in item) value = JSON.stringify(item.const);
```
so `const: 123` and/or `const: true` work too.

Happy to retest with my PoC after the tweak. lmk!

melloware:
OK how does that look i removed the check for `string` for const.  I didn't want to refactor the code too greatly.

k14uz:
lgtm!

PR #2831, opened 2026-01-22
fix(v7): CVE-2026-24132 Replace jsStringEscape with JSON.stringify

melloware:


k14uz:
FYI; One minor comment is the similar double-wrap issue mentioned under https://github.com/orval-labs/orval/pull/2830#issuecomment-3784710205, 

This double-wraps the result. Since `JSON.stringify` already returns a quoted string ("abc"), wrapping again produces '"abc"'. [scalar.ts#L259-L261](https://github.com/orval-labs/orval/blob/9aa2e52765cd5a31d7109603a9761c3f84e5b866/packages/mock/src/faker/getters/scalar.ts#L259-L261)

Should be:
```ts
value = JSON.stringify((item as SchemaObject31).const);
```
Same as the number/boolean branches.

melloware:
Yep!

PR #2837, opened 2026-01-22
feat: add axios instance injection support

### LINKED ISSUE #653: Pass in custom axios instance
Hey, first of all, this package seems really awesome and we currently have a very large manually maintained api package that has all of our typesafe code, which this package could completely make redundant, which would be a huge time saver!

I am struggling a bit with the docs here - If I wanted to have each of the endpoint-generated axios-functions to have a custom axios instance passed to it as a parameter, is that the job for a custom `mutator`?

What I am going for is something like this:

```ts
export const getPets = (axiosInstance: axiosInstance) => {
  /**
   * List all pets.
   * @summary GET pets
   */
  const petsIndex = <TData = AxiosResponse<PetsIndex200>>(
    params?: PetsIndexParams,
    options?: AxiosRequestConfig,
  ): Promise<TData> => {
    return axiosInstance.get(`/pets`, {
      ...options,
    });
  };
}
```

where `axiosInstance` is something I can pass into the request

Thanks

soufDev:
### Problem

Fix #653 

> This PR supersedes #2423 (originally by @jdrouet). We work together and I had to take over to complete the implementation. Related to issue #653.

Currently the generated code use global axios directly, which make it hard to:
- Add interceptors (auth, logging, etc)
- Mock in tests
- Use different configs per client

### Solution

Factory function now accept optional `axiosInstance` parameter with default `axios.create()`.

```typescript
// before
export const getPetApi = () => { ... }

// after
export const getPetApi = (axiosInstance: AxiosInstance = axios.create()) => { ... }
```

### Usage

```typescript
const myClient = axios.create({
  baseURL: 'https://api.example.com',
  headers: { Authorization: `Bearer ${token}` },
});

const petApi = getPetApi(myClient);
```

Backward compatible - existing code still works without changes.

### DIFF
--- a/samples/basic/api/endpoints/petstoreFromFileSpecWithConfig.ts
+++ b/samples/basic/api/endpoints/petstoreFromFileSpecWithConfig.ts
@@ -79,6 +79,8 @@ export type ListPetsNestedArrayParams = {
   limit?: string;
 };
 
+const axiosInstance = axios;
+
 /**
  * @summary List all pets
  */
@@ -86,7 +88,7 @@ export const listPets = <TData = AxiosResponse<PetsArray>>(
   params?: ListPetsParams,
   options?: AxiosRequestConfig,
 ): Promise<TData> => {
-  return axios.get(`/pets`, {
+  return axiosInstance.get(`/pets`, {
     ...options,
     params: { ...params, ...options?.params },
   });
@@ -99,7 +101,7 @@ export const createPets = <TData = AxiosResponse<void>>(
   createPetsBody: CreatePetsBody,
   options?: AxiosRequestConfig,
 ): Promise<TData> => {
-  return axios.post(`/pets`, createPetsBody, options);
+  return axiosInstance.post(`/pets`, createPetsBody, options);
 };
 
 /**
@@ -109,7 +111,7 @@ export const listPetsNestedArray = <TData = AxiosResponse<PetsNestedArray>>(
   params?: ListPetsNestedArrayParams,
   options?: AxiosRequestConfig,
 ): Promise<TData> => {
-  return axios.get(`/pets-nested-array`, {
+  return axiosInstance.get(`/pets-nested-array`, {
     ...options,
     params: { ...params, ...options?.params },
   });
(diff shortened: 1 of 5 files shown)


melloware:
you need to run `yarn update-samples`

packages/axios/src/index.ts
```diff
@@ -199,7 +200,7 @@ ${
     ? `type SecondParameter<T extends (...args: never) => unknown> = Parameters<T>[1];\n\n`
     : ''
 }
-  ${noFunction ? '' : `export const ${title} = () => {\n`}`;
+  ${noFunction ? '' : `export const ${title} = (axiosInstance: AxiosInstance = axios.create()) => {\n`}`;
```
melloware: Does this construct a new instance every time any method is called for backwards compatiblity

soufDev: Actually its not, `axios.create()` only runs once when you call the factory function, not on every method call.

```typescript
const api = getApi();      // axios.create() runs here (once)
api.getUser(1);            // reuses same instance
api.updateUser(1);         // reuses same instance
```

The instance is stored in a closure and shared by all methods. This keeps backwards compatibility - if you don't pass your own instance, you get a fresh isolated one automatically.

soufDev:
`yarn update-samples` is failing, even if I run it on `master`, I get the same error on both branches

Screenshot 2026-01-23 at 00 02 24

melloware:
Weird as main is passing?

Also I use global Axios are you sure this Axios create call won't override the global instance settings?

melloware:
I think i got the builds passing again

soufDev:
I just updated the PR, so its backward compatible with what we already have and support the axios global instance

melloware:
OK weird your other PR passed and built but this one is still failing

melloware:
@soufDev this is much better because 99% of our users have been using the global instance and we did not want to break it!

melloware:
OK you can see the test failures on my PR. The generated code is now breaking expected results

FR073N:
Hey @soufDev !

I saw you're working on axios instance injection. I've been using a workaround script for this for months and finally had time to contribute.

I just submitted an alternative approach in #2851 that you might want to check out. The main difference is in the injection pattern:

**Your approach (parameter on each function):**
```typescript
export const listPets = (params?, options?, axiosInstance = axios) => { ... }
// Usage: listPets(params, options, myAxios)
```

**My approach (factory with injection):**
```typescript
export const getPetsApi = (axios: AxiosInstance) => ({
  listPets: (params?, options?) => axios.get('/pets', { params }),
});
// Usage: const PetsApi = getPetsApi(myAxios); PetsApi.listPets(params);
```

Both approaches solve the same problem though. Happy to discuss or collaborate if you think combining ideas makes sense.

PR #2838, opened 2026-01-22
feat: add useNullForOptional option for JSON:API compatibility

soufDev:
## Description

Adds a `useNullForOptional` option that makes optional properties include `| null` in their type. This matches JSON:API behavior where optional fields can be `null`.

## Changes

- Added `useNullForOptional` option to the config
- Optional properties now get typed as `T | null` when enabled

## Example

**Before:**
```typescript
email?: string;
```

**After (with `useNullForOptional: true`):**
```typescript
email?: string | null;
```

## Usage

```typescript
export default {
  override: {
    useNullForOptional: true,
  },
};
```

### DIFF
--- a/docs/src/pages/reference/configuration/output.mdx
+++ b/docs/src/pages/reference/configuration/output.mdx
@@ -3139,6 +3139,50 @@ export default defineConfig({
 });
 ```
 
+### useNullForOptional
+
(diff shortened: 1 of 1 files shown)


melloware (changes requested):
Need to add this to the `/docs`

PR #2841, opened 2026-01-23
fix(axios): Allow axios instance per call

### LINKED ISSUE #653: Pass in custom axios instance

melloware:
Fix #653

### DIFF
--- a/packages/axios/src/index.ts
+++ b/packages/axios/src/index.ts
@@ -28,6 +28,7 @@ const AXIOS_DEPENDENCIES: GeneratorDependency[] = [
         values: true,
         syntheticDefaultImport: true,
       },
+      { name: 'AxiosInstance' },
       { name: 'AxiosRequestConfig' },
       { name: 'AxiosResponse' },
     ],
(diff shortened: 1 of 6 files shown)


melloware:
@soufDev  I copied your PR and ran `yarn update-samples` and this is what happened

Issue #2839, opened 2026-01-23
TypeScript error with an optional parameter and optionsParamRequired: true


n-e:
### Description

When optionsParamRequired is true, and the schema has an optional parameter, a function such as this one is generated : 

```ts
const apiUser = (
    params?: ApiUserParams,
 options: TTTT=> { ... }
```

This is wrong, since there is a required argument after an optional argument

### Configuration (orval.config)

```ts
import { defineConfig } from "orval";

export default defineConfig({
  lichess: {
    input: {
      target: "schema.yml",
    },
    output: {
      target: "./client.gen.ts",
      client: "axios",
      optionsParamRequired: true,
      override: { mutator: { path: "mutator.ts", extension: ".ts" } },
    },
  },
});
```

### Environment

  System:
    OS: macOS 15.5
    CPU: (8) arm64 Apple M2
    Memory: 66.64 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    orval: ^8.0.3 => 8.0.3 

### Expected behavior

The function should not have a required parameter after an optional one, for example:
```ts
const apiUser = (
    params: ApiUserParams | undefined,
 options: TTTT=> { ... }
```

### Actual behavior

TypeScript complains with

```
 A required parameter cannot follow an optional parameter.
```

### OpenAPI document (minimal, if applicable)

```yaml
openapi: 3.1.0
info:
  title: ""
paths:
  /api/apiUser:
    get:
      operationId: apiUser
      parameters:
        - in: query
          name: trophies
          schema:
            type: boolean
      responses:
        '200':
          description: The information of the user.
          content:
            application/json:
              schema:
                type: boolean
```

melloware:
@barrownicholas looks like you introduced this feature?

barrownicholas:
@melloware happy to take a crack at this!

melloware:
Ok assigned to you!

barrownicholas:
@melloware @n-e opened a PR: #2857

PR #2857, opened 2026-01-26
fix: #2839 (TypeScript error with an optional parameter and optionsParamRequired: true)

barrownicholas:
# Overview

Fixes: #2839 (TypeScript error with an optional parameter and optionsParamRequired: true)

# Detail

When `optionsParamRequired` is true, but body, header, etc., params are all optional, the `params` (and body) are optional. However, because they precede `options`, and `options` is now required, this error occurs:

``` A required parameter cannot follow an optional parameter.```

## Changed Behavior

Current Behavior:
```ts
const apiUser = (
    params?: ApiUserParams | undefined, // <-- note the '?'
    options: TTTT 
) => { ... }
```

New/Fixed Behavior:
```ts
const apiUser = (
    params: ApiUserParams | undefined, // <-- note the **absence** of '?'
    options: TTTT 
) => { ... }
```

### DIFF
--- a/docs/content/docs/reference/configuration/output.mdx
+++ b/docs/content/docs/reference/configuration/output.mdx
@@ -1195,7 +1195,7 @@ Enable URL encoding of path/query parameters.
 **Type:** `Boolean`
 **Default:** `false`
 
-Make the `options` parameter required.
+Make the `options` parameter required. Since the `options` parameter appears last in the parameter-set, any preceding parameters will also be required.
 
 ### propertySortOrder
 


packages/axios/src/index.ts
barrownicholas, line 1: This was a separate bug; it should have been fixed when `optionsParamRequired` was introduced, but I missed it.

melloware:
Looks like a merge conflict with the new docs

barrownicholas:
@melloware resolved that conflict!

PR #2874, opened 2026-01-28
fix(core): support contentMediaType in allOf/oneOf/anyOf for multipart/form-data (#2873)

### LINKED ISSUE #2873: using allOf in responseBody results in wrong types
### Description

When allOf is used to merge two schemas for request body params, some types are produced wrong. In my case it is `string` instead of `Blob` for string/binary type.

Normal case:
```yaml
        post:
            operationId: post_api_client_update
            parameters:
                -   name: client
                    in: path
                    required: true
                    schema:
                        type: string
            requestBody:
                required: true
                content:
                    multipart/form-data:
                        schema:
                            properties:
                                logo:
                                    type: string
                                    format: binary
                            type: object
```
Result:
```ts
export type PostApiClientUpdateBody = {
  logo?: Blob;
};
```

Error case:
```yaml
            operationId: post_api_client_update
            parameters:
                -   name: client
                    in: path
                    required: true
                    schema:
                        type: string
            requestBody:
                required: true
                content:
                    multipart/form-data:
                        schema:
                            allOf:
                                -   $ref: '#/components/schemas/ClientUpdateDto'
                                -   properties:
                                        logo:
                                            type: string
                                            format: binary
                                    type: object
```

```ts
import type { ClientUpdateDto } from './clientUpdateDto';

export type PostApiClientUpdateBody = ClientUpdateDto & {
  logo?: string;
};
```

---
Second problem: if orval is configured with `useNullsForOptional: true`, the model that is in allOf gets double "| null" type:

```ts
export interface ClientUpdateDto {
  name: string;
  /** @nullable */
  slug?: string | null | null;
  /** @nullable */
  website?: string | null | null;
}
```

```ts
export default defineConfig(<ConfigExternal>{
    dash: {
        input: './src/api/openapi.yaml',
        output: {
            target: './src/api/api.ts',
            schemas: './src/api/models',
            client: 'vue-query',
            override: {
                useNullForOptional: true,

            }
        },
    },
})
```

### Environment

```
  System:
    OS: Linux 6.6 Ubuntu 24.04.2 LTS 24.04.2 LTS (Noble Numbat)
    CPU: (16) x64 Intel(R) Core(TM) Ultra 9 285H
    Memory: 40.30 GB / 46.77 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  npmPackages:
    @tanstack/vue-query: ^5.92.9 => 5.92.9 
    orval: ^8.1.0 => 8.1.0 
    vue: ^3.4.21 => 3.4.30 
```

superxiao:
## Summary

Redesign multipart/form-data type generation to properly support OAS 3.1 schema composition.

**Problem**: `contentMediaType` was ignored when schemas used `allOf`, `oneOf`, or `anyOf` - file fields generated as `string` instead of `Blob`.

**Root cause**: The previous approach only handled schemas with direct `properties`. Schema composition (allOf/oneOf/anyOf) doesn't have direct properties - they're composed from member schemas.

**Solution**: Introduce `FormDataContext` discriminated union that tracks state during schema traversal:
- `{ atPart: false, encoding }` - navigating schema structure, carrying encoding map
- `{ atPart: true, partContentType }` - at a multipart part, ready for file type detection

## Changes

- **New `FormDataContext` type** in `object.ts` - state machine for form-data traversal
- **Context flows through entire schema resolution chain**: `resolveObject` → `getObject` → `combineSchemas` → `getArray` → `getScalar`
- **State transition at property iteration**: `atPart: false` → `atPart: true` with encoding lookup
- **Proper precedence**: Encoding Object's `contentType` > Schema Object's `contentMediaType`
- **Nested properties handled correctly**: Object fields serialize as JSON, nested `contentMediaType` ignored
- **Extract `utils/content-type.ts`**: `isBinaryContentType`, `getFormDataFieldFileType`

## Test Coverage

- Comprehensive test: encoding, contentMediaType, format binary, base64, arrays, nested properties, $ref
- Schema composition test: anyOf at root with nested oneOf, allOf with $ref, array containing oneOf

Fixes #2873

### DIFF
--- a/packages/core/src/getters/res-req-types.ts
+++ b/packages/core/src/getters/res-req-types.ts
@@ -1,7 +1,7 @@
 import { keyword } from 'esutils';
 import { uniqueBy } from 'remeda';
 
-import { createTypeAliasIfNeeded, resolveObject } from '../resolvers/object';
+import { resolveObject } from '../resolvers/object';
 import { resolveExampleRefs, resolveRef } from '../resolvers/ref';
 import {
   type ContextSpec,
@@ -14,7 +14,6 @@ import {
   type OpenApiResponseObject,
   type OpenApiSchemaObject,
   type ResReqTypesValue,
-  type ScalarValue,
 } from '../types';
 import { camel } from '../utils';
 import { isReference } from '../utils/assertion';
@@ -25,7 +24,6 @@ import {
 } from '../utils/content-type';
 import { getNumberWord } from '../utils/string';
 import type { FormDataContext } from './object';
-import { getObject } from './object';
 
 const formDataContentTypes = new Set(['multipart/form-data']);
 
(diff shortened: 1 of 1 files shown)


melloware:
looks like tests are failing?

melloware (approved):
This is a great refactor!

Issue #2876, opened 2026-01-29
OpenAPI `discriminator.propertyName` not reflected in generated `oneOf` member types


yekver:
### Description

After upgrading to v8, schemas using `oneOf` + `discriminator.propertyName` no longer get the discriminator property included in the generated member types. In v7 Orval generated a required `type` property for the `oneOf` variants; in v8 that property disappears.

This breaks narrowing on discriminated unions in TS.

### Configuration (orval.config)

```ts
tsconfig: './tsconfig.json'
client: 'axios',
httpClient: 'axios'
mode: 'tags-split'
```

### Environment

Orval: v8.1
OpenAPI: 3.0.3
NodeJS: 22.22

### Expected behavior

```ts
export type Cat = AnimalBase & {
  livesLeft: number;
} & Required<Pick<AnimalBase & { livesLeft: number }, 'livesLeft'>>;
```

### Actual behavior

```ts
export type Cat = AnimalBase & {
  livesLeft: number;
  type: 'CAT';
} & Required<Pick<AnimalBase & { livesLeft: number; type: 'CAT'; }, 'livesLeft' | 'type'>>;
```

### OpenAPI document (minimal, if applicable)

```yaml
openapi: 3.0.3
info:
  title: Discriminator regression reproduction 
  version: 1.0.0
components:
  schemas:
    Animal:
      type: object
      description: Base union
      discriminator:
        propertyName: type
        mapping:
          DOG: '#/components/schemas/Dog'
          CAT: '#/components/schemas/Cat'
        oneOf:
          - $ref: '#/components/schemas/Dog'
          - $ref: '#/components/schemas/Cat'

    AnimalBase:
      type: object
      description: Base union without discriminator
      properties:
        type:
          type: string

    Cat:
      allOf:
        - $ref: '#/components/schemas/AnimalBase'
      type: object
      required:
        - livesLeft
      properties:
        livesLeft:
          type: integer
          description: Number of lives left
          minimum: 0

    Dog:
      allOf:
        - $ref: '#/components/schemas/AnimalBase'
      type: object
      required:
        - breed
      properties:
        breed:
          type: string
          description: Breed of the dog 
paths: {}

```

### Additional context

Without `type`, code like `if (animal.type === 'CAT') { ... }` either fails to compile or loses type narrowing.

melloware:
@froggy1014 can you investigate this one?

froggy1014:
@melloware 

alright. assign me !

[...]

PR #2881, opened 2026-01-29
fix(core): dedupeUnionType only splits on top-level | #2880

### LINKED ISSUE #2880: dedupeUnionType corrupts inline combined types
### Description

When using v8's default inline combined types (no intermediate type aliases), the generated TypeScript can have syntax errors. This is because `dedupeUnionType` splits on all `|` characters, including those inside object types.

For example, an anyOf with two similar object branches:

```yaml
anyOf:
  - type: object
    properties:
      channel:
        enum: [email, sms]
  - type: object
    properties:
      channel:
        enum: [email, sms]
      id:
        type: integer
```

Generates corrupted output like:

```typescript
{ channel?: 'email' | 'sms' } | { channel?: 'email' | 'sms'; id: number }
// After dedupeUnionType incorrectly splits on nested |:
{ channel?: 'email' | 'sms' } | 'sms'; id: number }  // // Garbage - fragments deduped incorrectly
```

The fix is to only split on top-level `|`, tracking bracket depth to skip `|` inside `{}`, `()`, `[]`, and `<>`. This wasn't an issue in v7 because intermediate type aliases meant `dedupeUnionType` only saw simple type names like `ResponseAnyOf | ResponseAnyOfTwo`, never nested structures.

```ts
// paste your orval.config file here
```

### Environment

### Expected behavior

_No response_

### Actual behavior

_No response_

```yaml
openapi: 3.0.3
info:
  title: Demo
  version: 0.0.0
paths: {}
```

### Additional context

_No response_

superxiao:
Previously, dedupeUnionType split on all | characters, corrupting inline combined types containing nested unions like { a: 'x' | 'y' }. Now tracks bracket depth to only split at the top level.

Fixes #2880

### DIFF
--- a/packages/core/src/utils/string.test.ts
+++ b/packages/core/src/utils/string.test.ts
@@ -22,4 +22,27 @@ describe('dedupeUnionType', () => {
       expect(dedupeUnionType('A | B | A | C | B')).toBe('A | B | C');
     });
   });
+
+  describe('nested structures', () => {
+    it('should only split on top-level |', () => {
+      expect(dedupeUnionType("{ a: 'x' | 'y' } | { b: 'z' }")).toBe(
+        "{ a: 'x' | 'y' } | { b: 'z' }",
+      );
+      expect(dedupeUnionType("('a' | 'b')[] | string")).toBe(
+        "('a' | 'b')[] | string",
+      );
+      expect(dedupeUnionType("Array<'a' | 'b'> | string")).toBe(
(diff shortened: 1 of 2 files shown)


melloware:
@superxiao do you think it fixes this as well: https://github.com/orval-labs/orval/issues/2876 ?

superxiao:
No, that bug is not relevant to this. @melloware

melloware:
thank you !

PR #2882, opened 2026-01-29
fix(bug): CVE-2026-25141 jsStringEscape also escape JS comments `/*` and `*/`

melloware:
cc @k14uz @progfay thoughts?

### DIFF
--- a/packages/core/src/utils/string.ts
+++ b/packages/core/src/utils/string.ts
@@ -225,11 +225,13 @@ export function escape(str: string | null, char = "'") {
  * @param input String to escape
  */
 export function jsStringEscape(input: string) {
-  return input.replaceAll(/["'\\\n\r\u2028\u2029]/g, (character) => {
+  return input.replaceAll(/["'\\\n\r\u2028\u2029/*]/g, (character) => {
     switch (character) {
       case '"':
       case "'":
-      case '\\': {
+      case '\\':
+      case '/':
+      case '*': {
         return '\\' + character;
       }
       // Four possible LineTerminator characters need to be escaped:
--- a/packages/core/src/utils/string.test.ts
+++ b/packages/core/src/utils/string.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'vitest';
 
-import { dedupeUnionType } from './string';
+import { dedupeUnionType, jsStringEscape } from './string';
 
 describe('dedupeUnionType', () => {
   describe('edge cases', () => {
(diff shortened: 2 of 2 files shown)


packages/core/src/utils/string.ts
```diff
@@ -225,11 +225,13 @@ export function escape(str: string | null, char = "'") {
...
+  return input.replaceAll(/["'\\\n\r\u2028\u2029/*]/g, (character) => {
     switch (character) {
       case '"':
       case "'":
-      case '\\': {
+      case '\\':
+      case '/':
+      case '*': {
         return '\\' + character;
       }
```
k14uz: After giving this more thought, the current list wont escape characters like `/** hello */ }; JSFUCK; const a = { /* */` it scapes: `"'\ \n \r \u2028 \u2029 / *`but DOESN'T escape: { } ; [ ] ( ) ! (you can't directly target these). This seems to introduce more holes than closing up the gates. There are 2 options that come to mind at this point that might worth pondering and exploring more:
1. Stop using descriptions in comments entirely, something like:
```ts
export function getEnumImplementation(...) {
  // don't generate comments from x-enumDescriptions at all
  const comment = ''; // REMOVED
  
  return [...new Set(value.split(' | '))].reduce((acc, val, index) => {
    const name = names?.[index];
    // no description comment
    return acc + `  ${key}: ${val},\\n`;
  }, '');
}
```

2. Separate Metadata:
```ts
// generate enum WITHOUT descriptions
export const MyEnum = {
  ACTIVE: 'ACTIVE',
  PENDING: 'PENDING',
} as const;

// store descriptions separately as data (not code)
export const MyEnumMeta = {
  ACTIVE: { description: "User description here" },
  PENDING: { description: "Another description" },
} as const;
```

At this point it's worth mentioning that the JSF bypass cannot be prevented by escaping individual characters since it uses JS's structural syntax which are legitimate TS tokens. Attempting to escape these would break valid code generation. I think it's a wild hunt going after individual tokens (?). @melloware what do you think?

packages/core/src/utils/string.ts
```diff
@@ -225,11 +225,13 @@ export function escape(str: string | null, char = "'") {
...
 export function jsStringEscape(input: string) {
-  return input.replaceAll(/["'\\\n\r\u2028\u2029]/g, (character) => {
+  return input.replaceAll(/["'\\\n\r\u2028\u2029/*]/g, (character) => {
     switch (character) {
       case '"':
       case "'":
-      case '\\': {
+      case '\\':
+      case '/':
+      case '*': {
```
progfay, line 234: This change may cause the RegExp feature of `@orval/zod` to break.
https://github.com/orval-labs/orval/blob/6271e0269fe5c3ddb18240dbe41fd661e13ccb12/packages/zod/src/index.ts#L673

e.g.: `A*`

progfay (commented):
Thank you for the quick fix! I’ve confirmed that this specific vulnerability is now addressed.

However, I would like to suggest a more fundamental improvement for the code generation logic. Currently, the code is constructed using manual string concatenation and custom escape functions. This approach is often prone to edge cases, and there is a risk that other unescaped characters could still lead to similar issues in the future.

To make the code generation inherently secure, I highly recommend using the TypeScript Compiler API to build and print the code as an AST. This ensures that all syntax and escaping are handled correctly by the official compiler logic, rather than relying on manual string manipulation.

You can find more details and examples here: https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API#creating-and-printing-a-typescript-ast

This change would significantly improve the long-term maintainability and security of this package.

melloware:
Unfortunately changing away from string concatenation is a larger more invasive change across this large code base so to me that isn't really an option right now. 

Thanks for the report and the review!

packages/core/src/utils/string.ts
progfay: I was wrong!
It seems to be fine!

k14uz (approved):

PR #2914, opened 2026-02-03
Branded Types Support

### LINKED ISSUE #1222: Support replacing types with custom ones when transforming the input
(Not using the standard issue template as this is a feature request)

I'm migrating from `openapi-typescript` to orval. In my current setup I've added what could be called "nominal types" - so for instance instead of `type userId = number` it would use `type userId = UserId` with the following implementation:

```ts
const UserIdSymbol = Symbol('UserId');
type UserId = NominalType<typeof UserIdSymbol>;
```

Then I have a transformer on the API TS generation that modifies all instances of a `userId` to use this custom type, and injects an `import {UserId} from './types'`

Such that we end up with something like:

```ts
// Generated types from openapi-typescript
import { UserId } from './custom-types';

// ...

export type UserModel = {
  id: UserId;
  // ...
};
```

Arguments about the use-cases of this approach and such aside, would this be possible with `orval`?

I'm trying something like this in the orval config:

```ts
input: {
  override: {
    transformer: (api) => {
      // Just testing as an example
      const schema = api.components?.schemas?.ExampleSchema;

      if (schema && 'properties' in schema && schema.properties?.version && 'type' in schema.properties.version) {
          schema.properties.userId.type = UserIdSymbol as any;
      }

      return api;
    },
  },
},
```

But this just generates the type `userId: unknown` - so ideally there would be a supported way to do this.

When using `openapi-typescript` I just transform the type and basically set it to `"UserId"`, so that it then appears as the string content. And I inject the `import` statement at the top of the file manually (they have a way to inject anything at the top of the file)

spaansba:
## Branded Types Support

Fix #1222 

Adds branded type support based on [issue #1222](https://github.com/orval-labs/orval/issues/1222). It won't solve the full custom types problem (like importing from external modules), but it handles the branded/nominal types pattern.

You mark fields in your OpenAPI spec with `x-brand` (custom extension)

```yaml
 properties:                                                                                                                                                                                 
   id:
     type: integer
     x-brand: PetId
   email:
     type: string
     x-brand: Email
```

Enable it in your config:

```typescript
output: {
  override: {
    useBrandedTypes: true,
  },
}
```

And you get:

```typescript
type Branded<BaseType, Brand> = BaseType & { readonly __brand: Brand };

export type PetId = Branded<number, 'PetId'>;
export type Email= Branded<string, 'Email'>;

export interface Pet {
  id: PetId;
  email: Email;
}
```

For mocks you will get: 

```typescript
id: faker.number.int() as PetId,
```

It also works with zod as zod also has [branded types](https://zod.dev/api#branded-types): 

```typescript
id: zod.number().brand<'PetId'>(),
```

Now you can't accidentally pass a regular number where a PetId is expected, or mix up different ID types even though they're all numbers at runtime.

### DIFF
--- a/tests/configs/default.config.ts
+++ b/tests/configs/default.config.ts
@@ -542,20 +542,7 @@ export default defineConfig({
       },
     },
   },
-  'branded-types-with-dates': {
-    input: {
-      target: '../specifications/branded-types.yaml',
-    },
-    output: {
-      target: '../generated/default/branded-types-with-dates/endpoints.ts',
-      schemas: '../generated/default/branded-types-with-dates/model',
-      override: {
-        useBrandedTypes: true,
-        useDates: true,
-      },
-    },
-  },
-  'branded-types-with-bigint': {
+  'branded-types-with-type-generation': {
     input: {
       target: '../specifications/branded-types.yaml',
     },
@@ -565,6 +552,20 @@ export default defineConfig({
       override: {
         useBrandedTypes: true,
         useBigInt: true,
+        useDates: true,
+      },
+    },
+  },
+  'branded-types-tags-split': {
+    input: {
+      target: '../specifications/branded-types.yaml',
+    },
+    output: {
+      target: '../generated/default/branded-types-tags-split/endpoints.ts',
+      schemas: '../generated/default/branded-types-tags-split/model',
+      mode: 'tags-split',
+      override: {
+        useBrandedTypes: true,
       },
     },
   },
--- a/tests/configs/fetch.config.ts
+++ b/tests/configs/fetch.config.ts
@@ -397,4 +397,17 @@ export default defineConfig({
       target: '../specifications/petstore.yaml',
     },
   },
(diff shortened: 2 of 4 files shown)


melloware:
@spaansba can you update the docs too and find a place to document this what the use case is and how you use it?

spaansba:
> @spaansba can you update the docs too and find a place to document this what the use case is and how you use it?

Done!

A question from my side to you or other maintainers:

For collecting branded types I'm using a `Map` that gets passed through the `context` and populated during schema resolution. I do this to not recreate the same type and prevent type collisions. Is there something similar you already do elsewhere that I could hook into, or is this approach sufficient?

Also, thoughts on defaulting `generateBrandedTypes` to `true`? Adding `x-brand` to a spec is already intentional, so the config option feels a bit redundant.

melloware:
I feel liek the config always gives us a hook to disable it and it should be false by default lets make people be intentional about using this feature

Issue #2846, opened 2026-01-23
v8 migration with react-query, fetch, and custom client fails to generate valid code


ZebraFlesh:
### Description

Using Orval v7, we have a custom client that pulls an auth token from React context and shoves it into a request header. Everything uses native fetch; no axios in sight. Upon upgrading to v8, this just completely falls apart and Orval generates broken code the references non-existent custom hooks. (Those hooks were previously generated by Orval v7.)

Our custom client has this format:
```
export const useCustomClient = <T>({ ...config object... }): Promise<T> => {
  const { user } = useAuth();  // react hook from oidc-react-context package
  return async (config) => {
    const headers = new Headers(config.headers ?? {});
    if (user?.access_token) headers.set('Authorization', `Bearer: ${user.access_token}`);
    return await fetch(... details skipped...);
  }
}

export default useCustomClient;
```

This seems to just not work under v8. As best I can tell, v8 wants the custom client signature to be:

```
export const customClient = async<T>(url, config): Promise<T> { ...  details skipped ...}

export default customClient
```

But if that's true, then you can no longer use a React hook to obtain an access_token that is stored in React context. (React complains that the hook is being called from an async function instead of from a React component.) As best I can tell, the examples all [skip over where a token comes from](https://github.com/orval-labs/orval/blob/master/samples/react-query/custom-fetch/src/custom-fetch.ts#L35), but in React it's almost certain to come from React context.

What's the native fetch + React + React context solution? The only path forward I can see is to use axios, which I did not have to do under Orval v7.

### Configuration (orval.config)

```ts
import type { Config } from 'orval';

const config: Config = {
  client: {
    input: {
      target: './docs/my-api.yaml',
    },
    output: {
      clean: true,
      client: 'react-query',
      mode: 'tags-split',
      override: {
        mutator: {
          name: 'useCustomClient',
          path: '../mutator.ts',
        },
        query: {
          useInfinite: true,
          useMutation: true,
          usePrefetch: true,
          useQuery: true,
        },
        useTypeOverInterfaces: true,
      },
      prettier: true,
      schemas: './model',
      target: './handlers',
      workspace: '__generated__',
    },
  },
};

export default config;
```

### Environment

System:
  OS: macOS 15.7.3
  CPU: (12) arm64 Apple M2 Max
  Memory: 871.69 MB / 64.00 GB
  Shell: 5.9 - /bin/zsh
npmPackages:
  @tanstack/react-query: 5.90.18 => 5.90.18
[...]

PR #2919, opened 2026-02-06
DRAFT: refactor(query): introduce FrameworkAdapter pattern, eliminate redundant flags 

ScriptType:
> **Draft PR — looking for reviews and thoughts before finalizing.**
> This is a significant internal refactoring of `@orval/query`. The generated output is unchanged, but the generator internals are restructured. I'd appreciate feedback on the adapter interface design, naming, and whether anything was missed before I clean this up for merge.

Inspired by https://github.com/orval-labs/orval/issues/2846

### Problem

The `@orval/query` package generates TanStack Query code for 5 frameworks (React, Vue, Angular, Svelte, Solid) from a single codebase. All framework differences were handled via inline `if/else` chains threading boolean flags (`hasSvelteQueryV4`, `hasSvelteQueryV6`, `hasQueryV5`, `isAngularClient`, `isVue(outputClient)`, `isAngularHttp`, etc.) through deeply nested function calls. This made the code hard to follow — understanding what gets generated for a given framework required tracing flags through multiple indirection layers across several files.

### Solution

Introduce a `FrameworkAdapter` interface that encapsulates all framework-specific behavior behind a single dispatch point. Each framework gets its own adapter file in `packages/query/src/frameworks/`. The shared generators (`query-generator.ts`, `mutation-generator.ts`, `query-options.ts`) call adapter methods instead of branching on booleans.

A `withDefaults()` factory provides default implementations for 21 defaultable fields that are identical across the majority of frameworks, so each adapter only defines what makes it unique:

| Adapter | Methods defined | Methods from defaults |
|---------|----------------:|----------------------:|
| React   | 10              | 19                    |
| Solid   | 9               | 20                    |
| Vue     | 19              | 10                    |
| Svelte  | 16              | 13                    |
| Angular | 20              | 9                     |

### What changed

**New files:**
- `framework-adapter.ts` — `FrameworkAdapter` interface (6 readonly properties + 29 methods), context types for method arguments, and `FrameworkAdapterConfig` type with 21 defaultable fields
- `frameworks/index.ts` — `createFrameworkAdapter()` factory that resolves version flags and selects the right adapter; `withDefaults()` factory that fills in common implementations
- `frameworks/react.ts`, `solid.ts`, `vue.ts`, `svelte.ts`, `angular.ts` — one adapter per framework, each defining only its unique behavior

**Simplified files:**
- `query-generator.ts` — removed all flag-threading and `if (isAngular) / if (isVue)` branches; calls `adapter.*` methods directly. `generateMutationHook` simplified from ~12 params to 6
- `mutation-generator.ts` — removed `outputClient`, `hasQueryV5`, `hasSvelteQueryV4`, `hasSvelteQueryV6`, `isAngularHttp` from `MutationHookContext`; uses adapter properties directly
- `query-options.ts` — `getQueryOptionsDefinition` takes `prefix: string` instead of `hasSvelteQueryV4 + isAngularClient`. Deleted old `generateQueryArguments` function (dead code — all callers use `adapter.generateQueryArguments()`)
- `index.ts` — creates the adapter via `createFrameworkAdapter()` and passes it through

### Key design decisions

1. **`isAngularHttp` absorbed into adapter** — previously threaded as a cross-cutting concern. Investigation showed it's always `true` for Angular and `false` for all others, so it became a readonly adapter property with two methods (`getHttpFirstParam`, `getMutationHttpPrefix`) replacing inline computations.

2. **`forAngularInject` removed from interface** — the Angular adapter determines inject-style wrapping internally based on `type` and `forQueryOptions`, rather than receiving an external flag.

3. **`prefix` instead of boolean combinations** — `getQueryOptionsDefinition` now takes a `prefix: string` (`'Use'` or `'Create'`) from `adapter.getQueryOptionsDefinitionPrefix()`, replacing the `hasSvelteQueryV4 + isAngularClient` combination.

4. **Factory defaults over duplication** — `withDefaults()` provides implementations for methods shared across 3+ frameworks (identity transforms, common string builders, standard `enabled` option generation, etc.), so adapters remain focused on their unique behavior.

### DIFF
--- a/packages/query/DEVELOPMENT.md
+++ b/packages/query/DEVELOPMENT.md
@@ -1,0 +1,70 @@
+# Developing the Query Package
+
+Developer guidelines for working on `@orval/query` — the TanStack Query generator that targets five frameworks (React, Vue, Angular, Svelte, Solid) from a single codebase via the `FrameworkAdapter` strategy pattern.
+
+## How to Change Framework Behavior
+
+### Change one framework
+
+Edit its adapter file in `src/frameworks/<name>.ts`. The adapter method you override will only affect that framework.
+
+### Change all frameworks the same way
+
+Edit the default implementation in `withDefaults()` (`src/frameworks/index.ts`), or the shared generator that calls the adapter method (`src/query-generator.ts` / `src/mutation-generator.ts`).
+
+### Change all frameworks but one needs special handling
+
+Edit `withDefaults()` for the common case. Then override the method in the divergent adapter's file in `src/frameworks/`.
+
+## How to Add or Modify Adapter Methods
+
+**New defaultable method:** Add to `FrameworkAdapter` interface in `src/framework-adapter.ts`, add to `DefaultableFields` type, implement the default in `withDefaults()`. Adapters that need different behavior override it in their file.
+
+**New required method:** Add to `FrameworkAdapter` interface (NOT to `DefaultableFields`). Must be implemented in every adapter file.
+
+**Rule of thumb:** If most frameworks share the same implementation, make it defaultable. Only use a required method when each framework genuinely needs a distinct implementation.
+
+## How the `prefix` Pattern Works
+
+`src/query-options.ts` uses a string `prefix` (`'Use'` or `'Create'`) instead of framework booleans. Each adapter returns it from `getQueryOptionsDefinitionPrefix()`. This avoids framework conditionals inside shared code — the adapter decides the prefix, shared code just uses it.
+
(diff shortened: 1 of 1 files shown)


ScriptType:
@melloware Can we get some heavy weights in ? :)

melloware:
I love this idea i added all the core contributors on the review

AllieJonsson:
I think this is great!

melloware (dismissed):

melloware:
Awesome refactor

PR #2921, opened 2026-02-07
feat: generate separate functions for each requestBody content-type

### LINKED ISSUE #2812: multiple content type header caused incorrect generated code
### Description

It seems that if an operation has 2 content type headers, the generated code is ignoring the content type header

for example,

```
/v2/profile:
    patch:
      operationId: ABCD
      parameters: []
      responses:
        '200':
          description: The request has succeeded.
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/User'
      requestBody:
        required: true
        content:
          multipart/form-data:
            schema:
              $ref: '#/components/schemas/A'
          application/json:
            schema:
              $ref: '#/components/schemas/B'
```

the generated code are the following:

```
export const apiPatchV2Profile = (
  apiPatchV2ProfileBodies:
    | ASchema
    | BSchema,
  options?: SecondParameter<typeof customInstance>,
) => {
  return customInstance<UserSchema>(
    { url: `/v2/profile`, method: 'PATCH', data: apiPatchV2ProfileBodies },
    options,
  );
};
```

as you can see, the `content-type: multipart/form-data` is not accounted for.

```ts
output: {
      mode: 'split',
      prettier: true,
      client: 'react-query',
      target: outputTarget,
      override: {
        mutator: {
          path: './src/shared/api/customInstance.ts',
          name: 'customInstance',
        },
        formData: {
          path: './src/shared/api/customFormDataFn.ts',
          name: 'customFormDataFn',
        },
        useTypeOverInterfaces: true,
      },
    },
```

### Environment

  System:
    OS: macOS 15.4.1
    CPU: (8) arm64 Apple M2
    Memory: 151.75 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    @tanstack/react-query: ^5.90.17 => 5.90.17 
    axios: ^1.13.2 => 1.13.2 
    orval: 7.5.0 => 7.5.0 
    react: 18.3.1 => 18.3.1 
    zod: ^3.25.76 => 3.25.76 

### Expected behavior

Maybe there should be 2 functions for this?
`apiPatchV2ProfileWithApplicationJson` and `apiPatchV2ProfileWithMultipartFormData`

### Actual behavior

only `apiPatchV2Profile` is generated while the one that is multipart form data is missing

```yaml
paths: {
  /v2/profile:
    patch:
      operationId: ABCD
      parameters: []
      responses:
        '200':
          description: The request has succeeded.
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/User'
      requestBody:
        required: true
        content:
          multipart/form-data:
            schema:
              $ref: '#/components/schemas/A'
          application/json:
            schema:
              $ref: '#/components/schemas/B'
}
```

### Additional context

_No response_

froggy1014:
## Summary                                                                                                                                         
                                                                                                                                                     
Fixes #2812                                                                                                                                        
                                                                                                                                                     
Generates separate functions for each `requestBody` content-type instead of a single function with union type.                                     
                                                                                                                                                     
## Changes                                                                                                                                         
                                                                                                                                                     
**Before:**                                                                                                                                        
```typescript                                                                                                                                      
const updateProfile = (body: ProfileFormData | ProfileJson) => { ... }                                                                             
```
                                                                                                                                                     
**After:**                                                                                                                                             
```ts
const updateProfileWithFormData = (body: ProfileFormData) => {                                                                                     
  // ... proper FormData handling                                                                                                                  
}                                                                                                                                                  
                                                                                                                                                     
const updateProfileWithJson = (body: ProfileJson) => {                                                                                             
    return axiosInstance.patch('/profile', body);                                                                                                    
}                                                                                                                                                  
```                                                                                                                                                     
                                                                
                                                                                                                                                                                                                                                                                                                                                                                
  Test                                                                                                                                               
                                                                                                                                                     
  See reproduction [repo](https://github.com/froggy1014/2812)

### DIFF
--- a/packages/core/src/generators/verbs-options.ts
+++ b/packages/core/src/generators/verbs-options.ts
@@ -59,7 +59,10 @@ function getContentTypeSuffix(contentType: string): string {
     'text/plain': 'Text',
   };
 
-  return contentTypeMap[contentType] || pascal(contentType.replace(/[^a-zA-Z0-9]/g, '_'));
+  return (
+    contentTypeMap[contentType] ||
+    pascal(contentType.replace(/[^a-zA-Z0-9]/g, '_'))
+  );
 }
 
 export async function generateVerbOptions({
@@ -272,7 +275,10 @@ export function generateVerbsOptions({
     _filteredVerbs(verbs, input.filters),
     async (acc, [verb, operation]: [string, OpenApiOperationObject]) => {
       if (isVerb(verb)) {
-        const contentTypes = getRequestBodyContentTypes(operation.requestBody, context);
+        const contentTypes = getRequestBodyContentTypes(
+          operation.requestBody,
+          context,
+        );
 
         // If there are multiple content types, generate a separate operation for each
         if (contentTypes.length > 1) {


BramMusters:
This is a breaking change and results in significant errors in our codebase, since we are importing the "old" functions that existed in 8.2.0 and disappeared after upgrading to 8.3.0. 
Not sure how I should handle this..

boompttn:
@froggy1014 It seems that the generated functions still have bugs in it

The generated `WithFormData` doesn't have the content type headers

boompttn:
here's the repo https://github.com/boompikachu/bug-demo-orval

boompttn:
something to note, if there's a mutator / custom instance, there will be a header for `withJson` (application/json) but not for `withFormData`

https://github.com/boompikachu/bug-demo-orval/tree/with-custom-instance

Arnagos:
Is there a way to disable this behavior? This is a giant breaking change. And generally, having a WithJson, etc. suffix is not a desired behavior, in my opinion.

melloware:
@froggy1014 we should make this a configuration switch if pe`ople do want it the old way `const updateProfile = (body: ProfileFormData | ProfileJson) => { ... }

Arnagos:
@melloware Since this is already merged and there hasn't been a response from the original author, could you please re-open the issue and revert the PR? I would love to use the newer versions of Orval (especially because of the export order bug in 8.2), but this PR is preventing that.

melloware:
@Arnagos this bug was reverted try 8.4.2 and i reopened the original ticket?

Arnagos:
> @Arnagos this bug was reverted try 8.4.2 and i reopened the original ticket?

Oh, you're way faster than I expected. Thanks a lot!

Issue #2912, opened 2026-02-03
Disabling index files creates incorrect import paths for response and request body schemas


RinseV:
### Description

If you choose to [disable index files](https://orval.dev/docs/reference/configuration/output#indexfiles) in your generated code but your OpenAPI schemas references either `response` or `requestBody` `components`, the import paths for these generated models will be incorrect.

### Configuration (orval.config)

```ts
import { defineConfig } from 'orval';

export default defineConfig({
  petstore: {
    output: {
      mode: 'tags-split',
      target: 'gen/petstore.ts',
      schemas: 'gen/models',
      client: 'react-query',
      baseUrl: 'http://localhost:8000',
      clean: true,
      prettier: true,
      indexFiles: false,
    },
    input: {
      target: './petstore.yaml',
    },
  },
});

```

### Environment

Run the command above and paste the output here.
```
System:
    OS: macOS 15.7.3
    CPU: (8) arm64 Apple M3
    Memory: 212.16 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    @tanstack/react-query: ^5.90.20 => 5.90.20 
    orval: ^8.2.0 => 8.2.0 
```

### Expected behavior

The generated code should use the correct file names for importing the response and request body components.

### Actual behavior

The generated code strips out the last part of the generated file names causing an incorrect import.

### OpenAPI document (minimal, if applicable)

```yaml
openapi: '3.1.0'
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
servers:
  - url: http://petstore.swagger.io/v1
paths:
  /pets:
    get:
      summary: List all pets
      operationId: listPets
      tags:
        - pets
      parameters:
        - name: limit
          in: query
          description: How many items to return at one time (max 100)
          required: false
          schema:
            type: string
        - $ref: '#/components/parameters/offset'
      responses:
        '200':
          description: A paged array of pets
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/Pet'
        '404':
          $ref: '#/components/responses/NotFound'
    post:
      summary: Create a pet
      operationId: createPets
      tags:
        - pets
      requestBody:
        required: true
        content:
          application/json:
            schema:
              type: array
              items:
                type: object
                required:
                  - 'name'
                  - 'tag'
                properties:
                  name:
                    type: string
                  tag:
                    type: string
      responses:
        '200':
          description: Created Pet
          content:
[...]

PR #2924, opened 2026-02-08
chore: enable tsdown auto-generated package exports

froggy1014:
## Summary
- Enable `exports: true` in tsdown base config to auto-generate `exports`, `main`, `module`, `types` fields in package.json
- Normalize all package exports from manual nested conditions to tsdown-managed format
- Fixes incorrect exports structure (missing `"."` subpath) in some packages

## Motivation
Previously, each package manually defined `exports` with inconsistent formats — some used nested `import.types` conditions, others used flat strings. With tsdown's `exports` option, these fields are auto-generated on build, ensuring consistency across all packages and reducing manual maintenance.

### Reference

- https://tsdown.dev/options/package-exports

### DIFF
--- a/packages/core/src/generators/verbs-options.ts
+++ b/packages/core/src/generators/verbs-options.ts
@@ -5,6 +5,7 @@ import {
   getParams,
   getProps,
   getQueryParams,
+  getRequestBodyContentTypes,
   getResponse,
 } from '../getters';
 import type {
@@ -29,6 +30,7 @@ import {
   isVerb,
   jsDoc,
   mergeDeep,
+  pascal,
   sanitize,
 } from '../utils';
 import { generateMutator } from './mutator';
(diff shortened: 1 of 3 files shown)


melloware:
@froggy1014 can you check if this fixes this? https://github.com/orval-labs/orval/issues/2912 I saw your mention of incorrect export path

froggy1014:
> @froggy1014 can you check if this fixes this? #2912 I saw your mention of incorrect export path

@melloware 

This PR only changes the package-level exports field in each package.json for bundling.
 it doesn't affect the generated client code at all. 
 
 #2912 is about incorrect import paths in the generated output when indexFiles: false, which is a separate code-gen issue. This PR won't fix that.

melloware (approved):

PR #2932, opened 2026-02-09
fix(core,fetch): handle anyOf with null type in query parameters

### LINKED ISSUE #2904: Creating a circuit with explode = True for anyOf
### Description

If in the scheme
```
      - in: query
        name: affiliations
        required: false
        explode: true
        schema:
          anyOf:
          - format: uuid
            type: string
          - type: 'null'
```

Will not process due to anyOf

I had to do the implementation in the format
```
      - in: query
        name: affiliations
        required: false
        schema:
          explode: true
          items:
            $ref: '#/components/schemas/AffiliationEnum'
          nullable: true
          title: Affiliations
          type: array
```

Although both schemes are valid

```ts

export default defineConfig({
  replan: {
    input: '../../../backend/schema/v1.yaml',
    output: {
      mode: 'tags-split',
      workspace: '../../src/client/v1',
      target: 'api.ts',
      client: 'react-query',
      httpClient: 'fetch',
      override: {
        mutator: '../../../orval/v1/mutator.ts',
        fetch: {
          includeHttpResponseReturnType: false,
        },
      },
    },
    hooks: {
      afterAllFilesWrite: 'prettier --write',
    },
  },
});
```

### Environment

orval@7.17.0, orval@8.1.0, fetch

Neko1313:
Query parameters with anyOf schemas containing { type: 'null' } were not resolved as nullable. This adds nullable detection for anyOf schemas in the value resolver and extends explode parameter detection in the fetch client to support anyOf/oneOf/allOf containing array types.

Close #2904


Neko1313:
@melloware Could you watch this please

melloware:
Running the build now

PR #2939, opened 2026-02-10
fix(msw): mock generation bugs

### LINKED ISSUE #2934: MSW mock generation: invalid faker ranges, Blob typing, and @type property sanitization
### Summary

When using Orval with `client: 'angular'`, `mock: { type: 'msw' }`, and `schemas.type: 'zod'`, the generated MSW mock handler files contain several bugs that produce invalid TypeScript or runtime errors. We are having to implement post-processing workarounds for all of these in our Nx monorepo.

These are **three distinct bugs** in the MSW mock codegen, bundled here since they all affect the generated `.msw.ts` files.

### Versions / Environment

- orval: 8.2.0
- TypeScript: 5.9.3
- Angular: 20.x (`client: 'angular'`)
- @faker-js/faker: 9.x
- msw: 2.x
- pnpm: 10.28.2

```
Node           : 24.11.1
OS             : win32-x64
pnpm           : 10.28.2

nx                     : 22.4.4
@nx/angular            : 22.4.4
@nx/js                 : 22.4.4
typescript             : 5.9.3
```

We are implementing Orval in an **Nx Angular/frontend monorepo** and are having to carry custom post-processing code to work around these bugs. We'd love to remove these workarounds once Orval fixes them upstream.

### Config used

```ts
{
  input: { target: openAPIfile },
  output: {
    mode: 'split',
    target: `${specName}.client.ts`,
    schemas: { path: `${specName}.schemas`, type: 'zod' },
    client: 'angular',
    mock: { type: 'msw', delay: 0, useExamples: true },
    override: {
      zod: {
        strict: { response: true, query: true, param: true, header: true, body: true },
        generate: { param: true, body: true, response: true, query: true, header: true },
      },
    },
  },
}
```

---

### Bug 1: `faker.number.int({ min: undefined, max: undefined })` — invalid ranges

**What happens:**
When integer fields have no `minimum`/`maximum` constraints in the OpenAPI spec, Orval emits:

```ts
faker.number.int({ min: undefined, max: undefined })
faker.number.float({ min: undefined, max: undefined })
```

This is invalid for Faker.js — passing `undefined` for `min`/`max` is not the same as omitting them. With `exactOptionalPropertyTypes: true`, it also causes TypeScript compilation errors.

**Expected:**
```ts
faker.number.int()
faker.number.float()
```

**Workaround:**
Regex post-processing to remove `min: undefined` and `max: undefined` entries, and collapse empty option objects `({})` to no-argument calls `()`.

**Related:**
- #2353 — Previously reported and fixed for `multipleOf: undefined`, but `min: undefined` and `max: undefined` were not addressed in the same fix.
- Also visible in Orval's own MSW docs example: `id: faker.number.int({ min: undefined, max: undefined })`

---

### Bug 2: `Blob` response type generates invalid object literal assignments

**What happens:**
For endpoints with binary/blob responses, the MSW mock functions use `Blob` as the return type. But the mock function returns an object literal `{}`, which is not assignable to `Blob`:

```ts
export const getDownloadResponseMock = (overrideResponse?: Blob): Blob => ({
  // ... object fields — not assignable to Blob
})

export const getDownloadMockHandler = (
  overrideResponse?: Blob | ((info) => Promise<Blob> | Blob)
) => { ... }
```

TypeScript errors:
```
Type '{ ... }' is not assignable to type 'Blob'.
```

**Expected:**
For mock purposes, use `Record<string, unknown>` instead of `Blob` in mock generator/handler signatures:

```ts
export const getDownloadResponseMock = (
  overrideResponse?: Record<string, unknown>
): Record<string, unknown> => ({ ... })
```

**Workaround:**
Regex replacement of `Blob` → `Record<string, unknown>` in mock function signatures and return types.

---

### Bug 3: `@type` property name sanitized to `_type`

**What happens:**
OpenAPI schemas with property names starting with `@` (e.g., `@type` used in JSON-LD/HAL responses) are sanitized to `_type` in the generated MSW mock data:

```ts
export const getMyResourceResponseMock = () => ({
  "_type": faker.string.alpha(20),
  // ...
})
```

**Expected:**
```ts
export const getMyResourceResponseMock = () => ({
  "@type": faker.string.alpha(20),
  // ...
})
```

**Workaround:**
String replacement of `"_type"` → `"@type"` and `'_type'` → `'@type'` in generated MSW files.

**Note:** This same issue also affects Zod schema generation — see #2933.

---

### Possible Fixes

1. **Bug 1:** In `packages/mock/src/faker/getters/scalar.ts`, check that `min`/`max` values are not `undefined` before interpolating them into the options object. Similar to the fix applied for `multipleOf` in #2353.
2. **Bug 2:** In the MSW mock generator, when the response content type is `application/octet-stream` or similar binary type, use `Record<string, unknown>` instead of `Blob` for mock data signatures (since mock data is always a JSON object literal).
3. **Bug 3:** Preserve original property names (including `@`-prefixed) as string keys in generated mock object literals. Only sanitize JavaScript identifiers, not object property string keys.

### Related Issues

- #2353 — `multipleOf: undefined` fix (same pattern as Bug 1, but for `min`/`max`)
- #2930 — Angular client imports `DeepNonNullable` from `@orval/core`
- #2931 — Angular client: missing type-only imports + incorrect HttpClient params typing
- #2933 — Zod schema generation bugs (includes the same `@type` → `_type` issue)

MSW/mock-related issues (open):
- #2206 — `info` parameter in MSW handlers implicitly `any` (strict TS)
- #1853 — `required:true` not applied for nullable `anyOf` in MSW mocks

MSW/mock-related issues (closed/resolved):
- #2154 — missing type imports in MSW files with multiple content types
- #1764 — MSW file generation missing model import (also missing `JSON.stringify`)
- #2020 — MSW: `allOf` + object spread generates invalid syntax
- #1980 — MSW: `allOf` within referenced schema generates invalid syntax
- #1895 — MSW: `allOf` within `allOf` generates invalid syntax
- #2184 — MSW mock spread bug with `oneOf`
- #2580 — MSW: `date` format with `required` produces invalid mock type
- #2355 — MSW mocks break with numeric enums + `x-enumNames`

the-ult:
Closes #2934.

This PR fixes multiple MSW mock-generation bugs, modernizes handler output for MSW v2 helpers, and adds a configurable preferred content type for mixed content responses with tests and docs updates.

## Summary

- Fixes faker option objects to avoid `undefined` fields (exactOptionalPropertyTypes safe).
- Corrects binary/Blob mocks and response handling (ArrayBuffer + `HttpResponse.arrayBuffer`).
- Uses MSW v2 response helpers (`json`, `text`, `xml`, `html`, `arrayBuffer`).
- Adds `mock.preferredContentType` to choose the handler helper when multiple content types are present (fallback to spec order).
- Strengthens MSW test coverage and updates MSW docs.

### 1. `undefined` values in faker option objects
**Before:** `faker.number.int({ min: undefined, max: undefined })`
**After:** `faker.number.int()`

Options like `min`, `max`, `fractionDigits`, `multipleOf` are now only included when defined. This avoids TypeScript errors with `exactOptionalPropertyTypes` and prevents unexpected faker behavior.

Applies to: `faker.number.int()`, `faker.number.float()`, `faker.string.alpha()`, `Array.from({ length: ... })`.

### 2. Binary/Blob response handling
**Before:**
- Mock data: `new Blob(...)` — invalid in Node.js test environments
- Handler type: `overrideResponse?: Blob` — mismatches actual mock data
- Response: `new HttpResponse(JSON.stringify(blobData))` — serializes binary as JSON

**After:**
- Mock data: `new ArrayBuffer(faker.number.int({ min: 1, max: 64 }))`
- Handler type: `overrideResponse?: ArrayBuffer` (via mock return type replacement)
- Response: `HttpResponse.arrayBuffer(body, { headers: { 'Content-Type': '...' } })`

### 3. `@`-prefixed property keys (e.g. `@type`)
Already fixed in the PR’s base — the `getKey()` utility correctly quotes these as `'@type'` in generated mock objects. Verified with a dedicated test.

### MSW v2 response helpers
All generated handlers now use the idiomatic MSW v2 static helpers:

| Content type | Before | After |
|---|---|---|
| JSON (`application/json`) | `new HttpResponse(JSON.stringify(data), { headers: ... })` | `HttpResponse.json(data, { status })` |
| Text (`text/plain`, `text/*`, `*+xml`) | `new HttpResponse(data, { headers: ... })` | `HttpResponse.text(textBody, { status })` |
| XML (`application/xml`, `*+xml`) | `new HttpResponse(data, { headers: ... })` | `HttpResponse.xml(textBody, { status })` |
| HTML (`text/html`) | `new HttpResponse(data, { headers: ... })` | `HttpResponse.html(textBody, { status })` |
| Binary (`application/octet-stream`, `image/*`, `audio/*`, `video/*`, `font/*`, `application/pdf`) | `new HttpResponse(JSON.stringify(data), ...)` | `HttpResponse.arrayBuffer(body, { status, headers })` |
| No body (201, 204, etc.) | `new HttpResponse(null, ...)` | `new HttpResponse(null, { status })` (unchanged) |

### Preferred content type selection
New option: `mock.preferredContentType` to choose the response helper when multiple content types are present. If not set, the first matching content type in the spec is used. Typing is a literal union plus a `(string & {})` fallback for custom vendor types.

### Typed `info` parameter
Handler callbacks now explicitly type the `info` parameter:
```ts
// Before
(info) => { ... }  // implicitly typed

// After
(info: Parameters<Parameters<typeof http.get>[1]>[0]) => { ... }
```

### Broadened content type detection
- **Text responses:** `text/*`, `application/xml`, `*+xml`
- **Binary responses:** `application/octet-stream`, `application/pdf`, `image/*`, `audio/*`, `video/*`, `font/*`

## Docs
- MSW guide updated to document XML/HTML helpers and `mock.preferredContentType`.

## Tests
- Added MSW tests for preferred content type selection (JSON vs XML, text/html vs text/plain, binary header selection).

## Sample regen
- `yarn update-samples` regenerated MSW outputs across samples.

## Testing
- `yarn build`
- `yarn lint`
- `yarn test:ci`
- `yarn update-samples`
- `yarn test:samples`
- `yarn test:cli`


the-ult:
From generated `samples/angular-query/src/api/endpoints/pets/pets.msw.ts`: the download handler JSON.stringifies a `Blob` and sets `application/json`, and the mock returns `new Blob(...)` with `overrideResponse?: Blob`. This should be `ArrayBuffer` with `HttpResponse.arrayBuffer(...)` and a binary content-type (e.g. `application/octet-stream`), plus avoid `faker.number.int({ min: undefined, max: undefined })` options. Looks like this PR addresses those—flagging for verification/regeneration.

melloware (approved):

Issue #2930, opened 2026-02-09
Angular client output imports DeepNonNullable from @orval/core (breaks pnpm strict deps)


the-ult:
### Summary
Angular client generation emits `import type { DeepNonNullable } from '@orval/core'` in the generated service. Under pnpm strict dependency isolation, `@orval/core` is only a transitive dependency of the generator and is not available to the consuming project, causing a hard TypeScript error.

### Versions
- orval: 8.2.0
- pnpm: 10.x (strict dependency isolation)
- TypeScript: 5.9.x
- Angular client: `client: 'angular'`

### Steps to Reproduce
1. Generate Angular client code with `client: 'angular'` (and any schema config).
2. Build the consuming package in a pnpm workspace with strict deps.

### Actual Output
Generated service contains:
```ts
import type { DeepNonNullable } from '@orval/core';
```
Build fails with:
```
Cannot find module '@orval/core' or its corresponding type declarations.
```

### Expected Output
Generated code should not import from generator-internal packages. Options:
- Inline a local `DeepNonNullable` type in the generated file.
- Or emit a dependency on a public runtime package (e.g., document and add to dependencies), but ideally avoid external imports.

### Workaround
Post-process output to replace the import with `ts-essentials` or inline the utility type.

PR #2941, opened 2026-02-11
fix(angular): do not use/expose deepnonnullable

the-ult:
## Summary

- Remove `DeepNonNullable` usage from Angular (and Solid Start) generators to avoid downstream import issues.
- Filter Angular query params at runtime to drop `null`/`undefined` (including array entries) before passing to `HttpClient` or a custom params serializer.
- Update tests and regenerate Angular sample output.

## Changes

- **Core:** Add Angular params filtering helper and apply it in options/mutator generation; adjust query param typing to use raw types.
- **Angular/Solid Start:** Remove `DeepNonNullable` dependency; mark Angular mutators to use the new filtering path.
- **Tests:** Add/adjust unit tests for Angular params filtering and raw param typing.
- **Samples:** Regenerate Angular sample output to reflect new params handling.

## Why

Angular’s `HttpClient` params types don’t accept `null`/`undefined` and arrays must be primitive-only. Using `DeepNonNullable` in generated types forces an extra import that breaks pnpm strictness and still doesn’t prevent runtime issues. Filtering nullish values at runtime keeps types honest and avoids import churn.

## Verification
- `yarn update-samples`
- `yarn test:samples`
- `yarn test:cli`
- `yarn test:ci`

## Notes
- No functional changes for non-Angular clients.
- Angular params are now sanitized before passing to `HttpClient`/params serializer.

fixes:  #2930

### DIFF
--- a/packages/core/src/generators/options.test.ts
+++ b/packages/core/src/generators/options.test.ts
@@ -324,10 +324,10 @@ describe('generateAxiosOptions', () => {
       });
 
       expect(result).toContain(
-        'Object.entries({...params, ...options?.params}).reduce',
+        'for (const [key, value] of Object.entries({...params, ...options?.params}))',
       );
       expect(result).toContain(
-        '{} as Record<string, string | number | boolean | ReadonlyArray<string | number | boolean>>',
+        'const filteredParams = {} as Record<string, string | number | boolean | ReadonlyArray<string | number | boolean>>',
       );
       expect(result).toContain('Array.isArray(value)');
       expect(result).toContain("typeof item === 'string'");
@@ -349,10 +349,10 @@ describe('generateAxiosOptions', () => {
 
       expect(result).toContain('params: paramsSerializerMutator(');
       expect(result).toContain(
-        'Object.entries({...params, ...options?.params}).reduce',
+        'for (const [key, value] of Object.entries({...params, ...options?.params}))',
       );
       expect(result).toContain(
-        '{} as Record<string, string | number | boolean | ReadonlyArray<string | number | boolean>>',
+        'const filteredParams = {} as Record<string, string | number | boolean | ReadonlyArray<string | number | boolean>>',
       );
     });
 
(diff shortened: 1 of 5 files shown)


melloware:
OK now conflicts after the Lint PR was merged

the-ult:
I have addressed the review comments in the latest commit:

1.  **Strict Filtering**: Updated the params filtering logic to strictly allow only `string`, `number`, and `boolean` values (and arrays of them), filtering out `null`, `undefined`, and complex objects.
2.  **Compatibility**: Replaced `Object.fromEntries` with `Array.prototype.reduce` to ensure broader compatibility without polyfills.
3.  **Legacy Support**: Applied the filtering logic to the `requestOptions: false` path as well.
4.  **Tests**: Added a test case for `requestOptions: false` and updated existing tests to verify strict filtering and compatibility adjustments.

README.md
```diff
@@ -96,10 +96,11 @@ A typical development workflow would be:
 
 1. Make your code changes
 2. Run `yarn build` to compile your changes
-3. Run `yarn test` to ensure unit tests pass
-4. Run `yarn update-samples` to regenerate sample outputs
-5. Run `yarn test:samples` to verify samples work correctly
-6. Run `yarn test:cli` to validate TypeScript compilation
+3. Run `yarn lint` to catch lint issues early
+4. Run `yarn test:ci` to ensure unit tests pass
```
the-ult, line 100: changed from test to test:ci, since tests hangs/watches

README.md
```diff
@@ -96,10 +96,11 @@ A typical development workflow would be:
 
 1. Make your code changes
 2. Run `yarn build` to compile your changes
-3. Run `yarn test` to ensure unit tests pass
-4. Run `yarn update-samples` to regenerate sample outputs
-5. Run `yarn test:samples` to verify samples work correctly
-6. Run `yarn test:cli` to validate TypeScript compilation
+3. Run `yarn lint` to catch lint issues early
```
the-ult, line 99: added lint, since this is pretty important to do as well 😇

melloware:
might have to run `yarn update-samples`

PR #2946, opened 2026-02-12
fix(react-query): correctly type pre 5.14 and post 5.89 onSuccess

AllieJonsson:
When using React Query pre 5.14, the TContext parameter should be typed as optionally undefined. Post 5.89, the additional MutationFunctionContext parameter was added


melloware:
looksl ike build failed?

Issue #2688, opened 2025-12-26
Code generation broken for empty response body after `v8.0.0-rc.2`


heikkilamarko:
Code generation fails for a valid OpenAPI 3 specification where an endpoint defines a response with an empty response body (no `content`).

```
🍻 orval v8.0.0-rc.5 - A swagger client generator for typescript
🛑 api - Cannot read properties of undefined (reading 'schema')
🛑 One or more project failed, see above for details
```

This worked correctly in `v8.0.0-rc.2`, but fails in all later release candidates up to `v8.0.0-rc.5`.

Minimal OpenAPI spec to reproduce:

```yaml
openapi: 3.1.0
info:
  title: User API
  version: 1.0.0
  description: API for retrieving user information

paths:
  /api/v1/user/info:
    get:
      summary: Get user information
      description: Returns information about the user.
      responses:
        "401":
          $ref: "#/components/responses/Status401"

components:
  responses:
    Status401:
      description: Unauthorized
```

melloware:
@snebjorn is this similar to the other issue you just fixed?  Possibly even same issue?

snebjorn:
The provided spec works on current master

melloware:
@heikkilamarko this will be fixed in rc6

Issue #2681, opened 2025-12-23
[8.0.0-rc] Cannot read properties of undefined (reading 'schema')


mdoi2:
Since v8.0.0-rc.3, targeting Mailpit's swagger.json results in the following error:
> Cannot read properties of undefined (reading 'schema')

### Version causing this error
v8.0.0-rc.3 - rc.5

### swagger.json
https://raw.githubusercontent.com/axllent/mailpit/master/server/ui/api/v1/swagger.json

### orval.config.ts
```ts
import { defineConfig } from 'orval';

export default defineConfig({
  mailpit: {
    input: {
      target: 'https://raw.githubusercontent.com/axllent/mailpit/master/server/ui/api/v1/swagger.json',
    },
    output: {
      target: './output',
    },
  },
});
```

melloware:
@snebjorn can you look this looks related to the OpenAPI library updates.

PR #2967, opened 2026-02-16
fix(core): handle \$ref responses without content property

### LINKED ISSUE #2966: Generation for empty responses crash with "Cannot read properties of undefined (reading 'schema')"
### Description

Orval crashes with `TypeError: Cannot read properties of undefined (reading 'schema')` when an OpenAPI spec uses `$ref` to a response that has no `content` property.

This is valid OpenAPI 3.0/3.1 — a response can have only `description` without `content`.

This is the same bug as #2681 / #2688, which was fixed in PR #2686 by adding optional chaining (mediaType?.schema).

PR #2937 ("Fix(core) linting") removed those ?. operators because @typescript-eslint/no-unnecessary-condition flagged them

```ts

const definitionName = process.env.ORVAL_DEFINITION;

if (!definitionName) {
  throw new Error("ORVAL_DEFINITION environment variable is required");
}

export default defineConfig({
  [definitionName]: {
    output: {
      indexFiles: false,
      workspace: "./libs/shared/src/api/orval/generated",
      fileExtension: ".ts",
      client: "react-query",
      httpClient: "axios",
      headers: true,
      prettier: true,
      tsconfig: "./tsconfig.base.json",
      baseUrl: {
        getBaseUrlFromSpecification: true,
      },
      target: `${definitionName}/index.ts`,
      urlEncodeParameters: true,
      override: {
        mutator: "../../customInstance.ts",
        requestOptions: true,
        query: {
          useSuspenseQuery: true,
        },
      },
    },
    input: {
      target: `swagger/${definitionName}.yml`,
    },
  },
});

```

### Environment

- @tanstack/react-query: 5.90.12
- axios: 1.13.5 
- orval: 8.0.3
- react: 19.2.3 

### Expected behavior

_No response_

### Actual behavior

_No response_

```yaml
openapi: 3.1.0
info:
  title: Test
  version: 1.0.0

paths:
  /cancel:
    post:
      responses:
        200:
          $ref: '#/components/responses/OK'

components:
  responses:
    OK:
      description: OK
```

### Additional context

_No response_

tropnikov:
Fixes #2966 

$ref responses without a content property crash with TypeError: Cannot read properties of undefined (reading 'schema').

This is probably the same bug that was previously reported in #2681 / #2688 and fixed in #2686. PR #2937 reverted that fix.

### DIFF
--- a/packages/core/src/getters/res-req-types.test.ts
+++ b/packages/core/src/getters/res-req-types.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from 'vitest';
 
 import type {
   ContextSpec,
+  OpenApiReferenceObject,
   OpenApiRequestBodyObject,
   OpenApiResponseObject,
   OpenApiSchemaObject,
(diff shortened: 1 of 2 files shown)


melloware:
Running build now thanks for the PR!

Issue #2987, opened 2026-02-18
Zod: default values are not generated for required fields


oskargargas:
### Description

Zod generator doesn't provide defaults for fields marked as required. To be exact it is done in this line: https://github.com/orval-labs/orval/blob/25be7843bb21e271860fe0672429f16151419b06/packages/zod/src/index.ts#L761

Is this by design or a bug? For me it's a bug. Defining if field is required is separate from defining default values. Example: I want my booleans (both optional and required) to be `true` or `false` and by using default value make sure that it is never set to the `undefined`.

Also, if field has a `default` value and is not marked as `required` it should be `optional`. I guess that this if-else-if sequence should be made as independent checks. 
https://github.com/orval-labs/orval/blob/25be7843bb21e271860fe0672429f16151419b06/packages/zod/src/index.ts#L761-L769

### Configuration (orval.config)

`orval --client zod --mode single --output "GizmoSchema.ts" --input "Gizmo.yml"`

### Environment

```
  System:
    OS: macOS 26.3
    CPU: (12) arm64 Apple M2 Max
    Memory: 70.73 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    orval: ^8.4.0 => 8.4.0 
    zod: ^4.3.6 => 4.3.6 
```

### Expected behavior

```ts
export const gizmoResponseBooleanFlagDefault = false;
export const gizmoResponseNumberDefault = 10.11;
export const gizmoResponseIntegerDefault = 10;

export const GizmoResponse = zod.object({
  "booleanFlag": zod.boolean().default(gizmoResponseBooleanFlagDefault),
  "number": zod.number().default(gizmoResponseNumberDefault),
  "integer": zod.number().default(gizmoResponseIntegerDefault).optional()
})
```

### Actual behavior

```ts
export const gizmoResponseBooleanFlagDefault = false;
export const gizmoResponseNumberDefault = 10.11;
export const gizmoResponseIntegerDefault = 10;

export const GizmoResponse = zod.object({
  "booleanFlag": zod.boolean(),
  "number": zod.number(),
  "integer": zod.number().default(gizmoResponseIntegerDefault)
})
```

### OpenAPI document (minimal, if applicable)

```yaml
openapi: 3.1.0
info:
  title: GIZMO
  version: 1.0.0
  description: OpenAPI definition for a test entity with all possible attributes.
paths:
  /dummy-path:
    get:
      operationId: Gizmo
      responses:
        "200":
[...]

PR #2990, opened 2026-02-19
fix(zod): keep default values for required properties

the-ult:
## Summary

Fixes **#2987**: generate Zod `.default(...)` for OpenAPI properties with `default`, even when the property is `required`.

This builds on **#2501** (falsy defaults support via `schema.default !== undefined`) by fixing the remaining required-field case.

## Why

Default emission in `packages/zod/src/index.ts` was gated by `!required`, so `required + default` dropped `.default(...)`.

## What from the issue was correct

- `required` and `default` are separate concerns.
- Required properties with defaults should still emit `.default(...)`.
- The old `if / else if` coupling caused the bug.

## Zod semantics (clarification)

`.default(value)` applies when input is `undefined`, so defaulting is independent from requiredness.

Behavior after this PR:

- `required + default` → `.default(...)`
- `required + nullable + default` → `.nullable().default(...)`
- `optional + default` → `.default(...)` (no forced extra `.optional()`)
- `optional + nullable + default` → `.nullish().default(...)`

## Changes

- `packages/zod/src/index.ts`
  - decouple default emission from optional/nullable branching
  - add `hasDefault` for clarity
- `packages/zod/src/zod.test.ts`
  - add required+default and nullable+default tests
  - add end-to-end regression test for #2987
- `tests/specifications/zod-required-default-values.yaml` (new)
- `tests/configs/zod.config.ts` (fixture wiring)

## Validation

- `yarn workspace @orval/zod test --run src/zod.test.ts` ✅
- local CI/test run ✅


the-ult:
@oskargargas & @sentik would you mind to review?

melloware (approved):

PR #3008, opened 2026-02-21
bugfix(core)!: faker is now an optional peerDependency instead of being inlined

snebjorn:
`@faker-js/faker` is now an optional peerDependency 

This is a **BREAKING CHANGE** for users using the [Mock Option](https://orval.dev/docs/reference/configuration/output#mock-options) `locale`

Migration:
add `@faker-js/faker` to your project

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -46,5 +46,13 @@
     "remeda": "^2.33.6",
     "typedoc": "^0.28.15"
   },
+  "peerDependencies": {
+    "@faker-js/faker": ">=10"
+  },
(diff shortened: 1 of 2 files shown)


melloware:
OK iset this one for 8.5.0 since its a breaking change and will release 8.4.2

snebjorn:
That's fine. It's not a critical issue.
Though breaking changes usually cause a major version bump. So 9.0.0.
But it could be argued that this have a rather low impact so a minor version bump is fine. Your call 😃

melloware (approved):

PR #3010, opened 2026-02-21
chore(release): bump version in snapshots

snebjorn:


melloware:
Shouldn't we add this to me release action so this is done automatically?

snebjorn:
> Shouldn't we add this to me release action so this is done automatically?

- #3011

I made it in a separate PR so the change wouldn't be lost in these ~3500 changes :)

melloware (approved):

PR #3013, opened 2026-02-22
chore(ci): slightly optimized pull-request CI script

snebjorn:
Findings:

- `package-manager-cache: false` does not work, but `cache: 'yarn'` does.
- Apparently packages must be built for linting to pass.
- Github seems to be working on parallel steps. This would be a big speedup. Revisit when/if that is released.

https://github.blog/news-insights/product-news/lets-talk-about-github-actions/
> Moreover, we’ll start work on [parallel steps](https://github.com/orgs/community/discussions/14484), one of the most requested features across GitHub Actions. Our goal is to ship it before mid-2026.

### DIFF
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -1,15 +1,16 @@
-name: tests
+name: PR checks
 
 on:
   pull_request:
     branches:
       - 'master'
       - 'next'
+
 permissions:
   contents: read
 
 jobs:
-  tests:
+  pr-checks:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


snebjorn:
@melloware can you add "pr-checks" as required on PRs?

melloware:
It should always do it because the action has on on pull requests on main branch?

snebjorn:
> It should always do it because the action has on on pull requests on main branch?

That'll make it always run. 

But the person merging can click the merge button even if it fails. Making it required causes the merge button to be disabled until it passes.
Found these
- https://github.blog/changelog/2023-08-02-github-actions-required-workflows-will-move-to-repository-rules/
- https://docs.github.com/en/enterprise-cloud@latest/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-status-checks-before-merging

melloware:
No only anymanaix and myself can do that. Anyone else the build has to be passing and 1 approved review is required to merge.

melloware:
We already have branch protection rules turned on

snebjorn:
Seems they've changed it. Think it's now this

image

Issue #2984, opened 2026-02-17
feat: unified `formatter` option + oxfmt support (`prettier` | `biome` | `oxfmt` | `none`)


the-ult:
## Summary

Orval currently exposes two **separate boolean flags** for post-generation formatting:

```ts
output: {
  prettier: true,   // runs prettier on generated files
  biome: true,       // runs biome check --write on generated files
}
```

This has two problems:
1. **Mutually exclusive** — setting both `prettier: true` and `biome: true` would run both formatters sequentially (a bug waiting to happen). There's no enforcement that only one is active.
2. **Not extensible** — adding a third formatter (e.g. oxfmt) means adding yet another boolean, growing the API surface unboundedly.

This issue tracks:
- Replacing the two booleans with a single **`formatter` string enum**
- Adding **oxfmt** support as a third option

---

### `output.formatter`

```ts
output: {
  formatter: 'prettier' | 'biome' | 'oxfmt' | 'none'  // default: 'none'
}
```

- `'none'` — no post-formatting (current default, preserves backwards compat when neither flag is set)
- `'prettier'` — existing behaviour: try programmatic API, fall back to global CLI
- `'biome'` — existing behaviour: `biome check --write <paths>`
- `'oxfmt'` — new: `oxfmt --write <paths>` (see below)

### Backwards compatibility

The existing `prettier: boolean` and `biome: boolean` fields are **deprecated but kept** for one major version, resolved in `normalizeOptions` to the new enum:

```ts
// packages/orval/src/utils/options.ts
const legacyFormatter =
  outputOptions.prettier ? 'prettier'
  : outputOptions.biome  ? 'biome'
  : undefined;

formatter: outputOptions.formatter ?? legacyFormatter ?? formatter ?? 'none',
```

A deprecation warning is logged when the old fields are used:
```
⚠️  `output.prettier` is deprecated. Use `output.formatter: 'prettier'` instead.
⚠️  `output.biome` is deprecated. Use `output.formatter: 'biome'` instead.
```

### CLI flag

Replace `--prettier` and `--biome` flags with a single `--formatter <name>` flag (keep old flags as aliases with a deprecation warning):

```sh
orval --formatter prettier
orval --formatter biome
orval --formatter oxfmt
```

---

## oxfmt support

[Oxfmt](https://oxc.rs/docs/guide/usage/formatter.html) is a high-performance Prettier-compatible formatter from the [Oxc project](https://oxc.rs). Key facts:

[...]

PR #3056, opened 2026-03-03
feat: unified formatter option

hyperNURb:
## Summary
- Adds a new `oxfmt` option to format generated files using [oxfmt](https://oxfmt.dev), following the same pattern as the existing `prettier` and `biome` integrations
- Wired end-to-end: type surface, CLI flag, config normalization, and runtime execution
- Gracefully warns if `oxfmt` binary is not installed

## Test plan
- Verified existing tests pass (`@orval/mock`, `orval`)
- Verified builds pass (`@orval/core`, `@orval/mock`, `orval`)

## Summary by CodeRabbit

* **New Features**
  * Replaced separate prettier/biome flags with a unified formatter option accepting prettier, biome, or oxfmt; generation invokes the chosen formatter and shows warnings on failures.

* **Documentation**
  * Updated CLI, configuration, integration examples, samples, and README to use the new formatter option.

* **Tests**
  * Added tests covering formatter dispatch and error handling for missing formatter binaries.


melloware:
cc @snebjorn any thoughts?

melloware:
@the-ult you opened this ticket do you want to review and chime in?  https://github.com/orval-labs/orval/issues/2984

melloware:
@hyperNURb instead of a 3rd flag can you see @the-ult suggestion in the ticket https://github.com/orval-labs/orval/issues/2984

packages/orval/src/bin/orval.ts
```diff
@@ -74,6 +74,7 @@ cli
   .option('--clean [paths...]', 'Clean output directory')
   .option('--prettier', 'Prettier generated files')
   .option('--biome', 'biome generated files')
+  .option('--oxfmt', 'Format generated files with oxfmt')
```
snebjorn: I think we should make these mutually exclusive. Using multiple formatters doesn't make sense

packages/core/src/types.ts
```diff
@@ -40,6 +40,7 @@ export type NormalizedOutputOptions = {
   docs: boolean | OutputDocsOptions;
   prettier: boolean;
   biome: boolean;
+  oxfmt: boolean;
```
snebjorn: In context of the "mutually exclusivity" feedback. These should be turned into an enum, something like
```ts
formatter: SupportedFormatter;

...

enum SupportedFormatter {
  prettier,
  biome,
  oxfmt,
}
```

hyperNURb: Sounds good, thanks for the feedback! I'll update the PR and ping you back once done

snebjorn: I just read #2984.
I don't have a problem with having `--prettier` `--oxfmt` etc on the cli. We can handle that via `conflicts`. We already do that for some flags.
And once the cli input have been normalized it'll be a nice enum.

As for a "none" formatter option. I lean more towards having the `formatter` field be undefined. So 
```ts
formatter?: SupportedFormatter;
```

packages/core/src/types.ts
```diff
@@ -786,8 +803,13 @@ export interface ContextSpec {
 export interface GlobalOptions {
   watch?: boolean | string | string[];
   clean?: boolean | string[];
+  formatter?: SupportedFormatter;
+  /** @deprecated Use `formatter: 'prettier'` instead */
   prettier?: boolean;
+  /** @deprecated Use `formatter: 'biome'` instead */
   biome?: boolean;
+  /** @deprecated Use `formatter: 'oxfmt'` instead */
+  oxfmt?: boolean;
```
hyperNURb: I added it for consistency, since I assume it will be dropped at one point.

hyperNURb:
@snebjorn I hope I encapsulated the feedback accordingly.
I wasn't sure if `none` should be explicitly there or leave it as `undefined`. I decided to add it since it was mentioned in https://github.com/orval-labs/orval/issues/2984 but I'm good with dropping it as well.

melloware:
running build now

docs/content/docs/reference/configuration/output.mdx
```diff
@@ -374,19 +374,42 @@ export default defineConfig({
 
 Clean generated files before regenerating.
 
-## prettier
+## formatter
+
+**Type:** `'prettier' | 'biome' | 'oxfmt' | 'none'`
+
+Format generated files with the specified formatter. Only one formatter can be used at a time. Use `'none'` to explicitly disable formatting (useful to override a global formatter).
```
snebjorn: I'd prefer if users didn't have to write `formatter: 'none'`
```md
**Type:** `'prettier' | 'biome' | 'oxfmt' | undefined`
**Default:** `undefined`
```

As for the global formatter that a nice catch, I had forgotten about that.
But the GlobalOptions is just poorly named, it's the CLI options. CLI options should always override project config. 
So the none setting actually makes sense there. However I'm not sure anyone would use it. So why complicate things 😃

packages/core/src/types.ts
snebjorn: we don't really deprecated stuff, so just remove it :)

docs/content/docs/reference/configuration/output.mdx
hyperNURb: Sure I actually implemented without then I saw https://github.com/orval-labs/orval/issues/2984 :D

hyperNURb:
I decided to drop the boolean flags, code is much cleaner without it and using `--formatter` allows for any formatter to be plugged so I think this is correct.

packages/orval/src/bin/orval.ts
```diff
@@ -72,11 +72,17 @@ cli
   )
   .option('--mock', 'activate the mock')
   .option('--clean [paths...]', 'Clean output directory')
-  .option('--prettier', 'Prettier generated files')
-  .option('--biome', 'biome generated files')
+  .addOption(
+    new Option(
+      '--formatter <name>',
+      'Format generated files (prettier, biome, oxfmt)',
+    ).choices(['prettier', 'biome', 'oxfmt'] as const),
```
snebjorn: Does const do anything meaningful here? Actually can't you do `Object.values(SupportedFormatter)`? (or something like that)

packages/orval/src/bin/orval.ts
```diff
@@ -72,11 +72,17 @@ cli
...
+  )
   .option('--tsconfig <path>', 'path to your tsconfig file')
   .option('--verbose', 'Enable verbose logging')
   .action(async (options) => {
+    const { formatter } = options;
```
snebjorn: this seems redundant now

packages/orval/src/utils/options.ts
```diff
@@ -242,8 +244,7 @@ export async function normalizeOptions(
       mock,
       clean: outputOptions.clean ?? clean ?? false,
       docs: outputOptions.docs ?? false,
-      prettier: outputOptions.prettier ?? prettier ?? false,
-      biome: outputOptions.biome ?? biome ?? false,
+      formatter,
```
snebjorn:
```suggestion
      formatter: outputOptions.formatter ?? globalOptions.formatter,
```

docs/content/docs/reference/integration.mdx
```diff
@@ -63,9 +63,8 @@ const globalOptions: GlobalOptions = {
   // clean: ['./src/api', './types'],   // Clean specific directories (string[])
   output: './src/generated', // Override output directory
 
-  // Code formatting
-  prettier: true, // Format with Prettier
-  biome: true, // Format with Biome
+  // Code formatting (only one formatter can be used at a time)
+  formatter: 'prettier', // 'prettier' | 'biome' | 'oxfmt' | 'none'
```
snebjorn:
```suggestion
  formatter: 'prettier', // 'prettier' | 'biome' | 'oxfmt'
```

packages/orval/src/bin/orval.ts
```diff
@@ -146,11 +150,9 @@ cli
...
       let hasErrors = false;
       for (const [projectName, config] of configs) {
-        const normalizedOptions = await normalizeOptions(
-          config,
-          workspace,
-          options,
-        );
+        const normalizedOptions = await normalizeOptions(config, workspace, {
+          ...options,
+        });
```
snebjorn: no need to spread
```suggestion
        const normalizedOptions = await normalizeOptions(config, workspace, options);
```

soartec-lab:
@hyperNURb @snebjorn 
Why did you only remove `prettier`? `biome` is still there as the same option, right? Was this an oversight?
And this would be a breaking change, wouldn't it? I think we need to either maintain both options or guide users through the migration process for the existing option. Otherwise, users will be surprised when the formatting suddenly stops working.

snebjorn:
@soartec-lab the PR is labled with breaking change and was merged with commit prefix: feat!: ...

Where are you seeing biome? I'm only finding it on `SupportedFormatter`

soartec-lab:
@snebjorn 
Oops, my apologies. I realized that the `biome` option was also removed at the same time.

PR #3089, opened 2026-03-09
chore(http): Add rejectUnauthorized option to fetch requests

melloware:
So if you are accessing an internal OpenAPI URL in your company and it has self signed or bad certs it fails.  This would basically ignore SSL/TLS errors. I thought this was a safe default vs making it yet again another configuration?

@snebjorn @anymaniax @soartec-lab @ScriptType @the-ult  any thoughts on this?  

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL resolution to support skipping TLS certificate verification for requests during target validation.

### DIFF
--- a/packages/orval/src/utils/options.ts
+++ b/packages/orval/src/utils/options.ts
@@ -503,6 +503,7 @@ async function resolveFirstValidTarget(
             method: 'HEAD',
             headers,
             signal: controller.signal,
+            rejectUnauthorized: false,
           });
 
           if (response.ok) return target;
@@ -512,6 +513,7 @@ async function resolveFirstValidTarget(
               method: 'GET',
               headers,
               signal: controller.signal,
+              rejectUnauthorized: false,
             });
             if (getResponse.ok) return target;
           }


melloware:
according to the Typescript this isn't even valid even though all the docs I read says it is value for fetch?

snebjorn:
Does it work when you run it? I can't seem to find any up to date documentation on it :/

Node uses [undici for fetch](https://undici.nodejs.org/#/docs/api/Fetch.md).

I can only find something like this:
```ts
import { Agent } from 'undici';

const dispatcher = new Agent({
  connect: {
    rejectUnauthorized: false,
  },
});

await fetch(url, {
  dispatcher,
});
```
or setting env: `NODE_TLS_REJECT_UNAUTHORIZED=0`

But would using undici directly not break Bun, Deno support?

I'm leaning towards the `NODE_TLS_REJECT_UNAUTHORIZED=0` way. We can mention it in the docs.
Users would have to do
```jsonc
"scripts": {
  "gen-api": "NODE_TLS_REJECT_UNAUTHORIZED=0 oravl"
}
```

snebjorn:
and now I see the comments about `NODE_TLS_REJECT_UNAUTHORIZED=0` not working and your ticket https://github.com/APIDevTools/json-schema-ref-parser/issues/416.

That got me thinking. What part is failing? Is it `fetch` in `resolveFirstValidTarget` and/or https://github.com/orval-labs/orval/blob/43146fdfee6c1afa2bb99ba5779b43470a1a8746/packages/orval/src/import-specs.ts#L32-L34

Issue #3073, opened 2026-03-06
Zod: 8.5.0 fails generating zod for string with format email + regex


jandry:
### Description

When upgrading orval from 8.3.0 to 8.4.x or 8.5.x, it fail generation of zod client when openapi have email type and regex
My openapi has something like that
`
        owner:
          description: Owner of the team
          type: string
          format: email
          pattern: ^[a-zA-Z0-9._%+-]+@mycompany\.com$
`

8.5.0 generation do something like that
`
            owner: zod
                .email()
                .stringFormat('email', getTeamsResponseItemsItemOwnerRegExp)
                .optional()
                .describe('Owner of the team'),
`
but compilation says Property 'stringFormat' does not exist on type 'ZodEmail'.

Generation 8.3.0 was like this:

`
            owner: zod.email().regex(getTeamsResponseItemsItemOwnerRegExp).optional().describe('Owner of the team'),
`

### Configuration (orval.config)

```ts
export default {
    client: {
        input: './openapi.yml',
        output: {
            target: './client/src/zod.ts',
            client: 'zod',
            mode: 'single',
        },
    },
};
```

### Environment

  System:
    OS: macOS 26.3.1
    CPU: (8) arm64 Apple M3
    Memory: 90.95 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    zod: 4.3.6 => 4.3.6 

### Expected behavior

Generate working code compatible with zod

### Actual behavior

Generate non working code

### OpenAPI document (minimal, if applicable)

```yaml
openapi: 3.0.3
info:
  title: Demo
  version: 0.0.0
paths: {}

components:
  schemas:
    Empty:
      type: object
      properties: {}
      additionalProperties: false
    Team:
      type: object
      properties:
        teamId:
          type: string
          description: Team id
        displayName:
          description: Display name of the team
          type: string
        description:
          description: Description of the team
          type: string
          maxLength: 200
        owner:
          description: Owner of the team
          type: string
          format: email
          pattern: ^[a-zA-Z0-9._%+-]+@mycompany\.com$
        contactEmail:
          description: Contact email of the team
          type: string
          format: email
```

melloware:
@the-ult any interest in looking at this one?

melloware:
i "think" this PR fixes it https://github.com/orval-labs/orval/pull/3093

PR #3093, opened 2026-03-10
fix: code generation for predefined zod formats with custom regex pattern

Tarvion:
### Problem

When an OpenAPI string schema defines both a **predefined format** and a **regex pattern**, the generator can produce **invalid Zod code**.

For example, the following field schema

```json
"faultID": {
  "description": "Eindeutige ID des Fehlers",
  "type": "string",
  "pattern": "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-4[0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}$",
  "format": "uuid",
  "example": "834a5575-cdf5-4518-aad5-17b2e804db33"
}
```

generated this code

```ts
faultID: zod
  .uuid()
  .stringFormat("uuid", errorResponseCauseFaultIDRegExp)
  .describe("Eindeutige ID des Fehlers")
```

This is invalid because after calling `.uuid()` the function `.stringFormat()` does not exist on that resulting type. 

Instead, the expected output would be as follows:

```ts
faultID: zod
  .uuid()
  .regex(errorResponseCauseFaultIDRegExp)
  .describe("Eindeutige ID des Fehlers")
```

### Solution

This PR:
- Introduces a `PREDEFINED_FORMATS` set to centralize supported OpenAPI string formats.
- Ensures that regex patterns combined with predefined formats generate **`.regex()`** instead of `.stringFormat(...)`.
- Introduces the proper **`.hostname()`** call for the `hostname` format in Zod v4 (previously falling back to `.url()`).

## Summary by CodeRabbit

* **Refactor**
  * Centralized predefined string-format handling for Zod schema generation; clarified URI vs hostname behavior between Zod v3 and v4 and standardized date-like/other format handling.
* **Public API**
  * Exposed the predefined formats set for external use.
* **Tests**
  * Added tests covering v4-specific validators and v3 fallbacks, including hostname and UUID/regex interactions and tightened some regex assertions.

### DIFF
--- a/packages/zod/src/zod.test.ts
+++ b/packages/zod/src/zod.test.ts
@@ -12,8 +12,8 @@ import {
   generateZod,
   generateZodValidationSchemaDefinition,
   parseZodValidationSchemaDefinition,
-  type ZodValidationSchemaDefinition,
   predefinedZodFormats,
+  type ZodValidationSchemaDefinition,
 } from '.';
 import {
   getZodDateFormat,


melloware:
@Tarvion can you check this one too since it seems related: https://github.com/orval-labs/orval/issues/3073

Tarvion:
> @Tarvion can you check this one too since it seems related: #3073

It seems to be the same problem I had. I extended the test case a bit to cover all predefined zod format

melloware (dismissed):

melloware:
awesome thank you!

melloware:
looks like you need to sort the imports `Run autofix to sort these imports!  simple-import-sort/imports`

Tarvion:
Seems like there was a linting error. Just pushed a new commit after running eslint fix

PR #3094, opened 2026-03-10
fix(core): prevent enum key collisions from naming convention transforms

### LINKED ISSUE #638: Non unique enum names with `-` and `+` prefixes

1. Add an enumeration with two values that vary only with the `-` and `+` prefixes:

```
        - in: query
          name: sort
          required: false
          description: >
            The property to sort on. Direction is ascending by default. To 
            indicate descending, prefix property with `-`
          schema:
            type: string
            enum:
              - date
              - -date
```

## What happens?

Those special characters are ignored and the keys are identical:

```
  export const MyOperationSort = {
    date: '-date' as MyOperationSort,
    date: 'date' as MyOperationSort,
  };
```

The special characters are escaped so that we can use the enumeration.

## Any other comments?

This issue has been discovered while using the `-` and `+` convention to define the direction on a sorting parameter. It has been reported on the guideline owner too: https://github.com/belgif/rest-guide/issues/98

webjunkie:
**Problem**

Fix #638 

When `enumNamingConvention` is set (e.g. PascalCase) and an enum contains values that differ only by a leading special character, the naming convention strips the character and both values map to the same key. The second silently overwrites the first in the generated const/enum object.

Example — an ordering parameter enum with `["created_at", "-created_at", "email", "-email"]`:

```ts
// Before (bug): "-created_at" overwrites "created_at"
export const OrderingEnum = {
  CreatedAt: 'created_at',
  CreatedAt: '-created_at',  // silent overwrite
  Email: 'email',
  Email: '-email',           // silent overwrite
} as const;
```

This is common with DRF (Django REST Framework) ordering filters which auto-generate both `field` and `-field` enum values, but affects any enum with `+`/`-` prefixed values.

**Fix**

Replace special characters (`-`, `+`) with semantic words (`minus`, `plus`) before the naming convention transform, with a trailing underscore as word boundary:

`-created_at` → `minus_created_at` → `MinusCreatedAt`

```ts
// After (fix): distinct keys
export const OrderingEnum = {
  CreatedAt: 'created_at',
  MinusCreatedAt: '-created_at',
  Email: 'email',
  MinusEmail: '-email',
} as const;
```

Applied to both `getEnumImplementation` (const enums) and `getNativeEnumItems` (native enums). Only runs when `enumNamingConvention` is set — no change for users without convention config.

**Character map** is intentionally conservative (only `-` and `+`), can be extended incrementally for other special chars if needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of special characters in enum keys when naming conventions are applied, preventing duplicate or ambiguous keys and preserving original literals.

* **Tests**
  * Added comprehensive tests covering naming conventions (including PascalCase), collision disambiguation, uniqueness checks, and special-character cases (plus/minus and dash scenarios).

### DIFF
--- a/packages/core/src/getters/enum.test.ts
+++ b/packages/core/src/getters/enum.test.ts
@@ -21,8 +21,6 @@ describe('getEnumImplementation', () => {
         NamingConvention.PASCAL_CASE,
       );
 
-      // "-" is replaced with "Minus" before PascalCase, so both values
-      // produce distinct keys
       expect(result).toContain('CreatedAt');
       expect(result).toContain('MinusCreatedAt');
       expect(result).toContain('Email');
@@ -65,6 +63,20 @@ describe('getEnumImplementation', () => {
       expect(result).toContain('Active');
       expect(result).toContain('Inactive');
       expect(result).toContain('Pending');
+    });
+
+    it('should not change keys when dash values do not collide', () => {
+      // "-date" alone (no "date") should still produce "Date", not "MinusDate"
+      const result = getEnumImplementation(
+        "'-date' | 'name'",
+        undefined,
+        undefined,
+        NamingConvention.PASCAL_CASE,
+      );
+
+      expect(result).toContain('Date');
+      expect(result).toContain('Name');
+      expect(result).not.toContain('Minus');
     });
   });
 });
--- a/packages/core/src/getters/enum.ts
+++ b/packages/core/src/getters/enum.ts
@@ -124,6 +124,44 @@ const getTypeConstEnum = (
   return enumValue;
 };
 
+/**
+ * Derive the object/enum key for a single enum value.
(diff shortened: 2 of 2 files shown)


melloware:
@webjunkie i think this fixes this one as well right: https://github.com/orval-labs/orval/issues/638 ??

webjunkie:
> @webjunkie i think this fixes this one as well right: #638 ??



> @webjunkie i think this fixes this one as well right: #638 ??

Ah, yes, it fixes it as well

webjunkie:
One thing I was wondering is if I should compute all keys first, detect collisions, only apply the replacement for colliding keys. This would alleviate more backwards compatibility concerns I guess.

melloware:
@webjunkie up to you?

PR #3099, opened 2026-03-11
Chore/migrate oxlint oxfmt

anymaniax:


melloware:
we support all 3 now.

PR #3106, opened 2026-03-13
feat: add strict infinite hook generation

### LINKED ISSUE #3101: `useInfiniteQueryParam` should only be added to hooks with that query parameter present
### Description

Orval will generate a react query useInfinite hook with the `useInfiniteQueryParam` for every endpoint that has query parameters, even if `useInfiniteQueryParam` is not present in the query parameters for a certain endpoint. This will cause type error: `Property 'page' does not exist on type 'GetFloorplanImageParams'`, as GetFloorplanImage query parameters do not include `page`.  You could use endpoint specific overrides to prevent this, but it does not seem like an optimal solution, particularly for large OpenAPI specs.

```ts

export default defineConfig({
  backendQuery: {
    input: {
      target: 'my_swagger_here.json',
    },
    output: {
      workspace: '.',
      target: 'gen/apis',
      schemas: 'gen/models',
      client: 'react-query',
      httpClient: 'fetch',
      mode: 'tags-split',
      mock: true,
      urlEncodeParameters: true,
      override: {
        mutator: {
          path: 'src/custom-fetch.ts',
          name: 'customFetch',
        },
        query: {
          useQuery: true,
          useMutation: true,
          useInfinite: true,
          useInfiniteQueryParam: 'page',
          useSuspenseQuery: false,
          useSuspenseInfiniteQuery: false,
          usePrefetch: true,
          useInvalidate: true,
        },
      },
      prettier: true,
    },
  },
})
```

### Environment

```
  System:
    OS: macOS 26.3.1
    CPU: (12) arm64 Apple M4 Pro
    Memory: 419.77 MB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    msw: 2.10.4 => 2.10.4 
    zod: ^4.0.10 => 4.0.10 
orval: ^8.4.2 => 8.4.2
@tanstack/react-query: ^5.84.2 => 5.84.2
```

### Expected behavior

When generating useInfinite hooks for an endpoint with `useInfiniteQueryParam` present in the config, disqualify endpoints without the `useInfiniteQueryParam` value in their real query parameters from generating with the `useInfiniteQueryParam`.

### Actual behavior

Orval will generate a react query useInfinite hook with the `useInfiniteQueryParam` for every endpoint that has query parameters, even if `useInfiniteQueryParam` is not present in the query parameters for a certain endpoint. 

```yaml
openapi: 3.0.3
info:
  title: Seraphim Minimal Repro
  version: 1.0.0
paths:
  /api/facility/floors/{id}/image:
    get:
      operationId: getFloorplanImage
      summary: Get floorplan image by floor id
      parameters:
        - name: id
          in: path
          required: true
          schema:
            type: string
            format: uuid
        - name: variant
          in: query
          required: false
          schema:
            type: string
            enum:
              - cropped
              - original
      responses:
        '200':
          description: Image file
          content:
            image/png:
              schema:
                type: string
                format: binary
        '404':
          description: Not found

  /api/dashboard/dashboard:
    get:
      operationId: listDashboards
      summary: List dashboards with pagination
      parameters:
        - name: page
          in: query
          required: false
          schema:
            type: integer
            minimum: 1
            default: 1
        - name: pageSize
          in: query
          required: false
          schema:
            type: integer
            minimum: 1
            maximum: 200
            default: 20
      responses:
        '200':
          description: Paged dashboard list
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/DashboardPageDto'

components:
  schemas:
    DashboardPageDto:
      type: object
      properties:
        page:
          type: integer
        pageSize:
          type: integer
        items:
          type: array
          items:
            type: object
            additionalProperties: true
```

### Additional context

See `packages/query/src/frameworks/index.ts:85-93`, `packages/query/src/query-generator.ts:629-640`, `packages/query/src/query-generator.ts:202-210`, and `packages/query/src/query-generator.ts:384-412`.

This behaviour is also mirrored in other framework specific adapaters for angular and vue, respectively: `packages/query/src/frameworks/angular.ts:85-96` and `packages/query/src/frameworks/vue.ts:72-80`.

I activate useInfinite for all endpoints in our spec as we have a very large spec, and individually pulling and updating only paginated GET endpoints into the orval config to specifically override them with useInfinite would be a hassle. So i rely on tree shaking at build time to ensure those extra useInfinite hooks for non-paginated GET requests arent packed into the build. I realise its a little bit backwards in relation to orvals ethos.

the-inconvenience-store:
Introduce useInfiniteRequireQueryParam to gate infinite and suspense-infinite hook generation to endpoints that actually include the configured useInfiniteQueryParam

- useInfiniteRequireQueryParam to query option types and normalisation
- gate infinite/suspense-infite generation in query generator when strict mode is enabled
- update framework adapaters (shared, angular and vue) to support strict pageParam injection
- add unit tests for query option nroamlisation
- add regression spec and config scenarios for react,vue, and angular query
- include angular query in api snapshot harness

closes #3101

Not sure if there was a reason why `angular-query` was excluded for the snapshot testing harness, but this PR does re-introduce it as I was testing the changes in Angular. 

If desired I can push a commit with angular-query snapshots disabled and removed again.

Thanks

## Summary by CodeRabbit

* **New Features**
  * Added validation to only generate infinite query hooks when the specified query parameter exists in the operation schema.
  * Added configuration control for requiring query parameters in infinite query operations.

* **Documentation**
  * Documented side effects when using the infinite query parameter configuration option.


melloware:
@the-inconvenience-store looks like build is failing and CodeRabbit suggestions?

melloware:
@the-inconvenience-store is there any way to implement this without another new param?  Also if there is a new parameter it means you need to update the `/docs`

the-inconvenience-store:
> @the-inconvenience-store is there any way to implement this without another new param?  Also if there is a new parameter it means you need to update the `/docs`

yeah i just didn't want to make any changes that might confuse/break people's current setups. of course it could be automatic, so i suppose it's up to you if you want to release it as breaking change for v9 instead of a minor feat. 

let me know your preference and i'll make edits accordingly

cheers

melloware:
so my thoguht is if this is a string `useInfiniteQueryParam?: string;` could you just introduce a new mode there?

the-inconvenience-store:
> so my thoguht is if this is a string `useInfiniteQueryParam?: string;` could you just introduce a new mode there?

would you rather:
make the strict mode implicit if `useInfiniteQueryParam` is not undefined. if they provide a string then we implicitly apply the new strict mode behaviour. alternatively, change the type to an object for example:
```typescript
...
useInfiniteQueryParam : {
    queryParam: string // <-- current query param behaviour
    strictMode: bool // <-- toggle new behaviour
} | string // <-- backwards compat with current query param behaviour
```

melloware:
I think i like your implicitly apply this behavior option.

the-inconvenience-store:
> I think i like your implicitly apply this behavior option.

 🫡

will push my changes soon, cheers!

melloware:
looks like Snapshots are failing now

melloware:
looks like conflicts need to be resolved

the-inconvenience-store:
@melloware resolved, cheers :)

melloware:
Looks like SNAPSHOTs are failing now

the-inconvenience-store:
hi @melloware, can we take this out of draft state if it's still being considered since it's completed? thanks

melloware:
it has merge conflicts?

the-inconvenience-store:
those are new since i last pushed.
will fix

melloware:
looks like you need to run the formatter?

melloware:
@the-inconvenience-store this has been fixed by another PR!

PR #3115, opened 2026-03-17
fix(angular-query): scope angular filter params helper by tag

### LINKED ISSUE #2952: angular-query: emit `filterParams` helper only when generated operations use query params
## Summary
In Angular Query generated files, `filterParams` can be emitted even when no generated operation in that file uses query params (for example: `samples/angular-query/src/api/endpoints-zod/health/health.ts`).

This leaves dead/unused helper code in generated output.

## Context
Spotted during PR #2938 review:
- Review thread: https://github.com/orval-labs/orval/pull/2938#discussion_r2801447905

Current behavior comes from header-level helper emission being too broad for some generated files.

## Expected behavior
Only emit `filterParams` in a generated file when at least one emitted operation in that file needs query-param filtering.

## Proposed approach
- Tighten Angular Query helper emission condition in `packages/query/src/client.ts` (`getQueryHeader` path), so helper output is based on effective per-file operation usage.
- Keep existing nullable-query-param filtering behavior unchanged for files/operations that need it.

## Acceptance criteria
- [ ] `filterParams` is **not** emitted in generated files that have no query-param usage (e.g. health endpoints).
- [ ] `filterParams` **is** still emitted where required (e.g. pets endpoints with query params).
- [ ] Existing Angular/Angular Query generation tests pass.
- [ ] `yarn test:ci`, `yarn update-samples`, `yarn test:samples`, and `yarn test:cli` pass for the change.

## Notes
This is cleanup/optimization and does not change runtime behavior for endpoints without query params.

the-ult:
## Summary

Fix Angular Query tag-split generation so the `filterParams` helper is only emitted in files whose operations actually use query params.

This resolves the Angular Query `filterParams` over-generation issue tracked in **#2952**.

## Problem

When generating Angular Query clients in tag/tag-split mode, the generator could emit the Angular `filterParams` helper into a tag file even when that tag's operations did not use query params.

A concrete example was:

- `samples/angular-query/src/api/endpoints-zod/health/health.ts`

This file only contains the health-check endpoint and does **not** need query param filtering, but it still received the helper because header generation was checking all operations globally instead of only the operations relevant to the current tag.

## Root cause

In `packages/query/src/client.ts`, Angular Query header generation used:

- all `verbOptions` in the file-generation context
- without narrowing them to the current `tag`

That meant `hasQueryParams` could become `true` because of an operation from a different tag.

## Fix

Update Angular Query header generation to match the existing Angular generator behavior:

- when `params.tag` is present, filter `verbOptions` to only verbs belonging to that tag
- compare tags using the same normalized `camel(...)` pattern already used by the Angular generator
- emit `filterParams` only when those relevant verbs actually have `queryParams`

## Changes included

- `packages/query/src/client.ts`
  - make Angular Query header generation tag-aware
- `packages/query/src/client.test.ts`
  - add regression coverage for:
    - non-tagged Angular output with query params
    - tag files without query params
    - tag files with query params
    - normalized tag matching (`PetStatus` vs `pet-status`)
- regenerated sample output:
  - `samples/angular-query/src/api/endpoints-zod/health/health.ts`
- updated snapshot:
  - `samples/angular-query/__snapshots__/api/endpoints-zod/health/health.ts`

## Result

After regeneration:

- `health.ts` no longer includes the unused `filterParams` helper
- files like `pets.ts` still include it when query params are actually used

## Validation

Verified with:

- package build
- sample generation
- package tests
- snapshot tests
- sample tests, including Angular and Angular Query samples

### Commands run

- `bun run build`
- `bun run update-samples`
- `CI=1 bun run test`
- `CI=1 bun run test:snapshots`
- `CI=1 bun run test:samples`

## Issue

Closes #2952

## Summary by CodeRabbit

* **New Features**
  * Query header generation for Angular output is now tag-aware, producing more precise query parameter handling and emitting parameter-filtering only when relevant.

* **Tests**
  * Added tests validating Angular query header behavior and conditional emission of the parameter-filtering helper across tag scenarios.

* **Samples**
  * Sample Angular output updated to reflect conditional inclusion/removal of the parameter-filtering helper per tag.

### DIFF
--- a/packages/query/src/client.test.ts
+++ b/packages/query/src/client.test.ts
@@ -4,9 +4,83 @@ import { describe, expect, it } from 'vitest';
 import {
   generateRequestOptionsArguments,
   getHttpFunctionQueryProps,
+  getQueryHeader,
   getQueryOptions,
   getSignalDefinition,
 } from './client';
+
+describe('getQueryHeader', () => {
+  it('emits filterParams helper for Angular when a non-tagged file has query params', () => {
+    const header = getQueryHeader({
+      output: { httpClient: OutputHttpClient.ANGULAR },
+      verbOptions: {
+        listPets: {
+          tags: ['pets'],
+          queryParams: { schema: { name: 'ListPetsParams' } },
+        },
+        healthCheck: {
+          tags: ['health'],
+          queryParams: undefined,
+        },
+      },
+    } as never);
+
+    expect(header).toContain('function filterParams');
+  });
+
+  it('does not emit filterParams helper for a tag file with no query params', () => {
+    const header = getQueryHeader({
+      output: { httpClient: OutputHttpClient.ANGULAR },
+      tag: 'health',
+      verbOptions: {
+        listPets: {
+          tags: ['pets'],
+          queryParams: { schema: { name: 'ListPetsParams' } },
+        },
+        healthCheck: {
+          tags: ['health'],
+          queryParams: undefined,
+        },
+      },
+    } as never);
+
+    expect(header).toBe('');
+  });
+
+  it('emits filterParams helper for a tag file whose operations use query params', () => {
+    const header = getQueryHeader({
+      output: { httpClient: OutputHttpClient.ANGULAR },
+      tag: 'pets',
+      verbOptions: {
+        listPets: {
+          tags: ['pets'],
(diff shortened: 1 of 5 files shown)


the-ult:
@coderabbitai review

melloware (approved):

PR #3120, opened 2026-03-19
fix(mock): use one-sided bounds when only minLength/maxLength or minItems/maxItems is specified

### LINKED ISSUE #3121: Incorrect string/array min value in MSW mocks when only maxLength/maxItems is specified
## Description

When generating MSW mocks, if a schema specifies only `maxLength` (without `minLength`) or only `maxItems` (without `minItems`), the generator incorrectly falls back to the global `stringMin`/`arrayMin` default for the unspecified bound.

With the default `stringMin: 10`, a schema with `maxLength: 5` produces:

```ts
faker.string.alpha({ length: { min: 10, max: 5 } })
```

`min > max` causes faker to throw at runtime. Even when `maxLength >= 10`, it is semantically wrong — the schema implies no minimum constraint, but the global default is applied anyway. The same issue occurs symmetrically: `minLength` without `maxLength` causes the global `stringMax` to be used as the upper bound, which may be less than `minLength`.

```ts

export default defineConfig({
  api: {
    output: {
      target: './src/api/endpoints.ts',
      mock: { type: 'msw' },
    },
  },
});
```

## Environment

## Expected behavior

When only `maxLength: 5` is set, the generated mock should use only an upper bound:

```ts
faker.string.alpha({ length: { max: 5 } })
```

When only `minLength: 30` is set:

```ts
faker.string.alpha({ length: { min: 30 } })
```

Global defaults (`stringMin`, `stringMax`, `arrayMin`, `arrayMax`) should only be applied when **neither** bound is specified in the schema.

## Actual behavior

With `maxLength: 5` and the default `stringMin: 10`:

```ts
faker.string.alpha({ length: { min: 10, max: 5 } })
// ⚠️ faker throws — min > max
```

## OpenAPI document (minimal, if applicable)

```yaml
openapi: 3.0.3
info:
  title: Demo
  version: 0.0.0
paths:
  /items:
    get:
      operationId: getItems
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: object
                properties:
                  shortCode:
                    type: string
                    maxLength: 5
                  longDescription:
                    type: string
                    minLength: 30
                  tags:
                    type: array
                    maxItems: 2
                    items:
                      type: string
```

## Additional context

The bug is in `packages/mock/src/faker/getters/scalar.ts`. The fix is to conditionally apply global defaults only when neither bound is schema-specified:

```ts
// Before
const strMin = (item.minLength ?? safeMockOptions.stringMin) as number | undefined;
const strMax = (item.maxLength ?? safeMockOptions.stringMax) as number | undefined;

// After
const strMin = (item.minLength ?? (item.maxLength !== undefined ? undefined : safeMockOptions.stringMin)) as number | undefined;
const strMax = (item.maxLength ?? (item.minLength !== undefined ? undefined : safeMockOptions.stringMax)) as number | undefined;
```

Fix is available in PR #3120.

youngcm2:
Closes #3121

## Summary

When a schema specifies only `maxLength`/`maxItems` without a corresponding min, or only `minLength`/`minItems` without a corresponding max, the mock scalar generator was incorrectly falling back to global defaults (`stringMin: 10`, `arrayMin: 1`, etc.) for the unspecified bound. This could produce invalid faker calls like `faker.string.alpha({length: {min: 10, max: 5}})` where min > max.

## Changes

- Fix `strMin`/`strMax` fallback logic: global defaults now only apply when **neither** bound is schema-specified
- Fix `arrMin`/`arrMax` fallback logic with the same rule
- Add unit tests for `maxLength`-only and `minLength`-only string schema cases
- Add integration spec (`mock-constraints.yaml`) and snapshot for one-sided constraint scenarios

## Summary by CodeRabbit

* **New Features**
  * Added a new OpenAPI mock spec and generated MSW mock endpoint and model to produce constraint-aware sample responses.

* **Bug Fixes**
  * Mock generation now respects one-sided string and array constraints so generated values honor explicitly provided min/max bounds.

* **Tests**
  * Added tests and snapshots validating one-sided constraint handling in mock data generation.

* **Chore**
  * Added mock generation config for the new constraint-focused spec.


melloware:
Running build now.

melloware:
Looks like linter issues

melloware:
Looks like Linter is still failing build

melloware:
Looks like Snapshots are failing now

melloware (approved):

melloware:
@snebjorn can you chime in here the SNAPSHOTs seem to be frustrating users?

snebjorn:
Snapshots are working as expected. They're saying that `min: 1` was removed in the output

```diff
- { length: faker.number.int({ min: 1, max: 100 }) }
+ { length: faker.number.int({ max: 100 }) }
```

If this is intended, the snapshot should be updated.
If this was unintended, then a bug was found.

melloware:
@youngcm2 any update on this i agree with @snebjorn your code changes have made the snapshots now broken see his guidance.

melloware:
@wadakatu can you take over this PR please?

wadakatu:
> @wadakatu can you take over this PR please?

Certainly.

wadakatu:
Took over here in #3451 (per @melloware's request). The new PR keeps the original intent (no global default applied when only one bound is schema-specified) but uses a clamp strategy for strings to satisfy faker's required `length: { min, max }` pair, which was causing this PR's CI to fail silently on `test:cli` typechecking. @youngcm2 credited via `Co-authored-by`.

PR #3140, opened 2026-03-25
fix(core): strip const from discriminator merged property (#3139)

### LINKED ISSUE #3139: 8.3.0 broke type generation for some const discriminators without additional fields
### Description

Pin pointed this issue down to 8.3.0, 8.2.0 is fine

This shows itself initially as errors in generated msw code but other downstream code will also be wrong.
See how the DogValue has incorrectly been turned into an object

The old code is also a little strange the exports are two different things depending on if used as a type or runtime:
8.2.0 code:

Image

More info

Image

Image

Image

**If you add any extra fields to Dog the problem goes away**

Possible commit in 8.3.0 that broke it:
- https://github.com/orval-labs/orval/pull/2908 - https://github.com/orval-labs/orval/issues/2458

```ts

export default defineConfig({
  repro: {
    input: {
      target: "../../../orval-repro-discriminator-const-only.schema.json",
    },
    output: {
      mode: "tags-split",
      target: "_orval_repro_generated/default.ts",
      schemas: "_orval_repro_generated/types",
      client: "fetch",
      prettier: true,
      mock: { type: "msw", delay: 0 },
      override: {
        aliasCombinedTypes: true,
        fetch: { includeHttpResponseReturnType: false },
      },
    },
  },
});

```

### Environment

Had trouble running this

### Expected behavior

Generate type should not be forced into an object

Image

Image

Image

### Actual behavior

Image

```json
{
  "openapi": "3.1.0",
  "info": { "title": "Cats and dogs", "version": "1" },
  "paths": {
    "/": {
      "get": {
        "responses": {
          "200": {
            "description": "x",
            "content": {
              "application/json": {
                "schema": { "$ref": "#/components/schemas/Pet" }
              }
            }
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "Cat": {
        "type": "object",
        "additionalProperties": false,
        "required": ["type", "name"],
        "properties": {
          "type": { "const": "cat" },
          "name": { "type": "string" }
        }
      },
      "Dog": {
        "type": "object",
        "additionalProperties": false,
        "required": ["type"],
        "properties": {
          "type": { "const": "dog" }
          
        }
      },
      "Pet": {
        "oneOf": [
          { "$ref": "#/components/schemas/Cat" },
          { "$ref": "#/components/schemas/Dog" }
        ],
        "discriminator": {
          "propertyName": "type",
          "mapping": {
            "cat": "#/components/schemas/Cat",
            "dog": "#/components/schemas/Dog"
          }
        }
      }
    }
  }
}
```

### Additional context

_No response_

froggy1014:
## Summary

- Strip `const` keyword when merging discriminator properties to prevent broken type generation
- Objects with only a `const` discriminator property (no additional fields) were generating `const DogValue = { type: DogType }` instead of `interface Dog { type: DogType }`

## Problem

Fixes #3139

When a discriminator variant has only a `const` property and no extra fields:
```json
{
  "Dog": {
    "type": "object",
    "additionalProperties": false,
    "required": ["type"],
    "properties": {
      "type": { "const": "dog" }
    }
  }
}
```

**Before (broken):**
```ts
import { DogType } from './dogType';

export const DogValue = {
  type: DogType,    // runtime object { dog: 'dog' } ❌
} as const;
export type Dog = typeof DogValue;
```

**After (fixed):**
```ts
import type { DogType } from './dogType';

export interface Dog {
  type: DogType;    // type alias = 'dog' ✅
}
```

## Context

In #2897, `resolveDiscriminators` was changed to spread the existing property (`...schemaProperty`) to preserve metadata like `description`. This also preserved `const`, which triggered the all-const-properties check in `interface.ts` — generating a `const XValue` pattern where the enum object was used as a runtime value instead of a type.

No side effects: all downstream consumers (zod, mock) already prioritize `enum` over `const`, so the removed `const` was only affecting `interface.ts`.

## Testing

- [Reproduction](https://github.com/froggy1014/3139)
- `bun vitest run src/getters/discriminators.test.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed discriminator-driven schema merging so target properties no longer retain overly restrictive constant values; merged discriminator properties are represented as string enums with correct values and preserved descriptions.
* **Tests**
  * Added a test validating discriminator resolution for oneOf-style schemas (e.g., Pet → Cat/Dog) to ensure correct merge behavior.

### DIFF
--- a/packages/core/src/getters/discriminators.test.ts
+++ b/packages/core/src/getters/discriminators.test.ts
@@ -136,6 +136,59 @@ describe('resolveDiscriminators getter', () => {
     }
   });
 
+  it('strips const from discriminator property to prevent DogValue object bug (#3139)', () => {
+    const schemas: OpenApiSchemasObject = {
+      Pet: {
+        oneOf: [
+          { $ref: '#/components/schemas/Cat' },
+          { $ref: '#/components/schemas/Dog' },
+        ],
+        discriminator: {
+          propertyName: 'type',
+          mapping: {
+            cat: '#/components/schemas/Cat',
+            dog: '#/components/schemas/Dog',
+          },
+        },
+      },
+      Cat: {
+        type: 'object',
+        additionalProperties: false,
+        required: ['type', 'name'],
+        properties: {
+          type: { const: 'cat' },
+          name: { type: 'string' },
+        },
+      },
+      Dog: {
+        type: 'object',
+        additionalProperties: false,
+        required: ['type'],
+        properties: {
+          type: { const: 'dog', description: 'animal type' },
+        },
+      },
+    };
+
+    const result = resolveDiscriminators(structuredClone(schemas), context);
+    const dogSchema = result.Dog as NonNullable<OpenApiSchemasObject[string]>;
+    const dogProps = dogSchema.properties as
+      | Record<string, OpenApiSchemaObject | OpenApiReferenceObject>
+      | undefined;
+    const typeProp = dogProps?.type as OpenApiSchemaObject | undefined;
+
+    // The merged property must NOT retain `const` — otherwise interface.ts
+    // generates `const DogValue = { type: DogType }` where DogType is the
+    // runtime enum object instead of the string literal.
+    expect(typeProp).toBeDefined();
+    expect(typeProp).not.toHaveProperty('const');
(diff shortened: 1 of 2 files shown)


melloware:
Looks like a lint issue

PR #3163, opened 2026-03-26
fix(core): nullable-only schema generates type alias instead of invalid interface

### LINKED ISSUE #3149: Bug: nullable-only schema generates invalid TypeScript (`export interface Foo unknown | null`)
### Description

When an OpenAPI 3.0 schema has only `nullable: true` with no `type` field, orval v8 generates invalid TypeScript syntax:

```typescript
export interface CreateOrgUserResponse unknown | null
```

This should be:

```typescript
export type CreateOrgUserResponse = unknown | null;
```

The generated code fails TypeScript compilation with `TS1005: '{' expected`.

### OpenAPI schema (minimal reproduction)

```json
{
  "components": {
    "schemas": {
      "LogoutResponse": {
        "nullable": true
      },
      "CreateOrgUserResponse": {
        "nullable": true
      }
    }
  }
}
```

```ts

export default defineConfig({
  api: {
    input: {
      target: './openapi.json',
    },
    output: {
      target: './src/generated/rest-client.ts',
      schemas: './src/generated/model',
      client: 'fetch',
      mode: 'split',
      prettier: true,
    },
  },
});
```

### Environment

- orval: 8.6.2
- Node.js: 22.16.0
- TypeScript: 5.9.3
- OS: Windows 11 / Ubuntu (GitHub Actions)

### Expected behavior

Should generate `export type Foo = unknown | null;` (a type alias) since `unknown | null` is a union type and cannot be expressed as an `interface`.

### Actual behavior

Generates `export interface Foo unknown | null` which is syntactically invalid TypeScript.

Orval also logs a warning during generation but continues:
```
⚠️ Failed to format file .../createOrgUserResponse.ts: SyntaxError: '{' expected. (8:40)
```

### Workaround

Post-generation script to fix the output:

```js
// fix-orval-output.mjs
import { readdirSync, readFileSync, writeFileSync } from 'node:fs';
import { join } from 'node:path';

const GENERATED_DIR = join(import.meta.dirname, '..', 'src', 'generated');
const BROKEN_RE = /^export interface (\w+) (unknown \| null)$/gm;

function walk(dir) {
  for (const entry of readdirSync(dir, { withFileTypes: true })) {
    const full = join(dir, entry.name);
    if (entry.isDirectory()) walk(full);
    else if (entry.name.endsWith('.ts')) {
      const content = readFileSync(full, 'utf8');
      const fixed = content.replace(BROKEN_RE, 'export type $1 = $2;');
      if (fixed !== content) writeFileSync(full, fixed);
    }
  }
}

walk(GENERATED_DIR);
```

### Related issues

- #2755 (nullable:true broken in v8, but about `type`+`nullable` combo, not nullable-only)
- #2795 (nullable nested objects, different root cause)

vgpastor:
## Summary

- Fix invalid TypeScript generated for schemas with only `nullable: true` and no `type` field
- `getObject()` now sets `useTypeAlias: true` when `itemType !== 'object'`, so the generator emits `export type Foo = unknown | null;` instead of the invalid `export interface Foo unknown | null`
- Added `NullableOnly` test schema to the `null-type-v3-0.yaml` spec with corresponding snapshot

Closes #3149

## Test plan

- [ ] Verify `NullableOnly` snapshot generates `export type NullableOnly = unknown | null;`
- [ ] Verify existing null-type tests still pass (no regression)
- [ ] Verify TypeScript compilation succeeds on generated output

## Summary by CodeRabbit

* **New Features**
  * Support for OpenAPI schemas that are nullable without a type, generating union types like unknown | null.
  * Import pipeline now normalizes nullable-only schemas into equivalent nullable unions before processing.

* **Bug Fixes**
  * Improved nullability detection and union aliasing for combined/union schemas.

* **Tests**
  * Added tests and snapshots covering nullable-only schemas, normalization, and combiner behavior.


packages/core/src/getters/scalar.test.ts
```diff
@@ -60,3 +60,25 @@ describe('getScalar (contentMediaType: application/octet-stream)', () => {
     expect(result.value).toBe('string');
   });
 });
+
+describe('getScalar (nullable-only schema without type — GH-3149)', () => {
+  it('nullable: true without type → "unknown | null" with useTypeAlias', () => {
+    const schema: OpenApiSchemaObject = {
+      nullable: true,
```
snebjorn: all OAS is upgraded to 3.1 inside Orval. The nullable field no longer exists.

snebjorn (changes requested):

packages/core/src/getters/scalar.test.ts
vgpastor: Good catch! You're right — the nullable field shouldn't exist after the 3.1 upgrade. I investigated and found that @scalar/openapi-upgrader only converts nullable: true when schema.type is present. Typeless schemas like { nullable: true } survive the upgrade unchanged.

I've added a post-upgrade normalization step in import-specs.ts (normalizeLeftoverNullable) that rewrites any remaining { nullable: true, ...rest } (without type) into { anyOf: [{ ...rest }, { type: 'null' }] }, which is the proper OAS 3.1 representation.

snebjorn: hmm, that's a good find. I was unaware of this. 
Can you please file an issue with https://github.com/scalar/scalar so it gets fixed

vgpastor:
The root cause of this bug was traced back to OpenAPI specs that include `nullable: true` without an explicit `type` field — a valid pattern per the OpenAPI 3.0 spec, but one that can trip up code generators.

An issue has been opened in Scalar to raise awareness and request that their spec export always includes an explicit `type` alongside `nullable: true` for better ecosystem interoperability:

👉 scalar/scalar#8616

melloware:
Converting to draft until the build is passing

snebjorn:
@vgpastor seems https://github.com/scalar/scalar/issues/8616 was closed due to issue description being too vague and lack of feedback. Please try again. Thanks.

artnikbrothers:
Hi @vgpastor @snebjorn - I can confirm this is a real and significant issue in production. We run Orval 8.5.1 against a 600+ endpoint OAS 3.1 spec and hit the same root cause: `@scalar/openapi-upgrader` doesn't normalize `nullable` when `type` is missing.

In our case the pattern is slightly different - our spec emits:
```json
{ "type": "null", "allOf": [{ "$ref": "#/components/schemas/TimeIncrement" }] }
```
Orval reads `type: "null"` and collapses the entire union, generating `timeStep?: null` instead of `TimeIncrement | null`. This affected **182 occurrences** across our schema.

We built a pre-generation transformer as a workaround that rewrites these to OAS 3.0 form before Orval processes them:
```js
// { type: "null", allOf: [{ $ref }] }  ->  { nullable: true, allOf: [{ $ref }] }
if (schema.type === 'null' && Array.isArray(schema.allOf) && schema.allOf.length > 0) {
  delete schema.type;
  schema.nullable = true;
}
```
Recursing into properties, allOf, oneOf, anyOf, items, and additionalProperties. After applying this, our regen produced a 1,542-file diff restoring correct union types.

Regarding the stalled Scalar issue (#8616) - I'm happy to help re-open it with concrete examples if that would be useful. We have a large, real-world dataset that clearly demonstrates the upgrader gap.

Two technical observations on this PR:

1. `normalizeLeftoverNullable` catches `{ nullable: true }` without `type`, but the `{ type: "null", allOf/oneOf/anyOf: [...] }` pattern (valid OAS 3.1 nullable encoding) would still slip through. Both patterns should probably be handled for complete 3.1 coverage - either here or in a follow-up.

2. For the `{ anyOf: [...], nullable: true }` case, the current code wraps into `{ anyOf: [{ anyOf: [...] }, { type: "null" }] }` - double-nested anyOf. Does the combine getter flatten this correctly?

Great work tracking down the root cause. This fix (normalizing after the upgrader) is the right approach - it's essentially what we've been doing in production.

melloware:
@artnikbrothers can you try the latest version 8.12.3 lots of fixes since this version.

snebjorn:
Please file an issue with @scalar/openapi-upgrader to get the root cause fixed. Once fix is released we'll update our dependency.
Thanks

melloware:
@snebjorn the Scalar teamed marked it as "won't fix" and said downstream consumers should handle this scenario.  Can you open the ticket since they closed the last one as "vague" ?

melloware:
I just bumped to the latest Scalar libs in `main` as well

artnikbrothers:
Thanks @melloware - tried 8.12.3 (ships `@scalar/openapi-upgrader@0.2.6`, newer than what we run). My pattern still collapses. Minimal repro:

OAS 3.1 input:
```json
"timeStepBad":  { "type": "null", "allOf": [{ "$ref": "#/components/schemas/TimeIncrement" }] },
"timeStepGood": { "anyOf": [{ "$ref": "#/components/schemas/TimeIncrement" }, { "type": "null" }] }
```

Orval 8.12.3 output:
```ts
timeStepBad?: null;                  // ref lost
timeStepGood?: TimeIncrement | null; // correct
```

So Orval generates nullable refs correctly when they're encoded as `anyOf` - it's specifically the `{ type: "null", allOf: [$ref] }` encoding (valid OAS 3.1) that collapses to `null`. That's the gap from my earlier point: #3163's `normalizeLeftoverNullable` catches `{ nullable: true }` without `type`, but not the `type: "null"` + allOf/oneOf/anyOf form. Since Scalar marked the upgrader fix won't-fix, normalizing this downstream in Orval looks like the right path. I have a recursive transform handling it in production (182 occurrences across our spec) - happy to open a PR extending the normalization to cover this pattern if useful.

artnikbrothers:
Opened #3438 to fix this downstream in orval.

Root cause confirmed on our side: the `{ type: "null", allOf: [{ $ref }] }` form comes from .NET's Swashbuckle (`UseAllOfToExtendReferenceSchemas` + OpenAPI 3.1 output) for nullable object properties; `@scalar/openapi-upgrader` (bundled by orval) passes it through unchanged. Since Scalar marked the upgrader fix won't-fix, handling it downstream in orval is the practical path.

The PR treats `type: "null"` with a sibling `allOf`/`oneOf`/`anyOf` as `T | null` instead of collapsing to `null`. Canonical `anyOf: [{$ref}, {type:"null"}]` encodings are unchanged. It's complementary to this PR (which handles the typeless `{ nullable: true }` case) and covers the `type:"null"` + composition gap I flagged earlier.

PR #3167, opened 2026-03-30
feat(query): add useFlatInput option to merge params into a single object

### LINKED ISSUE #3164: feat(react-query): merge path params, query params, and body into a single object (flattened)
### Description

When an endpoint has multiple parameter types (`path params`, `query params`, `body`), the generated hooks require multiple positional arguments which is verbose and usually inconventient.

Proposing a new `useFlatInput?: boolean` config option under override that flattens all parameters into a single object arguments.

### Current behavior

Path params are separated from query params and body,
while our data is usually in contained in one flat object ready to be sent.

For example, currently we have something like:
`type PathParams = { userId: string }`
`type QueryParams = { withOrders: boolean }`
`useGetUser(paths: PathParams, params: QueryParams)`

I had to use post-processing to achieve the flattened object style, and it was very convenient for me.
Achieving something like:
`useGetUser(data: PathParams & QueryParams)`

### Proposed behavior

The user doesn't need to know which fields are for path params, query, or body.
The generated code will internally extract them inside the hook based on the OpenAPI spec.

### Proposed solution

- Add `useFlatInput?: boolean` under `override`                                                                                                                                                  - For queries: destructure path params by name, rest spread for query params. If body exists, query param fields are also destructured by name and body becomes the rest spread                
- For mutations: destructure path + query param fields by name, rest spread for body                                                                                                             - Uses TypeScript intersection types (e.g. `{ userId: number } & GetUserOrdersParams`)
- Throws error on duplicate field names across path/query params                                                                                                                               
- Throws error if used together with `useNamedParameters` (they conflict)
- Headers stay as a separate parameter
- Existing behavior unchanged when option is not set

I have a working prototype and can submit a PR.

inas-sirhan:
## Summary

Adds `useFlatInput?: boolean` config option under `override` that flattens path params, query params, and body into a single object argument for query hooks, query options functions, and mutations.

## Example

```ts
// orval.config.ts
override: {
  useFlatInput: true,
}
```

**Query (path + query params):**
```ts
// Before
useGetUserOrders(userId, params, options)

// After
useGetUserOrders({ userId, status, fromDate }, options)
```

**Mutation (query params + body):**
```ts
// Before
mutate({ data: { name: 'Rex' }, params: { limit: 10 } })

// After
mutate({ name: 'Rex', limit: 10 })
```

**Mutation (path + query + body):**
```ts
// Before
mutate({ data: { query: 'search' }, params: { page: 1, limit: 10 } })

// After
mutate({ orgId: '123', page: 1, limit: 10, query: 'search' })
```

## How it works

- Uses TypeScript intersection types (e.g. `{ userId: number } & GetUserOrdersParams`)
- Path params + query param fields destructured by name, body as rest spread
- Throws on duplicate field names across path/query params
- Throws if used together with `useNamedParameters` (they conflict)
- Headers stay as a separate parameter
- Existing behavior unchanged when option is not set

## Tests

Snapshot tests for:
- Petstore (mutations with query + body, headers)
- Multi-params (queries with path + query params)
- Path + query + body (all three combined in a mutation)

Closes #3164

## Summary by CodeRabbit

* **New Features**
  * Opt-in "flat input" mode for query/mutation hooks and option builders — pass path, query and body params as a single flattened object.
  * Generator now exposes query field names to support flat input and updated config entries to enable it.

* **Bug Fixes**
  * Runtime validation that errors on duplicate parameter names when flat input is enabled.

* **Tests / Examples**
  * Added generated React Query examples, models, specs and configs demonstrating flat-input scenarios.


inas-sirhan:
Working on the problems

melloware:
@anymaniax @snebjorn i have no idea if this use case makes sense to support.  Please review.

inas-sirhan:
@melloware
Can you think of downsides of such a feature?
Is there cases where it might be problematic?

I personally used it and prefer it (used post processing for it) in my react, zod, react-hook-form apps.

I also remember seeing it on Discord from long time ago, and a maintainer said a PR is welcome or so

If another solution (maybe less invasive?) is preferred I can try other approaches and see where we land

snebjorn (commented):
I didn't review the code, but rather the concept.

OAS 3.1 allows duplicate parameter names. This is valid

```yml
paths:
  /users/{id}:
    get:
      parameters:
        - name: id
          in: path
          required: true
          schema:
            type: integer
        - in: query
          name: id
          schema:
            type: string
```
and would result in `getUsers({id: ..., id: ...})` which doesn't work.

inas-sirhan:
@snebjorn 
the code will throw error in such cases, so it won't allow bugs
and i believe it's rare someone would use same field name in path and query/body (confusing).

snebjorn:
IMO throwing an error on valid OAS is a bug, no matter how poorly the OAS was written 😄

Furthermore I'd be cautious about adding more state/options to an already messy/complex codebase. I wouldn't make this an option but rather switch to it instead. But the duplicate key issue would have to be addressed. And I'm not convinced it would be a net benefit, given the friction it would cause the users.

inas-sirhan:
@snebjorn
How about for such cases we simply fallback to normal behavior? (i.e separated instead of flattened)
and document that behavior?

This would solve the issue cleanly without throwing an error!

It's your call but I believe it's a good and beneficial thing to have (from my personal experience), because it's just boilerplate to split the fields, especially when all fields are unique.
Those who want it simply opt in for it, and those who don't, get the normal Orval's behavior

Lemme know if you want this, I can try to minimize changes and prop drilling to make it more maintainable in the future

melloware:
@inas-sirhan i am going to close this in its current state for the reasons @snebjorn has listed.  I am just not sure this use case is worth supporting but a simpler PR that does not cause regression issues would be considered

PR #3179, opened 2026-04-01
feat: runtime baseUrl (optional imports)

### LINKED ISSUE #3071: Support runtime baseUrl (e.g. process.env) without custom mutator
### Description

Support runtime-resolved `baseUrl` (e.g. from `process.env`) without using a custom mutator, so that generated fetch clients work with `runtimeValidation`, `includeHttpResponseReturnType`, and other built-in options while using a single Docker image across multiple environments.

### Current behavior

- Setting `baseUrl: process.env.API_BASE_URL` in orval config resolves the value at **code generation time**, so the generated client is tied to the build-time environment. This forces us to build a separate Docker image per environment.
- We want to use the **same** Docker image in dev/staging/prod and inject the API base URL at **runtime** via environment variables.
- A workaround is to set `baseUrl: '${process.env.API_BASE_URL}'` (as a string). The generator then embeds this literal into template literals in the generated code, so at runtime `${process.env.API_BASE_URL}` is evaluated correctly. This is undocumented and may break if orval changes how it embeds `baseUrl`.
- Using a custom mutator to set the base URL at runtime is documented, but when we use `override.fetch` options such as `runtimeValidation`, `includeHttpResponseReturnType`, and `forceSuccessResponse`, we rely on the built-in fetch client. A custom mutator would replace that and we would lose those features.

### Proposed behavior

- `baseUrl` should support a way to emit a **runtime expression** (e.g. `process.env.API_BASE_URL`) into the generated code instead of a build-time string, so that the same generated client works in any environment by setting env vars at runtime.
- This should work with the built-in fetch client and all existing `override.fetch` options (e.g. `runtimeValidation`, `includeHttpResponseReturnType`, `forceSuccessResponse`) without requiring a custom mutator.

### Proposed solution

- **Option A:** Document and guarantee that when `baseUrl` is a string that looks like a JavaScript expression (e.g. `'process.env.API_BASE_URL'` or `'${process.env.API_BASE_URL}'`), the generator continues to embed it as-is into the generated template literals (no escaping or wrapping in quotes), so it is evaluated at runtime. This would make the current workaround officially supported.
- **Option B:** Extend the `baseUrl` config type, e.g. add a form like `baseUrl: { runtime: 'process.env.API_BASE_URL' }` or `baseUrl: { env: 'API_BASE_URL' }`, and in the fetch generator, emit that expression into the URL construction so it is clearly a first-class “runtime baseUrl” feature.

Either way, the generated code would keep constructing URLs with template literals that evaluate the expression at runtime, enabling a single Docker image across environments while keeping `runtimeValidation` and other built-in fetch options.

mikan3rd:
## Approach

Adds `baseUrl: { runtime, imports? }` so generated clients embed a runtime expression in URL template literals (Issue #3071). This avoids resolving the API host at codegen time and keeps the built-in fetch stack (`runtimeValidation`, `includeHttpResponseReturnType`, etc.) without a custom mutator.

Optional `imports` reuses `GeneratorImport` so bindings from other modules are emitted into generated files alongside the URL.

## Caveats

- `importPath` is relative to the generated client file (same idea as mutator paths).
- MSW host filtering remains `mock.baseUrl`, separate from `output.baseUrl`.

Closes #3071

## Summary by CodeRabbit

* **New Features**
  * Support runtime-configurable API base URLs via JavaScript expressions and emit required imports into generated clients so runtime expressions resolve.

* **Documentation**
  * Added “Runtime base URL” guidance to guides and reference docs; clarified how to supply runtime expressions/imports and that mock.baseUrl is separate.

* **Tests**
  * Added tests for runtime base URL resolution and import handling.

* **Chore**
  * Ensured generated fetch clients remain compatible with existing override.fetch options and prior examples.

### DIFF
--- a/docs/content/docs/reference/configuration/output.mdx
+++ b/docs/content/docs/reference/configuration/output.mdx
@@ -235,9 +235,71 @@ export default defineConfig({
 });
 ```
 
-### getBaseUrlFromSpecification
+### runtime {#runtime}
 
-Read URL from the OpenAPI `servers` field:
+Embed a JavaScript expression into generated request URLs so the same build can call different hosts at runtime (for example with Docker images and environment variables). The value is emitted inside template literals in generated clients; only use trusted expressions from your configuration.
+
+```ts title="orval.config.ts"
+export default defineConfig({
+  petstore: {
+    output: {
+      baseUrl: {
+        runtime: 'process.env.API_BASE_URL',
+      },
+    },
+  },
+});
+```
+
+#### runtime
+
+**Type:** `String`
+
+JavaScript expression used inside generated template literals for the request base URL. Set this to the expression only (for example `process.env.API_BASE_URL`), not including `` `${...}` ``; Orval wraps it for you.
+
+#### imports
+
+**Type:** `GeneratorImport[]`
+
+Optional. When `runtime` references a symbol from another module, list the imports Orval should emit into generated clients. Paths are relative to the generated file, same idea as mutator imports. The `runtime` expression must be valid where the generated code runs (after those imports).
+
+Use a default import:
+
+```ts title="orval.config.ts"
+export default defineConfig({
+  petstore: {
+    output: {
+      baseUrl: {
+        runtime: 'apiBase',
+        imports: [{ name: 'apiBase', importPath: '../config/api' }],
+      },
+    },
+  },
+});
+```
+
+Or a named export used as an object (for example `import { env } from '../../env'` and `env.API_BASE_URL` in application code) — set `runtime` to that property access and import the object under `name`:
+
+```ts title="orval.config.ts"
+export default defineConfig({
+  petstore: {
+    output: {
+      baseUrl: {
+        runtime: 'env.API_BASE_URL',
+        imports: [{ name: 'env', importPath: '../../env' }],
+      },
+    },
+  },
+});
+```
+
+Adjust `importPath` so it resolves from the generated client file to your module (the example assumes the client is nested deeper than `env.ts`).
+
+### getBaseUrlFromSpecification {#getbaseurlfromspecification}
+
+**Type:** `Boolean`
+
+Read the base URL from the OpenAPI `servers` field instead of a fixed string. When `true`, Orval resolves it from the spec’s `servers` entry (optionally with `variables` and `index` below).
 
 ```ts title="orval.config.ts"
 export default defineConfig({
@@ -254,11 +316,17 @@ export default defineConfig({
 });
 ```
 
-### index
+#### variables
+
+**Type:** `Record<string, string>`
+
+Values for variables used in server URL templates from the OpenAPI `servers` field.
+
+#### index
 
 **Type:** `Number`
 
-Select which server URL index to use (when `getBaseUrlFromSpecification: true`):
+Which `servers` entry to use (0-based) when multiple URLs are defined:
 
 ```ts title="orval.config.ts"
 export default defineConfig({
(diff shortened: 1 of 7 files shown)


soartec-lab:
Thanks! It may take some time, but we will definitely check it.

docs/content/docs/reference/configuration/output.mdx
```diff
@@ -235,6 +235,58 @@ export default defineConfig({
 });
 ```
 
+### runtime
```
soartec-lab: Please write the details for each parameter, nested within the document structure 👍 

```
### runtime
#### baseUrl
#### imports
```

packages/core/src/getters/route.test.ts
```diff
@@ -139,6 +145,37 @@ describe('getFullRoute getter', () => {
...
+      undefined,
+      { runtime: 'process.env.API_BASE_URL' },
+      '${process.env.API_BASE_URL}/pets',
+    ],
+    [
+      '/pets',
+      undefined,
+      { runtime: '  process.env.API_BASE_URL  ' },
+      '${process.env.API_BASE_URL}/pets',
+    ],
```
soartec-lab: Since there are duplicate cases, can I delete one of them?

```suggestion
    [
      '/pets',
      undefined,
      { runtime: 'process.env.API_BASE_URL' },
      '${process.env.API_BASE_URL}/pets',
    ],
```

docs/content/docs/reference/configuration/output.mdx
mikan3rd: Headings now follow config keys: `### runtime` + `#### runtime` / `#### imports`, and `### getBaseUrlFromSpecification` + `#### variables` / `#### index`. Anchors `#runtime` and `#getbaseurlfromspecification` kept.

packages/core/src/getters/route.test.ts
mikan3rd: Removed the trim/whitespace duplicate; kept the plain `process.env.API_BASE_URL` case.

docs/content/docs/reference/configuration/output.mdx
```diff
@@ -235,9 +235,75 @@ export default defineConfig({
 });
 ```
 
-### getBaseUrlFromSpecification
+### Runtime URL expression {#runtime}
```
soartec-lab: Match the configuration with other properties.

```suggestion
### baseUrl
```

docs/content/docs/reference/configuration/output.mdx
```diff
@@ -254,11 +320,17 @@ export default defineConfig({
...
 ```
 
-### index
+#### variables
+
+**Type:** `Record<string, string>`
+
+Values for variables used in server URL templates from the OpenAPI `servers` field.
+
+#### index {#index}
```
soartec-lab:
```suggestion
#### index
```

docs/content/docs/reference/configuration/output.mdx
mikan3rd: Using `### runtime` instead of `### baseUrl` to avoid repeating `## baseUrl`. Spec side is `### getBaseUrlFromSpecification` for the same reason.

docs/content/docs/reference/configuration/output.mdx
mikan3rd: Removed `{#index}` from the `#### index` heading.

mikan3rd:
@soartec-lab 
Thank you for your review!

PR #3180, opened 2026-04-01
feat(fetch): add `fetcher` option to support custom fetch function injection

### LINKED ISSUE #3119: Svelte-Query: Prefetch use provided fetch
### Description

When prefetching data in a svelte load function, you should be using the provided fetch function. Orvals usePrefetch generated functions do not allow for passing in this function. Orvals generated usePrefetch functions are not compatible with the [recommended way](https://tanstack.com/query/latest/docs/framework/svelte/ssr) of handing preFetching in svelte 

### Current behavior

currently you would think of doing something like this.
```js
export const load: PageLoad = async ({ parent,fetch }) => {
	const { queryClient } = await parent();
	await prefetchFooQuery(queryClient);
};
```
but this will throw a warning in the browser 
```
Loading /foo using `window.fetch`. For best results, use the `fetch` that is passed to your `load` function: https://svelte.dev/docs/kit/load#making-fetch-requests client.js:3299:15
```
the best you can do with orval's functions is 
```js
export const load: PageLoad = async ({ parent,fetch }) => {
	const { queryClient } = await parent();
		await queryClient.prefetchQuery({
		queryKey: getRestCircuitsActiveQueryKey(),
		queryFn: () => fetch(...),
	});
};
```
which negates needing to generate the prefetch functions in the first place.

### Proposed behavior

it would be nice to be able to do something like this
```js
export const load: PageLoad = async ({ parent,fetch }) => {
	const { queryClient } = await parent();
	await prefetchFooQuery(queryClient,fetch);
};
```
To retain the usefulness of the orval generated function without violating svelte rules.

### Proposed solution

Implement some way to allow the caller to pass a fetch client to at least prefetch functions

zeriong:
## Summary

Closes #3119

When using SSR frameworks like SvelteKit, the `load` function provides a context-specific `fetch` that should be used instead of `window.fetch`. Previously, orval's generated prefetch functions had no way to accept a custom fetch function, causing browser warnings:

> "Loading /foo using `window.fetch`. For best results, use the `fetch` that is passed to your `load` function"

This PR adds an optional `fetcher` property to the query options type for fetch-based clients (without custom mutator), allowing users to inject a custom fetch function.

### Changes

- **`@orval/fetch`**: Generated request functions now accept an optional `fetchFn` parameter, using `(fetchFn ?? fetch)()` instead of `fetch()`
- **`@orval/query`**: Query options type includes `fetcher?: typeof globalThis.fetch`, which is destructured and forwarded to request functions
- Fully backward-compatible — `fetcher` is optional and defaults to global `fetch`
- Custom mutator users are unaffected (no changes to their generated code)

### Usage

```typescript
// SvelteKit +page.ts
export const load = async ({ fetch }) => {
  const queryClient = new QueryClient();
  await prefetchListPetsQuery(queryClient, params, {
    fetcher: fetch, // SvelteKit's context-specific fetch
  });
  return { queryClient };
};
```

## Test plan

- [x] Updated existing `getQueryOptions` tests for fetch client
- [x] Added new tests for `getQueryArgumentsRequestType`, `getHookOptions`, `getQueryOptions` (11 new tests)
- [x] Verified generated output for default fetch client (`usePrefetchWithFunctionMutator`)
- [x] Verified custom mutator samples are unaffected
- [x] All 895 tests pass (3 pre-existing failures in `resolve-version.test.ts` are unrelated)

## Summary by CodeRabbit

* **New Features**
  * Added `useRuntimeFetcher` configuration option enabling runtime-injectable fetch functions for fetch-based HTTP clients.
  * Added `useGetQueryData` configuration option generating query data accessor functions for TanStack Query clients.

* **Documentation**
  * Added comprehensive documentation for new runtime fetcher and query data configuration options.

### DIFF
--- a/docs/content/docs/reference/configuration/output.mdx
+++ b/docs/content/docs/reference/configuration/output.mdx
@@ -1094,38 +1094,14 @@ Enable Zod runtime validation for fetch client responses. Requires `schemas: { t
 **Type:** `Boolean`
 **Default:** `false`
 
-Allow injecting a custom `fetch` function at runtime. When enabled:
-
-- Every generated request function gains an optional `fetchFn?: typeof globalThis.fetch` parameter and calls `(fetchFn ?? fetch)(...)` instead of `fetch(...)`.
-- For query clients (`react-query`, `vue-query`, `svelte-query`) with `httpClient: 'fetch'`, every generated query options helper, `useQuery` hook, `useMutation` hook, and prefetch function also gains a `fetcher?: typeof globalThis.fetch` field in its options type.
-
-Applies to the `fetch` client and all fetch-based query clients (`react-query`, `vue-query`, `svelte-query`). Has no effect on axios-based clients or any operation that uses a custom mutator.
-
-For the SWR client, the underlying request functions receive `fetchFn`, but SWR hook options do not include a `fetcher` field.
-
-The `angular` client (Angular HttpClient) is not affected.
+Allow injecting a custom `fetch` function at runtime. When enabled, generated request functions accept an optional `fetchFn` parameter and call `(fetchFn ?? fetch)(...)` instead of the global `fetch(...)`. For query client hooks, a `fetcher` field is added to the options type. Has no effect on operations that use a custom mutator.
 
 ```ts title="orval.config.ts"
-// With a query client (react-query / vue-query / svelte-query)
 export default defineConfig({
   petstore: {
     output: {
-      client: 'react-query',
+      client: 'react-query', // also works with 'fetch', 'vue-query', 'svelte-query'
       httpClient: 'fetch',
-      override: {
-        fetch: {
-          useRuntimeFetcher: true,
-        },
-      },
-    },
-  },
-});
-
-// With the standalone fetch client
-export default defineConfig({
-  petstore: {
-    output: {
-      client: 'fetch',
       override: {
         fetch: {
           useRuntimeFetcher: true,
@@ -1136,9 +1112,10 @@ export default defineConfig({
 });
 ```
 
-**Generated output (request function)**
+**Generated output example**
 
 ```ts
+// Request function — gains a fetchFn parameter
 export const listPets = async (
   params: ListPetsParams,
   options?: RequestInit,
@@ -1150,14 +1127,8 @@ export const listPets = async (
   });
   // ...
 };
-```
 
-**Generated output (query / mutation / prefetch options)**
-
-All generated helpers accept the same `fetcher` field:
-
-```ts
-// useQuery
+// Query hook — gains a fetcher field in options
 export const useListPets = (
   params: ListPetsParams,
   options?: {
@@ -1166,31 +1137,9 @@ export const useListPets = (
     fetcher?: typeof globalThis.fetch;
   },
 ) => { ... };
-
-// useMutation
-export const useCreatePets = (
-  options?: {
-    mutation?: UseMutationOptions<...>;
-    fetch?: RequestInit;
-    fetcher?: typeof globalThis.fetch;
-  },
-) => { ... };
-
-// prefetch
-export const prefetchListPetsQuery = async (
-  queryClient: QueryClient,
-  params: ListPetsParams,
-  options?: {
-    query?: UseQueryOptions<...>;
-    fetch?: RequestInit;
-    fetcher?: typeof globalThis.fetch;
-  },
-) => { ... };
 ```
 
-**Common use cases**
-
-This is useful in any context where you need to inject a custom `fetch` implementation — for example, SSR frameworks that provide a request-scoped `fetch`:
+**Usage — SSR with request-scoped fetch**
 
 ```ts
 // SvelteKit — +page.ts
@@ -1199,16 +1148,6 @@ export const load = async ({ fetch }) => {
   await prefetchListPetsQuery(queryClient, params, { fetcher: fetch });
   return { queryClient };
 };
-
-// Next.js App Router — page.tsx (server component)
-import { cache } from 'react';
-const getQueryClient = cache(() => new QueryClient());
-
-export default async function Page() {
-  const queryClient = getQueryClient();
-  await prefetchListPetsQuery(queryClient, params, { fetcher: fetch });
-  // ...
-}
 ```
 
 ---


soartec-lab:
Thanks. It may take some time, but we will definitely check it.

packages/fetch/src/index.ts
```diff
@@ -312,7 +312,9 @@ ${override.fetch.forceSuccessResponse && hasSuccess ? '' : `export type ${respon
     })
     .join(',');
 
-  const args = `${toObjectString(props, 'implementation')} ${isRequestOptions ? `options?: RequestInit` : ''}`;
+  const fetchFnParam =
+    isRequestOptions && !mutator ? ', fetchFn?: typeof globalThis.fetch' : '';
```
soartec-lab: Why not specify type `Promise<Response>` directly?

```suggestion
    isRequestOptions && !mutator ? ', fetchFn?: Promise<Response>' : '';
```

samples/hono/hono-with-fetch-client/__snapshots__/next-app/pets/pets.ts
```diff
@@ -227,8 +230,9 @@ export const getShowPetByIdUrl = (petId: string) => {
 export const showPetById = async (
   petId: string,
   options?: RequestInit,
+  fetchFn?: typeof globalThis.fetch,
```
soartec-lab: I want to simplify the function logic by making certain fields required and setting default values ​​when defining the function.

```suggestion
  fetcher: Promise<Response> = fetch,
```

packages/fetch/src/index.ts
zeriong: `Promise<Response>` is the **return type** of a fetch call, not the function type itself.

`fetchFn` needs to be callable — `fetchFn(url, init)` — but `Promise<Response>` is not callable, so this would cause a type error.

`typeof globalThis.fetch` correctly captures the full callable signature:
```ts
(input: RequestInfo | URL, init?: RequestInit) => Promise<Response>

File: samples/hono/hono-with-fetch-client/__snapshots__/next-app/pets/pets.ts (hunk 2 above)
Comment (zeriong) [reply samples/hono/hono-with-fetch-client/__snapshots__/next-app/pets/pets.ts]: Regarding making it required with a default value (`fetcher = fetch`): this would work syntactically, but it conflicts with the `override.fetch.useRuntimeFetcher` gate suggested above.

When the option is **disabled** (default), the `fetchFn` parameter won't appear in the signature at all — so required vs optional becomes irrelevant.

When **enabled**, keeping it optional (`fetchFn?`) is the better choice because:
- Existing callers don't need to change their code
- `(fetchFn ?? fetch)` already provides the default fallback in the function body
- A required parameter with a default value still occupies a positional slot, which can cause confusion when there are preceding optional parameters like `options?: RequestInit`

**Comment by zeriong** [comment]
Thank you for the detailed review, @soartec-lab!

## Responses to Review Comments

### 1. Configuration Gate (`override.fetch.useRuntimeFetcher`)

Completely agree — adding an extra parameter by default goes against the fetch client's simplicity principle. I'll gate the `fetchFn` parameter behind `override.fetch.useRuntimeFetcher` so it's only generated when explicitly opted in.

### 2. Type: `typeof globalThis.fetch` vs `Promise<Response>`

`Promise<Response>` is the **return type** of a fetch call, not the function type itself. Since `fetchFn` needs to be callable — `fetchFn(url, init)` — using `Promise<Response>` would cause a type error because a `Promise` is not callable.

`typeof globalThis.fetch` correctly captures the full callable signature:

```ts
(input: RequestInfo | URL, init?: RequestInit) => Promise<Response>
```

This ensures callers can pass SvelteKit's `event.fetch` or any spec-compliant fetch implementation with full type safety.

### 3. Required with Default Value (`fetcher = fetch`) vs Optional (`fetchFn?`)

With the `useRuntimeFetcher` gate in place, when the option is **disabled** (default), the parameter won't appear in the signature at all — so required vs optional becomes irrelevant.

When **enabled**, `fetchFn?` with `(fetchFn ?? fetch)` is a clean, standard pattern that keeps backward compatibility without requiring any change from existing callers.

## Proposed Implementation Plan

### Changes by file

| File | Change |
|------|--------|
| `packages/core/src/types.ts` | Add `useRuntimeFetcher` to `FetchOptions` / `NormalizedFetchOptions` |
| `packages/orval/src/utils/options.ts` | Normalize with default `false` |
| `packages/fetch/src/index.ts` | Gate `fetchFn?` parameter and `(fetchFn ?? fetch)()` call behind config |
| `packages/query/src/client.ts` | Gate `fetcher` in query hook options behind the same config |
| `packages/query/src/mutation-generator.ts` | Fix mutation hooks to forward `fetcherFn` to underlying API calls (CodeRabbit finding) |

### Generated output comparison

**`useRuntimeFetcher: false` (default) — no change from current behavior:**

```ts
  params?: ListPetsParams,
  options?: RequestInit,
    ...options,
    method: 'GET',
  });
  // ...
};
```

**`useRuntimeFetcher: true` — opt-in only:**

```ts
  params?: ListPetsParams,
  options?: RequestInit,
  fetchFn?: typeof globalThis.fetch,
  const res = await (fetchFn ?? fetch)(getListPetsUrl(params), {
    ...options,
    method: 'GET',
  });
  // ...
};
```

### Additional fixes

- **Mutation forwarding bug** (flagged by CodeRabbit): Mutation helpers currently declare `fetcher` in their options type but never forward it to the underlying request functions. This will be corrected so `fetchFn` is properly threaded through when `useRuntimeFetcher` is enabled.
- **SWR support**: SWR hooks will also be updated to support the fetcher parameter under the same config gate.

I'll update the PR with these changes. Thanks again for the feedback!

packages/fetch/src/index.ts
soartec-lab: i see, I understand perfectly. Thanks!

zeriong:
> @zeriong
> 
> That's great. One more thing, could you please write a guide about the newly added parameters?
> 
> https://orval.dev/docs/reference/configuration/output

@soartec-lab 
Thank you for understanding the intent behind the changes!
I'll get the documentation update done as soon as possible.

docs/content/docs/reference/configuration/output.mdx
```diff
@@ -1088,6 +1088,128 @@ Custom JSON reviver function (useful for date parsing).
...
+
+```ts
+// SvelteKit — +page.ts
+export const load = async ({ fetch }) => {
+  const queryClient = new QueryClient();
+  await prefetchListPetsQuery(queryClient, params, { fetcher: fetch });
+  return { queryClient };
+};
+
+// Next.js App Router — page.tsx (server component)
```
soartec-lab: Does this sample provide any explanation?

docs/content/docs/reference/configuration/output.mdx
```diff
@@ -1088,6 +1088,128 @@ Custom JSON reviver function (useful for date parsing).
...
+  // ...
+};
+```
+
+**Generated output (query / mutation / prefetch options)**
+
+All generated helpers accept the same `fetcher` field:
+
+```ts
+// useQuery
```
soartec-lab: If this applies to all cases, it will need to be updated as the number of cases increases in the future. How about summarizing them as a representative example?

docs/content/docs/reference/configuration/output.mdx
```diff
@@ -1088,6 +1088,128 @@ Custom JSON reviver function (useful for date parsing).
...
 Enable Zod runtime validation for fetch client responses. Requires `schemas: { type: 'zod' }`. When enabled, JSON responses are validated via `Schema.parse()` before being returned.
 
+### useRuntimeFetcher
+
+**Type:** `Boolean`
+**Default:** `false`
+
+Allow injecting a custom `fetch` function at runtime. When enabled:
+
+- Every generated request function gains an optional `fetchFn?: typeof globalThis.fetch` parameter and calls `(fetchFn ?? fetch)(...)` instead of `fetch(...)`.
```
soartec-lab: You only need to apply this effect to the extent that it is applicable; it doesn't need to cover all clients.

docs/content/docs/reference/configuration/output.mdx
zeriong: Added inline comments to clarify each generated output (e.g. // gains a fetchFn parameter, // gains a fetcher field in options), and the config example now notes which other clients are supported via a comment.

docs/content/docs/reference/configuration/output.mdx
zeriong: Good point. Consolidated the generated output into a single representative example (one request function + one query hook), merged the two config examples into one, and reduced the use case section to a single SvelteKit example.

docs/content/docs/reference/configuration/output.mdx
zeriong: Agreed. Removed the per-client breakdown (SWR, angular, etc.) and kept the description focused on the core behavior only.

PR #3193, opened 2026-04-03
fix: implement style:deepObject query parameter serialization

### LINKED ISSUE #2589: style=deepObject with an object as a parameter is not supported for proper URL formation
## Problem

I expect the pagination parameter to appear in the URL as `&pagination[limit]=10&pagination[offset]=0`.

However, the generated URL ends up as `&pagination=[object+Object]`, because object-type query parameters are not supported.

There is no configuration option to override or customize URL generation, which makes it impossible to use objects for forming query parameters.

More details in the documentation - https://spec.openapis.org/oas/v3.1.0.html#style-examples

Is there a plan to add `style` support?

## Scheme

```json
"parameters": [
	{
		"name": "pagination",
		"in": "query",
		"required": true,
		"style": "deepObject",
		"explode": true,
		"schema": {
			"type": "object",
			"title": "PaginationInDto",
			"properties": {
				"offset": {
					"type": "integer"
				},
				"limit": {
					"type": "integer"
				}
			},
			"required": [
				"offset",
				"limit"
			]
		}
	}
],
```

## Example of usage in code

```js
const { data } = useGetCatalogRubric({
	pagination: {
		limit: 10,
		offset: 0,
	},
});
```

leanderpeter:
## Summary

Implements `style: deepObject` query parameter serialization for `@orval/fetch` and `@orval/solid-start`.

When a query parameter declares `style: deepObject` with an object schema, the generated URL-building code now emits bracket notation (`key[field]=value`) instead of calling `.toString()` on the object, which produced `key=[object+Object]`.

Closes #2589

## Changes

- **`packages/fetch/src/index.ts`** — Added `deepObject` parameter detection and a template branch that serializes object values as `key[subKey]=value`. Decoded `%5B`/`%5D` from `URLSearchParams` output so servers receive raw bracket notation. Added `useDates` support for deepObject sub-values with `format: date-time`.
- **`packages/solid-start/src/index.ts`** — Same changes, following solid-start's existing conventions (path-level parameter merging, OpenAPI default explode semantics).
- **`packages/fetch/src/index.test.ts`** *(new)* — 9 tests covering deepObject serialization, bracket decoding, mixed param types, scalar fallback omission, and `useDates` handling.
- **`packages/solid-start/src/index.test.ts`** — 9 new tests in a `deepObject query parameters` describe block, plus bracket decoding and `useDates` coverage.

## Approach

Follows the existing pattern for `explode: true` array parameter handling:

1. Filter `parameterObjects` for `style === 'deepObject'` query params
2. Generate a template branch inside the `Object.entries(params).forEach(...)` loop
3. Emit `normalizedParams.append(\`${key}[${subKey}]\`, ...)` for each sub-key
4. Decode percent-encoded brackets in the serialized output (`.replace(/%5B/g, '[').replace(/%5D/g, ']')`)

Scoped to `deepObject` only — `pipeDelimited`, `spaceDelimited`, and other style values are not addressed.

## Test plan

- [x] `bun run build` — 12/12 packages
- [x] `bun run typecheck` — 12/12 packages
- [x] `bun run test` — 903 unit tests pass (9 new in fetch, 9 new in solid-start)
- [x] `bun run test:snapshots` — 3,481 snapshot tests pass, no updates needed

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Added support for OpenAPI `deepObject` query parameters, serializing nested values with bracketed notation.
  * DeepObject parameters now combine correctly with scalar and exploded array query parameters.
  * Nested date-time values can be converted to ISO strings when date handling is enabled.

* **Tests**
  * Added comprehensive coverage for serialization, parameter combinations, multiple deepObject parameters, and date handling.

### DIFF
--- a/packages/fetch/src/index.test.ts
+++ b/packages/fetch/src/index.test.ts
@@ -1,0 +1,470 @@
+import type {
+  ContextSpec,
+  GeneratorOptions,
+  GeneratorVerbOptions,
+  OpenApiParameterObject,
+  OpenApiReferenceObject,
+} from '@orval/core';
+import {
+  EnumGeneration,
+  FormDataArrayHandling,
+  NamingConvention,
+  OutputClient,
+  OutputHttpClient,
+  OutputMode,
+  PropertySortOrder,
+  Verbs,
+} from '@orval/core';
+import { describe, expect, it } from 'vitest';
+
+import { generateRequestFunction } from './index';
+
+type OpenApiParameterLike = OpenApiParameterObject | OpenApiReferenceObject;
+
+function makeOutput(useDates = false): ContextSpec['output'] {
+  return {
+    target: '',
+    namingConvention: NamingConvention.CAMEL_CASE,
+    fileExtension: '.ts',
+    mode: OutputMode.SINGLE,
+    client: OutputClient.FETCH,
+    httpClient: OutputHttpClient.FETCH,
+    clean: false,
+    docs: false,
+    formatter: undefined,
+    headers: false,
+    indexFiles: false,
+    allParamsOptional: false,
+    urlEncodeParameters: false,
+    unionAddMissingProperties: false,
+    optionsParamRequired: false,
+    propertySortOrder: PropertySortOrder.ALPHABETICAL,
+    override: {
+      title: undefined,
+      transformer: undefined,
+      mutator: undefined,
+      operations: {},
+      tags: {},
+      mock: undefined,
+      contentType: undefined,
+      header: false,
+      formData: {
+        disabled: false,
+        arrayHandling: FormDataArrayHandling.SERIALIZE,
+      },
+      formUrlEncoded: false,
+      paramsSerializer: undefined,
+      paramsSerializerOptions: undefined,
+      namingConvention: {},
+      components: {
+        schemas: { suffix: '', itemSuffix: '' },
+        responses: { suffix: '' },
+        parameters: { suffix: '' },
+        requestBodies: { suffix: '' },
+      },
+      hono: { compositeRoute: '', validator: false, validatorOutputPath: '' },
+      query: {
+        useQuery: false,
+        useSuspenseQuery: false,
+        useMutation: false,
+        useInfinite: false,
+        useSuspenseInfiniteQuery: false,
+        useInfiniteQueryParam: '',
+        usePrefetch: false,
+        useInvalidate: false,
+        useSetQueryData: false,
+        useGetQueryData: false,
+        shouldExportMutatorHooks: false,
+        shouldExportHttpClient: false,
+        shouldExportQueryKey: false,
+        shouldSplitQueryKey: false,
+        useOperationIdAsQueryKey: false,
+        signal: false,
+        version: 5,
+      },
+      angular: {
+        provideIn: 'root',
+        client: 'httpClient',
+        runtimeValidation: false,
+      },
+      swr: {},
+      zod: {
+        strict: {
+          param: false,
+          query: false,
+          header: false,
+          body: false,
+          response: false,
+        },
+        generate: {
+          param: false,
+          query: false,
+          header: false,
+          body: false,
+          response: false,
+        },
+        coerce: {
+          param: false,
+          query: false,
+          header: false,
+          body: false,
+          response: false,
+        },
+        generateEachHttpStatus: false,
+        useBrandedTypes: false,
+        dateTimeOptions: {},
+        timeOptions: { precision: 3 },
+      },
+      fetch: {
+        includeHttpResponseReturnType: false,
+        forceSuccessResponse: false,
+        runtimeValidation: false,
+      },
+      useDates,
+      enumGenerationType: EnumGeneration.UNION,
+      jsDoc: {},
+      requestOptions: true,
+      aliasCombinedTypes: false,
+    },
+  };
+}
+
+function makeContext(
+  parameters: OpenApiParameterLike[] = [],
+  useDates = false,
+): ContextSpec {
+  return {
+    target: '',
+    workspace: '',
+    spec: {
+      openapi: '3.1.0',
+      info: { title: 'Test' },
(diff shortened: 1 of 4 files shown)


soartec-lab:
We will check them in order, so please wait a moment.

packages/fetch/src/index.test.ts
```diff
@@ -0,0 +1,470 @@
...
+  PropertySortOrder,
+  Verbs,
+} from '@orval/core';
+import { describe, expect, it } from 'vitest';
+
+import { generateRequestFunction } from './index';
+
+type OpenApiParameterLike = OpenApiParameterObject | OpenApiReferenceObject;
+
+function makeOutput(useDates = false): ContextSpec['output'] {
```
soartec-lab, line 24: Could this definition be simplified using the core generator? The same applies to other definitions.

packages/fetch/src/index.test.ts
```diff
@@ -0,0 +1,470 @@
...
+          },
+        },
+      },
+    ];
+    const verbOptions = makeVerbOptions({ queryParams: STUB_QUERY_PARAMS });
+    const options = makeOptions(makeContext(parameters));
+
+    const implementation = generateImplementation(verbOptions, options);
+
+    expect(implementation).toContain('const deepObjectParameters = ["scope"]');
```
soartec-lab, line 254: Would it be difficult to use the entire set of functions instead of `toContain`?

packages/fetch/src/index.test.ts
```diff
@@ -0,0 +1,470 @@
...
+}
+
+describe('generateRequestFunction — deepObject query parameters', () => {
+  it('generates bracket notation for a style:deepObject query param', () => {
+    const parameters = [
+      {
+        name: 'scope',
+        in: 'query',
+        style: 'deepObject',
+        explode: true,
```
soartec-lab, line 240: What happens if the style is 'deepObject' but the explode method is not specified (false)?

leanderpeter: Current implementation treats style: 'deepObject' as sufficient to use deepObject serialization, even if explode is omitted or false.

I chose that because the practical intent behind this issue is to support bracket-style object query serialization, and the OpenAPI examples for deepObject are all shown with exploded object entries.

But the spec default for non-form styles makes the omitted/false case ambiguous. If you’d prefer strict handling here, I can change both generators to only apply this branch when style === 'deepObject' && explode === true, and add tests for the omitted/false cases (no strong opinion on this one)

packages/fetch/src/index.test.ts
leanderpeter: I mirrored the local helper setup from `packages/solid-start/src/index.test.ts` to keep the test style aligned with an existing package-level generator test

packages/fetch/src/index.test.ts
leanderpeter: I used toContain because that’s the dominant pattern in the existing solid-start generator tests, so I kept the fetch tests aligned with the current package-level style. If you’d prefer more behavioral assertions here, I can follow up by exercising the generated output more directly rather than checking structural snippets.

packages/fetch/src/index.test.ts
soartec-lab: I see, it's okay for now.

packages/fetch/src/index.test.ts
soartec-lab: Yes, it's unclear from this alone whether this is the correct behavior, so I'd like to change it.

packages/fetch/src/index.test.ts
soartec-lab: Yes, I want to strictly restrict it to only `style === 'deepObject' && explode === true`.

soartec-lab:
@leanderpeter 
Thank you. I replied to your comment.

artnikbrothers:
We use Orval with large OpenAPI specs in production, so I'd like to see this land. Checked out the branch - looks like the CodeRabbit issues from April are already fixed in the current code (path-level params are merged now, and brackets are not decoded across the whole query string anymore). So this seems to be waiting on a re-review and one decision, not on fixes.

On the explode question: careful with requiring `explode: true`. For deepObject, `explode` defaults to `false` when omitted (only `form` defaults to true - [spec](https://spec.openapis.org/oas/v3.1.0#parameter-object)). So a spec that just writes `style: deepObject` - which is the common case - would silently get `[object Object]` again. If you want strict, I'd at least print a warning at generation time for that case. Current behavior (style alone is enough) seems fine to me, the spec doesn't give deepObject any other meaning anyway.

One small thing I noticed while reading: nested objects still turn into `[object Object]`, e.g. `filter: { range: { min: 1 } }` -> `filter[range]=[object Object]`. Supporting one level only is fine, but maybe document it or warn instead of outputting that silently. Not a blocker.

soartec-lab:
@leanderpeter 
This PR is a great idea and I’d like to move forward with merging it, but are you planning to resume work on it?
If so, please keep the PR open, make the necessary corrections, and then submit a new request for review. If I don't hear back from you, I will close or I take over this PR.

packages/fetch/src/index.test.ts
soartec-lab: I have re-checked the specifications.

- The default for `explode` is "true only when `style: form`, and false for other styles."
- According to the spec, if only `style: deepObject` is specified, `explode` defaults to `false`.

Therefore, the stricter handling I mentioned earlier is unnecessary, and the current implementation is fine as is.

soartec-lab:
In other words, simply rebasing this PR should be enough. I'll go ahead and handle that.

melloware:
looks like a merge conflict now?

soartec-lab:
Yes, I'll fix this.

melloware:
sorry another one :)

soartec-lab (approved):
I rebased this and resolved the conflicts 👍

PR #3194, opened 2026-04-03
 fix(fetch): omit null query parameter values instead of serializing as "null"

### LINKED ISSUE #3192: fetch client: null query parameter values are serialized as the literal string "null"
### Description

The `getUrl()` functions generated by `httpClient: fetch` serialize `null` query parameter values as the literal string `"null"` instead of omitting them from the query string.

RFC 6570 Section 2.3 explicitly lists `null` as an example of an "undefined value", and Section 3.2.1 specifies that undefined variables should be skipped during expansion. `null` values should be omitted from the query string.

When a backend expects `offset` as an integer, receiving the string `"null"` causes a 400 Bad Request or parse error. Since the generated type is `offset?: number | null`, it is natural for frontend code to pass `null` expecting the parameter to be omitted.

The same issue occurs with `anyOf` + `{ type: 'null' }` schema definitions. Custom fetch mutators cannot work around this because the URL is already constructed inside `getUrl()` (related: Discussion #2368).

This behavior appears to be related to PR #2932.

### Output client

fetch

```ts
export default defineConfig({
  api: {
    output: {
      httpClient: "fetch",
    },
  },
});
```

### Environment

- orval: v8.5.2

### Expected behavior

`null` should be omitted from the query string, just like `undefined`:

```typescript
listCompanies({ offset: null })
// Expected: GET /companies  (parameter omitted)
```

Suggested fix — just add `&& value !== null` to the existing condition:

```diff
 Object.entries(params || {}).forEach(([key, value]) => {
-  if (value !== undefined) {
-    normalizedParams.append(key, value === null ? 'null' : value.toString())
+  if (value !== undefined && value !== null) {
+    normalizedParams.append(key, value.toString())
   }
 });
```

### Actual behavior

`null` is serialized as the literal string `"null"`:

```typescript
listCompanies({ offset: null })
// Actual: GET /companies?offset=null
```

```yaml
openapi: 3.0.3
info:
  title: Demo
  version: 0.0.0
paths:
  /companies:
    get:
      operationId: list_companies
      parameters:
        - name: offset
          in: query
          required: false
          schema:
            type: integer
            format: int32
            nullable: true
      responses:
        '200':
          description: OK
```

### Additional context

Rationale:
- **[RFC 6570 Section 2.3](https://datatracker.ietf.org/doc/html/rfc6570#section-2.3)**: Explicitly lists `null` as an example of an "undefined value"
- **[RFC 6570 Section 3.2.1](https://datatracker.ietf.org/doc/html/rfc6570#section-3.2.1)**: Specifies that undefined variables should be skipped during expansion
- **[OpenAPI 3.0.4](https://spec.openapis.org/oas/v3.0.4.html) / [OpenAPI 3.1.1](https://spec.openapis.org/oas/v3.1.1.html)**: Parameter serialization is based on RFC 6570. `nullable: true` indicates that null is a valid schema value — it does not prescribe sending the literal string `"null"` in the query string
- **Common REST API practice**: `?offset=null` cannot be parsed as an integer by most server frameworks

Related code: [`normalParamsImplementation` in `packages/fetch/src/index.ts`](https://github.com/orval-labs/orval/blob/master/packages/fetch/src/index.ts)

tk1024:
The `getUrl()` functions generated by `httpClient: fetch` serialize `null` query parameter values as the literal string `"null"` instead of omitting them from the query string.

## Problem

When a parameter type is `offset?: number | null`, passing `null` produces `?offset=null`. Most server frameworks cannot parse the string `"null"` as an integer, resulting in 400 Bad Request errors. Custom fetch mutators cannot work around this because the URL is already constructed inside `getUrl()`.

## Fix

- Skip `null` values in `normalParamsImplementation` by changing `value !== undefined` to `value !== undefined && value
  !== null`
- Skip `null` and `undefined` elements in `explodeArrayImplementation` by adding an early return guard

This aligns with [RFC 6570 Section 2.3](https://datatracker.ietf.org/doc/html/rfc6570#section-2.3), which explicitly lists `null` as an example of an "undefined value" that should be omitted during expansion.

Closes #3192

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed query parameter serialization to properly exclude `null` values instead of converting them to the literal string `'null'`. Generated request URLs no longer include parameters with null values, ensuring cleaner and more predictable API requests across all client implementations.

### DIFF
--- a/samples/hono/hono-with-fetch-client/__snapshots__/next-app/pets/pets.ts
+++ b/samples/hono/hono-with-fetch-client/__snapshots__/next-app/pets/pets.ts
@@ -71,8 +71,8 @@ export const getListPetsUrl = (params?: ListPetsParams) => {
   const normalizedParams = new URLSearchParams();
 
   Object.entries(params || {}).forEach(([key, value]) => {
-    if (value !== undefined) {
-      normalizedParams.append(key, value === null ? 'null' : value.toString());
+    if (value !== undefined && value !== null) {
+      normalizedParams.append(key, value.toString());
     }
   });
 
--- a/samples/hono/hono-with-fetch-client/next-app/app/gen/pets/pets.ts
+++ b/samples/hono/hono-with-fetch-client/next-app/app/gen/pets/pets.ts
@@ -71,8 +71,8 @@ export const getListPetsUrl = (params?: ListPetsParams) => {
   const normalizedParams = new URLSearchParams();
 
   Object.entries(params || {}).forEach(([key, value]) => {
-    if (value !== undefined) {
-      normalizedParams.append(key, value === null ? 'null' : value.toString());
+    if (value !== undefined && value !== null) {
+      normalizedParams.append(key, value.toString());
     }
   });
 
--- a/samples/mcp/petstore/src/http-client.ts
+++ b/samples/mcp/petstore/src/http-client.ts
@@ -107,8 +107,8 @@ export const getFindPetsByStatusUrl = (params?: FindPetsByStatusParams) => {
   const normalizedParams = new URLSearchParams();
 
   Object.entries(params || {}).forEach(([key, value]) => {
-    if (value !== undefined) {
-      normalizedParams.append(key, value === null ? 'null' : value.toString());
+    if (value !== undefined && value !== null) {
+      normalizedParams.append(key, value.toString());
     }
   });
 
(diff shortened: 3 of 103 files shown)


soartec-lab:
We will check them in order, so please wait a moment.

snebjorn:
Closing as per discussion in #3192

PR #3197, opened 2026-04-04
fix(query): fix mutationInvalidates missing args and wrong import path (#3090)

### LINKED ISSUE #3090: Bug: mutationInvalidates generates broken code with react-query + tags-split mode
## Bug Report

**orval version:** 8.5.3  
**client:** react-query  
**mode:** tags-split

---

### Bug 1: Generated `onSuccess` calls query key functions without required arguments

When `mutationInvalidates` is configured, the generated `onSuccess` callback calls query key functions with **no arguments**, even when those functions have required path parameters.

#### Configuration

```js
// orval.config.mjs
{
  output: {
    mode: 'tags-split',
    client: 'react-query',
    override: {
      query: {
        useInvalidate: true,
        mutationInvalidates: [
          { onMutations: ['postItems'], invalidates: ['getItems'] },
        ],
      },
    },
  },
}
```

#### Generated query key function

```typescript
// The query key function requires path parameters
export const getGetItemsQueryKey = (workspaceId: string, categoryId: string) => {
  return [`/items/${workspaceId}/categories/${categoryId}`] as const;
};
```

#### Generated `onSuccess` inside the mutation

```typescript
const onSuccess = (data, variables, onMutateResult, context) => {
  if (!options?.skipInvalidation) {
    // BUG: called without required arguments
    queryClient.invalidateQueries({ queryKey: getGetItemsQueryKey() });
  }
  mutationOptions?.onSuccess?.(data, variables, onMutateResult, context);
};
```

#### Impact

- **TypeScript error**: `Expected 2 arguments, but got 0. (TS2554)`
- **Silent runtime failure**: The key resolves to `["/items/undefined/categories/undefined"]`, which never matches any cached query. No invalidation actually occurs.

#### Expected behavior

The generated code should use a **partial / prefix match** so that all cached instances are invalidated regardless of params:

```typescript
// Option A: Use a partial key for fuzzy matching
await queryClient.invalidateQueries({ queryKey: ['/items'] });

// Option B: Map mutation variables to query params
await queryClient.invalidateQueries({
  queryKey: getGetItemsQueryKey(variables.workspaceId, variables.categoryId),
});
```

---

### Bug 2: `file` option generates incorrect import paths in `tags-split` mode

When a cross-tag invalidation is configured with the `file` option, orval generates wrong relative import paths.

#### Configuration

```js
mutationInvalidates: [
  {
    onMutations: ['deleteItem'],
    invalidates: [{ query: 'getItems', file: './items' }, 'getItem'],
  },
],
```

#### File layout (`tags-split`)

In `tags-split` mode, each tag gets its own subdirectory:

```
src/api/generated/
  item/
    item.ts         ← singular tag (delete/get/put mutations)
  items/
    items.ts        ← plural tag (list/create queries)
```

#### Generated import (`item/item.ts`)

```typescript
// WRONG — resolves to item/items.ts (does not exist)
import { getGetItemsQueryKey } from './items';
```

#### Expected

```typescript
// CORRECT — should resolve across tag directories
import { getGetItemsQueryKey } from '../items/items';
```

The `file` path is emitted as a relative path from the **consuming file's directory**, but it appears to be resolved from the **output root** instead, which breaks in `tags-split` mode where each tag is in its own subdirectory.

---

### Workaround

Using `useInvalidate: true` generated helpers manually in `onSuccess`. These correctly accept required parameters and produce type-safe exact-match keys:

```typescript
// Generated by useInvalidate: true — works correctly
export const invalidateGetItems = async (
  queryClient: QueryClient,
  workspaceId: string,
  categoryId: string,
  options?: InvalidateOptions,
) => {
  await queryClient.invalidateQueries(
    { queryKey: getGetItemsQueryKey(workspaceId, categoryId) },
    options,
  );
  return queryClient;
};
```

---

### Environment

| Item | Value |
|------|-------|
| orval | 8.5.3 |
| @tanstack/react-query | 5.x |
| TypeScript | 5.x |
| mode | tags-split |
| client | react-query |

zeriong:
## Summary

Closes #3090

### Bug 1: Missing required arguments in generated invalidation code

When `mutationInvalidates` targets a query with required path parameters (e.g. `showPetById`) but no `params` mapping is specified, the generated code called the query key function without arguments:

```typescript
// Before (broken): TS error + produces ['/pets/undefined']
queryClient.invalidateQueries({ queryKey: getShowPetByIdQueryKey() });
```

Now looks up the target query's route from the OpenAPI spec and generates predicate-based broad matching using the route prefix:

```typescript
// After (fixed): matches all /pets/* queries via startsWith
queryClient.invalidateQueries({
  predicate: (query) => {
    const key = query.queryKey[0];
    return typeof key === 'string' && key.startsWith('/pets');
  },
});
```

Queries **without** required path params (e.g. `listPets`) and queries **with** explicit `params` mapping continue to use `queryKey: getXxxQueryKey(...)` as before.

### Bug 2: Incorrect import path in tags-split mode

In `tags-split` mode, the `file` option for cross-tag invalidation generated import paths relative to the output root instead of the tag subdirectory:

```typescript
// Before (broken): resolves to item/items (wrong)
import { getGetItemsQueryKey } from './items';

// After (fixed): resolves to ../items/items (correct)
import { getGetItemsQueryKey } from '../items/items';
```

## Changes

| File | Change |
|------|--------|
| `packages/query/src/mutation-generator.ts` | Add `findOperationRoute` / `getRoutePrefix` helpers; generate `predicate`-based invalidation when params are missing for path-param queries |
| `packages/core/src/writers/split-tags-mode.ts` | Adjust `file` importPath from `./tag` to `../tag/tag` in tags-split directory structure |
| `tests/configs/react-query.config.ts` | Add `showPetById` without params to `createPets` invalidation for Bug 1 coverage |
| `tests/__snapshots__/react-query/invalidates/endpoints.ts` | Updated snapshot |

## Test plan

- [x] Unit tests pass: `bun vitest run` (2178/2181, 3 pre-existing failures in `resolve-version.test.ts`)
- [x] Snapshot tests pass: `bun run test:snapshots` (67/67 tasks)
- [x] Verified generated snapshot contains `predicate` + `startsWith('/pets')` for params-less `showPetById` invalidation
- [x] Verified `deletePetById` with `params: ['petId']` still generates `queryKey: getShowPetByIdQueryKey(variables.petId)`

## Summary by CodeRabbit

* **New Features**
  * Added `skipInvalidation` option to mutation APIs for finer cache control.
  * Enhanced cache invalidation to use predicate-based matching (broader route-prefix invalidation).
  * Added Angular client snapshots with multi-content accept support, filtered query-params handling, and new model types for tests.

* **Bug Fixes**
  * Fixed relative import path generation in split-tags mode so imports point to tag subdirectories.

### DIFF
--- a/packages/core/src/getters/route-extended.test.ts
+++ b/packages/core/src/getters/route-extended.test.ts
@@ -1,0 +1,1701 @@
+import { describe, expect, test } from 'vitest';
+
+import { getRoute, getRouteAsArray } from './route';
+
+/**
+ * Extended test suite for getRouteAsArray (Issue #883 fix verification)
+ * ~1000+ test cases across 20 categories.
+ *
+ * Two helpers:
+ *  - routeToArray: OpenAPI path → getRoute → getRouteAsArray (real usage flow)
+ *  - directArray:  template string → getRouteAsArray (unit-level)
+ */
+
+// Helper: full pipeline (OpenAPI path → template → array string)
+function routeToArray(openApiPath: string): string {
+  return getRouteAsArray(getRoute(openApiPath));
+}
+
+// Helper: direct getRouteAsArray call
+function directArray(templateRoute: string): string {
+  return getRouteAsArray(templateRoute);
+}
+
+// ============================================================
+// 1. Basic static paths (no parameters) — 40 cases
+// ============================================================
+describe('1. Basic static paths', () => {
+  test.each([
+    ['/api', "'api'"],
+    ['/api/v1', "'api','v1'"],
+    ['/api/v1/users', "'api','v1','users'"],
+    ['/api/v1/users/list', "'api','v1','users','list'"],
+    ['/health', "'health'"],
+    ['/a/b/c/d/e', "'a','b','c','d','e'"],
+    ['api/v1', "'api','v1'"],
+    ['users', "'users'"],
+    ['/api/v2/orders/items', "'api','v2','orders','items'"],
+    ['/graphql', "'graphql'"],
+    ['/api/v1/auth/login', "'api','v1','auth','login'"],
+    ['/api/v1/auth/logout', "'api','v1','auth','logout'"],
+    ['/api/v1/settings/preferences', "'api','v1','settings','preferences'"],
+    ['/webhooks/incoming', "'webhooks','incoming'"],
+    ['/internal/metrics', "'internal','metrics'"],
+    ['/api/v3/search/results', "'api','v3','search','results'"],
+    ['/static/assets/images', "'static','assets','images'"],
+    ['/rpc/call', "'rpc','call'"],
+    ['/feed/atom', "'feed','atom'"],
+    ['/sitemap.xml', "'sitemap.xml'"],
+    ['/robots.txt', "'robots.txt'"],
+    ['/favicon.ico', "'favicon.ico'"],
+    ['/manifest.json', "'manifest.json'"],
+    ['/openapi.yaml', "'openapi.yaml'"],
+    ['/swagger-ui', "'swagger-ui'"],
+    ['/api-docs', "'api-docs'"],
+    ['/healthz', "'healthz'"],
+    ['/readyz', "'readyz'"],
+    ['/livez', "'livez'"],
+    ['/metrics', "'metrics'"],
+    ['/status/check', "'status','check'"],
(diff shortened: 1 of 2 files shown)


melloware:
good CodeRabbit feedback and build did not pass.

PR #3218, opened 2026-04-11
fix: disable development export condition in tsdown base config

### LINKED ISSUE #3216: fix(packaging): published packages expose "development" export condition pointing to missing src/ files (v8.6.0+)
### Description

Since v8.6.0, all published `@orval/*` packages on npm contain a `"development"` export condition in their `package.json` that points to TypeScript source files (`./src/index.ts`). These source files are **not included** in the published tarball (the `files` field only includes `dist/`).

Any bundler or runtime that activates the `"development"` condition — such as **Vite in dev mode**, or a `tsconfig.json` with `"customConditions": ["development"]` — will resolve `@orval/core` (and other `@orval/*` packages) to `./src/index.ts`, which does not exist, causing a runtime error:

```
Error: Cannot find module '.../node_modules/@orval/core/src/index.ts'
```

**Root cause:** Commit `82fde637` changed `packages/tsdown.base.ts` from `exports: true` to `exports: { devExports: 'development' }`, which made tsdown auto-generate `"development": "./src/index.ts"` in the `exports` field of every package. A `publishConfig.exports` block was added to override this during publish, but the CI uses `bun publish` ([release-publish.yaml L60](https://github.com/orval-labs/orval/blob/82b0915e/.github/workflows/release-publish.yaml#L60)), which **does not apply `publishConfig.exports`** — unlike `npm publish`.

**Proof from the published tarball** (`npm pack @orval/core@8.7.0`):
```json
"exports": {
  ".": {
    "development": "./src/index.ts",
    "default": "./dist/index.mjs"
  }
}
```

In v8.5.3 this was simply:
```json
"exports": {
  ".": "./dist/index.mjs"
}
```

**Affected versions:** 8.6.0, 8.6.1, 8.6.2, 8.7.0
**Affected packages:** All 12 published packages in the monorepo (core, orval, angular, axios, fetch, hono, mcp, mock, query, solid-start, swr, zod).

### Output client

other

```ts
// Any config triggers the bug when using a bundler that activates
// the "development" export condition (e.g., Vite dev mode).
export default defineConfig({
  api: {
    input: { target: "./openapi.yaml" },
    output: {
      client: "react-query",
      target: "src/api",
    },
  },
});
```

### Environment

```
  System:
    OS: macOS 13.5
    CPU: Apple M1
    Node: v24.14.0
  npmPackages:
    orval: 8.7.0
```

### Expected behavior

All `@orval/*` packages should resolve to compiled `dist/` files regardless of the consumer's export conditions. The published `package.json` should not contain a `"development"` condition pointing to files that are excluded from the tarball.

### Actual behavior

```
Error: Cannot find module '.../node_modules/@orval/core/src/index.ts'
imported from .../node_modules/orval/dist/config-CO914zjc.mjs
```

Node/bundler follows the `"development"` export condition → `./src/index.ts` → file not found.

### Additional context

**Proposed fix:** Revert `packages/tsdown.base.ts` to `exports: true` (removes the `"development"` condition from all packages) and remove the now-unnecessary `publishConfig.exports` blocks. Dev-mode DX is already covered by `tsdown --watch`.

saschabuehrle:
Fixes #3216

This reverts `packages/tsdown.base.ts` to `exports: true`, which stops generating a `development` export condition that points to `src/index.ts` in published packages. With this change, package exports resolve to built `dist` output in consumers even when `development` custom conditions are active.

Greetings, saschabuehrle

## Summary by CodeRabbit

* **Chore**
  * Updated base configuration export settings to use default export mechanism instead of development-specific mapping.

### DIFF
--- a/packages/tsdown.base.ts
+++ b/packages/tsdown.base.ts
@@ -9,5 +9,5 @@ export const baseConfig: UserConfig = {
   dts: {
     sourcemap: true,
   },
-  exports: { devExports: 'development' },
+  exports: true,
 };


melloware:
Assigned to @snebjorn and @anymaniax to review.

snebjorn:
The `development` export haven't yet been removed in this PR. But removing it would degrade DX.
The real issue is the publishConfig isn't applied

pedrotainha:
@snebjorn Two possible fixes that keep the `development` condition for local DX:

**Option 1 — Switch to `npm publish`** (one-line change in `release-publish.yaml` L60):

```diff
-              if ! (cd "$dir" && bun publish --access public); then
+              if ! (cd "$dir" && npm publish --access public); then
```

`npm publish` already applies `publishConfig` fields correctly.

**Option 2 — Pre-publish step that applies `publishConfig.exports`** (package-manager agnostic):

Add a step before the publish loop that patches each `package.json` in-place:

```bash
for pkg in packages/*/package.json; do
  if jq -e '.publishConfig.exports' "$pkg" > /dev/null 2>&1; then
    jq '.exports = .publishConfig.exports' "$pkg" > tmp.$$.json && mv tmp.$$.json "$pkg"
  fi
done
```

This way `bun publish` stays, and the `exports` field is already correct before packing. The `publishConfig` blocks become a declarative source of truth for what gets published.

melloware:
i like option 2

pedrotainha:
@snebjorn @melloware any update? I would like to update the Orval package in the project, and I'm depending on this. Thanks

melloware:
I don't think the submitted fix is correct in its current state?

pedrotainha:
> I don't think the submitted fix is correct in its current state?

Yeah, I understood, so the question is:
- Can I open a new pr with the proposed fix I suggested or will you refactor this one to the suggestion?

Thanks

melloware:
absolutely submit the correct PR!

melloware:
@pedrotainha  @saschabuehrle  what is the latest with this?

melloware:
Closing this as not the right fix

pedrotainha:
> @pedrotainha @saschabuehrle what is the latest with this?

Sorry, I have done nothing, no spare time in the last few weeks

PR #3233, opened 2026-04-15
feat: 2163 mcp handlers accept options

### LINKED ISSUE #2163: MCP Handlers Accept Options
Hey yall!

I've implemented the below using patch-package because i needed it right away, but I think this should be the default behavior based off the spec.  Tools are passed a second parameter [`extra`](https://github.com/modelcontextprotocol/typescript-sdk/blob/6b3d326cf77cf212ea97d38572cf6e20ae047543/src/server/mcp.ts#L998) which includes `authInfo` (among other things) because its likely a user will need to pass a token or something else to a possible fetch call the tool may make, so the use case is pretty common.

At the moment the handlers that are generated don't accept an `options` parameter so you cannot pass headers etc to the eventual fetch call that the handler makes using the http-client. All the http-client functions accept `options` as the last argument, but we dont have a way to pass them through the generated handlers.

The contributor guide suggest opening an issue to get approval before doing a PR, so was looking to get agreement before I added my patch formally in a PR. In a nutshell adding `options` as an optional additional argument that handers accept and pass to the http-client.

soartec-lab:
## Status

**READY**

## Description

Allows MCP handlers to accept optional `options` argument that will be passed the the http-client

Fix https://github.com/orval-labs/orval/issues/2163

i didnt see any existing tests to add to for the mcp package-- if i missed em please point em out happy to add to them.


soartec-lab:
I have changed the response policy to #3235, so I am closing this thread.

PR #3239, opened 2026-04-16
fix(core): treat OpenAPI validation failures as warnings instead of errors

### LINKED ISSUE #3072: Core: Validation throws error when using fastapi.sse
### Description

I am trying to use the new SSE function in the latest version of FastAPI, the fastapi.sse will attach `itemSchema` field in the generated openapi endpoint definition like this:
```
{
  "openapi": "3.1.0",
  "info": {
    "title": "FastAPI",
    "version": "0.1.0"
  },
  "paths": {
    "/api/events/": {
      "post": {
        "tags": [
          "stream"
        ],
        "summary": "Sse Endpoint",
        "operationId": "sse_endpoint",
        "responses": {
          "200": {
            "description": "Successful Response",
            "content": {
              "text/event-stream": {
                "itemSchema": {
                  "type": "object",
                  "properties": {
                    "data": {
                      "type": "string"
                    },
                    "event": {
                      "type": "string"
                    },
                    "id": {
                      "type": "string"
                    },
                    "retry": {
                      "type": "integer",
                      "minimum": 0
                    }
                  }
                }
              }
            }
          },
          "400": {
            "description": "Bad Request",
            "content": {
              "text/event-stream": {
                "schema": {
                  "$ref": "#/components/schemas/ErrorResponseSchema"
                }
              }
            }
          },
          "401": {
...
```

And when the `itemSchema` appear, the orval will throw an error like this:
```
api - Validation failed
Cause: [
{
  "message": "if must match \"else\" schema",
  "path": "/paths/~1api~1events~1"
},                                                    
...                                                     
```

And I have to manually exclude the sse endpoints in the fastapi app to make orval normally generate other schemas. I have already excluded sse endpoints in the orval config.

```ts
import { defineConfig, type InputOptions, type OperationOptions } from "orval";

const operationSpecificConfig: Record<string, OperationOptions> = {
  fetch_models: {
    query: {
      useInfinite: false,
      useSuspenseInfiniteQuery: false,
    },
  },
};

const input: InputOptions = {
  target: "../openapi.json",
  filters: {
    mode: "exclude",
    tags: ["stream"],
  },
};

export default defineConfig({
  api: {
    input,
    output: {
      clean: true,
      target: "./src/api/generated/endpoints/",
      schemas: "./src/api/generated/schemas/",
      operationSchemas: "./src/api/generated/schemas/params/",
      client: "react-query",
      mode: "tags-split",
      namingConvention: "kebab-case",
      override: {
        useTypeOverInterfaces: true,
        fetch: {
          includeHttpResponseReturnType: false,
        },
        mutator: {
          path: "./src/api/orval-mutator/custom-fetch.ts",
          name: "fetchApi",
        },
        query: {
          version: 5,
          useQuery: true,
          useInfinite: true,
          useMutation: true,
          useSuspenseQuery: true,
          useSuspenseInfiniteQuery: true,
          useInfiniteQueryParam: "page",
        },

        operations: operationSpecificConfig,
      },
    },
    hooks: {
      afterAllFilesWrite: ["biome format --write --config-path config/biome.format.jsonc"],
    },
  },
});
```

### Environment

  System:
    OS: Windows 11 10.0.26200
    CPU: (16) x64 AMD Ryzen 7 6800HS Creator Edition
    Memory: 2.15 GB / 13.69 GB

### Expected behavior

Orval just ignore the sse endpoints and normally resolve others.

### Actual behavior

Orval immediately throws error and exit.

```json
{
  "openapi": "3.1.0",
  "info": {
    "title": "FastAPI",
    "version": "0.1.0"
  },
  "paths": {
    "/api/events/": {
      "post": {
        "tags": [
          "stream"
        ],
        "summary": "Sse Endpoint",
        "operationId": "sse_endpoint",
        "responses": {
          "200": {
            "description": "Successful Response",
            "content": {
              "text/event-stream": {
                "itemSchema": {
                  "type": "object",
                  "properties": {
                    "data": {
                      "type": "string"
                    },
                    "event": {
                      "type": "string"
                    },
                    "id": {
                      "type": "string"
                    },
                    "retry": {
                      "type": "integer",
                      "minimum": 0
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

### Additional context

_No response_

zeriong:
## Summary
- Change OpenAPI spec validation from hard error to warning, allowing code generation to proceed with non-standard fields (#3072)
- Add `input.validation` option (`boolean`, default `true`) to let users disable validation entirely
- Apply to both main spec validation (`import-specs.ts`) and transformer validation (`import-open-api.ts`)

## Problem
FastAPI SSE endpoints add a non-standard `itemSchema` field to response content. The `@scalar/openapi-parser` validator rejects this field, causing orval to throw and exit immediately. This blocks generation of all endpoints, even when the SSE endpoints are excluded via `input.filters`.

## Solution
Replace `throw new Error('Validation failed')` with `logWarning()` in both vlidation sites. When validation fails, orval now:
1. Prints a yellow warning with the validation errors
2. Suggests `input.validation: false` as an escape hatch
3. Continues with code generation

Users who want to completely skip validation can set:
```ts
input: {
  target: './openapi.json',
  validation: false,
}


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added an optional input setting to enable/disable OpenAPI validation (enabled by default).
  * Import flow now emits warnings with validation details for invalid specs instead of aborting the import.

* **Tests**
  * Added tests covering import behavior with validation enabled and disabled.

* **Chores**
  * Added an Angular Query generate script and introduced the Angular Query experimental dependency in samples/tests.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -942,7 +942,7 @@
       "@angular/forms": "21.2.2",
       "@angular/platform-browser": "21.2.2",
       "@angular/router": "21.2.2",
-      "@tanstack/angular-query-experimental": "5.90.16",
+      "@tanstack/angular-query-experimental": "5.90.25",
       "rxjs": "7.8.2",
     },
     "axios": {
@@ -1760,15 +1760,15 @@
 
     "@swc/helpers": ["@swc/helpers@0.5.15", "", { "dependencies": { "tslib": "^2.8.0" } }, "sha512-JQ5TuMi45Owi4/BIMAJBoSQoOJu12oOk/gADqlcUL9JEdHB8vyjUSsxqeNXnmXHjYKMi2WcYtezGEEhqUI/E2g=="],
 
-    "@tanstack/angular-query-experimental": ["@tanstack/angular-query-experimental@5.90.16", "", { "dependencies": { "@tanstack/query-core": "5.90.12" }, "optionalDependencies": { "@tanstack/query-devtools": "5.91.1" }, "peerDependencies": { "@angular/common": ">=16.0.0", "@angular/core": ">=16.0.0" } }, "sha512-ezXyxuaSA6kpwUxwrxo5Tf3B7KL7mb7cxhLnun/RMlw6h3ZQJzl1cJgrvN97+C0AeHoucmK7RwhEoIY12gY7WA=="],
+    "@tanstack/angular-query-experimental": ["@tanstack/angular-query-experimental@5.90.25", "", { "dependencies": { "@tanstack/query-core": "5.90.20" }, "optionalDependencies": { "@tanstack/query-devtools": "5.93.0" }, "peerDependencies": { "@angular/common": ">=16.0.0", "@angular/core": ">=16.0.0" } }, "sha512-8MmtvEq83kRbwtt6NNbDAWPMOFHLxSmvhScFtdmBP/Xo5gSSzAf4JHcblQDHV/CgWIGKOp7eOa6aXDiVIK+kbw=="],
 
     "@tanstack/directive-functions-plugin": ["@tanstack/directive-functions-plugin@1.121.21", "", { "dependencies": { "@babel/code-frame": "7.26.2", "@babel/core": "^7.26.8", "@babel/traverse": "^7.26.8", "@babel/types": "^7.26.8", "@tanstack/router-utils": "^1.121.21", "babel-dead-code-elimination": "^1.0.10", "tiny-invariant": "^1.3.3" }, "peerDependencies": { "vite": ">=6.0.0" } }, "sha512-B9z/HbF7gJBaRHieyX7f2uQ4LpLLAVAEutBZipH6w+CYD6RHRJvSVPzECGHF7icFhNWTiJQL2QR6K07s59yzEw=="],
 
     "@tanstack/match-sorter-utils": ["@tanstack/match-sorter-utils@8.19.4", "", { "dependencies": { "remove-accents": "0.5.0" } }, "sha512-Wo1iKt2b9OT7d+YGhvEPD3DXvPv2etTusIMhMUoG7fbhmxcXCtIjJDEygy91Y2JFlwGyjqiBPRozme7UD8hoqg=="],
 
-    "@tanstack/query-core": ["@tanstack/query-core@5.90.12", "", {}, "sha512-T1/8t5DhV/SisWjDnaiU2drl6ySvsHj1bHBCWNXd+/T+Hh1cf6JodyEYMd5sgwm+b/mETT4EV3H+zCVczCU5hg=="],
+    "@tanstack/query-core": ["@tanstack/query-core@5.90.20", "", {}, "sha512-OMD2HLpNouXEfZJWcKeVKUgQ5n+n3A2JFmBaScpNDUqSrQSjiveC7dKMe53uJUg1nDG16ttFPz2xfilz6i2uVg=="],
 
-    "@tanstack/query-devtools": ["@tanstack/query-devtools@5.91.1", "", {}, "sha512-l8bxjk6BMsCaVQH6NzQEE/bEgFy1hAs5qbgXl0xhzezlaQbPk6Mgz9BqEg2vTLPOHD8N4k+w/gdgCbEzecGyNg=="],
+    "@tanstack/query-devtools": ["@tanstack/query-devtools@5.93.0", "", {}, "sha512-+kpsx1NQnOFTZsw6HAFCW3HkKg0+2cepGtAWXjiiSOJJ1CtQpt72EE2nyZb+AjAbLRPoeRmPJ8MtQd8r8gsPdg=="],
 
     "@tanstack/react-query": ["@tanstack/react-query@5.62.16", "", { "dependencies": { "@tanstack/query-core": "5.62.16" }, "peerDependencies": { "react": "^18 || ^19" } }, "sha512-XJIZNj65d2IdvU8VBESmrPakfIm6FSdHDzrS1dPrAwmq3ZX+9riMh/ZfbNQHAWnhrgmq7KoXpgZSRyXnqMYT9A=="],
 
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: packages/core/src/types.ts [hunk 1]
```diff
@@ -206,6 +206,7 @@ export interface NormalizedOperationOptions {
 export interface NormalizedInputOptions {
   target: string | OpenApiDocument;
   override: OverrideInput;
+  validation?: boolean;
```
melloware: I think we need docs for this setting?  @snebjorn any disagreement with this i think you removed validation previously?

melloware (changes requested):

packages/core/src/types.ts
zeriong: Good point about the docs — I'll add documentation for the `input.validation` setting.

I also noticed that `validateComponentKeys()` was running unconditionally even when `validation: false`. I've moved it inside the `if (validation)` block so the escape hatch works completely.

Additionally:

- Made `NormalizedInputOptions.validation` non-optional (since `normalizeOptions()` always assigns it)
- Removed the redundant `?? true` fallback in `importSpecs`

Regarding previous validation removal — this PR doesn't remove validation. It just downgrades failures from hard errors to warnings, with an opt-out for specs that have non-standard extensions like FastAPI's `itemSchema`. Default behavior still validates.

snebjorn: The reason for removing the option to disable validation was basically that we can't reasonably parse the OAS if it's invalid.
Sure it works sometimes but that's, lets call it, an undocumented feature. And it might break in minor updates.

If the option is called `unsafeDisableValidationYolo` and prints a VERY noticeable warning that says that no bug reports are accepted with this flag on. Then I don't see a problem :)

zeriong: Thanks for the feedback @snebjorn! That makes total sense.

I'll update the PR with the following changes:

1. **Rename** `input.validation` → `input.unsafeDisableValidationYolo` (inverted semantics — `true` to disable)
2. **Restore hard error as default** — when validation fails and the flag is not set, orval will throw as before
3. **Add a very noticeable warning** when the flag is enabled:
   - Banner-style warning at startup
   - Clear message that no bug reports are accepted with this flag on
4. When `unsafeDisableValidationYolo: true`, skip all validation (both `validateSpec` and `validateComponentKeys`) and continue generation

This keeps the safe default while giving users with non-standard specs (like FastAPI's `itemSchema`) an explicit escape hatch they can't accidentally stumble into.

Will push the update shortly!

melloware: I think @snebjorn  was kidding about the name?

zeriong:
@snebjorn Review is welcome!

packages/core/src/types.ts
snebjorn: yes, it was a joking example. I've seen options like this in the wild but can't recall any right now.

IMO the option name should indicate that it's "turn on at your own risk" and it might break at any point

zeriong: Yeah! I actually understood it! 😁

packages/core/src/types.ts
```diff
@@ -326,6 +327,8 @@ export interface InputFiltersOptions {
 export interface InputOptions {
   target: string | string[] | Record<string, unknown> | OpenApiDocument;
   override?: OverrideInput;
+  /** Enable or disable OpenAPI spec validation. Default: true */
+  validation?: boolean;
```
snebjorn, line 331: IMO the option name should indicate that it's "turn on at your own risk" and it might break at any point.

Best I can come up with is `unsafeDisableValidation`, but feel free to play with this.

packages/orval/src/import-specs.ts
```diff
@@ -41,11 +43,24 @@ async function resolveSpec(
...
+  if (validation) {
+    validateComponentKeys(dereferencedData);
+
+    const { valid, errors } = await validateSpec(dereferencedData);
+    if (!valid) {
+      throw new Error(
+        `OpenAPI spec validation failed:\n${JSON.stringify(errors, undefined, 2)}\n\n` +
+          `If your spec uses non-standard extensions and you want to skip validation,\n` +
+          `set input.validation to false in your orval config.\n` +
+          `Note: no bug reports are accepted with validation disabled.`,
```
snebjorn, line 55: I don't think we should advertise this.

bun.lock
```diff
@@ -1758,15 +1760,15 @@
 
     "@swc/helpers": ["@swc/helpers@0.5.15", "", { "dependencies": { "tslib": "^2.8.0" } }, "sha512-JQ5TuMi45Owi4/BIMAJBoSQoOJu12oOk/gADqlcUL9JEdHB8vyjUSsxqeNXnmXHjYKMi2WcYtezGEEhqUI/E2g=="],
 
-    "@tanstack/angular-query-experimental": ["@tanstack/angular-query-experimental@5.90.25", "", { "dependencies": { "@tanstack/query-core": "5.90.20" }, "optionalDependencies": { "@tanstack/query-devtools": "5.93.0" }, "peerDependencies": { "@angular/common": ">=16.0.0", "@angular/core": ">=16.0.0" } }, "sha512-8MmtvEq83kRbwtt6NNbDAWPMOFHLxSmvhScFtdmBP/Xo5gSSzAf4JHcblQDHV/CgWIGKOp7eOa6aXDiVIK+kbw=="],
+    "@tanstack/angular-query-experimental": ["@tanstack/angular-query-experimental@5.90.16", "", { "dependencies": { "@tanstack/query-core": "5.90.12" }, "optionalDependencies": { "@tanstack/query-devtools": "5.91.1" }, "peerDependencies": { "@angular/common": ">=16.0.0", "@angular/core": ">=16.0.0" } }, "sha512-ezXyxuaSA6kpwUxwrxo5Tf3B7KL7mb7cxhLnun/RMlw6h3ZQJzl1cJgrvN97+C0AeHoucmK7RwhEoIY12gY7WA=="],
```
snebjorn, line 1763: seems some packages were downgraded, why?

snebjorn (changes requested):
The docs also need to be updated

packages/core/src/types.ts
zeriong: Agree. Renaming to unsafeDisableValidation and updating docs!

packages/orval/src/import-specs.ts
zeriong: Ok, Thanks for reiview!

bun.lock
zeriong: Good catch.
I pinned 5.90.16 in the catalog to match the sample's existing ^5.90.16, but the lockfile had already resolved to 5.90.25 before.

The pin caused the unintended downgrade.
I'll bump the catalog to the latest compatible version!

melloware (dismissed):

snebjorn (approved):

PR #3245, opened 2026-04-17
fix(core): support oneOf/anyOf in multipart/form-data code generation

### LINKED ISSUE #3242: Broken TypeScript generated for multipart/form-data request body with oneOf/anyOf
## Description

When an OpenAPI 3.1 spec uses `oneOf` (or `anyOf`) inside a `multipart/form-data` request body, Orval generates broken TypeScript code that fails type-checking. There are two distinct, independent bugs in `packages/core/src/getters/res-req-types.ts`:

- **Bug 1**: Variable name mismatch in `getResReqTypes`. The FormData code references a variable that doesn't exist. TypeScript compilation fails. Triggers as soon as a `oneOf` is present, even with a single schema inside.
- **Bug 2**: Duplicated `formData.append()` calls in `getSchemaFormDataAndUrlEncoded`. All `oneOf` branches run unconditionally, doubling shared fields at runtime. Triggers when the `oneOf` contains 2 or more schemas.

This is related to [#2747](https://github.com/orval-labs/orval/issues/2747) (fixed in 8.0.1), but that fix only addressed `multipart/form-data` **without** `oneOf`/`anyOf`. The `oneOf` code path was not updated and is still broken as of 8.8.0.

JSON `application/json` endpoints with `oneOf` are **not** affected because they use `JSON.stringify()` and skip the FormData generation entirely.

**Note:** `oneOf` with a single entry is a common pattern emitted by NestJS decorators (`@nestjs/swagger`, `zod-to-openapi`) for versioned request schemas. The spec is valid per OpenAPI 3.1.

### OpenAPI spec

```yaml
openapi: 3.1.0
info:
  title: Test API
  version: 1.0.0
paths:
  /upload:
    post:
      operationId: FileController_upload
      requestBody:
        required: true
        content:
          multipart/form-data:
            schema:
              oneOf:
                - $ref: "#/components/schemas/UploadFileDto"
      responses:
        "201":
          description: Created
  /upload-versioned:
    post:
      operationId: FileController_uploadVersioned
      requestBody:
        required: true
        content:
          multipart/form-data:
            schema:
              oneOf:
                - $ref: "#/components/schemas/UploadFileDtoV1"
                - $ref: "#/components/schemas/UploadFileDtoV2"
      responses:
        "201":
          description: Created
components:
  schemas:
    UploadFileDto:
      type: object
      properties:
        file:
          type: string
          format: binary
        label:
          type: string
        category:
          type: string
      required: [file, label, category]
    UploadFileDtoV1:
      type: object
      properties:
        file:
          type: string
          format: binary
        label:
          type: string
        category:
          type: string
      required: [file, label, category]
    UploadFileDtoV2:
      type: object
      properties:
        file:
          type: string
          format: binary
        label:
          type: string
        category:
          type: string
        tags:
          type: string
      required: [file, label, category, tags]
```

### Orval config

```ts

export default defineConfig({
  api: {
    output: {
      target: './generated.ts',
      client: 'fetch',
    },
    input: {
      target: './openapi.yaml',
    },
  },
});
```

### Steps to reproduce

1. Save the spec and config above
2. Run `npx orval --config orval.config.ts`
3. Run `npx tsc --noEmit --strict --lib es2020,dom generated.ts`
4. Observe compilation errors on both endpoints

### Root cause

Within `getResReqTypes`, there is a mismatch in how the variable name is resolved depending on whether the schema is a direct `$ref` or a `oneOf` wrapper:

- The **type resolution** path (via `getResReqContentTypes` / `resolveObject`) correctly walks through the `oneOf` and resolves the `$ref` inside. This produces the correct function parameter name (e.g. `uploadFileDto`).
- The **`effectivePropName`** path (used for the FormData variable) only checks `isReference(mediaType.schema)`. When the schema is `{ oneOf: [{ $ref: "..." }] }`, this check returns `false` because the top-level object is not a `$ref`. So `effectivePropName` falls back to the default name derived from the operation (e.g. `FileControllerUploadBody`).

Result: the function parameter is named `uploadFileDto` but the FormData code references `fileControllerUploadBody`, which was never declared.

### TypeScript compiler error

```
generated.ts(48,48): error TS2552: Cannot find name 'fileControllerUploadBody'. Did you mean 'fileControllerUpload'?
generated.ts(93,59): error TS2552: Cannot find name 'fileControllerUploadVersionedBody'. Did you mean 'fileControllerUploadVersioned'?
generated.ts(98,59): error TS2552: Cannot find name 'fileControllerUploadVersionedBody'. Did you mean 'fileControllerUploadVersioned'?
```

### Expected generated code

```typescript
export const fileControllerUpload = async (
  uploadFileDto: UploadFileDto
) => {
  formData.append('file', uploadFileDto.file);
  //                       ^^^^^^^^^^^^^ matches the parameter name
```

### Actual generated code

```typescript
export const fileControllerUpload = async (
  uploadFileDto: UploadFileDto
) => {
  formData.append('file', fileControllerUploadBody.file);
  //                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^ never declared
```

## Bug 2: Duplicated `formData.append()` calls (when `oneOf` contains 2+ schemas)

This is a separate bug from Bug 1. It manifests when there are 2+ schemas in the `oneOf`:

```yaml
oneOf:
  - $ref: "#/components/schemas/UploadFileDtoV1"
  - $ref: "#/components/schemas/UploadFileDtoV2"
```

### Root cause

In `getSchemaFormDataAndUrlEncoded`, when the schema has `combinedSchemas` (oneOf/anyOf), the code maps over each sub-schema and generates a `formData.append()` block for every field, then joins all blocks with `\n`. All blocks run unconditionally. There is no `if/else` or discriminator logic between them.

The TypeScript `as` casts (e.g. `param as UploadFileDtoV1`) are compile-time only and disappear at runtime, so both variables reference the same object. Every shared field gets appended to the FormData twice.

### Expected behavior

Each field should only be appended once at runtime, regardless of which `oneOf` variant is passed.

### Actual generated code (with 2 schemas in oneOf)

```typescript

// Block for V1 (always runs)
const uploadFileDtoV1 = (param as UploadFileDtoV1);
formData.append('file', uploadFileDtoV1.file);
formData.append('label', uploadFileDtoV1.label);
formData.append('category', uploadFileDtoV1.category);

// Block for V2 (also always runs, no condition!)
const uploadFileDtoV2 = (param as UploadFileDtoV2);
formData.append('file', uploadFileDtoV2.file);          // DUPLICATE
formData.append('label', uploadFileDtoV2.label);        // DUPLICATE
formData.append('category', uploadFileDtoV2.category);  // DUPLICATE
formData.append('tags', uploadFileDtoV2.tags);
```

## Workaround

We're using a `pnpm patchedDependencies` patch on `@orval/core@8.4.2` that fixes both bugs.

## Suggested fix direction

Two changes needed in `packages/core/src/getters/res-req-types.ts`:

### Bug 1: resolve `effectivePropName` through `oneOf`/`anyOf`

In `getResReqTypes`, after the existing `if (isReference(mediaType.schema))` block, add a branch that unwraps the `oneOf`/`anyOf`:

```typescript
// Existing code handles direct $ref:
if (isReference(mediaType.schema)) {
  // ... resolves effectivePropName correctly
}
// Add: resolve through oneOf/anyOf wrapper
else if (mediaType.schema) {
  const combinedRef = mediaType.schema.oneOf ?? mediaType.schema.anyOf;
  if (combinedRef) {
    const names = combinedRef
      .filter((ref) => isReference(ref))
      .map((ref) => resolveRef(ref, context).imports[0]?.name)
      .filter(Boolean);
    if (names.length) {
      effectivePropName = names.join('');
    }
  }
}
```

This resolves the `$ref`(s) inside the `oneOf` and uses their names. With a single schema, it gives the correct DTO name. With multiple schemas, it concatenates the names (same behavior as the existing JSON code path).

### Bug 2: generate a single FormData block instead of one per `oneOf` variant

In `getSchemaFormDataAndUrlEncoded`, for `oneOf`/`anyOf` (`shouldCast === true`), replace the `.map().join('\n')` pattern (which generates all variant blocks unconditionally) with a single `Object.entries()` loop that iterates over the object's actual properties at runtime:

```typescript
if (shouldCast) {
  form += `Object.entries(${propName}).forEach(([key, value]) => {\n`;
  form += `  if (value !== undefined) {\n`;
  form += `    if (value instanceof File || value instanceof Blob || (typeof Buffer !== 'undefined' && Buffer.isBuffer(value))) {\n`;
  form += `      ${variableName}.append(key, value);\n`;
  form += `    } else if (Array.isArray(value)) {\n`;
  form += `      value.forEach(v => ${variableName}.append(key, typeof v === 'object' ? JSON.stringify(v) : String(v)));\n`;
  form += `    } else if (typeof value === 'object' && value !== null) {\n`;
  form += `      ${variableName}.append(key, JSON.stringify(value));\n`;
  form += `    } else {\n`;
  form += `      ${variableName}.append(key, String(value));\n`;
  form += `    }\n`;
  form += `  }\n`;
  form += `});\n`;

  // Preserve imports from all sub-schemas
  combinedSchemas.forEach((subSchema) => {
    const { imports } = resolveRef(subSchema, context);
    if (imports[0]) additionalImports.push(imports[0]);
  });
} else {
  // allOf case: preserve original behavior (unchanged)
}
```

The type checks (File/Blob/Buffer, Array, Object, String) match the cases handled by `resolveSchemaPropertiesToFormData`, the function Orval already uses for multipart endpoints **without** `oneOf`.

**Known limitation**: the `Object.entries()` loop iterates dynamically at runtime, so it doesn't know each field's schema. Orval's existing static path (`resolveSchemaPropertiesToFormData`) generates one `formData.append()` per field and can skip `readOnly` fields, apply per-field `encoding` overrides, or serialize arrays differently. Our dynamic loop can't do that because it doesn't have that per-field information at runtime.

This works for our use case (no `readOnly`, no `encoding` overrides), but if you'd prefer a fix that preserves those features, we're open to a different approach.

## Environment

- **Orval version**: tested on 8.4.2 and reproduced on 8.8.0 (same broken output, same TS compilation errors)
- **Client**: `fetch` (the bug is in `getResReqTypes` / `getSchemaFormDataAndUrlEncoded` which are shared across clients, so `axios`, `react-query`, etc. are likely affected too, but we only tested `fetch`)
- **OpenAPI version**: 3.1.0

## Related issues

- [#2747](https://github.com/orval-labs/orval/issues/2747): Fixed `multipart/form-data` handling in 8.0.1, but only for schemas **without** `oneOf`/`anyOf`. The `oneOf` code path was not updated.

---

Happy to open a PR if this direction works for you. We have a working patch with this approach and can adapt it to the TS source with tests.

@melloware

nathansogut-sipios:
Fixes #3242

Two independent bugs in `getResReqTypes` and `getSchemaFormDataAndUrlEncoded` when a `multipart/form-data` request body uses `oneOf` or `anyOf`. Full repro, compiler errors, and expected output in the issue.

### Solution

- Bug 1: extend the ref-resolution branch in `getResReqTypes` with an `else if` that walks `oneOf`/`anyOf`, resolves the `$ref`(s) inside, and concatenates the resolved names.
- Bug 2: for `oneOf`/`anyOf` (`shouldCast === true`), replace the N per-variant `formData.append` branches with a single `Object.entries()` loop. `allOf` is unchanged.

### Known limitations (Bug 2)

The runtime approach drops some static-codegen features from the per-variant branches: `readOnly` filtering, `encoding` map, array handling modes, `getFormDataFieldFileType` binary/text detection, nullable handling. Only complex specs are affected.

### Testing

3 new tests in `res-req-types.test.ts` covering: oneOf with 1 `$ref` (Bug 1), oneOf with 2 `$refs` sharing a field (Bug 2), allOf regression. `bun run test` 1640 passing, lint + typecheck green. Tests were also run against the pre-fix source to confirm they fail as expected.

Same fix is also deployed as a `pnpm patch-package` on 8 TypeScript projects (NestJS + `zod-to-openapi`, 1-2 schema versions) in production - all compile and run correctly.

## Summary by CodeRabbit

* **Tests**
  * Added regression tests validating FormData generation for root-level compositions (oneOf/allOf/anyOf), optional request bodies, and arrays of binary files; updated snapshots to reflect the new serialization behavior.

* **Bug Fixes**
  * Use referenced DTO names when generating FormData entries.
  * Emit robust runtime iteration for mixed-composition request bodies (avoiding per-variant casts) while preserving per-field behavior for allOf.
  * Add guards for optional bodies and correct handling of binary arrays.

### DIFF
--- a/packages/core/src/getters/res-req-types.test.ts
+++ b/packages/core/src/getters/res-req-types.test.ts
@@ -725,6 +725,79 @@ bodyRequestBody.photos.forEach(value => formData.append(\`photos\`, value));
       );
     });
 
+    it('oneOf alongside direct properties: loop skips direct keys to avoid duplicate appends', () => {
+      const ctx: ContextSpec = {
+        ...context,
+        spec: {
+          components: {
+            schemas: {
+              VariantA: {
+                type: 'object',
+                properties: {
+                  kind: { type: 'string', enum: ['a'] },
+                  extraA: { type: 'string' },
+                },
+                required: ['kind'],
+              },
+              VariantB: {
+                type: 'object',
+                properties: {
+                  kind: { type: 'string', enum: ['b'] },
+                  extraB: { type: 'string' },
+                },
+                required: ['kind'],
+              },
+            },
+          },
+        },
+      };
+
+      const reqBody: [string, OpenApiRequestBodyObject][] = [
+        [
+          'requestBody',
+          {
+            content: {
+              'multipart/form-data': {
+                schema: {
+                  type: 'object',
+                  oneOf: [
+                    { $ref: '#/components/schemas/VariantA' },
+                    { $ref: '#/components/schemas/VariantB' },
+                  ],
+                  properties: {
+                    name: { type: 'string' },
+                    owner: { type: 'string' },
+                  },
+                  required: ['name'],
+                },
+              },
+            },
+            required: true,
+          },
+        ],
+      ];
+
+      const result = getResReqTypes(reqBody, 'Upload', ctx)[0];
+      const formData = result.formData;
+      if (!formData || !isString(formData)) {
+        throw new Error('Expected formData to be a defined string');
+      }
+
+      // The Object.entries loop skips keys that the direct-properties branch
+      // already appends (`name`, `owner`), so those fields are not appended
+      // twice at runtime.
+      expect(formData).toContain('Object.entries(');
+      expect(formData).toContain('["name", "owner"].includes(key)');
+      const nameAppendCount = (
+        formData.match(/formData\.append\(`name`/g) ?? []
+      ).length;
+      const ownerAppendCount = (
+        formData.match(/formData\.append\(`owner`/g) ?? []
+      ).length;
+      expect(nameAppendCount).toBe(1);
+      expect(ownerAppendCount).toBe(1);
+    });
+
     it('oneOf with an optional body: FormData guards against undefined', () => {
       const ctx: ContextSpec = {
         ...context,
--- a/packages/core/src/getters/res-req-types.test.ts
+++ b/packages/core/src/getters/res-req-types.test.ts
@@ -722,6 +722,112 @@ bodyRequestBody.photos.forEach(value => formData.append(\`photos\`, value));
         'formData.append(`logo`, uploadRequestBody.logo)',
       );
     });
+
+    it('oneOf with an optional body: FormData guards against undefined', () => {
+      const ctx: ContextSpec = {
+        ...context,
+        spec: {
+          components: {
+            schemas: {
+              OptionalBodyDto: {
+                type: 'object',
+                properties: {
+                  file: { type: 'string', format: 'binary' },
+                },
+              },
+            },
+          },
+        },
+      };
+
+      const reqBody: [string, OpenApiRequestBodyObject][] = [
+        [
+          'requestBody',
+          {
+            content: {
+              'multipart/form-data': {
+                schema: {
+                  oneOf: [{ $ref: '#/components/schemas/OptionalBodyDto' }],
+                },
+              },
+            },
+            required: false,
+          },
+        ],
+      ];
+
+      const result = getResReqTypes(reqBody, 'Upload', ctx)[0];
+      const formData = result.formData;
+      if (!formData || !isString(formData)) {
+        throw new Error('Expected formData to be a defined string');
+      }
+
+      // The runtime loop must defend against an undefined body so callers can
+      // safely pass nothing when the body is optional.
+      expect(formData).toMatch(/Object\.entries\([^)]*\?\?\s*\{\}\)/);
+    });
+
+    it('oneOf with an array of binary files: FormData appends Blob items directly', () => {
+      const ctx: ContextSpec = {
+        ...context,
+        spec: {
+          components: {
+            schemas: {
+              MultiUploadV1: {
+                type: 'object',
+                properties: {
+                  files: {
+                    type: 'array',
+                    items: { type: 'string', format: 'binary' },
+                  },
+                },
+              },
+              MultiUploadV2: {
+                type: 'object',
+                properties: {
+                  files: {
+                    type: 'array',
+                    items: { type: 'string', format: 'binary' },
+                  },
+                  tag: { type: 'string' },
+                },
+              },
+            },
+          },
+        },
+      };
+
+      const reqBody: [string, OpenApiRequestBodyObject][] = [
+        [
+          'requestBody',
+          {
+            content: {
+              'multipart/form-data': {
+                schema: {
+                  oneOf: [
+                    { $ref: '#/components/schemas/MultiUploadV1' },
+                    { $ref: '#/components/schemas/MultiUploadV2' },
+                  ],
+                },
+              },
+            },
+            required: true,
+          },
+        ],
+      ];
+
+      const result = getResReqTypes(reqBody, 'Upload', ctx)[0];
+      const formData = result.formData;
+      if (!formData || !isString(formData)) {
+        throw new Error('Expected formData to be a defined string');
+      }
+
+      // Array elements that are files must be appended directly. Without the
+      // per-element Blob/File/Buffer check, each binary item would have been
+      // serialized to "{}" via JSON.stringify.
+      expect(formData).toContain('value.forEach');
+      expect(formData).toContain('v instanceof Blob');
+    });
   });
 });
 
--- a/packages/core/src/getters/res-req-types.test.ts
+++ b/packages/core/src/getters/res-req-types.test.ts
@@ -657,9 +657,11 @@ bodyRequestBody.photos.forEach(value => formData.append(\`photos\`, value));
       // Shared fields used to be appended once per variant because TypeScript
       // casts are erased at runtime. The runtime loop appends each key once.
       expect(formData).toContain('Object.entries(');
-      const appendFileCount = (formData.match(/`file`/g) ?? []).length;
-      expect(appendFileCount).toBe(0);
       expect(formData).not.toMatch(/as UploadDtoV[12]/);
+      // Guard against a future refactor reintroducing the per-variant
+      // appends keyed off the shared field.
+      expect(formData).not.toContain('uploadDtoV1.file');
+      expect(formData).not.toContain('uploadDtoV2.file');
 
       // Variant types are still imported so they remain referenceable.
       const importNames = result.imports.map((i) => i.name);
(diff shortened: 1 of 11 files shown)


melloware (changes requested):
Good feedback from the AI code review please implement if possible

melloware:
looks like build failure now?

nathansogut-sipios:
Thanks for the thorough review @coderabbitai. Addressed all feedback in the latest commits:

- **Duplicate FormData entries** (critical, outside-diff on [4624a7ed5](https://github.com/orval-labs/orval/pull/3245/commits/4624a7ed5b1102bc26659c9d28fcfe9e82acfa73)): fixed in [9a4fb8de7](https://github.com/orval-labs/orval/pull/3245/commits/9a4fb8de7). The `shouldCast` branch now collects the outer schema's direct property keys and emits an early-return at the top of the `Object.entries` loop so those keys are handled exclusively by the properties branch that follows. Added a dedicated regression test (`'oneOf alongside direct properties: loop skips direct keys to avoid duplicate appends'`) that pins down the exact pattern (`oneOf: [VariantA, VariantB]` + `properties: {name, owner}`) and asserts each direct key is appended exactly once.
- **Buffer TS compile** (surfaced by the upstream `Verify generated code can build` job on the same commit): fixed in [9a4fb8de7](https://github.com/orval-labs/orval/pull/3245/commits/9a4fb8de7) and refined in [c25b3e394](https://github.com/orval-labs/orval/pull/3245/commits/c25b3e394). The Buffer path now wraps the value as `new Blob([Uint8Array.from(value)])`. I initially tried a `as unknown as BlobPart` cast but dropped it because `BlobPart` only lives in `lib.dom.d.ts` and would break consumers that compile the generated client without the DOM lib. `Uint8Array.from` is in the default ES lib and its signature returns `Uint8Array<ArrayBuffer>`, which satisfies the strict `BlobPart` signature without a cast.
- **Appendix nitpick** (weak `appendFileCount` assertion, [review 2](https://github.com/orval-labs/orval/pull/3245#pullrequestreview-PRR_kwDODjW7-c72C5bz)): replaced with per-variant reference checks (`not.toContain('uploadDtoV1.file')`, `not.toContain('uploadDtoV2.file')`) in [4624a7ed5](https://github.com/orval-labs/orval/pull/3245/commits/4624a7ed5b1102bc26659c9d28fcfe9e82acfa73).

Local verification on the equivalent Node 22 env:
- `bun run test` — 1643 passing
- `bun run --filter orval-tests build` — all 14 clients pass (this was the failing CI job)
- `bun run test:snapshots` — 3586 passing
- `bun run lint` / `typecheck` — clean

The one nitpick I did not act on is the import consolidation in the generated snapshots (`import type { Error, Pet } from './model'` + `import type { PetBase, PetExtended } from './model'` appearing as two separate statements). That's orthogonal to this fix and pre-existed in the codegen's import ordering, happy to split into a follow-up if you want it addressed here.

melloware (approved):

PR #3253, opened 2026-04-19
fix(core): baseUrl runtime imports missing in single mode without schemas

mikan3rd:
Fixes runtime `baseUrl.imports` not appearing in generated clients when `output.schemas` is unset (single-file mode previously passed an empty import list to the builder).

Also default those import specs to `values: true` so symbols used inside generated URL template literals emit as value imports (not `import type`).

Closes the report in https://github.com/orval-labs/orval/issues/3071#issuecomment-4274524995

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve explicit import value flags and default unspecified import values to true.
  * Always generate builder imports so runtime/importPath-based entries appear even when schema configuration is missing.

* **Tests**
  * Added and updated unit tests to validate import value behavior and unconditional generation of builder imports.

### DIFF
--- a/packages/core/src/getters/route.test.ts
+++ b/packages/core/src/getters/route.test.ts
@@ -250,9 +250,9 @@ describe('getBaseUrlRuntimeImports', () => {
 
   it('preserves explicit values: false on baseUrl imports', () => {
     const imports = [{ name: 'x', importPath: './x', values: false as const }];
-    expect(
-      getBaseUrlRuntimeImports({ runtime: 'x', imports }),
-    ).toEqual(imports);
+    expect(getBaseUrlRuntimeImports({ runtime: 'x', imports })).toEqual(
+      imports,
+    );
   });
 });
 
--- a/packages/core/src/getters/route.test.ts
+++ b/packages/core/src/getters/route.test.ts
@@ -235,7 +235,7 @@ describe('getBaseUrlRuntimeImports', () => {
         runtime: 'apiBase',
         imports,
       }),
-    ).toEqual(imports);
+    ).toEqual([{ ...imports[0], values: true }]);
   });
 
   it('returns imports for env object pattern (runtime uses property access)', () => {
@@ -245,6 +245,13 @@ describe('getBaseUrlRuntimeImports', () => {
         runtime: 'env.API_BASE_URL',
         imports,
       }),
+    ).toEqual([{ ...imports[0], values: true }]);
+  });
+
+  it('preserves explicit values: false on baseUrl imports', () => {
+    const imports = [{ name: 'x', importPath: './x', values: false as const }];
+    expect(
+      getBaseUrlRuntimeImports({ runtime: 'x', imports }),
     ).toEqual(imports);
   });
 });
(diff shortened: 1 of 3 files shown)


packages/core/src/getters/route.test.ts
```diff
@@ -245,6 +245,13 @@ describe('getBaseUrlRuntimeImports', () => {
...
       }),
+    ).toEqual([{ ...imports[0], values: true }]);
+  });
+
+  it('preserves explicit values: false on baseUrl imports', () => {
+    const imports = [{ name: 'x', importPath: './x', values: false as const }];
+    expect(
+      getBaseUrlRuntimeImports({ runtime: 'x', imports }),
     ).toEqual(imports);
   });
```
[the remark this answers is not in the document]
mikan3rd: fixed

packages/core/src/writers/single-mode.ts
```diff
@@ -94,9 +94,19 @@ export async function writeSingleMode({
...
+    // Use `.` only so `generateImportsForBuilder` receives a string; imports
+    // without `importPath` would be resolved relative to that path and can be
+    // wrong in no-`schemas` setups—typical single-file fetch + `baseUrl.imports`
+    // only uses entries with `importPath`. With `indexFiles`, schema buckets may
+    // list `dependency: '.'` and empty `exports`; `addDependency` then returns
+    // nothing and those entries are dropped from emitted import text.
+    const importsForBuilder = generateImportsForBuilder(
+      output,
+      normalizedImports,
+      schemasPath ?? '.',
```
mikan3rd: In-code note: why `schemasPath ?? '.'` is safe for the usual `baseUrl.imports` case, and what breaks if `importPath`-less schema imports appear without `output.schemas` (plus empty `exports` buckets when `indexFiles`).

packages/core/src/getters/route.ts
```diff
@@ -129,13 +129,20 @@ export function getFullRoute(
...
  */
 export function getBaseUrlRuntimeImports(
   baseUrl?: NormalizedOutputOptions['baseUrl'],
 ): GeneratorImport[] {
   if (!baseUrl) return [];
   if (!isBaseUrlRuntime(baseUrl)) return [];
-  return baseUrl.imports ?? [];
+  return (baseUrl.imports ?? []).map((imp) => ({
+    ...imp,
+    values: imp.values ?? true,
```
mikan3rd, line 144: JSDoc captures the `values: imp.values ?? true` intent (runtime URL literals) and when to set `values: false` explicitly.

mikan3rd (commented):
In-code notes added for the review caveats (`schemasPath ?? .` / `importPath`, empty schema buckets; `getBaseUrlRuntimeImports` `values` default).

packages/core/src/getters/route.ts
mikan3rd: fixed

packages/core/src/writers/single-mode.ts
mikan3rd: fixed

melloware (approved):

PR #3254, opened 2026-04-20
feat(core): only referenced schemas output when tags specified in `input.filter`

### LINKED ISSUE #1734: Feature Request: support removing unused schemas based on refs in paths
## What happens?

After a bunch of filtering, excluding and transforming, only parts of the original `paths` would be preserved and some schema definitions become irrelevant and stale after the referencing `paths` have been filtered out. 

1. There should be a configurable property in the `output` section of the config to specify whether or not to remove unused schemas.
2. If set to `true`, unused schemas will not be generated to the output file.

None.

## Any other comments?

For now, this can probably be implemented in `input.override.transformer`, but it's not efficient if every time a user wants the feature, they would need to implement it by themselves.

On top of that, if you think it's a desirable feature to add, I can help open a PR for it.

`7.3.0`

soartec-lab:
## Summary

fix #1734

This change makes orval automatically limit schema output to only those referenced (directly or transitively) by the matching operations, when `input.filters.tags` is set without `input.filters.schemas`.

## Motivation

When `input.filters.tags` is used to filter endpoints, all `#/components/schemas` entries were still emitted in the generated output. This caused unrelated schemas to appear as noise, even though the endpoints were already narrowed down by tags.

## Changes

| `filters.tags` | `filters.schemas` | Result |
| --- | --- | --- |
| ❌ | ❌ | All schemas (unchanged) |
| ❌ | ✅ | Explicit schemas only (unchanged) |
| ✅ | ❌ | **Only schemas referenced by matching operations (new)** |
| ✅ | ✅ | `filters.schemas` takes precedence (unchanged) |

Configuration:

```ts
export default defineConfig({
  petstore: {
    input: {
      target: './petstore.yaml',
      filters: {
        tags: ['pets'],
        // schemas not specified — only schemas referenced
        // by 'pets' operations are generated
      },
    },
    output: './generated/endpoints.ts',
  },
});
```

## Summary by CodeRabbit

* **New Features**
  * Output is automatically limited to schemas referenced by operations that match tag filters; explicit schema lists override this behavior
  * Generation pipeline now applies tag-based schema filtering so produced output reflects those filters

* **Documentation**
  * Clarified interaction between tag and schema filters with examples

* **Tests**
  * Added comprehensive tests covering tag filtering and referenced-component collection, including transitive refs and exclusion mode


melloware (approved):

soartec-lab:
@melloware 
Thanks for review. I'm done, when filtering by tags, now all component types (schemas, responses, parameters, requestBodies) are filtered to only what's needed. Indirect refs (e.g. schema refs inside a response component) are also followed, so the generated code compiles without missing types.

melloware (approved):

PR #3257, opened 2026-04-20
fix: add zip to the binary application types list

jmorel:
Prior to the [v8.7.0 release](https://github.com/orval-labs/orval/releases/tag/v8.7.0), responses that had a `{"type": "string", "format": "binary"}` response would have a `responseType: 'blob'` in the generated client.

```
"responses": {
  "200": {
    "description": "Successful Response",
    "content": {
      "application/zip": {
        "schema": {
          "type": "string",
          "format": "binary"
        }
      }
    }
  },
}
```

From 8.7.0 onwards, there is a strict list of application types that will trigger the blob response type. This PR adds `application/zip` to the list.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved binary file type detection to properly classify ZIP files as binary content.

* **Tests**
  * Extended test coverage for binary content type detection across multiple MIME types.

### DIFF
--- a/packages/core/src/getters/res-req-types.test.ts
+++ b/packages/core/src/getters/res-req-types.test.ts
@@ -78,6 +78,7 @@ describe('isBinaryContentType', () => {
   it('should return true for binary content types', () => {
     expect(isBinaryContentType('application/octet-stream')).toBe(true);
     expect(isBinaryContentType('application/pdf')).toBe(true);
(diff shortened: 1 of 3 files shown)


melloware:
@jmorel thanks this is a good improvement

melloware (approved):

PR #3261, opened 2026-04-21
fix(fetch): include error content types in parse strategy (follow-up to #3222) (#3088)

zeriong:
## Summary

Follow-up to #3222. That PR made the fetch client parse responses based on `Content-Type` instead of always calling `JSON.parse`, but the generation-time "is this operation mixed?" check only inspected 2xx success content types. When `forceSuccessResponse` is false the same parse block serves both success and error responses, so mixed success/error content types still emitted `body ? JSON.parse(body) : {}` and threw at runtime on the non-JSON branch.

@Huliiiiii (the reporter of #3088) pointed this out directly in their review on #3222:

> This vibe PR did not solve my issue. […] This will still parse text as json.

(with the exact `/data` spec — 200 `application/json` + 429 `text/plain` — that opened #3088). The root cause they also identified on #3222: "the new logic was added to `throwOnErrorImplementation`, which only gets executed when `forceSuccessResponse` is enabled." This PR fixes that by extending the detection to error content types when `forceSuccessResponse` is false.

## Change

One targeted change in `packages/fetch/src/index.ts`: the existing `hasMixedSuccessContentTypes` / `successAlwaysJson` detection now also considers error content types when `forceSuccessResponse` is false. Everything else — runtime dispatch shape, `toLowerCase`, `body` checks, snapshots — is already in place from #3222 and unchanged here.

## Scope

- Success-only operations: unchanged
- Operations with `forceSuccessResponse: true`: unchanged (errors go through the throw path which already has its own content-type handling)
- Operations with `forceSuccessResponse: false` (the default) that have mixed success/error content types: now emit the runtime `Content-Type` dispatch instead of a hard-coded `JSON.parse`

## Test plan

- [x] `bun run lint` passes
- [x] `bun run build` — 15/15 clients typecheck
- [x] `bun run test:snapshots` — 3703 passing
- [x] `tests/specifications/mixed-content-responses.yaml` `/data` (the exact case @Huliiiiii reported) now regenerates with the runtime header dispatch and no `JSON.parse(text)` path:

```ts
const contentType = (res.headers.get('content-type') ?? '').toLowerCase();
const data: getDataResponse['data'] = body
  ? contentType.includes('json')
    ? JSON.parse(body)
    : body
  : {};
```

Issue: #3088
Previous PR: #3222

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing for mixed success/error responses so content types from both are respected.
  * Fixed no-content responses (204/205/304) to return an empty object ({}) instead of an empty string.

* **New Features**
  * Response bodies are parsed as JSON only when the server Content-Type indicates JSON; non-JSON responses are returned as raw text.


soartec-lab:
@zeriong First of all, shouldn't you answer the question I asked in your initial PR?

zeriong:
@soartec-lab You're right, I apologize for the confusion. I did leave a reply on #3222:

> @melloware @soartec-lab
> I'm so sorry, the code mixed with the previous content went up the same way, and the PR was accepted.
> I'll open a new PR referring to this PR as soon as possible! 🥹

But I understand that wasn't sufficient, and I should have addressed your specific questions directly. I'll go through each of your comments on the original PR and reply to them one by one as soon as possible.

Thank you for your patience and for reviewing this.

tests/__snapshots__/default/file-extension-tags-split/health/health.generated.ts
```diff
@@ -97,9 +97,14 @@ export const healthCheck = async (
     method: 'GET',
   });
 
+  const contentType = (res.headers.get('content-type') ?? '').toLowerCase();
   const body = [204, 205, 304].includes(res.status) ? null : await res.text();
 
-  const data: healthCheckResponse['data'] = body !== null ? body : '';
+  const data: healthCheckResponse['data'] = body
```
soartec-lab, line 103: Since `health` is only `text/plain`, I don't think JSON identification is necessary. This PR doesn't need to be addressed, but could you please create a follow-up?

zeriong: @soartec-lab Before opening a follow-up — I want to flag that the runtime `contentType.includes('json')` check is actually required for `healthCheck`, not unnecessary. Wanted to double-check the premise with you first.

The spec ([tests/specifications/deprecated.yaml L154-L172](https://github.com/zeriong/orval/blob/d7bad64662b66ca838743a541242a98c34b346f4/tests/specifications/deprecated.yaml#L154-L172)) declares `200: text/plain` *and* `default: application/json` (the `Error` schema). In core, `groupedByStatus` puts anything not starting with `'2'` into `errors` ([packages/core/src/getters/response.ts L49-L55](https://github.com/zeriong/orval/blob/d7bad64662b66ca838743a541242a98c34b346f4/packages/core/src/getters/response.ts#L49-L55)), so `default application/json` becomes an *error* content type. With `forceSuccessResponse: false` (default), the same parse block serves both success and error responses, so `parseTimeContentTypes = [...successContentTypes, ...errorContentTypes] = ['text/plain', 'application/json']` ([packages/fetch/src/index.ts L218-L220](https://github.com/zeriong/orval/blob/d7bad64662b66ca838743a541242a98c34b346f4/packages/fetch/src/index.ts#L218-L220)). That makes `hasMixedSuccessContentTypes` `true` and the runtime `Content-Type` dispatch is emitted ([L528-L529](https://github.com/zeriong/orval/blob/d7bad64662b66ca838743a541242a98c34b346f4/packages/fetch/src/index.ts#L528-L529)). If we dropped the check here, a `default`-status JSON error body would be returned as a raw string instead of a parsed `Error` object.

So this isn't a `healthCheck`-only artifact — it's the intended behavior whenever success and error content types differ.

If you'd still like a follow-up to specialize this case, the options I see are:
1. `forceSuccessResponse: true` — routes errors through `throwOnErrorImplementation` and lets the success block use `body !== null ? body : ''`. No generator change needed; just a config-level choice per operation.
2. A new generator option to opt out of error-side parsing in the default path.

Happy to open the follow-up in either direction once you confirm which you'd prefer (or whether you'd like to leave the current behavior as-is given the above).

soartec-lab: @zeriong 
Ah, I see. So it was specifying JSON by default. You're absolutely right 👍

melloware (dismissed):

tests/__snapshots__/default/file-extension-tags-split/health/health.generated.ts
zeriong: @soartec-lab 
Thank you for always carefully reviewing the code!
I'm learning a lot thanks to you 🫡

melloware:
@soartec-lab where are we with this one?

melloware (approved):

soartec-lab:
@melloware @zeriong 
All issues have been resolved, so I will merge this PR 👍

soartec-lab (approved):

PR #3279, opened 2026-04-26
fix(core,mock): import not propagated to msw mock for nullable anyOf $ref types

### LINKED ISSUE #3269: mock: MSW file missing import for __-aliased types when using nullable $ref (anyOf: [$ref, {type: null}])
## Description

Starting in v8.8.1, schemas referenced via the OpenAPI 3.1 nullable-ref pattern (`anyOf: [{$ref: ...}, {type: null}]`) are imported into the main endpoint file with a `__`-prefixed alias (e.g. `Widget as __Widget`), and the type reference in the endpoint file uses the aliased name. However, the companion `.msw.ts` file **also uses the aliased name (`__Widget`) but never imports it**, producing `TS2304: Cannot find name '__Widget'` on compile.

This is a regression from v8.8.0, where both files imported the bare name `Widget`. The change appears to have been introduced by #3245 (`fix(core): support oneOf/anyOf in multipart/form-data code generation`), whose snapshot diff for `tests/__snapshots__/react-query/importFromSubdirectory/endpoints.msw.ts` adds `AnotherSchema as __AnotherSchema` imports. That PR's snapshot happens to include both bare and aliased imports on the msw side, but in the split-mode + nullable-anyOf-ref scenario below, the msw generator emits zero import for the aliased name.

The nullable-ref `anyOf` pattern is the default shape FastAPI/Pydantic emits for `Optional[Model]` response/request bodies, so this breaks a very common real-world case. Downgrading to `orval@8.8.0` against the exact same OpenAPI spec and orval config produces correct output.

## Output client

`react-query` (affects the MSW mock generator; likely any client combined with `mock: { type: 'msw' }`)

```js
// orval.config.mjs
export default {
  repro: {
    input: { target: './openapi.json' },
    output: {
      mode: 'tags-split',
      target: './generated',
      schemas: './generated/model',
      client: 'react-query',
      mock: { type: 'msw' },
      clean: true,
    },
  },
};
```

## Environment

```
System:
  OS: macOS 26.3.1 (darwin arm64) — also reproduces on Linux
  Node: v22+
npmPackages:
  orval: 8.8.1
  msw: 2.13.4
  typescript: 5.9.3
```

## Expected behavior

Generated with `orval@8.8.0` (working):

```ts
// generated/default/default.ts
import type { Widget } from '../model';
export type getWidgetResponse200 = { data: Widget | null; status: 200 };
```

```ts
// generated/default/default.msw.ts
import type { Widget } from '../model';
export const getGetWidgetResponseMock = (): Widget | null => (...)
```

## Actual behavior

Generated with `orval@8.8.1`:

```ts
// generated/default/default.ts — type imported with __ alias
import type { Widget as __Widget } from '../model';
export type getWidgetResponse200 = { data: __Widget | null; status: 200 };
```

```ts
// generated/default/default.msw.ts — uses __Widget but NEVER imports it
import { HttpResponse, http } from 'msw';
import type { RequestHandlerOptions } from 'msw';
// ^ no import of Widget or __Widget at all

export const getGetWidgetResponseMock = (): __Widget | null => (...)
//                                            ^^^^^^^^ TS2304
```

`tsc --noEmit` output:

```
generated/default/default.msw.ts(20,45): error TS2304: Cannot find name '__Widget'.
generated/default/default.msw.ts(23,60): error TS2304: Cannot find name '__Widget'.
generated/default/default.msw.ts(23,144): error TS2304: Cannot find name '__Widget'.
generated/default/default.msw.ts(23,163): error TS2304: Cannot find name '__Widget'.
```

## OpenAPI document (minimal)

```json
{
  "openapi": "3.1.0",
  "info": { "title": "Repro", "version": "0.0.1" },
  "paths": {
    "/widgets": {
      "get": {
        "operationId": "get_widget",
        "responses": {
          "200": {
            "description": "OK",
            "content": {
              "application/json": {
                "schema": {
                  "anyOf": [
                    { "$ref": "#/components/schemas/Widget" },
                    { "type": "null" }
                  ],
                  "title": "Response Get Widget"
                }
              }
            }
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "Widget": {
        "type": "object",
        "properties": {
          "id": { "type": "string" },
          "label": { "type": "string" }
        },
        "required": ["id", "label"],
        "title": "Widget"
      }
    }
  }
}
```

## Additional context

- Confirmed bisect: `orval@8.8.0` emits correct code from the identical config+spec; `orval@8.8.1` reproduces the TS2304.
- Suspected source: #3245. The snapshot diff adds `__`-aliased imports on the msw side, but the emission path for the split-mode + nullable-anyOf-ref case does not include the alias import.
- Also affects request bodies with `anyOf: [{$ref: Foo}, {type: null}]` schemas — in that case the body param is additionally renamed to `_fooNull` while the type alias `__Foo` is referenced without import in the msw mock.
- Affects any client when `mock: { type: 'msw' }` is enabled; reproduced against `react-query`, likely identical on `fetch` / `axios` / `swr`.
- Real-world occurrence: FastAPI/Pydantic emits exactly this shape for `Optional[Model]` response/request typing, so every endpoint with a nullable model body breaks compilation after upgrading to 8.8.1.

nathansogut-sipios:
Closes #3269.

## What users saw

8.8.0 → 8.8.1 regression. With tags-split + msw mode, an OpenAPI 3.1 schema using the FastAPI/Pydantic `Optional[Model]` shape (`anyOf: [{$ref: X}, {type: 'null'}]`) generates a `*.msw.ts` that references `__X` without importing it → `TS2304: Cannot find name '__X'`. Issue body has the minimal repro. Same bug shape on request bodies using the same schema.

### `packages/core/src/getters/res-req-types.ts` — over-broad scoping (introduced by #3245)

The `oneOf`/`anyOf` branch added in #3245 was unconditional on content-type, so it fired for every media type. Outside form-data and url-encoded paths, deriving `effectivePropName` from the inner `$ref` collides with the schema import in split mode and produces a `Foo as __Foo` alias that did not need to exist.

Narrowed the branch to `(isFormData || isFormUrlEncoded)`. Both share `getSchemaFormDataAndUrlEncoded` downstream and both need the schema name to align with the runtime variable iterated over by the generated `FormData` / `URLSearchParams` code. Other content types only use the param name to seed response type aliases.

### `packages/mock/src/msw/index.ts` — alias-blind import filter (pre-existing)

The filter at the `response.imports.filter` site searched the bare name with a word-boundary regex against the generated mock code. When an import is aliased, the code references the alias (`__Foo`), and `\bFoo\b` does not match `__Foo` (because `_` is a JS regex word-character), so the import was filtered out and the mock file ended up referencing an undeclared identifier.

Switched the filter to match alias OR name with `escapeRegExp`, mirroring the `addDependency` pattern in `core/src/generators/imports.ts`.

With (1) fixed, (2) no longer manifests on the #3269 shape specifically, but (2) is a real bug independently — any future code path that produces an aliased import in a context that emits only the alias hits it. Small, mirrored, tested fix.

## Tests

- 3 unit tests in `packages/mock/src/msw/index.test.ts` covering the alias filter: alias-only reference kept, unused alias dropped (with a positive-control sibling import to anchor the assertion), and dual bare + aliased entries both surviving when both forms appear in the code.
- New regression fixture `tests/specifications/issue-3269.yaml` plus a `issue3269` entry in `tests/configs/react-query.config.ts`. Covers response and request body `anyOf-nullable` in tags-split + msw mode end-to-end. Snapshots locked under `tests/__snapshots__/react-query/issue-3269/`.
- 3 `importFromSubdirectory` snapshots refreshed: the aliased imports they previously contained existed solely because of #3245's overreach in non-form-data paths and are no longer emitted.

## Verification

- 1779 unit tests pass (1662 `@orval/core`, 117 `@orval/mock` incl. the 3 new ones, 166 `@orval/zod`, 79 `orval`, all the rest)
- 3718 snapshot tests pass (was 3714 on stock 8.8.1 + 4 new from the fixture)
- 15/15 generated-client typecheck targets pass + `verify-mock-generated.mjs` OK
- Manual repros (response anyOf-nullable, request body anyOf-nullable) both produce clean `import type { X }` mock code with no `__X` references; `tsc --noEmit` clean

The #3245 multipart oneOf/anyOf tests in `res-req-types.test.ts` remain green, and the multipart snapshots (`react-query/form-data/endpoints.ts` etc.) still contain the `Object.entries(... ?? {}).forEach` runtime loop and DTO-derived FormData variable name — the original multipart fix is intact.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected request/response content type naming and aliasing behavior for form-related data types.

* **Tests**
  * Added regression tests for aliased response imports in mock generation.
  * Expanded test coverage for nullable anyOf schemas with tags-split and MSW configurations.


nathansogut-sipios:
@coderabbitai review

melloware (approved):

PR #3288, opened 2026-04-28
fix(core): wrap allOf intersection types in parens for array items

### LINKED ISSUE #3287: Array items using `allOf` generate `A & B[]` instead of `(A & B)[]`
### Description

Reproduction is available at https://github.com/sushichan044/reproductions/tree/main/workspace/orval-union-array

When an OpenAPI schema defines an array whose `items` uses `allOf` to combine multiple `$ref`s, Orval emits the intersection type **without parentheses**.

TypeScript operator precedence means `A & B[]` is parsed as `A & (B[])`, not as `(A & B)[]`.
The two types are fundamentally different:

| Expression  | Meaning                                                      |
| ----------- | ------------------------------------------------------------ |
| `A & B[]`   | an object that is simultaneously `A` **and** an array of `B` |
| `(A & B)[]` | an array whose every element satisfies both `A` and `B`      |

Per the [OpenAPI 3.0.3 specification §4.7.25 (Schema Object)](https://spec.openapis.org/oas/v3.0.3.html#schema-object) and the underlying [JSON Schema `allOf` semantics](https://json-schema.org/understanding-json-schema/reference/combining.html#allof), `allOf` is a logical AND: an instance must be valid against **all** listed subschemas simultaneously. Applied inside `items`, this means every array element must satisfy the intersection — so the correct TypeScript representation is `(A & B)[]`.

This is a **type generation** issue, independent of which output client or mock strategy is used.
The generated response type is semantically wrong regardless of whether mock generation is enabled.

### Output client

fetch

```ts

export default defineConfig({
  repro: {
    input: {
      target: "./openapi.yaml",
    },
    output: {
      target: "./generated/repro.ts",
      client: "fetch",
      formatter: "oxfmt",
    },
  },
});

```

### Environment

```
System:
  OS: macOS 26.4.1
  CPU: (16) arm64 Apple M3 Max
  Shell: 5.9 - /bin/zsh
npmPackages:
  orval: latest => 8.9.0
  typescript: latest => 6.0.3
```

### Expected behavior

For a response schema of `type: array` with `items.allOf: [$ref A, $ref B]`, the generated TypeScript type should be:

```ts
export type getOptionsResponse200 = {
  data: (UnionPartOne & UnionPartTwo)[];
  status: 200;
};
```

Or, if `aliasCombinedTypes` is in play, an equivalent alias:

```ts
export type UnionPartOneAndUnionPartTwo = UnionPartOne & UnionPartTwo;

export type getOptionsResponse200 = {
  data: UnionPartOneAndUnionPartTwo[];
  status: 200;
};
```

### Actual behavior

The generated type is missing parentheses:

```ts
export type getOptionsResponse200 = {
  data: UnionPartOne & UnionPartTwo[]; // ← wrong precedence
  status: 200;
};
```

`tsc` does not raise an error on the type definition itself because no value is assigned to `data` inside the generated file. The error surfaces at call-site as soon as application code tries to store an actual array value into `data`.

```yaml
openapi: 3.0.3
info:
  title: array allOf repro
  version: 1.0.0
paths:
  /options:
    get:
      operationId: getOptions
      responses:
        "200":
          description: OK
          content:
            application/json:
              schema:
                type: array
                items:
                  allOf:
                    - $ref: "#/components/schemas/UnionPartOne"
                    - $ref: "#/components/schemas/UnionPartTwo"
components:
  schemas:
    UnionPartOne:
      type: object
      required: [id, name]
      properties:
        id:
          type: integer
        name:
          type: string
    UnionPartTwo:
      type: object
      required: [deletable]
      properties:
        deletable:
          type: boolean
```

### Workaround

Setting `aliasCombinedTypes: true` avoids the incorrect output by routing through the V7-era alias-based code path:

```ts
output: {
  override: {
    aliasCombinedTypes: true,
  },
},
```

This option was introduced as a migration escape hatch in v8 (PR #2657) and has been suggested by a collaborator as a temporary workaround for related operator-precedence regressions (see #2791). It is not a fix — the default code path still emits the wrong type.

### Related issues

- #2791 — `oneOf inside allOf` generates incorrect precedence (labeled `bug`, closed at 8.0.3 milestone with `aliasCombinedTypes` workaround; the `items.allOf` case appears uncovered)
- #1508 — `items` + `allOf` + mock generation, invalid mock output
- #1807 — `allOf` array, invalid TypeScript syntax
- #2020 — `allOf` mock generation, invalid syntax
- #2029 — mock + `allOf`, required field handling

### Spec references

- [OpenAPI 3.0.3 §4.7.25 Schema Object](https://spec.openapis.org/oas/v3.0.3.html#schema-object) — defines `allOf` and `items`
- [JSON Schema Draft 4 §5.5.3 `allOf`](https://json-schema.org/draft-04/json-schema-validation#rfc.section.5.5.3) — base spec OAS 3.0 references
- [Understanding JSON Schema — Schema Composition (`allOf`)](https://json-schema.org/understanding-json-schema/reference/combining.html#allof) — "an instance must be valid against all schemas"

sushichan044:
fixes #3287 

> [!NOTE]
> I use AI for tasks including commits and PR creation, and I hereby declare that I review all such content.
I have also read the CONTRIBUTING.md file.

## What

`items.allOf: [$ref A, $ref B]` was generating `A & B[]` instead of `(A & B)[]`.
TypeScript operator precedence makes `A & B[]` mean `A & (B[])`, which is semantically wrong.

**Root cause:** `packages/core/src/getters/array.ts` already parenthesised values containing `|` before appending `[]`, but had no equivalent guard for `&`.

**Fix:** extend the condition to also check `includes('&')`.

## Breaking change risk

Generated types for any schema using `items.allOf` with `$ref` entries will change:

```diff
- data: UnionPartOne & UnionPartTwo[]   // wrong — was A & (B[])
+ data: (UnionPartOne & UnionPartTwo)[] // correct
```

Users who worked around the incorrect type (e.g. with `aliasCombinedTypes: true`) will see their output change. Downstream TypeScript errors that were previously hidden at declaration site may surface at call sites.

The `includes('&')` check also fires when `&` appears inside generic arguments (e.g. `Record<string, A & B>`), producing extra but semantically equivalent parens — not a correctness issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed TypeScript type generation for arrays containing intersected types to use proper parenthesized syntax `(A & B)[]` instead of incorrect `A & B[]`.

* **Tests**
  * Added comprehensive test coverage for array types composed of intersected components.

* **Chores**
  * Added new OpenAPI specification and configuration for testing array intersection scenarios.


packages/core/src/getters/array.test.ts
```diff
@@ -37,6 +37,45 @@ describe('getArray', () => {
...
+      },
+    };
+
+    const result = getArray({
+      schema,
+      name: 'MyArray',
+      context: contextWithSchemas,
+    });
+
+    expect(result.value).toEqual('(PartOne & PartTwo)[]');
```
sushichan044, line 76: This test case is a bit brittle to codegen changes, but it's intentional to ensure the union is correct. I prefer checking the generated code directly over regex, as it's easier to read and maintain.

melloware:
Running the build now

tests/specifications/all-of-array-items.yaml
```diff
@@ -0,0 +1,35 @@
+openapi: 3.0.3
```
soartec-lab: If the test for "deeply-nested-refs" has been updated, doesn't that mean it already exists?
Need to add new test cases?

if necessary, Instead of creating a new file, move to the file below.

https://github.com/orval-labs/orval/blob/master/tests/specifications/all-of.yaml

sushichan044: @soartec-lab 
Thank you for your review.
You are right that the existing cases and the new unit tests cover it sufficiently, so I have removed the test case that involved code generation in 87cd92b.

soartec-lab: Thanks!

soartec-lab (approved):

PR #3299, opened 2026-05-01
fix(packaging): drop "development" export condition leaking into tarball

### LINKED ISSUE #3216: fix(packaging): published packages expose "development" export condition pointing to missing src/ files (v8.6.0+)

zeriong:
## Summary

- Reverts packages/tsdown.base.ts to `exports: true`, removing the auto-injected `"development": "./src/index.ts"` condition from every package's `exports` field.
- Drops the now-unnecessary `publishConfig.exports` overrides from all 12 published packages.

## Why

Since v8.6.0, tsdown.base.ts used `exports: { devExports: 'development' }`, which makes tsdown auto-generate a `"development"` export condition pointing at `./src/index.ts`. A `publishConfig.exports` block was added to strip it at publish time — but the release workflow uses `bun publish`, and unlike `npm publish`, `bun publish` does NOT apply `publishConfig.exports`.

As a result, the published tarball (@orval/* 8.6.0+) shipped with an exports map containing `"development": "./src/index.ts"` alongside `"default": "./dist/index.mjs"`.

Any consumer whose resolver activates the `development` condition (Vite dev mode, tsconfig.json with `customConditions: ["development"]`, etc.) resolves to ./src/index.ts, which is excluded from the tarball by `files: ["dist"]` — producing: Error: Cannot find module '.../node_modules/@orval/core/src/index.ts'.

Affected versions: 8.6.0, 8.6.1, 8.6.2, 8.7.0 (and beyond until this fix lands). Affected packages: all 12 (core, orval, angular, axios, fetch, hono, mcp, mock, query, solid-start, swr, zod).

## Approach

Fix the source rather than the publish path: with `exports: true`, tsdown writes a simple dist-only `exports` field at build time, so the published package.json no longer carries any condition that can leak. The `publishConfig.exports` blocks become redundant and are removed.

Dev-mode DX for maintainers is unaffected — `bun run dev` (which runs `tsdown --watch`) keeps dist/ in sync during development, which is how internal @orval/* cross-package imports already resolved.

## Verification

- Clean build is idempotent: `bun run build` twice produces no package.json diff.
- `bun run test` (turbo, all packages) — passes.
- `bun run test:snapshots` — passes.
- `bun run typecheck` — passes.
- `bun run lint` — passes.
- `prettier --check` — passes.
- Inspected resulting `exports` in every packages/*/package.json — dist-only, no `development` condition, no residual `publishConfig`.
- orval package's multi-entry export (./bin/orval) is preserved.

## Test plan

- [ ] CI green on Ubuntu 22.x and Windows runners (build, typecheck, lint, unit, snapshots).
- [ ] After a release, `npm pack @orval/core@<next>` shows exports as `{ ".": "./dist/index.mjs", "./package.json": "./package.json" }` — no `"development"` key.
- [ ] Reproducer from the issue (Vite dev / tsconfig customConditions: ["development"]) no longer throws Cannot find module '.../src/index.ts'.

Fixes #3216

## Release Notes

* **Chores**
  * Simplified package export configurations across multiple packages to always resolve to built artifacts
  * Removed redundant publishing configuration overrides, streamlining export definitions


melloware:
Discussed internally with the dev team and they feel like this is not the correct fix. 

image

PR #3307, opened 2026-05-03
fix(core): support Map format for x-enumDescriptions and x-enumNames (#3232)

### LINKED ISSUE #3232: Bug: `x-enumDescriptions` only supports array format, not the `Map[string, string]` format defined by the OpenAPI extension spec
### Description

When `x-enumDescriptions` is defined as a `Map[string, string]`, Orval crashes during generation with:

`🛑 api - descriptions.map is not a function`

Orval internally calls `.map()` on `x-enumDescriptions`, assuming it is an array.  However, the correct format per Redocly's OpenAPI extension spec is `Map[string, string]`: https://redocly.com/docs/realm/content/api-docs/openapi-extensions/x-enum-descriptions

### Output client

vue-query

```ts

export default defineConfig({
    api: {
        output: {
            allParamsOptional: true,
            client: 'vue-query',
            headers: true,
            mode: 'single',
            httpClient: 'axios',
            override: {
                aliasCombinedTypes: true,
                components: {
                    requestBodies: {
                        suffix: 'Body',
                    },
                    parameters: {
                        suffix: 'Params',
                    },
                    schemas: {
                        suffix: 'Schema',
                    },
                    responses: {
                        suffix: 'Response',
                    },
                },
                mutator: {
                    path: './src/api/orval/index.ts',
                    name: 'customInstance',
                },
                formData: {
                    path: './src/api/orval/index.ts',
                    name: 'customFormData',
                },
                useTypeOverInterfaces: true,
            },
            prettier: true,
            target: 'src/api/orval/client.ts',
        },
        input: '../api/docs/scramble.json',
    },
});

```

### Environment

```
  System:
    OS: macOS 26.2
    CPU: (8) arm64 Apple M3
    Memory: 71.47 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    @tanstack/vue-query: ^5.96.2 => 5.96.2 
    axios: ^1.15.0 => 1.15.0 
    orval: ^8.6.2 => 8.6.2 
    vue: ^3.5.32 => 3.5.32 
```

### Expected behavior

Orval should handle `x-enumDescriptions` as both an `Array` (current) and a `Map[string, string]` (object), since both formats exist in the wild and the Redocly spec explicitly defines it as `Map[string, string]`.

### Actual behavior

Currently only handles `x-enumDescriptions` as `Array`, which causes `descriptions.map is not a function`.

```yaml
openapi: 3.0.3
info:
  title: Demo
  version: 0.0.0
paths: {}
components:
  schemas:
    Color:
      type: string
      enum:
        - red
        - green
        - blue
      x-enumDescriptions:
        red: The color red.
        green: The color green.
        blue: The color blue.
```

### Additional context

_No response_

zeriong:
## Summary
  - Handle both array (existing) and `Map[string, string]` (Redocly spec) formats for `x-enumDescriptions`, `x-enumdescriptions`, `x-enum-descriptions`,
  `x-enumNames`, `x-enumnames`, and `x-enum-varnames`.
  - When an object is provided, keys are matched against the schema's `enum` values to produce a positional array with `undefined` for unmatched entries.
  - Update type signatures throughout `enum.ts` to accept `(string | undefined)[]` for partial coverage support.
  - Previously, object format caused `descriptions.map is not a function` crash.

  ## Test plan
  - [x] `bun run build` — 12 successful
  - [x] `bun vitest run` — 2318 passed (3 pre-existing failures in `resolve-version.test.ts`, unrelated); 15 new tests covering array format, object/Map format,
  partial descriptions, numeric enum values, special character escaping, all three casing variants, and integration with `getEnumImplementation`
  - [x] `bun run test:snapshots` — 3758 passed, no snapshot changes

  Closes #3232

## Summary by CodeRabbit

* **Tests**
  * Extended coverage for enum names/descriptions across array and object/map forms, partial/missing entries, numeric keys, hyphenated/lowercase extension keys, special-character escaping, and conditional JSDoc generation.

* **New Features**
  * Support for partial/missing enum values in object-format extensions; consumers now receive undefined for missing mapped entries.

* **Style**
  * Reflowed docblock summary comments in generated snapshots.


melloware (approved):

melloware (changes requested):
Looks like Lint issues

zeriong:
@melloware
I think the react-query snapshot was broken in the previous merge.
It also included changes in that part, so if you merge this pr, it is expected to be supplemented right away.

melloware (approved):

PR #3308, opened 2026-05-03
fix(fetch): respect OpenAPI default explode: true for array query parameters (#3231)

### LINKED ISSUE #3231: Fetch client does not respect OpenAPI default `explode: true` for array query parameters
### Description

The fetch client currently treats array query parameters without an explicit `explode` field as non-exploded (comma-separated), but the [OpenAPI 3.1.0 specification](https://spec.openapis.org/oas/v3.1.0#parameter-object) defines the defaults for query parameters as `style: form` and `explode: true`.

From the [Swagger serialization documentation](https://swagger.io/docs/specification/v3_0/serialization/#query-parameters):

> The default serialization method is `style: form` and `explode: true`.

This means that when neither `style` nor `explode` is explicitly specified on a query parameter, the spec-compliant behavior is `style: form` + `explode: true`, resulting in repeated key-value pairs (e.g. `?ids=1&ids=2&ids=3`).

### Current behavior

When a query parameter is defined as an array type without explicitly setting `explode`, Orval's fetch client serializes it as a comma-separated string:

````
?ids=1,2,3
````

This happens because `parameterObject.explode` evaluates to `undefined` (falsy) in [`packages/fetch/src/index.ts`](https://github.com/orval-labs/orval/blob/3e92b827ddb105cc2a30e1ba341a022dcf9da859/packages/fetch/src/index.ts#L123), so the parameter is excluded from the `explodeParameters` list and falls through to `normalParamsImplementation`, where JavaScript's `Array.toString()` joins values with commas.

### Proposed behavior

Per the OpenAPI 3.0/3.1 spec, the default serialization for query parameters is `style: form` + `explode: true`. Array parameters without an explicit `explode` field should be serialized as repeated key-value pairs:

````
?ids=1&ids=2&ids=3
````

### Proposed solution

Change the explode check to use nullish coalescing, so that `undefined` (the spec default) is treated as `true`:

```typescript
// Before
parameterObject.in === 'query' && isArrayLike && parameterObject.explode

// After
parameterObject.in === 'query' && isArrayLike && (parameterObject.explode ?? true)
```

#### Note on `style`

It is also worth noting that the current implementation does not reference the `style` property at all — the fetch client only checks `explode` and completely ignores `style`. According to the OpenAPI spec, query parameter serialization is determined by the combination of `style` and `explode`, and other styles such as `spaceDelimited`, `pipeDelimited`, and `deepObject` are not handled. Full `style` support is out of scope for this issue, but as a middle ground, it might be useful to provide a configuration option to specify the default `style` for query parameters.

#### Breaking Change Consideration

This is a breaking change for users who rely on the current comma-separated behavior without explicitly setting `explode: false`. It may be worth introducing a configuration option to preserve the legacy behavior, or treating this as a major version change.

zeriong:
## Summary
  - Use nullish coalescing (`parameterObject.explode ?? true`) so that `undefined` defaults to `true` per the OpenAPI 3.0/3.1 spec instead of being treated as
  falsy.
  - Previously, array query parameters without an explicit `explode` field were serialized as comma-separated (`?ids=1,2,3`) instead of repeated key-value pairs
  (`?ids=1&ids=2&ids=3`).

  ## Breaking Change
  Users relying on the implicit comma-separated behavior without setting `explode: false` will see a change in serialization output. To preserve the old behavior,
  explicitly set `explode: false` on the parameter.

  ## Test plan
  - [x] `bun run build` — 12 successful
  - [x] `bun vitest run` — 2303 passed (3 pre-existing failures in `resolve-version.test.ts`, unrelated)
  - [x] `bun run test:snapshots` — 3746 passed; 2 snapshot files updated (`fetch/parameters`, `fetch/dateParams`)
  - [x] Verified `solid-start` package already handled this correctly, no other clients affected

  Closes #3231

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed query-parameter explode behavior so array-valued query parameters follow style-based defaults (form-style defaults to explode).
  * Arrays in query parameters are now serialized as repeated entries when exploded (e.g., q and tag), producing correct URLs and request behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/orval-labs/orval/pull/3308?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


soartec-lab:
@zeriong 

In v3.1, the default value is `false`.

https://swagger.io/specification/

idiodoneo-chris:
According to the Swagger spec, for query parameters the default is, in fact, `true` if `style` is not specified or is set to `form`. This is the case when FastAPI produces an api spec (and FastAPI does then require exploded parameters in the query which broke with Orval 8).

Here's a typical example produced by FastAPI:

```json
  {
    "/api/v1/interests": {
      "get": {
        "summary": "List all interests",
        "description": "Lists all available interests with localized names.",
        "operationId": "list_interests_endpoint",
        "parameters": [
          {
            "name": "l",
            "in": "query",
            "required": false,
            "schema": {
              "anyOf": [
                {
                  "type": "array",
                  "items": {
                    "type": "string"
                  }
                },
                {
                  "type": "null"
                }
              ],
              "description": "Language codes to prioritize for text",
              "title": "L"
            },
            "description": "Language codes to prioritize for text"
          }
        ],
      }
    }
  }
```

soartec-lab:
@idiodoneo-chris 
Thank you for your comment. The solution would be to have a process to calculate the value of `explode` from the values ​​of `style` and `form`.

zeriong:
@soartec-lab 
Good point on involving!
`style`. Kept the PR scope but added a guard so default-`true` only applies when `style` is `form` or unspecified — leaves room for future non-form style handling without changing current behavior:

```ts
const style = parameterObject.style ?? 'form';

return (
  parameterObject.in === 'query' &&
  isArrayLike &&
  style === 'form' &&
  (parameterObject.explode ?? true)
);
```

Snapshots unchanged.

packages/fetch/src/index.ts
```diff
@@ -120,8 +120,13 @@ export const generateRequestFunction = (
           | undefined) ?? []
       ).some((s) => resolveSchemaRef(s, context).schema.type === 'array');
 
+    const style = parameterObject.style ?? 'form';
```
soartec-lab: Note that the default value may change depending on the type of parameter.

> Describes how the parameter value will be serialized depending on the type of the parameter value. Default values (based on value of in): for "query" - "form"; for "path" - "simple"; for "header" - "simple"; for "cookie" - "form".

https://swagger.io/specification/

zeriong: @soartec-lab 
Thanks for check, please one more review!

soartec-lab:
@zeriong 
I apologize, but I'm spending too much time on this, so please give me a little more time to review it.
It appears that the AI ​​has made a change of direction midway through the process and is performing unnecessary and redundant processing.

https://github.com/orval-labs/orval/blob/master/CONTRIBUTING.md#a-note-about-ai

zeriong:
@soartec-lab

Apologies — I think I misinterpreted your earlier note about per-`in` defaults as a hint to refactor further, but rereading it, it was just clarifying the spec. The extra helpers in the latest commit are out of scope for #3231 and don't change behavior.

Shall I revert to the 2nd commit (6cb04e1) with just `explode ?? true` and the `style === 'form'` guard?

I'll also be more careful in reviewing AI-assisted output before pushing. Thanks for your patience.

packages/fetch/src/index.ts
```diff
@@ -103,38 +180,9 @@ export const generateRequestFunction = (
...
-        (schemaObject.allOf as
-          | (OpenApiSchemaObject | OpenApiReferenceObject)[]
-          | undefined) ?? []
-      ).some((s) => resolveSchemaRef(s, context).schema.type === 'array');
-
-    return (
-      parameterObject.in === 'query' && isArrayLike && parameterObject.explode
-    );
-  });
+  const explodeParameters = parameterObjects.filter((parameterObject) =>
```
soartec-lab: What is the significance of extracting this function? It just looks like the code has been moved.

packages/fetch/src/index.ts
```diff
@@ -39,6 +39,83 @@ const getStatusCodeType = (key: string): string => {
...
+    parameterObject.schema,
+    context,
+  );
+
+  if (!isArrayLikeSchema(schemaObject, context)) {
+    return false;
+  }
+
+  const style = parameterObject.style ?? getDefaultStyle(parameterObject.in);
+  // Only `form` style supports explode semantics in a way orval currently
```
soartec-lab: / Only `form` style supports explode semantics in a way orval currently
  // emits (repeated `key=value` pairs). Other styles (spaceDelimited,
  // pipeDelimited, deepObject) are intentionally not exploded here.

packages/fetch/src/index.ts
```diff
@@ -39,6 +39,83 @@ const getStatusCodeType = (key: string): string => {
...
+    }
+    default: {
+      return 'form';
+    }
+  }
+};
+
+// Per OpenAPI 3.0/3.1, the default `explode` is `true` when `style` is `form`,
+// otherwise `false`. https://swagger.io/specification/
+const getDefaultExplode = (style: string): boolean => style === 'form';
```
soartec-lab: Similarly, what is the significance of extracting this variable?

packages/fetch/src/index.ts
```diff
@@ -39,6 +39,83 @@ const getStatusCodeType = (key: string): string => {
   return key;
 };
 
+// Per OpenAPI 3.0/3.1, the default `style` depends on `in`:
+// query, cookie -> "form"; path, header -> "simple".
+// https://swagger.io/specification/
+const getDefaultStyle = (parameterIn: string | undefined): string => {
```
soartec-lab: It seems unlikely that the default value will be reached. I thought a ternary operator would suffice.

zeriong: You're right. getDefaultStyle is only ever called from shouldExplodeArrayQueryParameter, which already returns early when in !== 'query'. So it's always called with 'query' and always returns 'form' — the cookie/path/header/default branches are unreachable here. I'll drop the switch entirely and just default style to 'form'.

packages/fetch/src/index.ts
zeriong: Agreed. By the time this is reached, style === 'form' is already guaranteed by the guard above, so getDefaultExplode(style) always returns true. The extraction adds nothing — I'll inline it to parameterObject.explode ?? true.

packages/fetch/src/index.ts
zeriong: Fair point — it's a single-use extraction that just relocates the original inline filter without reducing complexity. I'll fold the logic back into the .filter() callback.

packages/fetch/src/index.ts
zeriong: @soartec-lab
I think the annotation is simply copied, is there any other meaning?

zeriong:
@soartec-lab
Thanks for the thorough review — you're right on all four points, and I want to be transparent about how this happened. After you cited the spec's per-in style/explode defaults on May 11, I read it as a request to encode that full table explicitly, and added getDefaultStyle / getDefaultExplode / a named predicate to do so. The intent was spec fidelity, but I missed that this code path is query- and form-only by construction (orval's explodeParameters mechanism only emits repeated key=value, which is form-style query behavior). So the per-in and non-form branches I added are not just unused — they're unreachable and untestable through the actual generation path, which makes them a liability rather than future-proofing. I'll collapse these back to the minimal form from 6cb04e1 (inline array-like check, a style ?? 'form' / style === 'form' guard, and explode ?? true); behavior and snapshots are identical. I'll keep isArrayLikeSchema since it dedups the three repeated oneOf/anyOf/allOf blocks, but happy to inline that too if you'd prefer the PR to stay strictly at 6cb04e1.

packages/fetch/src/index.ts
soartec-lab: If that's what you think, then stop extracting functions and revert to the original state.

packages/fetch/src/index.ts
soartec-lab: Is this comment a copy of a comment already posted elsewhere?
If so, please delete both the original comment and the copied comment.

soartec-lab (commented):
That's great. However, there are still some issues that need to be addressed, so please check them.

zeriong:
@soartec-lab 
Thanks for the follow-up. Confirming the plan: I'll revert the function extractions back to the original (6cb04e1) state per your comment on line 183, and remove the duplicate/redundant comment block per your comment on line 109. I'll push the revert, re-run the snapshot tests, and request another review.

zeriong:
@soartec-lab
Work is done! please one more review

melloware:
Build is failing?

zeriong:
@melloware 
I made a mistake! I'll fix it right away.

melloware (approved):

PR #3319, opened 2026-05-05
fix: order exports with 'en' locale to prevent different ordering in different environments

### LINKED ISSUE #3317: Exports in index.ts ordering is not deterministic
### Description

Exports ordering depends on OS locale, causing noisy index.ts file in teams with different locales. The issue seems to be on line 438 in packages/core/src/writes/schemas.ts, which calls localeCompare without specifying a locale. In our example, this causes 'aa' to be interpreted as 'å' in environments with Norwegian Windows, while it's interpreted as 'aa' in English.
In order to reproduce, generate a client in a Norwegian and a non-Norwegian (e.g. US english) environments and compare the output, specifically index.ts

### Output client

react-query

```ts
module.exports = {
  api: {
    output: {
      mode: 'tags-split',
      target: 'src/api/endpoints',
      schemas: 'src/api/model',
      client: 'react-query', // Generates useQuery/useMutation hooks
      clean: true
    },
    input: {
      target: '../orval-demo.json',
    },
  },
};
```

### Environment

  System:
    OS: Windows 11 10.0.26100
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
    Memory: 1.82 GB / 15.73 GB
  npmPackages:
    @tanstack/react-query: ^5.100.6 => 5.100.6
    orval: ^8.9.0 => 8.9.1
    react: ^19.2.5 => 19.2.5

### Expected behavior

index.ts should be exactly the same, regardless of the OS locale

### Actual behavior

'aa' is interpreted as 'å' and moved after 'z' alphabetically in Norwegian environments. The same presumably happens for 'oe' -> 'ø' and 'ae' -> 'æ'

```yaml
{
  "openapi": "3.1.1",
  "info": {
    "title": "orvalDemo | v1",
    "version": "1.0.0"
  },
  "paths": {
    "/api/aa": {
      "get": {
        "responses": {
          "200": {
            "description": "OK",
            "content": {
              "text/plain": {
                "schema": {
                  "type": "array",
                  "items": {
                    "$ref": "#/components/schemas/aaModel"
                  }
                }
              },
              "application/json": {
                "schema": {
                  "type": "array",
                  "items": {
                    "$ref": "#/components/schemas/aaModel"
                  }
                }
              },
              "text/json": {
                "schema": {
                  "type": "array",
                  "items": {
                    "$ref": "#/components/schemas/aaModel"
                  }
                }
              }
            }
          },
          "401": {
            "description": "Unauthorized",
            "content": {
              "text/plain": {
                "schema": {
                  "$ref": "#/components/schemas/ProblemDetails"
                }
              },
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/ProblemDetails"
                }
              },
              "text/json": {
                "schema": {
                  "$ref": "#/components/schemas/ProblemDetails"
                }
              }
            }
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "aaModel": {
        "type": "integer"
      },
      "ProblemDetails": {
        "type": "object",
        "properties": {
          "type": {
            "type": [
              "null",
              "string"
            ]
          },
          "title": {
            "type": [
              "null",
              "string"
            ]
          },
          "status": {
            "type": [
              "null",
              "integer"
            ],
            "format": "int32"
          },
          "detail": {
            "type": [
              "null",
              "string"
            ]
          },
          "instance": {
            "type": [
              "null",
              "string"
            ]
          }
        }
      }
    }
  },
  "tags": [
  ]
}
```

### Additional context

_No response_

computaserik:
Fix #3317

localeCompare produces different orders in different environments in certain cases. For example, 'aa' is ordered as 'å' in Norwegian environments, but not in e.g. English. Using an explicit locale produces consistent ordering in different environments

## Summary by CodeRabbit

* **Chores**
  * Export ordering for schema names is now locale-aware and case-insensitive, producing a more consistent, deterministic order when names differ only by case or diacritics. This reduces ordering surprises in generated index lists and improves predictability for downstream tooling and consumers.

### DIFF
--- a/packages/core/src/writers/schemas.ts
+++ b/packages/core/src/writers/schemas.ts
@@ -435,7 +435,7 @@ export async function writeSchemas({
       // Create export statements
       const currentExports = uniqueSchemaNames
         .map((schemaName) => `export * from './${schemaName}${ext}';`)
-        .toSorted((a, b) => a.localeCompare(b));
+        .toSorted((a, b) => a.localeCompare(b, 'en', { sensitivity: 'base' }));
 
       const exports = currentExports.join('\n');
 
--- a/packages/core/src/writers/schemas.ts
+++ b/packages/core/src/writers/schemas.ts
@@ -435,7 +435,7 @@ export async function writeSchemas({
       // Create export statements
       const currentExports = uniqueSchemaNames
         .map((schemaName) => `export * from './${schemaName}${ext}';`)
-        .toSorted((a, b) => a.localeCompare(b, 'en', { sensitivity: 'base' }));
+        .toSorted((a, b) => a.localeCompare(b, 'en'));
 
       const exports = currentExports.join('\n');
 


melloware:
Looking at all the other places we do localeCompare i wonder if we should make them all consistent. Its all about alpha ordering imports, schemas, objects etc?

PR #3325, opened 2026-05-06
fix(query,swr): use null-check for auto-generated `enabled` option (#3241)

### LINKED ISSUE #3241: 0 as param causes query to be disabled
### Description

Image

If this value is `0` the query does not run

### Output client

angular-query

should not be relevant to this issue

### Environment

should not be relevant to this issue

### Expected behavior

_No response_

### Actual behavior

_No response_

should not be relevant to this issue

### Additional context

orval 8.8.0

zeriong:
## Summary

- Auto-generated `enabled` option used `!!(param)`, which treats falsy-but-valid values like `0`, `""`, and `false` as "missing" and silently disables the query. This is the root cause of #3241 (a `0` path/query parameter never fires the request).
- Switch to `param != null` so only `null` / `undefined` short-circuit the query, matching the standard TanStack Query dependent-query pattern.
- Applies to every TanStack Query client (react / vue / svelte / solid / angular) via the shared default adapter and the Vue adapter's `computed(() => unref(p) != null)` form.
- @orval/swr had the same bug in its `isEnabled = swrOptions?.enabled !== false && !!(...)` derivation; fixed in the same PR for consistency (issue comment: "this probably affects all tanstack query clients").
- Docs guides (react / vue / svelte / solid / swr) updated to mirror the new generated output.

## Why `!= null` is the only correct choice

OpenAPI parameters can legitimately be `0`, `""`, or `false`. The "value is missing" condition that should disable a dependent query is strictly `null` or `undefined` — nothing else. The fix has to express exactly that, and `!= null` is the only single expression that does it without losing information:

| Input | `!!(param)` (before) | `param != null` (after) | Correct? |
| --- | --- | --- | --- |
| `undefined` | `false` (disable) | `false` (disable) | ✅ |
| `null` | `false` (disable) | `false` (disable) | ✅ |
| `0` | `false` (disable) | `true` (enable) | ✅ (was bug) |
| `""` | `false` (disable) | `true` (enable) | ✅ (was bug) |
| `false` | `false` (disable) | `true` (enable) | ✅ (was bug) |
| any other value | `true` | `true` | ✅ |

Per the ECMAScript abstract equality rules, `x != null` is exactly equivalent to `x !== null && x !== undefined` (because `null == undefined` is the only cross-type loose-equality case in the spec). It is the canonical idiom for "is this value present?" and the only legitimate use of `==` / `!=`.

### Alternatives considered and rejected

- `!!(param)` — current behaviour. Conflates "absent" with "falsy". Breaks `0`, `""`, `false`. This is the bug.
- `param !== undefined` — misses the `null` case. OpenAPI/JS treat `null` as a valid "explicit absence" form, so a `null` path param would incorrectly enable the query.
- `typeof param !== 'undefined'` — verbose, still misses `null`, no upside.
- Type-aware generation (emit `!!` for strings, `!== undefined` for numbers, etc.) — `GetterParam` only carries `{ name, required, definition, ... }` and no resolved schema type (`packages/core/src/types.ts`). Threading schema type through the query generator would be a much larger architectural change for a one-line semantic fix, and it would still need `!= null` for `null`-able params. Rejected on DX/scope grounds.

`!= null` is therefore the minimal, type-agnostic, semantically exact fix.

## Before / After

Generated output (react / svelte / solid / angular query):

```diff
- enabled: !!petId,
+ enabled: petId != null,
```

```diff
- enabled: !!(version && petId),
+ enabled: version != null && petId != null,
```

Vue adapter (reactive `computed`):

```diff
- enabled: computed(() => !!(unref(petId))),
+ enabled: computed(() => unref(petId) != null),
```

SWR `isEnabled` derivation:

```diff
- const isEnabled = swrOptions?.enabled !== false && !!(petId);
+ const isEnabled = swrOptions?.enabled !== false && petId != null;
```

### Concrete user-facing scenario (the bug from #3241)

```ts
// petId: number — a perfectly valid id of 0
const [petId, setPetId] = useState<number | null>(null);

useGetPet(petId);   // null  -> disabled (correct, both before and after)
setPetId(0);        // 0     -> BEFORE: still disabled (bug). AFTER: fires the request.
setPetId(42);       // 42    -> enabled (unchanged)
```

`undefined` (e.g. a route param that hasn't resolved yet) keeps disabling the query as before, so existing dependent-query setups are unaffected:

```ts
const { id } = useParams<{ id?: string }>();
useGetUser(id); // id === undefined -> still disabled (no behaviour change)
```

## Test plan

- [x] New unit tests in `packages/query/src/query-options.test.ts` cover single param, multi param, the Vue `computed(() => unref(p) != null)` form, and the user-supplied-`enabled` opt-out path.
- [x] `bun vitest run` — query + swr suites green.
- [x] `bun run update-samples` + `bun run test:snapshots:update` regenerated; `bun run test:snapshots` green.
- [x] `bun run typecheck` and `bun run lint` green.
- [x] Manually verified generated output for react-query, angular-query, vue-query, svelte-query, and swr snapshots.

Fixes #3241.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened query enablement across frameworks to use explicit null/undefined checks so falsy-but-valid values (e.g., 0, empty string) no longer incorrectly disable queries.

* **Documentation**
  * Updated framework guides and generated examples (React Query, Vue, Svelte, Angular, SWR) to show the null-safe enablement pattern.

* **Tests**
  * Added tests covering query enablement behavior for various parameter scenarios.

### DIFF
--- a/docs/content/docs/guides/react-query.mdx
+++ b/docs/content/docs/guides/react-query.mdx
@@ -60,7 +60,7 @@ export const useShowPetById = <
   const query = useQuery<AsyncReturnType<typeof queryFn>, TError, TData>(
     queryKey,
     queryFn,
-    { enabled: !!petId, ...queryOptions },
+    { enabled: petId != null, ...queryOptions },
   );
 
   return {
--- a/docs/content/docs/guides/react-query.mdx
+++ b/docs/content/docs/guides/react-query.mdx
@@ -60,7 +60,10 @@ export const useShowPetById = <
   const query = useQuery<AsyncReturnType<typeof queryFn>, TError, TData>(
     queryKey,
     queryFn,
-    { enabled: petId != null, ...queryOptions },
+    {
+      enabled: petId !== null && petId !== undefined,
+      ...queryOptions,
+    },
   );
 
   return {
--- a/docs/content/docs/guides/solid-query.mdx
+++ b/docs/content/docs/guides/solid-query.mdx
@@ -65,7 +65,7 @@ export const createShowPetById = <
     () => ({
       queryKey,
       queryFn,
-      enabled: !!petId,
+      enabled: petId != null,
       ...queryOptions,
     }),
   );
--- a/docs/content/docs/guides/solid-query.mdx
+++ b/docs/content/docs/guides/solid-query.mdx
@@ -65,7 +65,7 @@ export const createShowPetById = <
     () => ({
       queryKey,
       queryFn,
-      enabled: petId != null,
+      enabled: petId !== null && petId !== undefined,
       ...queryOptions,
     }),
   );
--- a/docs/content/docs/guides/svelte-query.mdx
+++ b/docs/content/docs/guides/svelte-query.mdx
@@ -60,7 +60,7 @@ export const useShowPetById = <
   const query = useQuery<AsyncReturnType<typeof queryFn>, TError, TData>(
     queryKey,
     queryFn,
-    { enabled: !!petId, ...queryOptions },
+    { enabled: petId != null, ...queryOptions },
   );
 
   return {
--- a/docs/content/docs/guides/svelte-query.mdx
+++ b/docs/content/docs/guides/svelte-query.mdx
@@ -60,7 +60,10 @@ export const useShowPetById = <
   const query = useQuery<AsyncReturnType<typeof queryFn>, TError, TData>(
     queryKey,
     queryFn,
-    { enabled: petId != null, ...queryOptions },
+    {
+      enabled: petId !== null && petId !== undefined,
+      ...queryOptions,
+    },
   );
 
   return {
--- a/docs/content/docs/guides/swr.mdx
+++ b/docs/content/docs/guides/swr.mdx
@@ -99,7 +99,7 @@ Disable the hook conditionally:
 ```tsx
 const { data } = useShowPetById(petId, {
   swr: {
-    enabled: !!petId, // Only fetch when petId is defined
+    enabled: petId != null, // Only fetch when petId is defined
   },
 });
 ```
--- a/docs/content/docs/guides/swr.mdx
+++ b/docs/content/docs/guides/swr.mdx
@@ -54,7 +54,10 @@ export const useShowPetById = <TError = Error>(
 ) => {
   const { swr: swrOptions, axios: axiosOptions } = options ?? {};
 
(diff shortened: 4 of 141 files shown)


melloware:
@zeriong i was thinking it needs to be more than just `!= null` i think we want to handle undefined etc as well but treat 0 NOT as falsy. Here is what I have done in other projects

```ts
static isEmpty(value) {
        return value === null || value === undefined || value === '' || (Array.isArray(value) && value.length === 0) || (!(value instanceof Date) && typeof value === 'object' && Object.keys(value).length === 0);
}

static isNotEmpty(value) {
        return !this.isEmpty(value);
}
```

zeriong:
@melloware
Thanks for the feedback!
Before extending the PR I'd like to confirm the intended semantics, since the current PR description explicitly
treats `''`, `false`, and `0` as "valid → enable".

Your `isEmpty` helper would additionally disable `''`, `[]`, and `{}`,
which is a semantic change beyond the original #3241 scope (where the
reported bug is specifically `0` being silently disabled). A few options:

1. **Keep this PR as-is** (`param != null`) — keeps the existing logic
   and shape of the generated output, ships as a minimal regression
   fix for #3241 with all CI checks (build / typecheck / lint / unit
   tests / snapshot verification) green. Matches TanStack's
   dependent-query idiom (`id != null` / `Boolean(id)`). No follow-up
   churn for users already on this branch.
2. **Introduce a generated `isQueryParamReady` helper** mirroring your
   `isEmpty`, and emit `enabled: isQueryParamReady(param)` in the
   generated output. This requires:
   - deciding where the helper lives (inline per-file vs. shared
     import vs. mutator-style injection),
   - updating the PR description table, the docs guides
     (react/vue/svelte/solid/swr), and regenerating all 130+
     snapshots.
3. **Same as 2, but as a follow-up PR** so this one stays a clean
   regression fix for the `0`-as-falsy bug.

Which direction do you prefer? Happy to implement (2) or (3) once
the desired semantics is confirmed.

melloware:
yeah maybe we just do `value === null || value === undefined` ?

zeriong:
@melloware
Sounds good, I'll go with that.
Pushing the changes(generator + Vue/SWR variants + docs + regenerated snapshots)

zeriong:
@melloware
Work is done!

melloware (approved):

PR #3361, opened 2026-05-15
fix: file extension is not added to imported files when using NodeNext moduleResolution

### LINKED ISSUE #2284: File extension is not added to imported files when using `NodeNext` moduleResolution
`tsconfig.json`:

```json
{
  "compilerOptions": {
    "target": "ESNext",
    "module": "NodeNext",
    "allowImportingTsExtensions": true,
  }
}
```
`orval.config.ts`:
```typescript
import type { OptionsExport, Tsconfig } from 'orval';
import path from 'path';
import tsconfig from '#tsconfig.json' with { type: 'json' };

export default {
  input: {
    target: 'foo.yaml',
  },
  output: {
    fileExtension: '.gen.ts',
    indexFiles: false,
    mode: 'tags',
    override: {
      mutator: {
        extension: '.ts',
        name: 'foo',
        path: 'foo.ts',
      },
    },
    schemas: 'models',
    target: 'endpoints',
    tsconfig: tsconfig as Tsconfig,
    workspace: process.cwd(),
  },
} satisfies OptionsExport;
```

With such config orval generates correct import of my mutator:

```typescript
import { foo } from 'foo.ts';
```

However model imports still uses extensionless syntax:

```typescript
import type { FooResponse } from '../models/fooResponse';
```

Is it possible for orval to respect more options from `tsconfig.json`, and use `.js` in case just `moduleResolution` (`module`) is set to `ESNext`, and `.ts` in case both `moduleResolution` set to `ESNext` and `allowImportingTsExtensions` set to `true`? (or maybe just use the provided type similar to what was done in `mutator` case)

rkday-pro:
Fixes https://github.com/orval-labs/orval/issues/2284.

This basically takes the suggestion from that issue and runs with it:

> Is it possible for orval to respect more options from tsconfig.json, and use .js in case just moduleResolution (module) is set to ESNext, and .ts in case both moduleResolution set to ESNext and allowImportingTsExtensions set to true? 

- `interface Tsconfig` now includes the relevant options
- `getImportExtension` no longer just strips .ts - it takes the tsconfig as an argument and adds .js when those options are set
- the `tsconfig` has to be plumbed through to more places, but I think that's unavoidable

I've run this in my project, where the `orval`-generated files previously failed typechecking with `error TS2835: Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Did you mean './datapoint.js'?`, and they now pass.

## Summary by CodeRabbit

* **New Features**
  * Generated imports now respect additional TypeScript compiler options (module, moduleResolution, allowImportingTsExtensions) when resolving emitted extensions.
  * Import path extensions are computed consistently and can preserve .ts-style extensions when allowed.
  * Schema generation and index exports now emit imports using the configured extension logic.

* **Tests**
  * Added tests covering import/export suffix behavior across TypeScript configuration scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/orval-labs/orval/pull/3361)

### DIFF
--- a/packages/core/src/generators/imports.ts
+++ b/packages/core/src/generators/imports.ts
@@ -13,11 +13,13 @@ import { escapeRegExp } from '../utils/string';
 interface GenerateImportsOptions {
   imports: readonly GeneratorImport[];
   namingConvention?: NamingConvention;
+  importExtension?: string;
 }
 
 export function generateImports({
   imports,
   namingConvention = NamingConvention.CAMEL_CASE,
+  importExtension = '',
 }: GenerateImportsOptions) {
   if (imports.length === 0) {
     return '';
@@ -37,7 +39,8 @@ export function generateImports({
   ).map((imp) => ({
     ...imp,
     importPath:
-      imp.importPath ?? `./${conventionName(imp.name, namingConvention)}`,
+      imp.importPath ??
+      `./${conventionName(imp.name, namingConvention)}${importExtension}`,
   }));
 
   const grouped = groupBy(normalized, (imp) =>
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -990,6 +990,9 @@ export interface Tsconfig {
     exactOptionalPropertyTypes?: boolean;
     paths?: Record<string, string[]>;
     target?: TsConfigTarget;
+    module?: string;
+    moduleResolution?: string;
+    allowImportingTsExtensions?: boolean;
   };
 }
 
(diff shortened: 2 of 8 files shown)


melloware (changes requested):
Build is failing

melloware (approved):

mauriceackel:
@rkday-pro, thank you so much for this PR. I just realized that there is one place missing for change to take action and this is when writing zod specs. In particular, [this line](https://github.com/orval-labs/orval/blob/b059ca64e239ad5551a17546d14a35c74c66b844/packages/orval/src/write-zod-specs.ts#L186) needs the changes to be applied too.

Any chance this can be added? I'd assume this could be labeled as a bug.

CC: @melloware

PR #3362, opened 2026-05-16
feat: add throwOnError option to generate function

### LINKED ISSUE #2290: Expose throwing `generate` function from `orval` package
When trying to generate with orval using it's exposed `generate` function, we get the same functionality as we get from orval tool itself. It's not always desirable, especially for error handling. 

Is it possible to expose lower level functions or some wrapper around `generateSpec` that still throws errors, that we can handle later manually? For example something like this API:

```typescript
import { generate, logError } from 'orval';
import orvalConfig from './orval.config.ts';

try {
  await generate(orvalConfig);
} catch (error) {
  logError(error);
  // handle appropriately
}
```

chu0119:
## Summary
Added `throwOnError` option to the `generate` function that allows callers to opt into error-throwing behavior.

## Why
When using orval programmatically, errors are caught internally and only logged, making it difficult for callers to implement custom error handling.

## What changed
- Added `throwOnError?: boolean` to `GlobalOptions`
- Updated three `try/catch` blocks in `generate.ts` to re-throw when enabled

Closes #2290

## Release Notes

* **New Features**
  * Added optional `throwOnError` configuration setting to control error handling behavior during generation operations. When enabled, errors are re-thrown for improved error visibility and control.

[Review Change Stack](https://app.coderabbit.ai/change-stack/orval-labs/orval/pull/3362?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -980,6 +980,7 @@ export interface GlobalOptions {
   input?: string | string[];
   output?: string;
   failOnWarnings?: boolean;
+  throwOnError?: boolean;
 }
 
 export interface Tsconfig {
--- a/packages/orval/src/generate.ts
+++ b/packages/orval/src/generate.ts
@@ -39,6 +39,9 @@ export async function generate(
         await generateSpec(workspace, normalizedOptions, projectName);
       } catch (error) {
         hasErrors = true;
+        if (options?.throwOnError) {
+          throw error;
+        }
         logError(error, projectName);
       }
 
@@ -88,6 +91,9 @@ export async function generate(
   try {
     await generateSpec(workspace, normalizedOptions);
   } catch (error) {
+    if (options?.throwOnError) {
+      throw error;
+    }
     logError(error);
   }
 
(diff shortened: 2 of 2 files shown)


melloware (changes requested):
Build is failing and documentation needed based on your other ticket

ded-furby (commented):
One behavior worth documenting in this PR: when `generate()` loads a config file with multiple named projects, `throwOnError` now fails fast on the first `generateSpec` error and exits the `for...of` loop before later projects run. That is probably the right tradeoff for programmatic callers, but it is a semantic change from the current "log and continue" behavior, not just a different error surface.

I’d call that out explicitly in `docs/content/docs/reference/integration.mdx` alongside the new option example so consumers know enabling `throwOnError` changes multi-project runs from best-effort to fail-fast.

melloware:
@ded-furby any chance you want to resubmit this PR with the documentation change? It seems the OP has let this branch go dead...

melloware:
Superseded by another PR

Issue #398, opened 2022-05-18
References with tokens that contains placeholders cannot be found


bug-brain:
## What are the steps to reproduce this issue?
Try to generate code for this schema.
```yaml
openapi: 3.0.3
info: {}
paths:
  /{id}:
    get:
      parameters:
        - in: path
          name: id
          schema:
            type: string
      responses:
        200:
          content:
            application/json:
              schema:
                $ref: "#/paths/~1%7Bid%7D/get/parameters/0/schema"
```

## What happens?
Orval fails because the reference cannot be resolved.

## What were you expecting to happen?
The reference to be resolved and code to be generated.

## Any logs, error output, etc?
```
🛑  <name> - Oups... 🍻. Ref not found: #/paths/~1%7Bid%7D/get/parameters/0/schema
```

## Any other comments?
You get this kind of reference when [bundling](https://apitools.dev/json-schema-ref-parser/docs/ref-parser.html#bundleschema-options-callback) a schema. Maybe the problem occurs with escaped characters in general.

## What versions are you using?

**Package Version:** 6.7.1

PR #3382, opened 2026-05-18
fix(orval): decode escaped JSON Pointer tokens in external $refs (#3380)

### LINKED ISSUE #3380: External path-item $ref fails with "Can't resolve reference" (JSON Pointer escapes not decoded)
## Description

When a Path Item Object uses a `$ref` to a path defined in **another file** (e.g. `$ref: "common.yaml#/paths/~1pets"`), orval aborts with an `INVALID_REFERENCE` validation error.

External path-item `$ref` is valid per the OpenAPI 3.x spec, and the JSON Pointer `~1pets` correctly encodes the path key `/pets`. The same problem affects external `$ref`s into any non-`components/schemas` location whose JSON Pointer contains escaped characters.

## Output client

other (not client-specific — the failure happens during spec validation, before code generation).

```js
const path = require('path');
module.exports = {
  api: {
    input: { target: path.join(__dirname, 'api.yaml') },
    output: { target: path.join(__dirname, 'gen/endpoints.ts') },
  },
};
```

## Environment

- orval: 8.11.0
- Node: 24.15.0
- OS: macOS (Darwin 24.6.0)

## Expected behavior

orval resolves the external path-item `$ref` and generates the client, treating `common.yaml`'s `/pets` as if it were inline.

## Actual behavior

```
🛑 api - OpenAPI spec validation failed:
[
  {
    "code": "INVALID_REFERENCE",
    "message": "Can't resolve reference: #/x-ext/829a1f2/paths/~1pets"
  }
]
```

## OpenAPI document (minimal)

`api.yaml`:

```yaml
openapi: "3.0.0"
info:
  title: API
  version: 1.0.0
paths:
  /pets:
    $ref: "common.yaml#/paths/~1pets"
components: {}
```

`common.yaml`:

```yaml
openapi: "3.0.0"
info:
  title: Common
  version: 1.0.0
paths:
  /pets:
    get:
      operationId: listPets
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                type: array
                items:
                  type: string
components: {}
```

## Additional context (root cause analysis)

`@scalar/json-magic`'s bundler does not dereference `$ref`s; it fetches external files under `x-ext` and rewrites refs to `#/x-ext/<key>/...`. `dereferenceExternalRef` → `replaceXExtRefs` in `packages/orval/src/import-specs.ts` then resolves those.

For non-`components/schemas` refs, `replaceXExtRefs` walks the external document by splitting the pointer on `/` and indexing each segment directly (around `import-specs.ts` L405):

```ts
const parts = pathStr.split('/');   // ['paths', '~1pets']
// ...
for (const p of parts) {
  if (refObj && (isObject(refObj) || Array.isArray(refObj)) && p in refObj) {
    refObj = refObj[p];
  } else {
    refObj = undefined;
    break;
  }
}
```

The segment `~1pets` is **not JSON Pointer-decoded** (`~1` → `/`, `~0` → `~`), so `'~1pets' in doc.paths` is `false` (the real key is `/pets`). Resolution fails, the `#/x-ext/...` ref is left in the document, and validation rejects it.

Since every path key begins with `/`, **every external path-item `$ref` is affected**. Percent-encoded characters (e.g. `%7B`/`%7D` for `{}` in templated paths such as `/pets/{petId}`) need decoding as well. Component-schema refs happen to work because their pointers contain no escaped characters.

This is the external-file counterpart of #398 — the same class of "escaped tokens in `$ref` not resolved" bug. #398 was fixed for same-document bundler refs in the core resolver (`resolveValue`, PR #3355), but `import-specs.ts` / `replaceXExtRefs` was not covered.

wadakatu:
## Summary

Fixes #3380.

A cross-file path-item `$ref` (e.g. `$ref: "common.yaml#/paths/~1pets"`) aborts generation with an `INVALID_REFERENCE` validation error. External path-item `$ref` is valid OpenAPI 3.x, and the same problem affects external `$ref`s into any non-`components/schemas` location whose JSON Pointer contains escaped characters.

## Root cause

`@scalar/json-magic` bundles external files under `x-ext` and rewrites refs to `#/x-ext/<key>/...`. `replaceXExtRefs` (`packages/orval/src/import-specs.ts`) then resolves those by splitting the pointer on `/` and indexing the external document segment by segment.

The segments are used **verbatim**, without decoding:

- JSON Pointer escapes — `~1` for `/`, `~0` for `~` (RFC 6901)
- Percent-encoding — e.g. `%7B`/`%7D` for `{`/`}` in templated paths

So the segment `~1pets` never matches the real `/pets` key, the `#/x-ext/...` ref is left unresolved, and validation rejects it. Because every path key starts with `/`, every external path-item `$ref` is affected. Component-schema refs happen to work only because their pointers contain no escaped characters.

This is the external-file counterpart of #398 — #398 fixed escaped-token resolution in the core resolver (`resolveValue`, PR #3355), but the x-ext bundling path in `import-specs.ts` was not covered.

## Fix

Add `decodeRefToken`, applied to each pointer segment before walking the external document: percent-decoding first (outer layer; malformed sequences fall back to the raw token instead of throwing), then JSON Pointer unescaping (`~1` → `/`, `~0` → `~`). The change is scoped to the inline-walk branch, which covers path items and any other non-`components/schemas` external ref.

## Tests

- **Unit** (`import-specs.test.ts`) — `dereferenceExternalRef` now resolves an x-ext path-item ref containing both `~1` and `%7B`/`%7D`. Verified red before the fix, green after.
- **Integration** (`tests/specifications/issue-3380/`) — a spec whose path items are cross-file `$ref`s (`~1pets` and `~1pets~1%7BpetId%7D`); a focused assertion in `api-generation.spec.ts` confirms both operations are generated, alongside the snapshot.

All checks pass locally: `format:check`, `lint`, `typecheck`, `test`, `test:snapshots`, generated typecheck, and mock verification.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed dereferencing of external path-item references with percent-encoded and JSON Pointer-escaped tokens, restoring correct handling of templated paths, inlined operations, and removal of intermediate extension markers after resolution.

* **Tests**
  * Added regression tests, fixtures and a generated snapshot asserting cross-file external path refs, encoded/escaped pointer segments, and templated path parameter resolution (includes emitted client endpoints for listPets/getPet).

[Review Change Stack](https://app.coderabbit.ai/change-stack/orval-labs/orval/pull/3382?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


wadakatu:
Thanks for the review! Addressed the two nitpicks:

- **`~0` test coverage** — added a `/pets~dogs` case (`~1pets~0dogs`) to the unit test in 49b6cf18e, so the test now exercises all three escapes the helper handles: `~1` → `/`, `~0` → `~`, and percent-encoding.
- **JSDoc on `decodeRefToken`** — leaving as-is. `decodeRefToken` already carries a JSDoc block documenting the two-layer decoding contract (percent-encoding outer, JSON Pointer inner) and the malformed-input fallback. Adding `@param`/`@returns` tags would be inconsistent with the sibling functions in `import-specs.ts` (`scrubUnwantedKeys`, `updateInternalRefs`, `replaceXExtRefs`, `dereferenceExternalRef`), which all use prose-only JSDoc.

melloware (approved):

PR #3384, opened 2026-05-18
fix(fetch): type error caused by useDates date param string coerce

ivoberger:
## Problem

With `useDates: true` and a query params interface where every field is typed `Date` (every param is `format: date-time`), the generated fetch URL builder fails strict TypeScript compilation. Repro spec at `tests/specifications/usedates-only-date-params.yaml`.

We currently have a post-process script to fix this but seems worth getting a fix upstream.

The actual changes are in this commit: https://github.com/orval-labs/orval/pull/3384/changes/9dcf070da96b3909e0d0224829d53cd799cdb18d
The rest are just snapshot updates

## Cause

URL builder emits `value === null ? 'null' : value instanceof Date ? value.toISOString() : value.toString()`.
After the `value !== undefined` guard, `value === null` (impossible for `Date`), and `value instanceof Date`, the else branch is narrowed to type `never`, which does not have a `toString` function and thuse causes a type error.

## Fix

Swap the non-Date fallback `value.toString()` → `String(value)` (and `v.toString()` → `String(v)`) in `packages/fetch/src/index.ts` and `packages/solid-start/src/index.ts`.
`String()`  has no problem with `never` and does the same thing at runtime.

## Tests

- New fixture `tests/specifications/usedates-only-date-params.yaml` + fetch config entry, generated to `tests/generated/fetch/usedates-only-date-params/`.
- New regression test in `tests/api-generation.spec.ts` mirrors pattern of existing ones, asserts the fallback is `String(value)` and not `value.toString()`.
- `bun --filter orval-tests build` (strict tsc over generated output) passes all 15 clients.

## Summary by CodeRabbit

* **Bug Fixes**
  * Safer query-parameter stringification in generated clients: non-null values are coerced reliably and nulls remain serialized as "null", preventing runtime errors.

* **New Features**
  * Added a fetch client fixture that supports date-time query parameters (useDates).

* **Tests**
  * Added tests covering date-time query parameter serialization behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/orval-labs/orval/pull/3384?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ivoberger:
fixed that failing CI job. not sure how I missed those on the first try but should be passing now

ivoberger:
humm I don't know what I'm doing wrong.
I ran `bun run test:snapshots:update` and then `bun run test:snapshots` and thats all green for me locally. But CI is still unhappy

melloware:
did you also run `bun run update-samples` ?

ivoberger:
okay, I did a `nuke:all` and then re-ran the snapshot update and the samples update, I think that did it

melloware (approved):

PR #3388, opened 2026-05-19
test(query): add regression coverage for react-query onMutate option (#1177)

### LINKED ISSUE #1177: React Query: onMutate is missing from generated mutateOptions

1. Have OpenAPI with any endpoint that would generate mutations.
2. Try to pass onMutate callback function

## What happens?
 Only onSuccess/onError/onSettled is there. But no onMutate.

I want to have onMutate option too. [Example usage.](https://tanstack.com/query/v5/docs/react/guides/optimistic-updates#updating-a-list-of-todos-when-adding-a-new-todo)

No

## Any other comments?
No
…

6.23.0

**Operating System:** Linux Mint 21.1 Cinnamon
**Package Version:** 6.23.0
**Browser Version:** 120.0.6099.129

wadakatu:
## Goals

Issue #1177 reported that `onMutate` was missing from the options of generated TanStack Query mutation hooks. On current orval this is **already fixed** — the generated `useXxx` hook exposes its `mutation` option as `UseMutationOptions<...>`, which (extending TanStack's `MutationObserverOptions`) includes `onMutate`. However, there was no regression test guarding it: `query-options.test.ts` only string-matches the `UseMutationOptions<` type name, which would still pass if the type were narrowed (e.g. `Omit<UseMutationOptions, 'onMutate'>`).

This PR adds a real, compile-time regression test:

- **`tests/regressions/react-query.ts`** — adds a consumption test that calls a generated mutation hook (`useCreatePets`) with `mutation.onMutate`. If orval ever narrows the mutation option type (as it did when #1177 was filed), this file stops compiling. A `@ts-expect-error` line also asserts the `onMutate` `variables` argument stays strongly typed.
- **`tests/scripts/typecheck-generated.mjs`** — wires the existing `tests/regressions` directory into the per-client `tsconfig` `include`. This directory already held hand-written files importing generated code, but was not referenced by any `tsconfig`, so it was never type-checked. It is now checked in CI alongside `generated/` and `mutators/`.

Verified that removing `onMutate` from the generated mutation option type makes `bun run --filter orval-tests build` fail in `tests/regressions/react-query.ts`, and that all 15 generated clients type-check with the change in place.

Closes #1177

## Summary by CodeRabbit

* **Tests**
  * Enhanced TypeScript regression testing for generated React Query mutation hooks to validate callback support and type safety. Expanded test coverage to include the `regressions` directory for comprehensive compile-time validation of generated types.

[Review Change Stack](https://app.coderabbit.ai/change-stack/orval-labs/orval/pull/3388?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tests/scripts/typecheck-generated.mjs
+++ b/tests/scripts/typecheck-generated.mjs
@@ -34,7 +34,10 @@ for (const folder of folders) {
 
   const config = {
     extends: './tsconfig.json',
-    include: [`generated/${folder}`, 'mutators'],
+    // `regressions` holds hand-written compile-time tests that import generated
+    // code and exercise its public types (e.g. the #1177 onMutate regression),
+    // so a narrowed option type fails the typecheck.
+    include: [`generated/${folder}`, 'mutators', 'regressions'],
   };
 
   // Bun's flat node_modules makes the MCP SDK resolve `zod` to the project's v3.25
--- a/tests/regressions/react-query.ts
+++ b/tests/regressions/react-query.ts
@@ -1,6 +1,7 @@
 import { keepPreviousData } from '@tanstack/react-query';
 
 import {
+  useCreatePets,
   useListPets,
   useListPetsInfinite,
 } from '../generated/react-query/mutator/endpoints';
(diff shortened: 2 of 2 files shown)


tests/scripts/typecheck-generated.mjs
```diff
@@ -34,7 +34,10 @@ for (const folder of folders) {
 
   const config = {
     extends: './tsconfig.json',
-    include: [`generated/${folder}`, 'mutators'],
+    // `regressions` holds hand-written compile-time tests that import generated
+    // code and exercise its public types (e.g. the #1177 onMutate regression),
+    // so a narrowed option type fails the typecheck.
+    include: [`generated/${folder}`, 'mutators', 'regressions'],
```
wadakatu, line 40: @melloware 
I wired the previously-orphaned tests/regressions/ directory into the per-client typecheck so the #1177 test actually runs in CI. This compiles every regressions/ file under every client's tsconfig, mirroring how mutators is already shared. An alternative would be a dedicated, narrowly-scoped directory just for these compile-time type tests. I went with reusing regressions/ since it already exists for exactly this purpose, but I'm happy to switch approaches if you'd prefer — let me know.

melloware (approved):

PR #3393, opened 2026-05-19
fix(core): handle multi-line descriptions in JSDoc generation

### LINKED ISSUE #3392: Multi-line `info.description` breaks JSDoc header formatting in generated files
## Problem

When an OpenAPI spec's `info.description` contains newlines (e.g. YAML folded `>` or literal `|` block scalars), the generated file header produces a malformed JSDoc comment. Lines after the first description line are missing the ` * ` prefix:

```ts
/**
 * Generated by orval v8.11.0 🍺
 * Do not edit manually.
 * Swagger Petstore - OpenAPI 3.0
 * This is a sample Pet Store Server based on the OpenAPI 3.0 specification.  You can find out more about
Swagger at [https://swagger.io](https://swagger.io). In the third iteration of the pet store, we've switched to the design first approach!
You can now help us improve the API whether it's by making changes to the definition itself or to the code.
That way, with time, we can improve the API in general, and expose some of the new features in OAS3.

Some useful links:
- [The Pet Store repository](https://github.com/swagger-api/swagger-petstore)
- [The source API definition for the Pet Store](https://github.com/swagger-api/swagger-petstore/blob/master/src/main/resources/openapi.yaml)
 * OpenAPI spec version: 1.0.27-SNAPSHOT
 */
```

Note lines 5–10 have no ` * ` prefix — they appear as bare text inside the `/** ... */` block.

## Root Cause

The header is assembled in `getDefaultFilesHeader()` (`packages/orval/src/utils/options.ts:1052`), which puts the raw `description` string as a single array element. That array is then passed to `jsDoc()` (`packages/core/src/utils/doc.ts:67`), which joins the `description` lines with `\n * `:

```ts
// doc.ts:136
doc += ` ${lines.join('\n * ')}`;
```

However, if an individual `lines` entry itself contains `\n` characters (from a multi-line YAML description), those inner newlines are never prefixed with ` * `. The `join` only adds the prefix between *array elements*, not within them.

## Expected Behavior

Each line of the description should be prefixed with ` * ` to produce a well-formed JSDoc block:

```ts
/**
 * Generated by orval v8.11.0 🍺
 * Do not edit manually.
 * Swagger Petstore - OpenAPI 3.0
 * This is a sample Pet Store Server based on the OpenAPI 3.0 specification.  You can find out more about
 * Swagger at [https://swagger.io](https://swagger.io). In the third iteration of the pet store, we've
 * switched to the design first approach!
 * You can now help us improve the API whether it's by making changes to the definition itself or to the code.
 * That way, with time, we can improve the API in general, and expose some of the new features in OAS3.
 * ...
 * OpenAPI spec version: 1.0.27-SNAPSHOT
 */
```

## Suggested Fix

Split description strings on newlines before joining. Either:

1. In `jsDoc()` (`doc.ts:97–101`), split each description string on `\n` so multi-line content becomes separate array entries, **or**
2. In `getDefaultFilesHeader()` (`options.ts:1065`), split the description on newlines before adding it to the array.

Option 1 fixes the problem for all JSDoc emission (not just file headers), while option 2 is the more targeted fix.

## Reproduction

Any spec with a multi-line `info.description` triggers this. Existing samples on `master` already demonstrate the bug:

- `samples/mcp/petstore/__snapshots__/http-schemas/pet.ts` (lines 4–11)
- `samples/mcp/custom-server/__snapshots__/http-schemas/pet.ts` (lines 4–11)

Both use the multi-line `info.description` from `samples/mcp/petstore/petstore.yaml`.

aqeelat:
## Problem

Multi-line `info.description` in OpenAPI specs (e.g. YAML folded `>` or literal `|` block scalars) produces malformed JSDoc comments where inner newlines lack the `* ` prefix.

## Root Cause

In `jsDoc()` (`packages/core/src/utils/doc.ts`), description strings containing `\n` were joined with `\n * ` between *array elements*, but newlines *within* a single element were never prefixed. The `join` only adds the prefix at element boundaries.

## Fix

Added `.flatMap((line) => line.split(/\r?\n/))` to the lines processing pipeline in `jsDoc()`, so multi-line description strings are split into individual entries before joining. The CRLF-aware regex (`/\r?\n/`) ensures correctness across platforms.

## Changes

- `packages/core/src/utils/doc.ts`: Added `.flatMap((line) => line.split(/\r?\n/))` to split multi-line strings before the escape step
- `packages/core/src/utils/doc.test.ts`: Added 4 new tests covering:
  - Single-string multi-line descriptions
  - Array multi-line descriptions
  - CRLF line endings
  - `*/` escaping in multi-line descriptions
- 359 snapshot files updated to reflect the corrected JSDoc output

Fixes #3392

### DIFF
--- a/packages/core/src/utils/doc.test.ts
+++ b/packages/core/src/utils/doc.test.ts
@@ -44,6 +44,57 @@ describe('jsDoc', () => {
 `);
   });
 
+  it('prefixes each line of a multi-line description with *', () => {
+    expect(
+      jsDoc({
+        description: 'line one\nline two\nline three',
+      }),
+    ).toBe(`/**
+ * line one
+ * line two
+ * line three
+ */
+`);
+  });
+
+  it('prefixes each line of multi-line array descriptions with *', () => {
+    expect(
+      jsDoc({
+        description: ['first\nsecond', 'third'],
+      }),
+    ).toBe(`/**
+ * first
+ * second
+ * third
+ */
+`);
+  });
+
+  it('handles CRLF line endings in description', () => {
+    expect(
+      jsDoc({
+        description: 'line one\r\nline two\r\nline three',
+      }),
+    ).toBe(`/**
+ * line one
+ * line two
+ * line three
+ */
+`);
+  });
+
+  it('escapes */ in multi-line descriptions', () => {
+    expect(
+      jsDoc({
+        description: 'line with /* comment\nline with */ end',
+      }),
+    ).toBe(`/**
+ * line with /* comment
+ * line with *\\/ end
+ */
+`);
+  });
+
   it('stops traversing circular item schemas', () => {
     interface CircularItems {
       [key: string]: unknown;
--- a/packages/core/src/utils/doc.test.ts
+++ b/packages/core/src/utils/doc.test.ts
@@ -88,9 +88,9 @@ describe('jsDoc', () => {
       jsDoc({
         description: 'line with /* comment\nline with */ end',
       }),
-    ).toBe(`/**
+    ).toBe(String.raw`/**
  * line with /* comment
- * line with *\\/ end
+ * line with *\/ end
  */
 `);
   });
--- a/packages/core/src/utils/doc.ts
+++ b/packages/core/src/utils/doc.ts
@@ -98,7 +98,9 @@ export function jsDoc(
     Array.isArray(description)
       ? description.filter((d) => !d.includes('eslint-disable'))
       : [description ?? '']
-  ).map((line) => line.replaceAll(regex, replacement));
+  )
+    .flatMap((line) => line.split(/\r?\n/))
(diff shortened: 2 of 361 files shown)


daugvinasr:
This breaks current snapshots `bun run test:snapshots`. Please run `bun run test:snapshots:update` and inspect if generated output is expected.

aqeelat:
@daugvinasr rabbit is telling that I'm failing the docstring coverage. Should this config be tuned?

Also, snapshots are updated and rabbit nits were addressed.

aqeelat:
@coderabbitai review --dir packages

snebjorn (dismissed):
Good change, thanks

Just need to address the linting error

aqeelat:
> Good change, thanks
> 
> Just need to address the linting error

fixed.

melloware (approved):

Issue #3244, opened 2026-04-17
Zod: TypeScript error for object-literal defaults


tothandras:
### Description

Zod outputs should always have `as const` on default values.

follow-up to #2427 / #2498

### Output client

zod

### Configuration (orval.config)

```ts
import { defineConfig } from 'orval'

export default defineConfig({
  openmeter: {
    input: {
      target: './openapi.yaml',
    },
    output: {
      formatter: 'biome',
      clean: true,
      client: 'zod',
      mode: 'single',
      namingConvention: 'PascalCase',
      override: {
        useDates: true,
        zod: {
          coerce: {
            body: true,
            header: false,
            param: true,
            query: true,
            response: false,
          },
          generate: {
            body: true,
            header: false,
            param: true,
            query: true,
            response: false,
          },
        },
      },
      propertySortOrder: 'Alphabetical',
      target: './src/zod/index.ts',
      tsconfig: './tsconfig.json',
    },
  },
})
```

### Environment

```
  System:
    OS: macOS 26.4.1
    CPU: (16) arm64 Apple M3 Max
    Memory: 288.73 MB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    orval: 8.7.0 => 8.7.0 
    react: 19.2.5 => 19.2.5 
    zod: 4.3.6 => 4.3.6 
```

### Expected behavior

```ts
/**
 * Generated by orval v8.7.0 🍺
 * Do not edit manually.
 * Example API
 * OpenAPI spec version: 1.0.0
 */
import { z as zod } from 'zod'

export const EnumPropertiesObjectDefault = { enabled: true, value: 'a' } as const
export const EnumPropertiesObject = zod
  .object({
    enabled: zod.coerce.boolean(),
    value: zod.enum(['a', 'b', 'c']),
  })
  .default(EnumPropertiesObjectDefault)

export type EnumPropertiesObject = zod.input<typeof EnumPropertiesObject>
export type EnumPropertiesObjectOutput = zod.output<typeof EnumPropertiesObject>

```

### Output

```ts
/**
 * Generated by orval v8.7.0 🍺
 * Do not edit manually.
 * Example API
 * OpenAPI spec version: 1.0.0
 */
import { z as zod } from 'zod'

export const EnumPropertiesObjectDefault = { enabled: true, value: 'a' }
export const EnumPropertiesObject = zod
  .object({
    enabled: zod.coerce.boolean(),
[...]

PR #3400, opened 2026-05-20
fix(zod): narrow string literals per-value in object defaults (#3399)

### LINKED ISSUE #3399: Zod: object defaults containing arrays produce TS2769 since 8.10.0 ("readonly cannot be assigned to mutable type", regression from #3339)
### Description

Since `8.10.0` (PR #3339, follow-up to #3244), the zod generator emits object defaults with `as const`. When the schema has a property whose default is an empty array, the array literal is inferred as `readonly never[]`, which is no longer assignable to the mutable array type that `zod@4.x`'s `.default()` overloads expect.

Result: every `.default({...})` call where the default object contains an empty array fails to type-check with `TS2769: No overload matches this call` — `Type 'readonly []' is 'readonly' and cannot be assigned to the mutable type '{ value: string; }[]'`.

This appears to be a regression / incomplete fix from #3339. The previous output (`{ ... }` without `as const`) typechecked fine.

### Output client

zod

```ts

export default defineConfig({
  repro: {
    output: {
      mode: 'single',
      client: 'zod',
      target: './src/repro.zod.ts',
    },
  },
});
```

### Environment

```
  System:
    OS: macOS 26.5
    CPU: (8) arm64 Apple M1 Pro
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.15.0
  npmPackages:
    orval: 8.11.0 => 8.11.0
    zod: ^4.4.3 => 4.4.3
    typescript: ^5.6.0 => 5.6.x
```

### Expected behavior

Generated zod schemas type-check cleanly when an object default contains arrays or other writable structures. Either:

1. Emit defaults without `as const`, or
2. Cast the default expression so it satisfies zod's `.default()` parameter type (e.g. use a typed annotation rather than `as const`), or
3. Wrap the default in a thunk (`.default(() => ({ ... }))`).

### Actual behavior

`tsc --noEmit` fails:

```
src/repro.zod.ts(21,12): error TS2769: No overload matches this call.
  Overload 1 of 2, '(def: { checklist?: ... | null | undefined; available_indexes?: { value: string; }[] | undefined; }): ZodDefault<...>', gave the following error.
    Argument of type '{ readonly checklist: null; readonly available_indexes: readonly []; }' is not assignable to parameter of type '{ checklist?: ... | null | undefined; available_indexes?: { value: string; }[] | undefined; }'.
      Types of property '"available_indexes"' are incompatible.
        The type 'readonly []' is 'readonly' and cannot be assigned to the mutable type '{ value: string; }[]'.
  Overload 2 of 2, '(def: () => { ... }): ZodDefault<...>', gave the following error.
    Argument of type '{ readonly checklist: null; readonly available_indexes: readonly []; }' is not assignable to parameter of type '() => { ... }'.
      Type '{ readonly checklist: null; readonly available_indexes: readonly []; }' provides no match for the signature '(): { ... }'.
```

```yaml
openapi: 3.1.0
info:
  title: Minimal repro
  version: "1.0.0"
paths:
  /thing:
    get:
      operationId: getThing
      responses:
        "200":
          description: ok
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Thing"
components:
  schemas:
    Thing:
      type: object
      properties:
        settings:
          type: object
          properties:
            checklist:
              type: object
              nullable: true
              additionalProperties:
                type: object
                properties:
                  required_templates:
                    type: array
                    items:
                      type: object
                      properties:
                        id:
                          type: string
                      required: [id]
            available_indexes:
              type: array
              items:
                type: object
                properties:
                  value:
                    type: string
                required: [value]
          default:
            checklist: null
            available_indexes: []
```

### Generated output (orval 8.11.0)

```ts
// src/repro.zod.ts
import * as zod from 'zod';

export const getThingResponseSettingsDefault = { checklist: null, available_indexes: [] } as const;

export const GetThingResponse = zod.object({
  "settings": zod.object({
    "checklist": zod.record(zod.string(), zod.object({
      "required_templates": zod.array(zod.object({
        "id": zod.string()
      })).optional()
    })).nullish(),
    "available_indexes": zod.array(zod.object({
      "value": zod.string()
    })).optional()
  }).default(getThingResponseSettingsDefault)  // ❌ TS2769
});
```

### Additional context

- Pre-`8.10.0` orval did not append `as const`, and the same spec type-checked successfully.
- The root cause is in `@orval/zod` (see `packages/zod/src/index.ts`, the branch added by PR #3339):
  ```js
  defaultValue = entries.length === 0
    ? `{} as const`
    : `{ ${entries} } as const`;
  ```
  Any default containing an empty/literal array becomes `readonly never[]` / `readonly T[]`, which zod's `.default()` rejects.
- Issue #3244 was closed by #3339 — but that fix solved enum literal narrowing while introducing this regression for default objects that include arrays. Both cases need to coexist.

A simple workaround is to omit `as const` for objects whose values include arrays, or to switch to the thunk form `.default(() => ({ ... }))`, which side-steps the variance check.

zeriong:
## Summary

- Closes #3399 — zod object defaults containing arrays produced `TS2769` since 8.10.0 because the whole-object `as const` from PR #3339 made nested arrays `readonly`, which zod v4's `.default()` rejects against its mutable parameter type.
- Switches the emitted default to apply `as const` only to individual string literal values (including strings inside arrays), instead of wrapping the whole object/array literal. This preserves the enum literal narrowing that #3244 / #3339 needed while keeping arrays and objects mutable.

### Before (`8.10.0` – `8.11.0`, regression)

```ts
// Whole-object as const → nested [] becomes `readonly never[]` → TS2769
export const settingsDefault = { checklist: null, available_indexes: [] } as const;
```

### After

```ts
// Per-value narrowing → arrays/objects stay mutable, strings keep literal types
export const settingsDefault = { checklist: null, available_indexes: [] };
// Enum literal narrowing still works for string properties:
export const enumPropertiesObjectDefault = { enabled: true, value: "a" as const };
// And for enum-array properties:
export const taggedObjectDefault = { tags: ["a" as const, "b" as const] };
```

I verified end-to-end against zod `4.4.3` (the version in the repro) that:
- the minimal repro from #3399 (object default with empty array) now type-checks,
- the original #3244 case (object default with enum-typed string property) still type-checks,
- the combined case (object default with enum-typed array property) also type-checks.

## Test plan

- [x] `bun vitest run packages/zod` — 170 tests pass, including the new regression test for #3399 and the updated #3244 test
- [x] `bun run --filter @orval/zod lint`
- [x] `bun run build`
- [x] `bunx turbo run test:snapshots --filter='./samples/**' --filter=orval-tests` — 4070 snapshot tests pass, no snapshot drift
- [x] Verified the emitted output against `zod@4.4.3` with `tsc --noEmit --strict` using the issue's exact repro plus the #3244 and enum-array cases
- [x] No new/untracked files under `tests/__snapshots__`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced default value generation for object-based validation schemas to better preserve literal types for string properties and enum-derived values
  * Fixed type handling of arrays within object defaults to maintain proper mutability while ensuring full Zod v4 compatibility

[Review Change Stack](https://app.coderabbit.ai/change-stack/orval-labs/orval/pull/3400?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -344,15 +344,19 @@ export const generateZodValidationSchemaDefinition = (
       // OpenApiSchemaObject defines default as 'any'
       defaultValue = `new Date("${escape(schema.default)}")`;
     } else if (isObject(schema.default)) {
+      // Narrow string literals individually with `as const` so `zod.enum([...])`
+      // properties accept the emitted default (#3244). Whole-object/array
+      // `as const` would make nested arrays `readonly`, which zod v4's
+      // `.default()` rejects against its mutable parameter type (#3399).
       const entries = Object.entries(schema.default)
         .map(([key, value]) => {
           if (isString(value)) {
-            return `${key}: "${escape(value)}"`;
+            return `${key}: "${escape(value)}" as const`;
           }
 
           if (Array.isArray(value)) {
             const arrayItems = value.map((item) =>
-              isString(item) ? `"${escape(item)}"` : `${item}`,
+              isString(item) ? `"${escape(item)}" as const` : `${item}`,
             );
             return `${key}: [${arrayItems.join(', ')}]`;
           }
@@ -366,10 +370,7 @@ export const generateZodValidationSchemaDefinition = (
             return `${key}: ${value}`;
         })
         .join(', ');
-      // `as const` preserves literal types so `zod.enum([...])` properties
-      // accept the emitted default (#3244).
-      defaultValue =
-        entries.length === 0 ? `{} as const` : `{ ${entries} } as const`;
+      defaultValue = entries.length === 0 ? `{}` : `{ ${entries} }`;
     } else {
       // OpenApiSchemaObject defines default as 'any'
       const rawStringified = stringify(schema.default);
--- a/packages/zod/src/zod.test.ts
+++ b/packages/zod/src/zod.test.ts
@@ -1574,7 +1574,7 @@ describe('generateZodValidationSchemaDefinition`', () => {
           ['default', 'testObjectDefaultDefault'],
         ],
         consts: [
-          'export const testObjectDefaultDefault = { name: "Fluffy", age: 3 } as const;',
+          'export const testObjectDefaultDefault = { name: "Fluffy" as const, age: 3 };',
         ],
       });
 
@@ -1589,7 +1589,7 @@ describe('generateZodValidationSchemaDefinition`', () => {
         'zod.object({\n  "name": zod.string().optional(),\n  "age": zod.number().optional()\n}).default(testObjectDefaultDefault)',
       );
       expect(parsed.consts).toBe(
-        'export const testObjectDefaultDefault = { name: "Fluffy", age: 3 } as const;',
+        'export const testObjectDefaultDefault = { name: "Fluffy" as const, age: 3 };',
       );
     });
 
@@ -2329,7 +2329,7 @@ describe('generateZodValidationSchemaDefinition`', () => {
       );
     });
 
-    it('appends "as const" to object defaults so enum properties keep literal types (#3244)', () => {
+    it('narrows string literals in object defaults so enum properties keep literal types (#3244)', () => {
       const schemaWithObjectEnumDefault: OpenApiSchemaObject = {
         type: 'object',
         required: ['enabled', 'value'],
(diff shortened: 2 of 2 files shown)


zeriong:
@melloware
Opened #3400. Switches to per-value as const on string literals only, so #3244's enum narrowing is preserved while nested arrays stay mutable for zod v4's .default(). Verified with zod@4.4.3 against the repro here plus the #3244 / enum-array cases. Review appreciated 🙏

melloware (approved):

PR #3419, opened 2026-05-21
fix(core): handle type arrays (["array","null"]) in form-urlencoded serialization

### LINKED ISSUE #3418: form-urlencoded array serialization broken when transformer adds nullable (v8.11.0+)
## Bug Description

In Orval v8+, `application/x-www-form-urlencoded` request bodies with nullable array-typed properties generate TypeScript code that fails to compile:

```
error TS2345: Argument of type 'UpdatePetRequestTagsItem[]' is not assignable to parameter of type 'string'.
```

This affects any OpenAPI 3.0 spec using `nullable: true` on an array property in a form-urlencoded request body. It also affects OpenAPI 3.1 specs using `type: ["array", "null"]`.

## Reproduction

**Schema (OpenAPI 3.0):**

```yaml
openapi: "3.0.3"
info:
  title: Petstore
  version: "1.0.0"
paths:
  /pet/update:
    post:
      operationId: updatePet
      requestBody:
        content:
          application/x-www-form-urlencoded:
            schema:
              $ref: "#/components/schemas/UpdatePetRequest"
      responses:
        "200":
          description: OK
components:
  schemas:
    UpdatePetRequest:
      type: object
      required:
        - petId
        - name
      properties:
        petId:
          type: string
        name:
          type: string
        tags:
          type: array
          nullable: true
          items:
            type: object
            required:
              - tagId
              - label
            properties:
              tagId:
                type: string
              label:
                type: string
```

**Orval config:**

```ts

export default defineConfig({
  petstore: {
    input: {
      target: "./petstore.yaml",
    },
    output: {
      target: "./output/petstore.ts",
      client: "fetch",
      httpClient: "fetch",
      baseUrl: "/api",
    },
  },
});
```

**Generated code (v7.3.0 — correct):**

```ts
updatePetRequest.tags.forEach(value =>
  formUrlEncoded.append('tags', JSON.stringify(value))
);
```

**Generated code (v8.0.0+ — broken):**

```ts
formUrlEncoded.append(`tags`, updatePetRequest.tags);
// TS2345: Argument of type 'UpdatePetRequestTagsItem[]' is not assignable to parameter of type 'string'.
```

## Root Cause

Orval v8 introduced an internal `upgrade()` step (via `@scalar/openapi-upgrader`) that converts all specs to OpenAPI 3.1 before code generation. This converts:

```json
{ "type": "array", "nullable": true }
```

Into:

```json
{ "type": ["array", "null"] }
```

However, `resolveSchemaPropertiesToFormData` in `@orval/core` checks `property.type === "array"` using strict string equality. After the upgrade, `property.type` is `["array", "null"]` (an array), so the check fails. Execution falls through to the generic `else` branch which generates `formUrlEncoded.append('tags', updatePetRequest.tags)` — passing the raw array directly to a function that expects a `string`.

The same issue likely applies to nullable `object`-typed properties (`property.type === "object"` check) and nullable numeric/boolean properties.

## Proposed Fix

In `resolveSchemaPropertiesToFormData` (packages/core/src/getters/form-data.ts), the type checks should account for `type` being an array (the 3.1 format):

```ts
const isType = (property: SchemaObject, type: string): boolean => {
  if (property.type === type) return true;
  if (Array.isArray(property.type) && property.type.includes(type)) return true;
  return false;
};
```

Then replace:
```ts
else if (property.type === "object") { ... }
else if (property.type === "array") { ... }
```

With:
```ts
else if (isType(property, "object")) { ... }
else if (isType(property, "array")) { ... }
```

The same pattern should be applied to the number/integer/boolean type checks in the same function.

## Versions Tested

| Version | Result |
|---------|--------|
| 7.3.0   | ✅ Correct — generates `.forEach(value => append('tags', JSON.stringify(value)))` |
| 8.0.0   | ❌ Broken — generates `append('tags', updatePetRequest.tags)` |
| 8.8.1   | ❌ Broken |
| 8.10.0  | ❌ Broken |
| 8.11.0  | ❌ Broken |
| 8.12.1  | ❌ Broken |

## Environment

- Node: v24.15.0
- TypeScript: 5.7.3
- Client: `fetch`
- Platform: macOS

jakiestfu:
## Summary

Fixes #3418

After `@scalar/openapi-upgrader` converts an OpenAPI 3.0 spec to 3.1 format, `nullable: true` arrays are transformed from:

```json
{ "type": "array", "nullable": true }
```

to:

```json
{ "type": ["array", "null"] }
```

The `resolveSchemaPropertiesToFormData` function in `res-req-types.ts` used strict string equality (`property.type === 'array'` and `property.type === 'object'`), so it fell through to the generic `else` branch and generated:

```ts
formUrlEncoded.append(`tags`, updatePetRequest.tags);
// TS2345: Argument of type 'UpdatePetRequestTagsItem[]' is not assignable to parameter of type 'string'.
```

## Fix

Extended both `'object'` and `'array'` type checks to also match when `property.type` is an array that includes the respective type string — the same pattern already used for `number`/`integer`/`boolean` checks in the same function.

## Test

Added a regression test in `res-req-types.test.ts` that sets `type: ["array", "null"]` directly (mimicking the post-upgrade schema) and asserts that a `forEach` loop is generated rather than a bare `append` of the raw array.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fix handling of nullable array/object type declarations so URL-encoded request bodies serialize nullable arrays/objects correctly at runtime.

* **Tests**
  * Added test coverage validating URL-encoded request bodies with nullable array properties to prevent regressions.

* **Chores**
  * Updated generated-file headers to reflect the newer tool version.

[Review Change Stack](https://app.coderabbit.ai/change-stack/orval-labs/orval/pull/3419?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


jakiestfu:
@melloware This should do it!

melloware (approved):

melloware:
let me cut a new release

melloware:
8.12.2 is in NPM!

PR #3430, opened 2026-05-24
fix(zod): parser generation when content-type contains charset precision

melloware:
Fix #3420

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved content-type matching to correctly handle charset/parameterized media types (e.g., `application/json; charset=utf-8`) and multipart form-data.

* **Tests**
  * Added regression test covering multipart/form-data and JSON content keys with charset suffixes.

* **Chores**
  * Generated mock outputs updated to emit missing/nullable values using an equivalent sentinel (e.g., `void 0`).
  * Added generated response schemas for the ListPets endpoint in snapshots.

[Review Change Stack](https://app.coderabbit.ai/change-stack/orval-labs/orval/pull/3430?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -1350,7 +1350,7 @@ const parseBodyAndResponse = ({
     ),
   );
   const formDataContent = contentEntries.find(
-    isMediaType('multipart/form-data'),
+    isMediaType(String.raw`^multipart\/form-data$`),
   );
   const [contentType, mediaType] = jsonContent
     ? (['application/json', jsonContent[1]] as const)
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -1335,13 +1335,27 @@ const parseBodyAndResponse = ({
     | OpenApiRequestBodyObject;
 
   // Only handle JSON and form-data; other content types (e.g., application/octet-stream)
-  // are skipped - unclear if this is correct behavior for root-level binary/text bodies
-  const jsonMedia = resolvedRef.content?.['application/json'];
-  const formDataMedia = resolvedRef.content?.['multipart/form-data'];
-  const [contentType, mediaType] = jsonMedia
-    ? (['application/json', jsonMedia] as const)
-    : formDataMedia
-      ? (['multipart/form-data', formDataMedia] as const)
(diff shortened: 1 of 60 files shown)


melloware:
cc @wadakatu i must be doing something wrong here i have run everything locally including a `nuke:all` but this keeps failing on GitHub with the same flip flopping failure about `void 0` vs `undefined`?

wadakatu:
Hey @melloware — debugged locally; CI is only upset about a narrow slice of the diff.

**Fine:** the zod fix in `packages/zod/src/index.ts` and the 20 snapshot updates under `tests/__snapshots__/.../zod.ts`. Verified by applying just the zod patch on top of clean master — those are legitimate drift from the fix and CI accepts them.

**Breaking CI:** the 38 `undefined` → `void 0` substitutions across `samples/**/*.msw.ts` (and their `__snapshots__/` counterparts). The orval mock generator only emits the literal `'undefined'` — there's no `void 0` anywhere in `packages/mock/`:

```
$ git grep -nE "void 0" packages/mock/
(no results)
```

`test:snapshots` depends on `generate-api` in `turbo.json`, so CI regenerates `samples/*/src/api/.../*.msw.ts` before comparing. The regenerated files come out with `undefined`, but the committed `__snapshots__/` now have `void 0` — mismatch.

I couldn't tell from here where those `void 0`s came from on your side (orval itself doesn't emit them). Resetting `samples/**` and regenerating clean should surface the source:

```bash
git checkout origin/master -- samples/
bun run nuke:all && bun install
bun run --filter '*' generate-api
git diff samples/
```

If it's easier on your side, happy to open a separate PR with the same fix (the zod change + the legitimate `tests/__snapshots__/` updates only) — full credit stays with you for the fix. Totally fine either way 🙏

melloware:
Yes please i just ran all your steps and its still thinks there is snapshot fix. Can you open a PR with this fix?

melloware:
One caveat is I am running under Windows WSL not sure if that makes a difference or not?

wadakatu:
Opened #3433 with just the zod fix + the legitimate `tests/__snapshots__/` updates — both you and @pierre-isabel-bbc are tagged as co-authors. Feel free to close this PR once #3433 is merged 🙏

PR #3462, opened 2026-05-27
fix(zod): emit reusable schemas for sanitized component names

### LINKED ISSUE #3460: generateReusableSchemas imports reusable schemas it never generates
### Description

When `output.client: "zod"` is used with `override.zod.generateReusableSchemas: true`, Orval emits an `import { … } from "./schemas"` (and a reference to it) for component schemas that it never actually generates.

The bug triggers for a component schema that is **both**:

  1. named with characters that the naming convention transforms — e.g. `_` or `-`
     (so the camelCased reference name differs from the raw component name), **and**
  2. referenced **only** at the top level of an operation request/response — i.e. it
     is *not* also nested as a `$ref` inside another component schema.

So Orval has two passes that disagree:
  - the **reference pass** emits `import { <conventionName(rawName)> } from "./schemas"`
    and `export const <Op>Response = <conventionName(rawName)>` unconditionally;
  - the **schema-file pass** that writes `./schemas/*.ts` skips top-level-only schemas
    whose raw name isn't already a clean identifier.

### Output client

zod

```ts

export default defineConfig({
  repro: {
    input: {
      target: "./openapi.yaml",
    },
    output: {
      mode: "split",
      client: "zod",
      target: "./out/zod",
      schemas: "./out/zod/schemas",
      fileExtension: ".ts",
      override: {
        zod: {
          generateReusableSchemas: true,
        },
      },
    },
  },
});
```

### Environment

  System:
    OS: Linux 7.0 Arch Linux
    CPU: (22) x64 Intel(R) Core(TM) Ultra 7 155H
    Memory: 13.41 GB / 30.84 GB
    Container: Yes
    Shell: 4.7.1 - /bin/fish 
 Environment
  - orval: 8.13.0
  - @orval/zod: 8.13.0
  - zod: 4.3.6
  - node: v22.12.0

### Expected behavior

Orval should generate a reusable schema for Page_Item_ in ./schemas (e.g. export const pageItem = …) so the import resolves, the same way it already does when the schema is nested inside another component, and the same way the TypeScript model client handles it.

### Actual behavior

Running orval generates `out/zod/schemas/` with files for `CleanName`, `Parent`, and `Child_Item_`, but **no** file/export for `Page_Item_`. Despite that, the main generated file imports and references it:

  ```ts
  import {
    cleanName,
    pageItem,   // never generated in ./schemas
    parent,
  } from "./schemas";

  export const GetGenericResponse = pageItem;
```

The output therefore does not compile: out/zod/orvalReusableSchemasRepro.ts(9,3): error TS2305: Module './schemas' has no exported member 'pageItem'.

```yaml
openapi: 3.1.0
info:
  title: Orval Reusable-Schemas Repro
  version: 1.0.0

paths:
  /clean:
    get:
      operationId: getClean
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/CleanName'

  /generic:
    get:
      operationId: getGeneric
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Page_Item_'

  /parent:
    get:
      operationId: getParent
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Parent'

components:
  schemas:
    CleanName:
      type: object
      properties:
        id:
          type: integer
      required:
        - id

    # BUG: special name, referenced only top-level -> imported as `pageItem` but never generated
    Page_Item_:
      type: object
      properties:
        total:
          type: integer
      required:
        - total

    Parent:
      type: object
      properties:
        child:
          $ref: '#/components/schemas/Child_Item_'
      required:
        - child

    # Same special name shape, but nested under Parent -> generated fine (proves trigger is top-level-only, not the chars)
    Child_Item_:
      type: object
      properties:
        value:
          type: integer
      required:
        - value
```

### Additional context

_No response_

titivermeesch:
Closes #3460 

## Problem

With `zod` client + `override.zod.generateReusableSchemas: true`, orval emits `import { … } from "./schemas"` for component schemas it never generates, producing uncompilable output

`writeZodSchemasReusable` built its root refs from `builder.schemas[].name`, the sanitized model identifier (e.g. `PaginatedResponse_Asset_` becomes `PaginatedResponseAsset`), then filtered them against the raw-keyed `spec.components.schemas`. Every schema whose name needs sanitizing failed that filter and was dropped from the roots. Schemas reached as nested `$ref`s still surfaced via `nameToRef`, but ones
referenced only at the top level of an operation never did.

## Fix

Derive root refs straight from the raw `spec.components.schemas` keys, matching the names operation files import.

## Tests

Added a regression case (`builder.schemas` name `PageItem` vs raw key `Page_Item_`);

### DIFF
--- a/packages/orval/src/write-zod-specs.test.ts
+++ b/packages/orval/src/write-zod-specs.test.ts
@@ -375,6 +375,51 @@ describe('writeZodSchemas with generateReusableSchemas', () => {
 
     await fs.remove(root);
   });
+
+  it('emits schemas whose raw name differs from the sanitized model name', async () => {
+    const root = await fs.mkdtemp(path.join(tmpdir(), 'orval-zod-reuse-raw-'));
+    const schemasPath = path.join(root, 'schemas');
+
+    const builder = {
+      spec: {
+        components: {
+          schemas: {
+            Page_Item_: {
+              type: 'object',
+              properties: { total: { type: 'number' } },
+              required: ['total'],
+            },
+          },
+        },
+      },
+      target: '',
+      schemas: [
+        {
+          name: 'PageItem',
+          schema: { $ref: '#/components/schemas/Page_Item_' },
+        },
+      ],
+    } satisfies Parameters<typeof writeZodSchemas>[0];
+
+    const options = createOutputOptions();
+    (options.override.zod as Record<string, unknown>).generateReusableSchemas =
+      true;
+
+    await writeZodSchemas(builder, schemasPath, '.ts', '', options);
+
+    const fileExists = await fs.pathExists(
+      path.join(schemasPath, 'PageItem.ts'),
+    );
+    expect(fileExists).toBe(true);
+
+    const content = await fs.readFile(
+      path.join(schemasPath, 'PageItem.ts'),
+      'utf8',
+    );
+    expect(content).toContain('export const PageItem = ');
+
+    await fs.remove(root);
+  });
 });
 
 describe('writeZodSchemasFromVerbs with generateReusableSchemas', () => {
--- a/packages/orval/src/write-zod-specs.ts
+++ b/packages/orval/src/write-zod-specs.ts
@@ -447,17 +447,22 @@ async function writeZodSchemasReusable(
     output: output as ContextSpec['output'],
   };
 
-  // Roots = every component schema in the builder list. This matches today's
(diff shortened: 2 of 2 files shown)


z4o4z (approved):
Looks good to me, thanks!

melloware (approved):

PR #3475, opened 2026-05-28
feat: add option to inject params in generated zod schemas

### LINKED ISSUE #3425: Allow injecting custom params into generated Zod schemas IL8N
### Description

Generated Zod schemas use Zod's built-in English error messages, and there is no way to customize them without forking or post-processing the generated files.

Zod itself supports a params argument on every validator (e.g. zod.string().email({ error: 'My message' })), but Orval has no option to thread anything into it.

**I'm new to this repo so bear with me if I misinterpreted certain systems Orval proposes. It's a feature I would like to work on if you think this is a good idea, but it will take some take to get used to the codebase**

### Current behavior

- Every validator Orval emits is called without a params argument: zod.string().email(), zod.number().min(), etc.
- The only user-provided function that can be threaded into the generated Zod code is override.zod.preprocess, which does not fit this usecase
- Zod's global error map (z.config({ customError })) works at runtime and receives issue.path, but issue.path is relative to the parsed schema. Fields with the same name across different operations collide on the same key, so it cannot fully replace generator-level customization.

### Proposed behavior

A new override.zod.params option lets users provide a function that receives codegen-time context for each emitted validator and returns a Zod params object. Orval bakes a call to that function as the trailing argument of the validator.

```javascript
// orval.config.ts
override: {
  zod: {
    params: { path: './zod-params.ts', name: 'zodParams' },
  },
 }
```

```javascript
// zod-params.ts
import { i18n } from './i18n';
export const zodParams = (ctx) => ({
  error: (issue) =>
    i18n.t(
      `errors.${ctx.schemaName}.${ctx.fieldPath.join('.')}.${ctx.validator}`,
      { value: issue.input },
    ),
});
```

Generated output:

```javascript
import { zodParams } from './zod-params';

zod.string().email(zodParams({
  operationId: 'createUser',
  location: 'body',
  schemaName: 'CreateUserBody',
  fieldPath: ['email'],
  validator: 'email',
}))
```

Key properties:
- zodParams(...) runs once when the schema is built, so it sees full codegen-time context (operationId, fieldPath, etc.).
- The error callback it returns runs at parse time
- Users who only want static strings just return { error: 'My message' }.
- If zodParams returns undefined, the validator is emitted unchanged.

### Proposed solution

Reuse the existing Mutator pattern (the same one preprocess uses)

1. Extend ZodOptions / NormalizedZodOptions with params?: Mutator.
2. Resolve the mutator the same way preprocess is resolved in generateZodRoute
3. Thread it down through parseZodValidationSchemaDefinition into parseProperty, where each validator call is serialized. When the mutator is configured, append a call to it as the trailing argument with the ZodParamsContext for that call site.

titivermeesch:
Closes #3425 

Adds override.zod.params to inject a Zod params argument (e.g. { error: ... }) into each generated validator. The option takes a Mutator (same shape as the existing preprocess) pointing to a function. Orval calls that function once per validator at schema construction time with codegen-time context (operation, location, schema name, field path, validator name), and emits the return value as the trailing argument of the call.

## Summary by CodeRabbit

* **New Features**
  * Added a Zod params option to inject build-time parameters into generated validators (global, per-operation/tag, and component schemas); re-exported the Zod params context type.

* **Documentation**
  * Expanded Zod output docs with examples, where injection applies or is skipped, fieldPath semantics, and supported return shapes.

* **Tests**
  * Added coverage for injection placement, field-path propagation, special-merge rules, reusable-schema behavior, and no-injection regression.

[Review Change Stack](https://app.coderabbit.ai/change-stack/orval-labs/orval/pull/3475?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/content/docs/reference/configuration/output.mdx
+++ b/docs/content/docs/reference/configuration/output.mdx
@@ -1080,6 +1080,65 @@ Control which schemas are generated.
 
 Add preprocess functions to schemas.
 
+### params
+
+**Type:** [`Mutator`](#mutator)
+
+Inject a Zod `params` argument (e.g. `{ error: ... }`) into every generated validator. The referenced function is called once per validator at schema construction time and receives codegen-time context (operation, location, schema name, field path, validator name). Whatever it returns is passed as the trailing argument of the call.
+
+Useful for i18n error keys, branded error messages, or any field-aware customisation that Zod's global error map cannot disambiguate on its own (because `issue.path` does not carry operation/schema identity).
+
+```ts title="orval.config.ts"
+export default defineConfig({
+  petstore: {
+    output: {
+      override: {
+        zod: {
+          params: { path: './zod-params.ts', name: 'zodParams' },
+        },
+      },
+    },
+  },
+});
+```
+
+```ts title="zod-params.ts"
+import type { ZodParamsContext } from 'orval';
+import { i18n } from './i18n';
+
+export const zodParams = (ctx: ZodParamsContext) => ({
+  error: (issue: { input: unknown; path: PropertyKey[] }) =>
+    i18n.t(
+      `errors.${ctx.schemaName}.${ctx.fieldPath.join('.')}.${ctx.validator}`,
+      { value: issue.input },
+    ),
+});
+```
+
+The `'schema'` location is used for shared component schemas emitted under [`generateReusableSchemas`](#generatereusableschemas). Component schemas have no single owning operation, so `operationId` is the empty string in that case — branch on `ctx.location === 'schema'` if your error keys need to fall back to a schema-only namespace.
+
+Generated output (excerpt):
+
+```ts
+import { zodParams } from './zod-params';
+
+export const CreateUserBody = zod.object({
+  email: zod
+    .string(zodParams({ operationId: 'createUser', location: 'body', schemaName: 'CreateUserBody', fieldPath: ['email'], validator: 'string' }))
+    .email(zodParams({ operationId: 'createUser', location: 'body', schemaName: 'CreateUserBody', fieldPath: ['email'], validator: 'email' })),
+});
+```
+
+Injection scope:
+
+- Applied to base types (`string`, `number`, `boolean`, `bigint`, `date`, `integer`), constraints (`min`, `max`, `gt`, `lt`, `multipleOf`, `regex`, `length`), formats (`email`, `url`, `uuid`, `hostname`, `datetime`, `time`), and `literal`, `enum`, `instanceof`, `stringFormat`.
+- Skipped on modifiers (`optional`, `nullable`, `nullish`, `default`, `describe`) and structural calls (`object`, `array`, `tuple`, `union`, `rest`, `passthrough`, `strict`).
+- `fieldPath` only includes object property names, mirroring Zod's own `issue.path`. Array indices and tuple positions are not appended — the inner element of `{ tags: array<string> }` and a top-level `tags: string` both see `fieldPath: ['tags']`. Use the `validator` field to distinguish a container (`'array'`, `'tuple'`) from its element (`'string'`, `'number'`).
+
+For static messages, return an object with a string `error`: `return { error: 'My message' }`. The function may return `undefined` to fall back to Zod defaults for a specific call.
+
+> The `{ error }` shape is Zod v4-only — on v3 it is silently ignored and the default message is used. If your project supports both Zod v3 and v4, return `{ message: 'My message' }` instead, which works on both.
+
 ### dateTimeOptions / timeOptions
 
 **Type:** `Object`
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -744,6 +744,18 @@ export interface ZodOptions {
     body?: Mutator;
     response?: Mutator;
   };
+  /**
+   * Mutator referencing a function called once per emitted validator at schema
+   * construction time. It receives codegen-time context (operation, location,
+   * schema name, field path, validator name) and returns a Zod `params` object
+   * (e.g. `{ error: ... }`) that is appended as the trailing argument.
+   *
+   * The plural name follows Zod's own term for the validator's second argument
+   * (`z.string(params)`) and is unrelated to the singular `param` key used by
+   * `generate` / `coerce` / `preprocess` above, which refers to the path-parameter
+   * location.
+   */
+  params?: Mutator;
   dateTimeOptions?: ZodDateTimeOptions;
   timeOptions?: ZodTimeOptions;
   generateEachHttpStatus?: boolean;
@@ -802,6 +814,7 @@ export interface NormalizedZodOptions {
     body?: NormalizedMutator;
     response?: NormalizedMutator;
   };
+  params?: NormalizedMutator;
   generateEachHttpStatus: boolean;
   useBrandedTypes: boolean;
   generateReusableSchemas: boolean;
--- a/packages/orval/src/write-zod-specs.ts
+++ b/packages/orval/src/write-zod-specs.ts
@@ -3,6 +3,7 @@ import path from 'node:path';
 import {
   type ContextSpec,
   conventionName,
+  type GeneratorMutator,
   getRefInfo,
   isComponentRef,
   type NamingConvention,
@@ -123,6 +124,29 @@ interface WriteZodSchemasFromVerbsContext {
   spec: ContextSpec['spec'];
   target: string;
   workspace: string;
+}
+
+/**
+ * Render the `import { ... } from '...'` line for a resolved
+ * `GeneratorMutator`. Mirrors the format produced by
+ * `generateMutatorImports` in `@orval/core` but inlined to avoid pulling in
+ * its full surface area for a single statement.
+ */
+function buildMutatorImportStatement(mutator: GeneratorMutator): string {
+  const importClause = mutator.default ? mutator.name : `{ ${mutator.name} }`;
+  return `import ${importClause} from '${mutator.path}';`;
+}
+
+/**
+ * Whole-word substring check for a resolved mutator alias inside generated
+ * code. Plain `String.includes` would false-positive when the user names the
+ * mutator something like `min` against `.min(1)`.
+ */
+function bodyReferencesMutator(
+  body: string,
+  mutator: GeneratorMutator,
+): boolean {
+  return new RegExp(String.raw`\b${mutator.name}\b`).test(body);
 }
 
 function generateZodSchemaFileContent(
@@ -401,12 +425,20 @@ export function generateZodSchemasInline(
   builder: WriteZodSchemasInput,
   output: WriteZodOutputOptions,
   includeZodImport = true,
+  paramsMutator?: GeneratorMutator,
+  includeParamsImport = false,
 ): string {
   const useReusableSchemas =
     output.override.zod.generateReusableSchemas === true;
 
   if (useReusableSchemas) {
-    return generateZodSchemasInlineReusable(builder, output, includeZodImport);
+    return generateZodSchemasInlineReusable(
+      builder,
+      output,
+      includeZodImport,
+      paramsMutator,
+      includeParamsImport,
+    );
   }
 
   const schemasWithOpenApiDef = builder.schemas.filter((s) => s.schema);
@@ -472,6 +504,8 @@ function generateZodSchemasInlineReusable(
   builder: WriteZodSchemasInput,
   output: WriteZodOutputOptions,
   includeZodImport = true,
+  paramsMutator?: GeneratorMutator,
+  includeParamsImport = false,
 ): string {
   const isZodV4 = !!output.packageJson && isZodVersionV4(output.packageJson);
   const strict = output.override.zod.strict.body;
@@ -506,6 +540,7 @@ function generateZodSchemasInlineReusable(
     isZodV4,
     coerce,
     generateMeta: output.override.zod.generateMeta,
+    paramsMutator,
   });
 
   const rewritten = rewriteReusableSchemas(entries);
@@ -519,7 +554,21 @@ function generateZodSchemasInlineReusable(
 
   // Omit the zod import when concatenated into a file that already imports it
   // (inline single-mode output where the zod client emits `import * as zod`).
-  const prefix = includeZodImport ? `import { z as zod } from 'zod';\n\n` : '';
+  const zodImport = includeZodImport ? `import { z as zod } from 'zod';\n` : '';
+  // In split modes (`split` / `tags-split`) the inline schemas are written to
+  // a separate `.schemas` file with no other imports, so the params-mutator
+  // import has to be emitted here. In `single` / `tags` modes the schemas are
+  // concatenated into the operation file, which already imports the same
+  // mutator via its per-verb mutators array (see `generateZodRoute`) — we
+  // skip emitting it again to avoid a duplicate import line.
+  const paramsImport =
+    paramsMutator &&
+    includeParamsImport &&
+    bodyReferencesMutator(body, paramsMutator)
+      ? `${buildMutatorImportStatement(paramsMutator)}\n`
+      : '';
+  const prefix =
+    zodImport || paramsImport ? `${zodImport}${paramsImport}\n` : '';
   return `${prefix}${body}\n`;
 }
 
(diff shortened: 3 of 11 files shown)


packages/zod/src/index.ts
```diff
@@ -2054,7 +2164,10 @@ export const ${operationResponse} = zod.array(${operationResponse}Item)${
         ];
       }),
     ].join('\n\n'),
-    mutators: preprocessResponse ? [preprocessResponse] : [],
+    mutators: [
+      ...(preprocessResponse ? [preprocessResponse] : []),
+      ...(paramsMutator ? [paramsMutator] : []),
+    ],
```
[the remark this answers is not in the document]
titivermeesch: This was preexisting to this PR. There are some more of these inconsistencies upstream, I would opt to address this in a different PR

titivermeesch:
Fixed relevant CodeRabbit comments

melloware:
cc @wadakatu thoughts?

melloware:
@z4o4z wouldn't mind your thoughts on this too

z4o4z (commented):
Solid PR — the API mirrors the existing `preprocess` Mutator pattern, the modifier / nullary skip lists are well-targeted, and the `iso.datetime` / `iso.time` merge special-case is exactly right. Docs include a concrete example and there are 7 unit tests covering the parser-level injection.

One structural item I'd want resolved before merge (left inline near `makeParamsInjection`): `override.zod.params` is currently wired only into operation-schema generation (`generateZodRoute`). Component (reusable-schema) generation has no parallel call, so when paired with `generateReusableSchemas: true` the named component validators emit without injection — quietly missing the case where the feature compounds the most.

The other items left inline are improvements (signature length, generated-output verbosity, a fieldPath note for array/tuple/rest, missing test shapes), not blockers.

packages/zod/src/index.ts
```diff
@@ -1852,13 +1932,38 @@ const generateZodRoute = async (
...
+        output,
+        mutator: override.zod.params,
+        name: `${operationName}ZodParams`,
+        workspace: context.workspace,
+        tsconfig: context.output.tsconfig,
+      })
+    : undefined;
+
+  const pascalOperationName = pascal(operationName);
+  const makeParamsInjection = (
```
z4o4z, line 2023: **Major** — this `makeParamsInjection` (and the `paramsMutator` setup above) is wired only into `generateZodRoute`. Component / reusable-schema generation — `generateReusableSchemaSet` in `packages/orval/src/reusable-schemas.ts`, plus `generateZodSchemasInlineReusable` / `writeZodSchemasReusable` in `packages/orval/src/write-zod-specs.ts` — has no parallel call. So when a user pairs `override.zod.params` with `override.zod.generateReusableSchemas: true`, the operation wrappers get `zodParams(...)` injected but the actual named component validators (`Pet`, `Owner`, …) — referenced by N operations — emit without it. That's a quiet hole exactly where the feature's value compounds.

Two reasonable resolutions:
- Plumb a `ZodParamsInjection` through the reusable writer (likely with `operationId: ''` / a `location: 'schema'` sentinel so user-side `zodParams` can branch on it), mirroring how `GenerateReusableSchemaSetOptions` already threads `strict` / `coerce` / `isZodV4`.
- Or scope it out explicitly in the docs (the current copy reads 'into every generated validator' without qualification).

packages/zod/src/index.ts
```diff
@@ -1172,7 +1233,7 @@ ${Object.entries(objectArgs)
     if (fn === 'array') {
       const arrayArgs = args as ZodValidationSchemaDefinition;
       const value = arrayArgs.functions
-        .map((prop: [string, unknown]) => parseProperty(prop))
+        .map((prop: [string, unknown]) => parseProperty(prop, fieldPath))
```
z4o4z, line 1313: Minor — `array` (here), `tuple` (~L1252), and `rest` (~L1262) all recurse with `parseProperty(prop, fieldPath)` unchanged, so `{ tags: array<string> }` and `{ tags: string }` both yield the inner string's `fieldPath: ['tags']`. The `validator` field disambiguates the container (`'array'`) from the element (`'string'`), but two tuple positions can't be distinguished, and a user-side i18n key can't tell element-level constraints from container-level ones.

Probably acceptable — it matches Zod's own `issue.path`, which also drops array indices — but worth a sentence in the docs alongside the existing scope explanation, or a future-proofing `[]` / `0` segment behind an option.

packages/zod/src/index.ts
```diff
@@ -953,13 +953,49 @@ export const generateZodValidationSchemaDefinition = (
...
+]);
+
 export const parseZodValidationSchemaDefinition = (
   input: ZodValidationSchemaDefinition,
   context: ContextSpec,
   coerceTypes: boolean | ZodCoerceType[] = false,
   strict: boolean,
   isZodV4: boolean,
   preprocess?: GeneratorMutator,
+  paramsInjection?: ZodParamsInjection,
```
z4o4z, line 1054: Nit — `paramsInjection` is the 7th positional arg to `parseZodValidationSchemaDefinition`. The signature is already long (`coerceTypes, strict, isZodV4, preprocess, paramsInjection`); each new opt-in adds another. Not a blocker for this PR, but at this point an options-object overload would age better and let future fields land without breaking call sites.

packages/zod/src/index.ts
```diff
@@ -1205,14 +1268,31 @@ ${Object.entries(objectArgs)
         ? coerceTypes.includes(fn as ZodCoerceType)
         : COERCIBLE_TYPES.has(fn));
 
+    const formattedArgs = formatFunctionArgs(args);
+    const paramsArg = buildParamsArg(fn, fieldPath);
+    let combinedArgs: string;
```
z4o4z, line 1350: Two small nits:

1. `buildParamsArg` (above, ~L1037–1049) inlines `JSON.stringify(ctx)` verbatim at every call site, so for a 10-field schema with three constraints each the same `operationId` / `location` / `schemaName` prefix is repeated 30+ times in the emitted file. Functionally fine, but a follow-up that hoists a per-property const (`const _ctx_email = { … }`) would shrink output noticeably.

2. The `combinedArgs` branching here is correct but a small helper (`combineParamsArg(formattedArgs, paramsArg, fn)` returning the merged string) would localize the `PARAMS_MERGE_INTO_OPTIONS_VALIDATORS` special-case knowledge and let the `.coerce.${fn}(...)` and `.${fn}(...)` returns stay symmetric.

packages/zod/src/zod.test.ts
```diff
@@ -107,6 +107,310 @@ describe('parseZodValidationSchemaDefinition', () => {
...
+      ctx,
+      false,
+      false,
+      false,
+    );
+
+    expect(zod).not.toContain('zodParams(');
+    expect(zod).toContain('zod.string().email()');
+  });
+});
```
z4o4z, line 585: Coverage is good for the parser-level injection, but a few realistic shapes aren't exercised:

- No end-to-end test through `generateZodRoute` — the new wiring (the `makeParamsInjection('body', 'Body')` family, the `[...preprocess, paramsMutator]` aggregation in `mutators`, the `${pascalOperationName}${schemaSuffix}` derivation) is all untested.
- No reusable-schema test (which would surface the structural gap I flagged on `index.ts:1946`).
- No `tuple` / `rest` / `additionalProperties` test — only `array` and `union` exercise structural `fieldPath` propagation.
- No `regex` / `length` / `multipleOf` test — the docs claim these are in scope but only `min` / `max` are checked.

packages/zod/src/index.ts
titivermeesch: Agreed that this is indeed lacking. I'll make the required changes to also include the param injection in those places

packages/zod/src/index.ts
titivermeesch: I wouldn't touch this for now as it will increase the complexity of this PR beyond it's original scope, but happy to pick it up later on

packages/zod/src/index.ts
titivermeesch: I'll specify this in the docs

docs/content/docs/reference/configuration/output.mdx
```diff
@@ -1080,6 +1080,67 @@ Control which schemas are generated.
...
+      },
+    },
+  },
+});
+```
+
+```ts title="zod-params.ts"
+import { i18n } from './i18n';
+
+type ZodParamsContext = {
```
wadakatu: Could this context type be exported from orval instead of hand-written here? Everywhere else orval hands a user function an orval-specific object, the shape ships as an exported type (e.g. `override.operationName` → `OpenApiOperationObject` / `Verbs` via `@orval/core`). `preprocess` is the exception only because its signature is plain Zod. Exporting a `ZodParamsContext` would give users completion + surface breakage when the context evolves, and let this doc `import type` it instead of duplicating the shape.

docs/content/docs/reference/configuration/output.mdx
```diff
@@ -1080,6 +1080,67 @@ Control which schemas are generated.
...
+    .email(zodParams({ operationId: 'createUser', location: 'body', schemaName: 'CreateUserBody', fieldPath: ['email'], validator: 'email' })),
+});
+```
+
+Injection scope:
+
+- Applied to base types (`string`, `number`, `boolean`, `bigint`, `date`, `integer`), constraints (`min`, `max`, `gt`, `lt`, `multipleOf`, `regex`, `length`), formats (`email`, `url`, `uuid`, `hostname`, `datetime`, `time`), and `literal`, `enum`, `instanceof`, `stringFormat`.
+- Skipped on modifiers (`optional`, `nullable`, `nullish`, `default`, `describe`) and structural calls (`object`, `array`, `tuple`, `union`, `rest`, `passthrough`, `strict`).
+
+For static messages, return an object with a string `error`: `return { error: 'My message' }`. The function may return `undefined` to fall back to Zod defaults for a specific call.
```
wadakatu, line 1138: The `{ error }` shape is zod v4-only. orval supports both v3 and v4 and this injection runs on both paths, but I checked against zod 3.25.76 / 4.3.6: on v3 `{ error }` is silently ignored — no build- or parse-time error, the message just stays the default (`.email({ error: 'x' })` → `"Invalid email"`). `{ message }` works on both. Might be worth a one-liner noting the v3/v4 difference and suggesting `{ message }` for anyone supporting both versions.

packages/core/src/types.ts
```diff
@@ -743,6 +743,13 @@ export interface ZodOptions {
     body?: Mutator;
     response?: Mutator;
   };
+  /**
+   * Mutator referencing a function called once per emitted validator at schema
+   * construction time. It receives codegen-time context (operation, location,
+   * schema name, field path, validator name) and returns a Zod `params` object
+   * (e.g. `{ error: ... }`) that is appended as the trailing argument.
+   */
+  params?: Mutator;
```
wadakatu, line 758: Tiny naming flag (defer to you): `generate` / `coerce` / `preprocess` all key the path-parameter location as `param` (singular), and this context's `location` can be `'param'` too. The new `params` (plural) next to them leans on singular-vs-plural to separate two unrelated concepts. Keeping `params` is defensible since it's Zod's term for the validator's 2nd argument — just flagging in case `validatorParams` reads cleaner.

wadakatu (commented):
Thanks for the ping. Overall this fills a real gap — generator-time context (operationId / schemaName / fieldPath) is exactly what Zod's global error map can't disambiguate, the `preprocess`-style Mutator wiring keeps it consistent, and it's non-breaking when unset. +1 to @z4o4z's reusable-schema point as the thing to resolve before merge. A few smaller notes inline.

titivermeesch:
@wadakatu @z4o4z I believe I addressed all your comments. Do you mind looking at it again?

docs/content/docs/reference/configuration/output.mdx
titivermeesch: Added a section to the docs to specify this

packages/core/src/types.ts
titivermeesch: Clarified in the docs as I would like to keep the zod semantics here

packages/orval/src/write-zod-specs.ts
```diff
@@ -417,6 +437,10 @@ function generateZodSchemasInlineReusable(
   // Omit the zod import when concatenated into a file that already imports it
   // (inline single-mode output where the zod client emits `import * as zod`).
   const prefix = includeZodImport ? `import { z as zod } from 'zod';\n\n` : '';
+  // The `zodParams` import is registered as a mutator on the operation file
+  // (via `generateZodRoute`) and on the standalone single-mode wrapper (via
+  // `getInlineZodSchemasMutators`), so the import line is emitted by the
```
z4o4z: This comment references `getInlineZodSchemasMutators`, but that function doesn't exist anywhere in the repo — so the mechanism it describes ("the import line is emitted by the outer file builder") was never implemented for the inline path.

Concretely: the only caller of `generateZodSchemasInline` (`write-specs.ts:489`) passes just 3 args, so `paramsMutator` is always `undefined` here in inline-reusable mode. Component schemas inlined via this function therefore get **no** `zodParams(...)` injection — even though operations in the same file do (via `generateZodRoute`). Separate-files mode (`writeZodSchemasReusable`) injects correctly, so behavior diverges by output mode.

Either wire the mutator through at `:489` (hoist `schemasParamsMutator` so both branches share it) or, if inline injection is out of scope, remove the dead param and this comment and document the limitation.

packages/orval/src/write-zod-specs.ts
```diff
@@ -301,12 +313,18 @@ export function generateZodSchemasInline(
   builder: WriteZodSchemasInput,
   output: WriteZodOutputOptions,
   includeZodImport = true,
+  paramsMutator?: GeneratorMutator,
```
z4o4z, line 428: Dead parameter: `generateZodSchemasInline`'s only production caller (`write-specs.ts:489`) never passes a 4th arg, so `paramsMutator` is always `undefined` and the value threaded into `generateZodSchemasInlineReusable` → `generateReusableSchemaSet` is too. The plumbing is in place but nothing feeds it. Pass `schemasParamsMutator` from `write-specs.ts` to close the inline-reusable gap (see the comment below at line 442).

z4o4z (commented):
Re-review against `a6e9a223`. The previous round is well-handled — thanks for the thorough follow-up.

**Prior comments — all addressed ✅**
- The Major one (injection only in `generateZodRoute`, missing component schemas) is wired through `reusable-schemas.ts` + `write-specs.ts` for separate-files mode and covered by two new `reusable-schemas.test.ts` cases.
- `array`/`tuple`/`rest` `fieldPath` semantics are now documented in `output.mdx`.
- The end-to-end test gap is closed — 11 parser cases plus a real `generateZod` e2e test. Nice.
- Bonus: CodeRabbit's `operationId`-vs-`operationName` Major is genuinely fixed (the e2e test asserts the raw `createCat`), and `paramsMutator` is now in the returned `mutators` array.

**One new finding — Major: inline-reusable mode silently skips injection.**

The component-schema injection only fires in **separate-files** mode. In **inline-reusable** mode it's dead-wired:

- `shouldGenerateZodSchemasInline` (`write-specs.ts:288`) returns `true` for the common config `client: 'zod'` + `generateReusableSchemas: true` with no `output.schemas`.
- That path calls `generateZodSchemasInline(builder, output, includeZodImport)` at **`write-specs.ts:489`** with **no 4th arg**, so `paramsMutator` is `undefined` all the way into `generateZodSchemasInlineReusable` → `generateReusableSchemaSet({ paramsMutator: undefined })` → no params on the inlined component schemas.
- Net: in inline-reusable mode the operation request/response schemas get `zodParams(...)` (via `generateZodRoute`) but the shared `export const Pet = …` definitions they reference do **not** — whereas separate-files mode injects both. The documented option behaves differently depending on output mode.

Small fix: hoist the `schemasParamsMutator` resolution out of the `isZodSchemas` block and pass it at `:489`:

```ts
generateSchemasInline: needZodSchemasInline
  ? () => generateZodSchemasInline(builder, output, includeZodImport, schemasParamsMutator)
  : undefined,
```

Or, if inline-mode injection is intentionally out of scope, drop the dead param + comment and note the limitation in the docs — right now it reads as supported. See the two inline notes.

**Recommendation:** tests and prior feedback are solid; the inline-reusable gap is the one thing I'd want resolved (or explicitly documented) before merge — it's a real per-mode inconsistency for a mainstream config, and it ships with a comment pointing at a function that doesn't exist.

titivermeesch:
@z4o4z I believe I addressed your concerns

melloware:
@titivermeesch sorry just merged some stuff and you have 1 merge conflict!

titivermeesch:
Probably conflicting with my own changes. Looking into it

titivermeesch:
@melloware fixed the conflicts

melloware (approved):

Issue #1853, opened 2025-01-28
MSW mocks seem to "not consequently apply" "required:true" to anyOf elements that could be null


snext-twilde:
Hi,

not an expert in this, but stumbled upon this behavior:

with config
`output: { ..., override: { required: true }, ... }`

with openapi spec (from a pydantic/fastapi project with an Optional[str] attribute) :
```
"message": {
            "anyOf": [{ "type": "string" }, { "type": "null" }],
            "title": "Message"
          },
```

behavior: MSW mocks are null 50% of the time:
`faker.helpers.arrayElement([faker.string.alpha(20), null])`

Now I'd like to have a mock that is always populated and have many of these fields where e.g. some data is only returned on some conditions and the difference between [] and null is relevant.

Would it be consequent in your understanding of "required" to make required:true force the null types out of the anyOf for mocks? Or am I looking at this the wrong way? LMK if you'd like a PR, in case you like the idea.

Best,
Thomas

soartec-lab:
Hi @snext-twilde , We are currently rewriting the relevant source code in #1815, so we recommend that you wait for this PR to be merged before checking again.

AllieJonsson:
The way the code works now, `override.required` only affects if the item is required or not, not if it is nullable or not (making a difference between `undefined` and `null`). I'm unsure if setting `override.required = true` should make the item not nullable as well? 

Maybe we should add another override option for removing nullable from the mock? @soartec-lab

soartec-lab:
@AllieJonsson @melloware 

It is a make sense to have a property contain null only if `nullable: true` is specified.

melloware:
Yes that does make sense

snext-twilde:
true, but this - as far as I understand the approach - makes the msw mocks drastically less useful if you wrap your returned data in a generic "ApiResponse" that adds typical params like pagination, error code, error message and carries the data payload nested in a property that... yeah, in rare cases can be null (primarily in error cases).

For this scenario, the data payload is populated in only 50% of cases, in a nested payload structure with an optional parent property you see these properties only populated 25% of times. Using these mocks for tests means that you focus 50% on testing trivial error scenarios where no data is returned.

But thanks for the input, I can rephrase my question now: Would you consider it a useful feature to override nullable for specific properties in mock generation to force certain fields to always be populated?

AllieJonsson:
I think that is something we should add!

soartec-lab:
@snext-twilde 
[...]

PR #3476, opened 2026-05-28
feat(mock): add override.mock.nonNullable option

### LINKED ISSUE #3477: feat(mock): add override.mock.nonNullable to avoid random null in faker/MSW mocks
## Summary

Add an opt-in `override.mock.nonNullable` flag so faker and MSW mocks generate realistic non-null values for nullable OpenAPI fields by default, while TypeScript types stay `T | null` and tests can still pass `null` via `overrideResponse`.

Related: #1853 (discusses null randomization with `required: true`; this proposes a dedicated flag instead of overloading `required`).

## Problem

With `mock: { generators: [{ type: 'faker' }] }`, nullable fields are often generated as:

```ts
duration: faker.helpers.arrayElement([
  faker.helpers.arrayElement([faker.string.alpha({ length: { min: 10, max: 20 } }), null]),
  null,
]),
```

`override.mock.required: true` only forces **optional** properties to appear on the object. It does **not** stop nullable fields (or nullable array items) from randomly being `null`.

That leads to flaky frontend tests (Cypress, Vitest, Storybook) when forms or lists expect populated data.

There is no global way today to say "generate mocks without null" except per-property `override.mock.properties` or manual `overrideResponse` in every test.

## Proposed solution

```ts
override: {
  mock: {
    required: true,      // existing — always include optional properties
    nonNullable: true,   // new — never wrap mock values in arrayElement([value, null])
  },
}
```

| Flag | Behavior |
|------|----------|
| `required: false` (default) | Optional properties may be omitted via `arrayElement([value, undefined])` |
| `required: true` | All properties always emitted on the mock object |
| `nonNullable: false` (default) | Nullable fields use `arrayElement([value, null])` (current behavior) |
| `nonNullable: true` | Emit value only — no `null` branch in generated faker code |

`required` and `nonNullable` stay **independent**:

- `required` → presence (optional vs required in OpenAPI)
- `nonNullable` → value shape (skip null in mock generation only)

`override.mock.properties` overrides and `...overrideResponse` at the end of factories should continue to allow explicit `null` when a test needs it.

### Scope

- Nullable **properties** (OpenAPI 3.0 `nullable: true` and 3.1 `type: [..., 'null']`)
- Nullable **array items** (`items: { type: string, nullable: true }` / `type: ['string', 'null']`)
- Property-level mock overrides (`override.mock.properties`) should respect `nonNullable`
- Applies to both **faker** and **MSW** generators (shared mock pipeline)
- Default `false` — no breaking change

## Use case

Large Angular (or React) apps using Orval-generated faker mocks for:

- **E2E (Cypress)** — stable API responses so forms load with all fields set
- **Unit / component tests** — deterministic fixtures without `null` unless testing null explicitly
- **Storybook** — always-populated props for default stories

Example:

```ts
// orval.config.ts
export default defineConfig({
  petstore: {
    output: {
      mock: { generators: [{ type: 'faker' }] },
      override: {
        mock: {
          required: true,
          nonNullable: true,
        },
      },
    },
  },
});
```

```ts
// Still possible when testing null handling
getPetMock({ duration: null });
```

## Acceptance criteria

- [ ] `override.mock.nonNullable` documented next to `required`
- [ ] Unit tests for optional + nullable properties (default vs `nonNullable: true`)
- [ ] `required: true` + `nonNullable: true` → all keys present, no null randomization
- [ ] Nullable array items covered (OAS 3.0 and 3.1)
- [ ] `override.mock.properties` still works; runtime `overrideResponse` can set `null`
- [ ] Default `false` — existing consumers unchanged

## Implementation notes

Likely touch points:

- `packages/mock/src/faker/getters/object.ts` — property loop, optional/nullable branches
- `packages/mock/src/faker/getters/scalar.ts` — `getNullable` / nullable detection
- `packages/mock/src/faker/resolvers/value.ts` — `getNullable`, property overrides
- `packages/core/src/types.ts` — `OverrideMockOptions`

Hypenate:
## Summary

- Adds `override.mock.nonNullable` to skip `faker.helpers.arrayElement([value, null])` wrapping for nullable OpenAPI properties in faker and MSW mocks.
- Keeps `required` and `nonNullable` independent: `required` controls optional property presence; `nonNullable` only affects null randomization in generated code.
- Runtime `overrideResponse` can still pass explicit `null` values in tests.

Closes #3477

Related: #1853

## Test plan

- [x] `bun run lint` (all packages)
- [x] `bunx vitest run src/faker` in `@orval/mock`
- [x] Unit tests for default nullable behavior, `nonNullable: true`, and `required: true` + `nonNullable: true`
- [x] OpenAPI 3.0 / 3.1 nullable array items
- [x] `@orval/mock:typecheck`


melloware:
cc @wadakatu thoughts?

wadakatu:
Thanks @Hypenate — the opt-in `override.mock.nonNullable` itself looks solid and the test coverage is thorough. One thing I'd flag before this lands, though: with `nonNullable` **unset** (i.e. existing users on defaults), this changes the generated output for OpenAPI **3.0** `nullable: true` schemas.

The cause is widening the nullable check in `scalar.ts` from 3.1-only to `isNullableSchema` (3.0 `nullable: true` + 3.1 `type: [..., 'null']`):

```diff
- const isNullable = Array.isArray(item.type) && item.type.includes('null');
+ const isNullable = isNullableSchema(item);
```

Now the scalar layer wraps 3.0 nullable values too, on top of the existing wrapping in `object.ts`. Comparing generated mocks on `master` vs this branch, **both with default options**:

```
// optional, 3.0 nullable: true
master:  {tag: faker.helpers.arrayElement([faker.string.alpha(), null])}
#3476 :  {tag: faker.helpers.arrayElement([faker.helpers.arrayElement([faker.string.alpha(), null]), null])}   // double-wrapped (~75% null)

// required, 3.0 nullable: true
master:  {tag: faker.string.alpha()}                                       // never null
#3476 :  {tag: faker.helpers.arrayElement([faker.string.alpha(), null])}   // now null ~50%
```

So on defaults, 3.0 specs actually get *more* nulls (the opposite of the feature's intent), and required-nullable fields that were always populated can now be `null`.

The feature only needs `nonNullableOption` threaded through to suppress the wrapping — it doesn't need the detection widened. I'd suggest keeping `isNullable` as the 3.1-only check here so default behavior is unchanged. The 3.0 `nullable: true` detection gap (and the resulting double-wrap) is pre-existing behavior, so it's probably cleaner to address it in a separate PR rather than couple it to this one.

(Reproduced locally; happy to share the assertion-based test if useful.)

Hypenate:
@wadakatu 

I reverted scalar nullable detection to 3.1 null unions only so default output matches master for `nullable: true`. `nonNullableOption` still threads through to suppress wrapping when opted in. 
Added regression tests for the cases you described.

I HOPE this what you meant? :)

wadakatu:
Yes, exactly what I meant — thanks for the quick turnaround! 🙏

I pulled the branch and verified locally:
- **Defaults now match `master`** for 3.0 `nullable: true` — optional stays single-wrapped `arrayElement([..., null])`, and required-nullable is back to never-null.
- **The opt-in still works** — with `nonNullable: true`, 3.0 optional becomes `arrayElement([..., undefined])` (omittable but never null) and 3.1 required drops the null branch entirely.
- Existing mock tests + your new regression tests all green.

LGTM 👍 The pre-existing 3.0 detection gap / 3.1 double-wrap on defaults is unrelated to this PR, so good call keeping it separate.

wadakatu:
Filed the pre-existing 3.1 double-wrap as a separate issue for tracking: #3484 (reproduced end-to-end on `master` 8.13.0).

melloware (approved):

Issue #3478, opened 2026-05-28
Operation response wrapper collides with imported component schema name with generateReusableSchemas


titivermeesch:
### Description

With client: 'zod' and override.zod.generateReusableSchemas: true, orval emits operation files that fail to compile when the operation's pascalized name plus Response matches a component schema it imports.

Trigger: pascal(operationId) + 'Response' === <imported component schema name>, with generateReusableSchemas: true.

### Output client

zod

### Configuration (orval.config)

```ts
 import { defineConfig } from 'orval';

  export default defineConfig({
    petstore: {
      input: './spec.yaml',
      output: {
        target: './src/zod/petstore.ts',
        client: 'zod',
        mode: 'split',
        schemas: './src/zod/schemas',
        override: {
          zod: {
            generateReusableSchemas: true,
          },
        },
      },
    },
  });
```

### Environment

System:
OS: Linux 7.0 Arch Linux
CPU: (22) x64 Intel(R) Core(TM) Ultra 7 155H
Memory: 13.41 GB / 30.84 GB
Container: Yes
Shell: 4.7.1 - /bin/fish

Environment:
orval: 8.13.0
@orval/zod: 8.13.0
zod: 4.3.6
node: v22.12.0

### Expected behavior

When the wrapper name collides with an imported component, orval renames the wrapper. The Schema suffix matches a convention orval already uses elsewhere:

```typescript
import { ListPetsResponse } from "./schemas";

export const ListPetsResponseSchemaItem = ListPetsResponse;
export const ListPetsResponseSchema = zod.array(ListPetsResponseSchemaItem);
```

The import keeps its meaning. Non-colliding operations do not change.

### Actual behavior

 Generated operation file:

```typescript
import { ListPetsResponse } from "./schemas";
// ...

export const ListPetsResponseItem = ListPetsResponse;
export const ListPetsResponse = zod.array(ListPetsResponseItem);
```

tsc output: 'ListPetsResponseItem' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer. ts(7022)

### OpenAPI document (minimal, if applicable)

```yaml
  openapi: 3.1.0
  info:
    title: Petstore repro
    version: 0.0.0
  paths:
    /pets:
      get:
        operationId: listPets
        responses:
          '200':
            description: ok
            content:
              application/json:
                schema:
                  type: array
                  items:
                    $ref: '#/components/schemas/ListPetsResponse'
  components:
    schemas:
[...]

PR #3485, opened 2026-05-29
fix: deduplicate name collisions with generateReusableSchema (#3478)

titivermeesch:
Two zod + `generateReusableSchemas` bugs. Both make `tsc` reject the generated output.

### 1. Operation wrapper shadows imported component schema (#3478)

When `pascal(operationId)` plus a wrapper suffix matches a sibling component schema's import name, the wrapper's `export const` shadows the import with a self-referential initializer:

  ```ts
import { ListPetsResponse } from "./schemas";

export const ListPetsResponseItem = ListPetsResponse;            // self-ref
export const ListPetsResponse = zod.array(ListPetsResponseItem); // shadows import
```

tsc rejects this with TS7022 ("implicitly has type 'any' because it is referenced ... in its own initializer"). generateZodRoute now runs each operation export name through a collision check against allUsedRefs. On collision it appends Schema (with a numeric counter for chained collisions) so the import survives:

```ts
import { ListPetsResponse } from "./schemas";

export const ListPetsResponseSchemaItem = ListPetsResponse;
export const ListPetsResponseSchema = zod.array(ListPetsResponseSchemaItem);
```

### 2. Recursive reusable schema drops type-only imports

When generateReusableSchemas emits a recursive component to its own file, renderReusableSchemaEntry renders the TS type via the model generator (resolveValue) and pins the schema to it as `const X: zod.ZodType<X>`.

A name that the TS type body needs but the zod runtime collapses never enters usedRefs. The clearest case: `propertyNames: { $ref: KeyType }` becomes `zod.record(zod.string(), ...)` in the runtime expression, dropping KeyType. The rendered TS type still references KeyType, so tsc fails

Example from a JSON-Schema spec with propertyNames constraining a Record<> key:

```yaml
 Pet:
    type: object
    required: [id, name]
    properties:
      id: { type: integer }
      name: { type: string }
      relatives:
        type: object
        propertyNames: { $ref: '#/components/schemas/PetRelation' }
        additionalProperties:
          type: array
          items: { $ref: '#/components/schemas/Pet' }
  PetRelation:
    type: string
    enum: [parent, sibling, offspring]
```

```ts
// Before
import { z as zod } from 'zod';
// PetRelation: never imported

export type Pet = {
  id: number;
  name: string;
  relatives?: Partial<Record<PetRelation, Pet[]>>; // TS2304
};

export const Pet: zod.ZodType<Pet> = zod.object({ ... });
```

```ts
// After
import { z as zod } from 'zod';
import { PetRelation } from './petRelation';

export type Pet = {
  id: number;
  name: string;
  relatives?: Partial<Record<PetRelation, Pet[]>>;
 };

export const Pet: zod.ZodType<Pet> = zod.object({ ... });
```

#3467 introduced the recursive-schema emission path; this bug was already present and surfaced while verifying #3478 against a real spec.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented generated export name collisions that could shadow imported component schemas or cause self-referential initializer errors; export names now adapt to avoid redeclaration.
  * Fixed recursive/reusable-schema emission so required type-only imports are included in split mode and no unresolved reference sentinels appear in outputs.

* **Tests**
  * Added regression tests for naming collisions (including double-collision fallbacks) and recursive-schema generation across single-file and split modes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/orval-labs/orval/pull/3485?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/orval/src/generate-spec.test.ts
+++ b/packages/orval/src/generate-spec.test.ts
@@ -268,6 +268,98 @@ describe('generateSpec - generateReusableSchemas recursive ($ref to self)', () =
       await rm(workspace, { recursive: true, force: true });
     }
   });
+
+  // The split-mode writer merges `extraImports` (TS-body refs the zod runtime
+  // collapses, e.g. a `propertyNames` $ref) into per-file imports. Inline
+  // single-file mode resolves everything in the same file and discards
+  // `extraImports`; a regression that started threading those names into the
+  // inline output as bogus `from './X'` lines wouldn't be caught by the
+  // split-mode test alone.
+  const RECURSIVE_PROPERTY_NAMES_SPEC: OpenApiDocument = {
+    openapi: '3.1.0',
+    info: { title: 'RecursivePropertyNames', version: '1.0.0' },
+    paths: {
+      '/trees': {
+        get: {
+          operationId: 'listTrees',
+          responses: {
+            '200': {
+              description: 'ok',
+              content: {
+                'application/json': {
+                  schema: {
+                    type: 'array',
+                    items: { $ref: '#/components/schemas/Tree' },
+                  },
+                },
+              },
+            },
+          },
+        },
+      },
+    },
+    components: {
+      schemas: {
+        Tree: {
+          type: 'object',
+          properties: {
+            children: {
+              type: 'object',
+              propertyNames: { $ref: '#/components/schemas/RelationType' },
+              additionalProperties: {
+                type: 'array',
+                items: { $ref: '#/components/schemas/Tree' },
+              },
+            },
+          },
+        },
+        RelationType: {
+          type: 'string',
+          enum: ['parent', 'child', 'sibling'],
+        },
+      },
+    },
+  };
+
+  it('emits no sibling imports in single mode even when the recursive TS body references another component', async () => {
+    const workspace = await createTempWorkspace();
+    const targetFile = path.join(workspace, 'zod.ts');
+
+    try {
+      const options = await normalizeOptions(
+        {
+          input: { target: RECURSIVE_PROPERTY_NAMES_SPEC },
+          output: {
+            target: './zod.ts',
+            mode: 'single',
+            client: 'zod',
+            override: { zod: { generateReusableSchemas: true } },
+          },
+        },
+        workspace,
+      );
+
+      await generateSpec(workspace, options);
+
+      const content = await fs.readFile(targetFile, 'utf8');
+
+      // Both component schemas are emitted inline...
+      expect(content).toContain(
+        'export const Tree: zod.ZodType<Tree> = zod.object(',
+      );
+      expect(content).toContain('export const RelationType = zod.enum(');
+      // ...and the recursive type references the inline RelationType through
+      // the `Record<>` key.
+      expect(content).toMatch(/Record<\s*RelationType/);
+      // No sibling imports leaked through `extraImports`: every component
+      // (including `RelationType`) lives in this same file.
+      expect(content).not.toMatch(/from '\.\/RelationType'/);
+      expect(content).not.toMatch(/from '\.\/Tree'/);
+      expect(content).not.toContain('__REF_');
+    } finally {
+      await rm(workspace, { recursive: true, force: true });
+    }
+  });
 });
 
 describe('generateSpec - generateReusableSchemas inline (pure-$ref operations)', () => {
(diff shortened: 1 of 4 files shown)


titivermeesch:
@z4o4z FYI

packages/zod/src/index.ts
```diff
@@ -2052,42 +2052,77 @@ const generateZodRoute = async (
...
+  // their PascalCase name from a sibling schemas module. When an operation's
+  // own pascalized wrapper name (e.g. `ListPetsResponse` from operationId
+  // `listPets`) matches an imported ref, the generated `export const` shadows
+  // the import and produces a self-referential initializer that TS rejects
+  // with TS7022. Detect the collision and append `Schema` (with a counter for
+  // further collisions) so the import keeps its meaning. For the array case
+  // both the wrapper and its `Item` companion are checked together so they
+  // stay in sync. The original name is preserved when there is no collision,
+  // so non-colliding operations are unaffected.
+  const allocateExportName = (baseName: string, hasItem: boolean): string => {
```
z4o4z, line 2073: Minor — `allocateExportName` checks only `allUsedRefs`, not the names it has itself handed out earlier in this same `generateZodRoute` call. Up to 6 wrappers are allocated per call (`Params`, `QueryParams`, `Header`, `Body`, one per response status); each is resolved independently and the chosen name isn't reserved. Same-call collision is unlikely (the suffixes are distinct), but it's an implicit assumption. A local reservation set closes it:

```ts
const localTaken = new Set(allUsedRefs);
const allocateExportName = (baseName, hasItem) => {
  const collides = (n) => localTaken.has(n) || (hasItem && localTaken.has(`${n}Item`));
  // … after resolving `candidate`:
  localTaken.add(candidate);
  if (hasItem) localTaken.add(`${candidate}Item`);
  return candidate;
};
```

While here: a one-line comment that the component side stays canonical and the *wrapper* accepts the `Schema` suffix would document the rename-policy asymmetry (it's the right call — components are shared, wrappers are local — but currently implicit).

packages/orval/src/write-zod-specs.ts
```diff
@@ -195,22 +210,31 @@ function renderReusableSchemaEntry(
...
-      `${consts}export type ${entry.name} = ${typeBody};\n\n` +
-      `export const ${entry.name}: zod.ZodType<${entry.name}> = ${entry.zod};\n\n` +
-      `export type ${entry.name}Output = zod.output<typeof ${entry.name}>;`
-    );
+    const resolved = schema
+      ? resolveValue({ schema, name: entry.name, context })
+      : undefined;
+    const typeBody = resolved ? resolved.value : 'unknown';
+    const extraImports = new Set<string>();
+    for (const imp of resolved?.imports ?? []) {
```
z4o4z, line 237: Minor — `resolveValue.imports` entries are `{ name, schemaName, alias? }`, and this harvest uses `imp.name` only. An aliased import (rare — e.g. `generateVerbImports`' `Error → ErrorSchema` path) would emit `import { name }` instead of `import { name as alias }`. Doesn't affect the `propertyNames` repro, but it's a latent footgun if `resolveValue` ever takes an aliasing path for a recursive body. Carrying `alias` through `extraImports` (or asserting none is expected here) would be safer.

packages/orval/src/generate-spec.test.ts
```diff
@@ -343,3 +343,97 @@ describe('generateSpec - generateReusableSchemas inline (pure-$ref operations)',
...
+          properties: {
+            id: { type: 'string', format: 'uuid' },
+            name: { type: 'string' },
+          },
+        },
+      },
+    },
+  };
+
+  it('renames the operation wrapper to avoid shadowing the imported component schema', async () => {
```
z4o4z, line 488: Test gap — only the single `Schema` fallback is exercised. A spec with **both** a `ListPetsResponse` component and a `ListPetsResponseSchema` component would prove the `Schema1` numeric-counter path in `allocateExportName` actually fires (right now nothing covers `counter >= 1`).

z4o4z (commented):
Two distinct `tsc`-rejecting bugs in the `generateReusableSchemas` + zod pipeline, both with verifiable repros, well-scoped fixes, and end-to-end tests against the right writers. LGTM overall — I'd merge after one defensive tweak and one extra test, neither a blocker.

Strong moves worth calling out: the `hasItem`-aware companion check lives in the helper rather than at each call site; the numeric counter handles chained collisions; the `componentNames` filter on `extraImports` guards against incidental scalar identifiers leaking through `resolveValue.imports`; and the self-ref filter keeps `Tree.ts` from importing `./Tree`.

Inline notes below — all minor or test-coverage. The collision fix correctly renames the operation *wrapper* (local) rather than the *component* (shared), which is the right policy.

packages/orval/src/write-zod-specs.ts
```diff
@@ -410,8 +434,11 @@ function generateZodSchemasInlineReusable(
 
   const rewritten = rewriteReusableSchemas(entries);
 
+  // Single-file inline mode emits every component schema in one file, so
```
z4o4z, line 513: The inline single-file writer correctly discards `extraImports` (everything resolves in one file) — good. But that choice is currently unanchored by a test: a regression that started leaking spurious sibling imports into the inline output wouldn't be caught. A small `generate-spec` test (mode `single` + a recursive schema with a `propertyNames` `$ref`) asserting the file compiles with no stray `import … from './X'` would pin it, complementing the split-mode test you added in `write-zod-specs.test.ts`.

packages/orval/src/write-zod-specs.test.ts
```diff
@@ -482,6 +482,80 @@ describe('writeZodSchemas with generateReusableSchemas', () => {
 
     await fs.remove(root);
   });
+
+  it('imports type-only refs that the recursive TS body uses but the zod runtime drops', async () => {
```
z4o4z, line 487: Nice end-to-end coverage of `extraImports` merging, the self-ref filter, and the `componentNames` guard — this is the test that exercises bug #2's real path. Two small additions would round it out: assert `RelationType.ts` itself is still emitted (not just imported), and a negative assertion that the runtime `zod.record(zod.string(), …)` did collapse the key (i.e. `RelationType` is absent from the `zod.object({...})` body) — that's what makes the type-only import necessary in the first place, so pinning it documents the bug's mechanism.

titivermeesch:
@z4o4z thanks for the review. I addressed your points. Please have a look again.
@melloware any thoughts on this?

packages/zod/src/index.ts
```diff
@@ -2052,42 +2052,93 @@ const generateZodRoute = async (
...
+    }
+    let counter = 0;
+    let candidate = `${baseName}Schema`;
+    while (collides(candidate)) {
+      counter += 1;
+      candidate = `${baseName}Schema${counter}`;
+    }
+    reserve(candidate);
+    return candidate;
+  };
```
z4o4z, line 2092: 🟡 **The wrapper rename isn't reflected in the clients that reconstruct these names.** `allocateExportName` renames `Params`/`QueryParams`/`Header`/`Body`/`…Response` on collision, but the **hono** and **mcp** clients re-derive those exact names from `pascalOperationName` and import them from the generated zod file:
```
packages/hono/src/index.ts:226  zValidator('response', ${pascalOperationName}Response)
packages/hono/src/index.ts:282  specifiers.push(`${pascalOperationName}Response`)   // + Params/QueryParams/Header/Body
packages/mcp/src/index.ts:198   pathParams: ${pascalOperationName}Params            // + QueryParams/Body/Header
```
In the intersection {`generateReusableSchemas` on} × {a component named identically to a wrapper} × {client = hono|mcp}, the zod file now exports e.g. `…ResponseSchema`, while hono still emits `import { …Response }` / `zValidator('response', …Response)`. Since the collision is *with* a component ref, `…Response` then resolves to the **component schema** instead of the operation wrapper — hono validates against the wrong schema (or the import dangles). Not a regression (pre-fix that case was TS7022 and didn't compile), but the fix is incomplete for those clients and fails silently.

Could you confirm whether `generateReusableSchemas` can be active for hono/mcp? If so, the resolved wrapper names should be surfaced to those generators rather than each re-deriving them, plus a hono+reusable+collision regression test. If reusable schemas is effectively zod-standalone, a one-line note (or guard) here would close the loop. (`effect` emits its own identical `${pascalOperationName}Body` wrappers inline at index.ts:1511 — same latent collision, separate generator, candidate follow-up.)

Minor, same block: `paramsName`/`queryParamsName`/`headerName`/`bodyName` are allocated unconditionally even when their `*.zod` is empty and nothing is emitted — harmless (reserved names can't shadow a later distinct-suffix wrapper), just a couple of wasted reservations.

packages/orval/src/write-zod-specs.ts
```diff
@@ -195,27 +224,98 @@ function renderReusableSchemaEntry(
...
+  importExt: string;
+}): string {
+  const importsByName = new Map<string, ExtraImport>();
+  for (const name of usedRefs) {
+    if (name === entryName) continue;
+    importsByName.set(name, { name });
+  }
+  for (const imp of extraImports) {
+    if (imp.name === entryName || !componentNames.has(imp.name)) continue;
+    importsByName.set(imp.name, imp);
```
z4o4z, line 307: 🟢 **The "aliased `extraImports` entry overrides a bare `usedRefs` entry" rule is the wrong tiebreaker for a (currently impossible) case.** If the same component name appeared in both `usedRefs` (unaliased — the zod runtime expression references it bare) and `extraImports` (aliased), this `set` would emit `import { X as Y }`, leaving the runtime's bare `X` unbound (TS2304). The JSDoc correctly notes no `resolveValue` producer surfaces aliases today, so it can't happen — but if you want it robust to a future aliasing producer, skip the alias when `name` is also present in `usedRefs` (the runtime binding has to win), or assert the precondition. As-is, a comment acknowledging it would do.

z4o4z (commented):
Solid, well-tested bugfix — both TS-error fixes are correct and there's no regression for the non-reusable path (`localTaken` seeds from `allUsedRefs`, which is empty unless `useReusableSchemas`, so names are unchanged otherwise). The test authorship is excellent: the regression tests pin the *mechanism* (e.g. asserting `zod.record(zod.string())` to prove why the sibling is absent from `usedRefs`), cover single + split mode and the double-collision counter branch. Two notes inline; the first is the one I'd want addressed or explicitly acknowledged before merge.

packages/zod/src/index.ts
titivermeesch: `generateReusableSchemas` is only for zod right? You implemented it originally :)

I would not touch it if it's only to be in line with other generators. I would only pick it up if there is a need for those generators. Let me know if you still want something to be addressed here.

z4o4z (approved):

packages/orval/src/write-zod-specs.ts
titivermeesch: The case you're describing can't fire today, so the override is dead code either way. The original ask was to carry `alias` through, which is implemented now. The tiebreaker for an overlap that no producer reaches feels like bikeshedding to me.

melloware (approved):

PR #3498, opened 2026-05-31
fix(core): preserve original input as cache key in pascal() to fix cache misses for uppercase inputs

### LINKED ISSUE #3493: bug(core): cache key mutation in `pascal()` causes cache misses for uppercase inputs
## Description

In `packages/core/src/utils/case.ts`, the `pascal()` function mutates the variable `s` (via `s = low(s)`) before writing the result into the `pascalMemory` cache. The initial cache lookup uses the original value of `s`, but the write uses the lowercased value. This means that for all-uppercase inputs (e.g. `'SNAKE_CASE'`), a cache hit will never occur on subsequent calls.

## Steps to Reproduce

Call `pascal('SNAKE_CASE')` twice — the second call will not return from cache.

## Expected Behavior

Repeated calls with the same input should return the cached result.

## Related

Identified during review of #3488 by @daugvinasr.

_Backlink:

xxiaoxiong:
## Problem

Calling `pascal('SNAKE_CASE')` twice results in a cache miss on the second call because `pascalMemory[s]` caches using the **mutated** `s` (lowercased via `s = low(s)` on line 105), not the original input. Since `'SNAKE_CASE' !== 'snake_case'`, the second lookup falls through and recomputes the result.

## Root Cause

In `packages/core/src/utils/case.ts`, the `pascal()` function mutates the parameter `s` (via `s = low(s)`) before writing to the cache at `pascalMemory[s] = pascalWithUnderscore`. After mutation, `s` holds the lowercased value, so the cache key stored is `'snake_case'` instead of `'SNAKE_CASE'`. Subsequent calls with the original uppercase input miss the cache.

## Changes

- Preserve the original `s` value in a `cacheKey` variable before mutation
- Use `pascalMemory[cacheKey]` instead of `pascalMemory[s]` for cache storage

## Tested

- The fix is a minimal 2-line change (one line added, one line modified)
- Cache now correctly returns cached results for all-uppercase inputs
- Existing pascal behavior is unchanged since the mutation flow is preserved

Closes #3493

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal caching mechanism in utility functions to improve performance and code maintainability without affecting external behavior.

### DIFF
--- a/packages/core/src/utils/case.ts
+++ b/packages/core/src/utils/case.ts
@@ -100,6 +100,7 @@ export function pascal(s = '') {
   }
 
   const isStartWithUnderscore = s.startsWith('_');
+  const cacheKey = s;
 
   if (regexps.upper.test(s)) {
     s = low(s);
@@ -113,7 +114,7 @@ export function pascal(s = '') {
     ? `_${pascalString}`
     : pascalString;
 
-  pascalMemory[s] = pascalWithUnderscore;
+  pascalMemory[cacheKey] = pascalWithUnderscore;
 
   return pascalWithUnderscore;
 }


melloware (approved):

Exagone313:
@melloware Hello, you have merged a PR made by a bot.

It is sending PR on a lot of projects, this is completely automated: <https://github.com/search?q=author%3Axxiaoxiong+is%3Apr&s=created&type=pullrequests>

This could be used to gain access to insecure GitHub Actions setups. By default, approval on pull requests is required only for first-time contributors. If that is the case for this project, you might need to change the setting `Approval for running fork pull request workflows from contributors`.

Thanks.

PR #3501, opened 2026-05-31
fix(mock): avoid double-wrapping null branch for required nullable scalars

### LINKED ISSUE #3484: Mock: OpenAPI 3.1 `type: [..., 'null']` is double-wrapped with `null` in faker mocks (`arrayElement([arrayElement([value, null]), null])`)
## Description

When a property uses an OpenAPI 3.1 nullable type union (`type: ['string', 'null']`), the faker/MSW mock wraps the value in `faker.helpers.arrayElement([..., null])` **twice**, producing a nested `arrayElement([arrayElement([value, null]), null])`. This is pre-existing behavior on `master` (8.13.0) and is independent of `override.mock.nonNullable`.

Surfaced during review of #3476, but it reproduces on plain `master` with default options.

## Output client

`react-query` with `mock: true`. The mock pipeline is shared, so this affects faker and MSW output for any client.

```ts
export default defineConfig({
  petstore: {
    input: './api.yaml',
    output: { target: './gen.ts', mock: true, client: 'react-query' },
  },
});
```

## Environment

- orval: `8.13.0` (current `master`)
- No `override.mock.*` options set (defaults)

## Expected behavior

A single null branch:

```ts
tag: faker.helpers.arrayElement([faker.string.alpha(), null])
```

## Actual behavior

Running orval `8.13.0` on the spec below generates a nested double wrap (null probability ~75% instead of 50%):

```ts
export const getGetPetResponseMock = (overrideResponse: Partial<Extract<Pet, object>> = {}): Pet =>
  ({tag: faker.helpers.arrayElement([faker.helpers.arrayElement([faker.string.alpha({length: {min: 10, max: 20}}), null]), null]), ...overrideResponse});
```

## OpenAPI document (minimal)

```yaml
openapi: 3.1.0
info: { title: repro, version: 1.0.0 }
paths:
  /pet:
    get:
      operationId: getPet
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema: { $ref: '#/components/schemas/Pet' }
components:
  schemas:
    Pet:
      type: object
      required: [tag]
      properties:
        tag:
          type: [string, 'null']
```

## Additional context

Root cause: two layers each detect the 3.1 null union independently and each wraps the value.

- `packages/mock/src/faker/getters/scalar.ts:148` — `const isNullable = Array.isArray(item.type) && item.type.includes('null');` → wraps the leaf value via `getNullable` → inner `arrayElement([value, null])`.
- `packages/mock/src/faker/getters/object.ts:208-216` — the property-level branch re-checks `Array.isArray(prop.type) && prop.type.includes('null')` and wraps the already-wrapped value again → outer `arrayElement([..., null])`.

**Correction — both 3.0 and 3.1 are affected.** An earlier draft of this issue claimed OpenAPI 3.0 `nullable: true` wraps only once. That is wrong: orval upgrades 3.0 inputs to 3.1 via `@scalar/openapi-parser` **before** the mock getters run, so a 3.0 `nullable: true` scalar arrives as `type: [..., 'null']` and trips **both** layers exactly like a native 3.1 union. (The earlier claim assumed `scalar.ts` still saw `nullable: true` instead of the upgraded array — it does not. The #3476 scalar-check change is a separate, since-reverted matter.) Reconfirmed on current `master` (`8.14.0`) for both 3.0 and 3.1.

Scope: only **required** nullable scalars double-wrap. An *optional* nullable scalar instead nests under the `undefined` (omit) branch — `arrayElement([arrayElement([value, null]), undefined])` — which is a valid `value | null | undefined` distribution and is left as-is.

Likely fix direction: avoid re-wrapping at the property layer when the scalar layer has already applied the null branch (let a single layer own the null wrapping).

wadakatu:
## What

Fixes the double-wrapped `null` branch in faker/MSW mocks for **required** nullable scalar properties.

A required property typed as an OpenAPI 3.1 nullable union (`type: [<scalar>, 'null']`) — the same shape an OpenAPI 3.0 `nullable: true` scalar is upgraded to by `@scalar/openapi-parser` before mock generation — was wrapped with a `null` branch **twice**, pushing `null` to ~75% instead of ~50%:

```ts
// before
tag: faker.helpers.arrayElement([
  faker.helpers.arrayElement([faker.string.alpha(), null]),
  null,
])
// after
tag: faker.helpers.arrayElement([faker.string.alpha(), null])
```

## Why

Two layers each detected the null union independently and each added a branch:

- `packages/mock/src/faker/getters/scalar.ts` wraps the leaf value via `getNullable` → inner `arrayElement([value, null])`.
- `packages/mock/src/faker/getters/object.ts` re-checks the property type and wraps the already-wrapped value again → outer `arrayElement([..., null])`.

## How

Let the scalar getter own the null branch. It now flags the returned `MockDefinition` with a new optional `nullWrapped` field when it has already applied `getNullable`, and the object property layer skips its own wrap when that flag is set. Boolean (and number `enum`/`const`, where the scalar getter does not keep the wrap) stay bare, so the object layer still contributes their single null branch — no behavior change for those.

Only the **required** path is changed; the *optional* nullable branch (`arrayElement([value, undefined])` / `arrayElement([value, null])` for omit) is intentionally untouched, as its `value | null | undefined` nesting is a valid distribution.

## Tests

- New regression spec `tests/specifications/issue-3484.yaml` (required nullable `string` / `integer` / `string`-enum / `boolean`) wired into `mock.config.ts`, with a focused assertion in `api-generation.spec.ts`.
- Two existing snapshots that already exhibited the bug now collapse to a single wrap: `default/all-of` (`category`) and `mock/recursive-discriminator-allof` (`BaseProp`).
- `format:check`, `build`, `typecheck`, `lint`, `test`, `test:snapshots`, and the generated-output typecheck + mock verification (`orval-tests build`) all pass.

Closes #3484

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed double-wrapped null handling in generated mocks so nullable scalars now produce a single null choice and redundant null-wrapping is removed.

* **Tests**
  * Added a regression test ensuring single null-branch behavior and updated/added generated mock snapshots.

* **Chores**
  * Added a mock generation config entry and a new OpenAPI fixture to reproduce and validate the regression.

* **Samples**
  * Updated generated MSW sample mocks to reflect corrected nullable-value selection.


melloware (approved):

Hypenate:
Good catch! 

I ran into this issue at work :D 
Looking forwards to a new release (again?!)

PR #3507, opened 2026-06-01
fix(core): use jsStringEscape for enum/const values to prevent backslash corruption

### LINKED ISSUE #3505: Generated enum/const values are not JS-escaped (backslash values corrupt or break generated output)
### Description

Generated enum/`const` **values** are emitted with the quote-only `escape()` helper instead of `jsStringEscape()`. Any enum value containing a backslash (e.g. PHP class-string discriminators like `App\Models\Document`) is emitted with single backslashes:

```ts
export const VisitableType = {
  'App\Models\Document': 'App\Models\Document',
  'App\Models\Template': 'App\Models\Template',
} as const;
```

Two effects:

1. **Silent value corruption** — `'App\Models\Document'` evaluates at runtime to `"AppModelsDocument"` (`\M`/`\D` are not valid escapes, so the backslash is dropped), so it never equals the real API value. Equality checks against the enum silently fail.
2. **Invalid / unparseable output** — a value ending in a backslash (e.g. a Windows path prefix `C:\logs\`) is emitted as `'C:\logs\'`, where the final `\'` escapes the closing quote, so the string literal does not terminate and the file fails to parse.

orval already exposes `jsStringEscape()` and applies it to enum names/descriptions (`getEnumNames`/`getEnumDescriptions`), but the value emission paths still use the quote-only `escape()`.

Reproduced on orval 8.14.0. A minimal OpenAPI document that reproduces the issue is provided below.

### Output client

other

```ts
export default {
  poc: {
    input: './openapi.json',
    output: {
      target: './src/api.ts',
      client: 'fetch',
      mode: 'single',
    },
  },
};
```

### Environment

```
System:
  OS: macOS 26.5
  CPU: (12) arm64 Apple M3 Pro
  Shell: 5.9 - /bin/zsh
npmPackages:
  orval: 8.14.0 => 8.14.0
```

### Expected behavior

Enum/`const` values containing backslashes should be JS-escaped so the generated literal equals the source value at runtime and the output is always valid TypeScript, e.g.:

```ts
export const VisitableType = {
  'App\\Models\\Document': 'App\\Models\\Document',
} as const;
```

### Actual behavior

- `'App\Models\Document'` -> runtime value `"AppModelsDocument"` (backslashes dropped, mismatch vs API).
- A value ending in a backslash emits `'C:\logs\'`, which escapes the closing quote and produces invalid, unparseable output.

```yaml
openapi: 3.0.3
info:
  title: orval enum value escaping repro
  version: 1.0.0
paths:
  /ping:
    get:
      operationId: ping
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                type: object
                properties:
                  visitableType:
                    $ref: '#/components/schemas/VisitableType'
                  directoryPrefix:
                    $ref: '#/components/schemas/DirectoryPrefix'
components:
  schemas:
    VisitableType:
      type: string
      enum:
        - 'App\Models\Document'
        - 'App\Models\Template'
    DirectoryPrefix:
      type: string
      enum:
        - 'C:\logs\'
        - 'C:\tmp\'
```

### Additional context

Suggested fix: switch the enum/`const` **value** sites from `escape()` to the already-exported `jsStringEscape()` (the helper orval already uses for names/descriptions):

- `packages/core/src/getters/enum.ts` — `getEnumUnionFromSchema` (~L318)
- `packages/core/src/getters/object.ts` — enum union values (~L43), const property value (~L52), const literal (~L353), const value (~L560)
- `packages/core/src/getters/scalar.ts` — scalar enum items (~L203, ~L329)

enekesabel:
## Summary

Fixes #3505.

Enum and const string values were emitted using the quote-only `escape()` helper, which does not handle backslashes. This caused two issues:

- **Silent value corruption** — `'App\Models\Document'` evaluates at runtime to `"AppModelsDocument"` (backslashes dropped), so it never equals the real API value.
- **Invalid output** — a value ending in a backslash (e.g. `'C:\logs\'`) escapes the closing quote, producing an unterminated string literal that fails to parse.

The fix switches all enum/const value emission sites from `escape()` to the already-exported `jsStringEscape()`, which properly escapes backslashes, quotes, and other special characters.

### Changed files (core fix)

| File | Sites | What changed |
|------|-------|-------------|
| `packages/core/src/getters/enum.ts` | 1 | `getEnumUnionFromSchema` |
| `packages/core/src/getters/object.ts` | 4 | `getPropertyNamesEnumKeyType` (enum + const), `getObject` const literal + const return |
| `packages/core/src/getters/scalar.ts` | 2 | `getScalar` string + integer enum items |

### Before / After

**Input:** `enum: ['App\Models\Document']`

| | Generated output | Runtime value |
|---|---|---|
| Before | `'App\Models\Document'` | `"AppModelsDocument"` |
| After | `'App\\Models\\Document'` | `"App\Models\Document"` |

## Test plan

- [x] 3 unit tests added in `enum.test.ts` — verify `getEnumUnionFromSchema` and `getEnumImplementation` produce double-escaped backslashes
- [x] 1 integration test added in `api-generation.spec.ts` — verifies end-to-end generated output for the issue's repro spec
- [x] Spec file + config entry + snapshots added for `issue-3505`
- [x] All 1945 core unit tests pass
- [x] All 4362 snapshot/integration tests pass
- [x] Build succeeds, prettier + commitlint pass

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected escaping of backslashes in enum/string values so generated TypeScript emits valid JS-escaped string literals.

* **Tests**
  * Added test(s) validating proper double-escaping of backslashes in generated types and constants.

* **Chores**
  * Added a new OpenAPI test fixture and corresponding generation config to cover enum values containing backslashes.


melloware (changes requested):
Looks like Lint failures

enekesabel:
Yeah, had to apply some trickery to make the linter happy :D 
Locally, I reverted the fix after changing the tests to verify they would still catch the issue - they do.

melloware:
Running the build again.

melloware (approved):

melloware:
@enekesabel i am going to revert this change

PR #3514, opened 2026-06-01
feat(mock): add arrayItems faker option for reusable array item mocks

### LINKED ISSUE #3513: feat(faker): export reusable mock factories for array item schemas in operation responses
## Problem

When Orval generates faker mocks for paginated/list responses (e.g. `ResponseWrapperOfTenantResponseModelDto` with a `value: TenantResponseModelDto[]` field), the array item object mock is inlined inside `.map(() => ({ ... }))`.

This makes it hard to reuse a single tenant mock in unit tests, Storybook, or other fixtures without copying faker logic or manually extracting helpers.

`schemas: true` only emits factories for entries under `components/schemas`. In practice, array item types often appear only inside response wrappers, or consumers want item-level factories without opting into consolidated schema mocks for every component schema.

## Proposed solution

Add an opt-in faker generator option `arrayItems: true` that lifts object-like array item schemas into exported factories:

- `$ref` items → `get<SchemaName>Mock` (shared across operations)
- inline object items → `get<OperationId>Response<PropertyName>ItemMock` typed as Orval's generated item alias (e.g. `GetTenants200ValueItem`)

Operation response factories then delegate to these item factories inside `.map()`.

## Example

```ts
mock: {
  generators: [{ type: 'faker', arrayItems: true }],
}
```

```ts
export const getTenantResponseModelDtoMock = (override = {}) => ({ ... });
export const getGetTenantsByRefResponseMock = () => ({
  value: Array.from(...).map(() => ({ ...getTenantResponseModelDtoMock() })),
  count: faker.number.int(),
});
```

## Notes

- Complements `schemas: true` (which still preferred for `$ref` items when consolidated schema factories are desired)
- Primitive array items (string, number, etc.) are unchanged

Hypenate:
## Summary

- Adds opt-in faker generator option `arrayItems: true` to export reusable mock factories for object-like array item schemas in operation responses
- `$ref` array items emit `get<SchemaName>Mock` (deduplicated across operations); inline object items emit `get<OperationId>Response<PropertyName>ItemMock` typed to match Orval's generated item aliases (e.g. `GetTenants200ValueItem`)
- Operation response mocks delegate to these item factories inside `.map()` instead of inlining the full object body

Closes #3513

## Test plan

- [x] Unit tests for array item factory extraction (`array-item-factory.test.ts`)
- [x] All `@orval/mock` package tests pass
- [x] Added `faker-array-items` spec + snapshot covering inline and `$ref` array items
- [ ] CI green

## Usage

```ts
mock: {
  generators: [{ type: 'faker', arrayItems: true }],
}
```

```ts
import { getTenantResponseModelDtoMock } from './api/tenants.faker';

const tenant = getTenantResponseModelDtoMock({ name: 'Acme' });
```

## Summary by CodeRabbit

* **New Features**
  * Added an `arrayItems` option to the Faker generator to emit reusable mock factory functions for object-like array items in responses.

* **Documentation**
  * Updated the Faker guide with an "Array Item Factories" section and options table entry explaining behavior and interactions.

* **Tests / Examples**
  * Added tests, OpenAPI spec, snapshots and mock config showcasing array-item factories.

* **Improvements**
  * More stable factory naming, deduplication, and item-context handling for generated mocks.


melloware (changes requested):
Looking like linter issue

melloware:
cc @wadakatu

wadakatu:
Thanks for the ping @melloware — took a look at the mock side.

Nice feature 👍 One correctness issue though: the `$ref` array-item factories aren't actually deduplicated across operations. When the same `$ref` is used as an array item by more than one operation, the generated file ends up with duplicate `export const get<Schema>Mock` declarations and fails to compile.

**Repro** (`arrayItems: true`, no `schemas: true`):

```yaml
paths:
  /tenants-a:
    get:
      operationId: getTenantsA
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema: { $ref: '#/components/schemas/TenantListResponse' }
  /tenants-b:
    get:
      operationId: getTenantsB
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema: { $ref: '#/components/schemas/TenantListResponse' }
components:
  schemas:
    TenantResponseModelDto:
      type: object
      required: [id, name]
      properties: { id: { type: string }, name: { type: string } }
    TenantListResponse:
      type: object
      properties:
        value: { type: array, items: { $ref: '#/components/schemas/TenantResponseModelDto' } }
        count: { type: integer }
```

Generated output:

```ts
// from getTenantsA
export const getTenantResponseModelDtoMock = (...) => ({ ... });
export const getGetTenantsAResponseMock = (...) => ({ value: [...].map(() => ({ ...getTenantResponseModelDtoMock() })), ... });

// from getTenantsB — same name again
export const getTenantResponseModelDtoMock = (...) => ({ ... });
export const getGetTenantsBResponseMock = (...) => ({ ... });
```

```
error TS2451: Cannot redeclare block-scoped variable 'getTenantResponseModelDtoMock'.
```

**Root cause:** the dedup guard in `extractArrayItemMock` only checks the `splitMockImplementations` array, but that array is re-initialized per operation (`packages/mock/src/msw/index.ts:398`), and the per-operation mock bodies are then concatenated without any file-level dedup (`packages/core/src/writers/target.ts:73`). So two operations never see each other's factories.

**Scope:**
- Only the `$ref` case (shared `get<Schema>Mock` name). Inline object items are fine — they're named per operation (`get<OperationId>Response<Prop>ItemMock`).
- `schemas: true` is also fine — `hasConsolidatedSchemaFactory` skips extraction and delegates to the consolidated factory.
- So it hits exactly the `$ref` + multiple-operations + no `schemas: true` combo, which is the use case the docs highlight ("item factories without emitting every `components/schemas` entry"). Sharing a `$ref` across operations is very common, so it's easy to run into.

The current `faker-array-items` spec only has a single operation using the `$ref`, which is why the snapshots stay green — might be worth adding a second operation sharing the same `$ref` as a regression test once the dedup is file-scoped.

wadakatu:
One more thing — this one isn't a functional bug but a scope question.

The `single-mode.ts` import-merge change in this PR seems to alter the output for **all single-mode + mock users**, not just those who opt into `arrayItems`. This PR updates **70 existing snapshots** (angular / axios / react-query / swr / vue-query / fetch …); the bulk of the diff is mock-only type imports being hoisted from the trailing block into the main import block.

Most of that is harmless consolidation, but a few outputs now get **spurious unused imports**. The new filter matches an import name (and alias) against the combined `implementation + mock` source text via a word-boundary regex, which produces false positives on name collisions and common identifiers.

Example 1 — `tests/__snapshots__/angular/custom-client/endpoints.ts`:
```ts
import { HttpClient, HttpResponse as AngularHttpResponse } from '@angular/common/http';
```
`AngularHttpResponse` is never used in the body, but its search term `HttpResponse` matches msw's `HttpResponse.json(...)` in the mock code, so it gets imported (master correctly dropped it).

Example 2 — `tests/__snapshots__/angular/http-resource-zod/endpoints.ts`:
```ts
import { ..., map } from 'rxjs';
```
The rxjs `map` operator isn't called anywhere — the only `map` is the `Array.prototype.map` in the mock body. The search term `map` matches the `map` inside `.map(`, so the rxjs import is pulled in.

CI stays green because the snapshot tests are string comparisons (no `noUnusedLocals` typecheck on `tests/__snapshots__`), so these dead imports don't fail compilation — but they're still incorrect output and could trip `noUnusedLocals` / eslint `no-unused-vars` in consumer projects.

Since this import consolidation doesn't seem strictly required for `arrayItems` itself (the factory type imports can flow through the existing mock-import path), it might be cleaner to **split it into its own PR**, or at least **fix the false-positive matching first**. That would also keep this PR's diff focused on the new feature and easier to review.

Hypenate:
@wadakatu I've made the requested changes.

wadakatu:
@Hypenate thanks for the quick turnaround 🙏 Confirmed the single-mode duplicate is gone and the regression test you added covers it. Reverting `single-mode.ts` also dropped the unrelated snapshot churn 👍

One follow-up though: the new file-level dedup breaks the other way in `tags-split` / `tags` modes. The `context.arrayItemMockFactories` `Set` is shared across **all per-tag files** of the output target, so when operations in different tags (= different files) reference the same `$ref`, the factory is only emitted into the first file and the second file ends up with a dangling reference.

**Repro** (`mode: 'tags-split'`, `arrayItems: true`, two tags sharing the same `$ref`):

```yaml
paths:
  /a:
    get: { operationId: getA, tags: [alpha], responses: { '200': { description: ok, content: { application/json: { schema: { $ref: '#/components/schemas/TenantListResponse' } } } } } }
  /b:
    get: { operationId: getB, tags: [beta], responses: { '200': { description: ok, content: { application/json: { schema: { $ref: '#/components/schemas/TenantListResponse' } } } } } }
components:
  schemas:
    TenantResponseModelDto:
      type: object
      required: [id, name]
      properties: { id: { type: string }, name: { type: string } }
    TenantListResponse:
      type: object
      properties:
        value: { type: array, items: { $ref: '#/components/schemas/TenantResponseModelDto' } }
        count: { type: integer }
```

**Generated:**
- `alpha/alpha.faker.ts` → defines `export const getTenantResponseModelDtoMock` ✅
- `beta/beta.faker.ts` → references it without defining or importing it:

```ts
...Array.from(...).map(() => ({ ...getTenantResponseModelDtoMock() }))...
```
```
beta/beta.faker.ts(16,258): error TS2304: Cannot find name 'getTenantResponseModelDtoMock'.
```

**Behavior per mode** (all verified):

| mode | result |
|---|---|
| `single` | ✅ |
| `split` | ✅ (mock goes into a single `endpoints.faker.ts`) |
| `tags-split` | 🔴 TS2304 |
| `tags` | 🔴 TS2304 |

CI stays green because the `faker-array-items` test only exercises single-mode, so the multi-file modes aren't covered (same blind spot as the duplicate issue).

For a direction: scoping the dedup **per output file rather than per `context`** would fix it, or for the split modes, hoisting the shared factory into an importable shared file (the way `schemas: true` emits `index.faker.ts`) and importing it from each tag file. Might also be worth adding a `tags-split` case to the spec so this stays covered.

Hypenate:
@wadakatu another try  😃

wadakatu:
Confirmed the tags-split / tags issue is fixed on the latest head 👍 While I was at it I found a few more gaps when `arrayItems: true` is enabled.

None of these are regressions — `master` and `arrayItems: false` generate compiling code for all of the specs below. They're cases where the new extraction logic emits broken (uncompilable, or wrong-data) code for certain array-item shapes. Plain object items, `$ref`-to-object items, and inline `allOf` items are all fine.

**(1) `$ref` to a scalar/primitive schema**
```yaml
names: { type: array, items: { $ref: '#/components/schemas/Name' } }
# components.schemas.Name: { type: string, format: email }
```
```ts
export const getNameMock = (overrideResponse: Partial<Name> = {}): Name =>
  ({ ...faker.internet.email(), ...overrideResponse });   // Name = string
```
```
error TS2322: Type '{}' is not assignable to type 'string'.
error TS2698: Spread types may only be created from object types.
```
At runtime it also returns `{0:'a',1:'@',...}` instead of the string. `isObjectLikeArrayItem` returns `true` for any `$ref` without resolving the target, so scalar refs get wrapped in an object factory.

**(2) inline `oneOf` / `anyOf` array item**
```yaml
things: { type: array, items: { oneOf: [ { $ref: '#/.../Cat' }, { $ref: '#/.../Dog' } ] } }
```
```
error TS2305: Module './model' has no exported member 'GetThings200ThingsItem'.
```
The item factory imports/types itself as `GetThings200ThingsItem`, but orval doesn't emit that alias for a composition item (the model only has `Cat` / `Dog` / `GetThings200`). The `<Parent><Prop>Item` name assumption only holds for plain object / `allOf` items.

**(3) nullable object array item**
```yaml
rows: { type: array, items: { type: object, nullable: true, properties: { id: { type: string } } } }
```
```
error TS2322: Type '{ id?: string | undefined; }[]' is not assignable to type 'GetNullable200RowsItem[]'.
```
`GetNullable200RowsItem` is `{ id: string } | null`, but the extracted factory's body / return type doesn't account for the nullable item.

**(4) same property name under two parents in one operation**
```yaml
outer: { properties: { items: { type: array, items: { type: object, properties: { a: { type: string } } } } } }
inner: { properties: { items: { type: array, items: { type: object, properties: { b: { type: number } } } } } }
```
Only one factory is emitted and reused for both:
```ts
export const getGetCollideResponseItemsItemMock = (...): OuterItemsItem => ({ ...{ a: ... }, ... });
// both outer.items and inner.items call getGetCollideResponseItemsItemMock()
```
So `inner.items` is silently mocked with the `outer` shape (`{ a }` instead of `{ b }`) — that part doesn't even error. There's also `error TS2305: ... has no exported member 'OuterItemsItem'` (the type name is derived from the immediate parent `OuterItemsItem`, while orval emits `GetCollide200OuterItemsItem`). The factory name is `get<OperationId>Response<Prop>ItemMock`, which collides when two array properties share a name in the same operation.

**On the fix:** in all four, turning `arrayItems` off falls back to inline mocks that compile fine, so a conservative guard would handle them — only extract for plain object inline items, `$ref`-to-object items, and `allOf` (all verified OK), and skip scalar / scalar-`$ref`, `oneOf` / `anyOf`, nullable, and name-collision cases (fall back to inline). Adding these shapes to the spec would lock it in.

Hypenate:
@wadakatu Good catches!!

Deeper into the 🐇 :hole: I go 😆

Hypenate:
@wadakatu Can you review it again please?

melloware (approved):

wadakatu:
Re-reviewed at `f18b2bf` — this all looks good now 🎉

Re-ran all four cases plus a few neighbours and everything compiles cleanly:
- `$ref`→scalar, `oneOf`/`anyOf`, nullable, and the same-prop collision all fall back to inline mocks now ✅
- `$ref`→composition schema and a `$ref` response wrapper with an inline item are handled too ✅
- the happy path (plain object inline, `$ref`→object, `allOf`) still extracts + dedups correctly across `single` / `split` / `tags-split` / `tags` ✅
- `@orval/mock` unit tests pass and the edge shapes are now in the spec 👍

Nice work chasing all of these down. LGTM from the mock side 👍

_(Minor, non-blocking: an inline object array item inside a `$ref`'d response wrapper — e.g. a reusable `ListResponse` with `items: {…}[]` — now skips extraction because the parent name doesn't include the operationId, so it falls back to inline. Totally safe, just a missed factory opportunity if you ever want to widen it later.)_

PR #3292, opened 2026-04-30
fix(hono): refresh handler preamble and fix import paths for hono.handlers option

zeriong:
## Summary

Closes #2989.

Two bugs caused generated Hono handler files to drift out of sync with the rest of the output:

1. When `override.hono.handlers` was set, the zod/context module paths used the `tags-split` layout (`dirname/tag/tag.zod`) for every non-`tags` mode. In `split`/`single` modes the generated imports therefore pointed at `../<tag>/<tag>.context`, while `generateContextFiles` actually emitted `../<filename>.context`. `getHonoHeader` had the symmetrical mistake for `tags` mode, joining a tag sub-directory that does not exist on disk.
2. Once a handler file existed, the generator skipped it entirely and only appended missing handlers. Stale imports, wrong casing, or paths left over from earlier configurations were never corrected — exactly the behavior reported in the issue.

This PR splits path resolution by mode (`tags` / `tags-split` / split-or-single) so it mirrors where `generateZodFiles` and `generateContextFiles` actually write, and rewrites the existing-file branch to always regenerate the file header, imports, and `const factory = createFactory();` preamble while preserving each `factory.createHandlers(...)` body verbatim via a paren-aware extractor. New handlers are appended at the end as before.

## Changes

- `packages/hono/src/index.ts`: per-mode `zodModule` / `contextModule` resolution inside the `override.hono.handlers` branch; corrected `getHonoHeader` per-tag join (only `tags-split` joins a sub-directory); `generateHandlerFile` now refreshes preamble and preserves bodies via a new exported `extractExistingHandlers` helper.
- `packages/hono/src/index.test.ts`: unit tests for `extractExistingHandlers` (multi-handler files, nested parens in user-edited bodies, empty input).
- `tests/configs/hono.config.ts`: three new regression fixtures exercising `override.hono.handlers` with `mode: 'split'`, `mode: 'tags'`, and `mode: 'tags-split'`.
- `docs/content/docs/guides/hono.mdx`: brief note that re-running orval refreshes the preamble while preserving handler bodies.

## Test plan

- [x] `bun vitest run` — new `extractExistingHandlers` tests pass.
- [x] `bun run test:snapshots` — 3776 snapshots pass, including the three new `petstore-*-with-handlers` fixtures.
- [x] `bun run lint` and `bun run typecheck` — clean.
- [x] `tests/scripts/typecheck-generated.mjs` — all 15 generated clients (including the new hono fixtures) typecheck.
[...]

PR #3518, opened 2026-06-01
feat(hono): setup new propery handler generation strategy

anymaniax:
### Summary
Handler files under `override.hono.handlers` are where users put business logic.
Since #3292 (shipped in 8.11.x), regeneration rebuilt the preamble + validator
chain from the spec and spliced back only the `async (c) => {…}` body — silently
deleting custom imports, middleware (`authenticate()`, `rateLimit()`), and
top-level helpers on every run. In one project a single regeneration stripped 66
handlers (~2,880 deletions), removing auth middleware from protected endpoints.

This PR replaces that with a configurable, non-destructive strategy.

### New option: `override.hono.handlerGenerationStrategy`
Controls how an **existing** handler file is treated on regeneration (a
non-existent file is always generated fresh):

- **`smart`** (default) — parse the file (TypeScript compiler API) and reconcile
  only orval-owned regions: its own imports (names, module paths,
  camelCase→PascalCase casing) and the `zValidator(...)` arguments, and append
  handlers for new operations. Custom imports, middleware, bodies, and top-level
  helpers are preserved.
- **`skip`** — leave an existing file byte-for-byte unchanged.
- **`full`** — previous behavior: rebuild header/imports/validators, splice back
  the body (drops custom imports/middleware/helpers).

`smart` is a general reconcile: adds missing imports/validators, removes stale
validators, migrates renamed schemas, appends new-operation stubs.

### Also fixed
- **multipart/form-data**: bodies with `multipart/form-data` or
  `application/x-www-form-urlencoded` now emit `zValidator('form', …)` + `form:`
  context instead of `'json'`.

### Implementation
- New module `packages/hono/src/handler-merge.ts` (AST reconcile). Only rewrites a
  plain named import whose names are all orval-owned; locates imports by module
  specifier (no fuzzy name matching); resolves an aliased `zValidator` scoped to
  the validator module; on parse failure falls back to `skip`.
- `typescript` is an **optional `peerDependency`**, loaded lazily via
  `import('typescript')`. If absent, `smart`/`full` degrade to `skip` with a
  warning. The compiler is not bundled.

### ⚠️ Breaking change
Default changes from the 8.11.x `full` (destructive) to `smart` (non-destructive).
Set `handlerGenerationStrategy: 'full'` for the old behavior.

> Note: if `output.clean` is enabled and the handlers dir is under the output
> target, handler files are deleted before generation, defeating preservation.

### Tests
- `handler-merge.test.ts` + `index.test.ts`: preservation of
  imports/middleware/helpers/bodies; camelCase→PascalCase migration (incl. mixed
  imports); add/remove/rename validators; new-operation append;
  namespace/default/aliased import safety; the `getCommunityFilterContext` clobber
  regression; inline-middleware handling; CRLF; typescript-unavailable fallback.
- `options.test.ts`: default + explicit strategy.
- `tests/handler-preservation.spec.ts`: end-to-end `generate()` ×2 with an edit
  between (clean disabled).
- All existing generation snapshots unchanged.

### Docs
- `output.mdx`: documents `handlerGenerationStrategy` (+ typescript-peer & `clean`
  callouts).
- `hono.mdx`: replaces the stale "refresh header/imports, preserve body" text.

## Summary by CodeRabbit

* **New Features**
  * Added `handlerGenerationStrategy` for Hono handlers: `smart` (default, non-destructive), `skip` (freeze), and `full` (regenerate wrapper and splice back bodies). Request-body validator now uses `form` for multipart/form-data and x-www-form-urlencoded.

* **Documentation**
  * Expanded guides and configuration docs to describe regeneration strategies and the optional TypeScript peer dependency and its fallback behavior.

* **Tests**
  * Added extensive unit and end-to-end tests covering reconciliation, preservation, strategy variations, and TypeScript-absent fallbacks.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -140,6 +140,12 @@
         "vitest": "catalog:",
         "zod": "catalog:zod3",
       },
+      "peerDependencies": {
+        "typescript": ">=5",
+      },
+      "optionalPeers": [
+        "typescript",
+      ],
     },
     "packages/mcp": {
       "name": "@orval/mcp",
--- a/docs/content/docs/guides/hono.mdx
+++ b/docs/content/docs/guides/hono.mdx
@@ -78,10 +78,13 @@ export const listPetsHandlers = factory.createHandlers(
 
 Add your business logic to the generated handlers:
 
-> Re-running orval refreshes the file header, imports, and `factory` line so
-> changes to your config (paths, validators, etc.) take effect — but the body
-> of each `factory.createHandlers(...)` call is preserved verbatim, so your
-> business logic is safe across regeneration.
+> Re-running orval reconciles your handler files according to
+> [`override.hono.handlerGenerationStrategy`](/docs/reference/configuration/output#handlergenerationstrategy)
+> (default `smart`). With `smart`, orval only updates the parts it owns — its own
+> imports (names, paths, casing) and the `zValidator(...)` arguments, and appends
+> handlers for new operations — while preserving your custom imports, middleware,
+> handler bodies, and top-level helpers. Use `skip` to freeze a file, or `full`
+> for the legacy behavior that rebuilds the wrapper and keeps only the body.
 
 ```ts title="src/handlers/listPets.ts"
 export const listPetsHandlers = factory.createHandlers(
--- a/docs/content/docs/reference/configuration/output.mdx
+++ b/docs/content/docs/reference/configuration/output.mdx
@@ -1400,6 +1400,7 @@ export default defineConfig({
       override: {
         hono: {
           handlers: 'src/handlers',
+          handlerGenerationStrategy: 'smart',
           validatorOutputPath: 'src/validator.ts',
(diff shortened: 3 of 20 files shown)


anymaniax:
Just flagging a pre-existing bug I noticed while working on this (not introduced
by this PR): the `hono.validator` option isn't actually applied to the generated
handlers. Whether it's set to `true`, `'hono'`, or `false`, the handlers come out
the same — always importing `zValidator` from the local `.validator` file and
including all validators.

The reason is that handler generation always receives the `.validator.ts` path,
so it never sees the `'hono'` (use `@hono/zod-validator` directly) or `false`
(no validators) intent. With `'hono'`/`false` the handler even imports from a
validator file that isn't emitted.

It's out of scope here so I'll fix it in a follow-up PR — just wanted to note it
so it's on the radar.

melloware (approved):

PR #3524, opened 2026-06-02
factory methods: discriminator aliases and missing parent properties - fix

mironbalcerzak:
### Problem
OpenAPI specifications defining polymorphic interfaces often use a combiner (`oneOf`/`anyOf`/`allOf`) alongside the parent schema's own `properties`.
However, Orval's factory generator treated combiners and `properties` as mutually exclusive.
When a combiner was found, the generator returned early and completely ignored the parent's own properties.

Relates to: https://github.com/orval-labs/orval/pull/3523

### Result of the bad code
The generated parent factory simply returned the variant's factory output, which lacked the parent's required properties.
This caused TypeScript compiler errors (e.g., **TS2322: Type 'Variant' is not assignable to type 'Variant & { parentProp: string }'**) when trying to assign the factory output to the generated intersection type.

### Example OpenAPI Spec causing the issue
```yaml
openapi: 3.1.0
components:
  schemas:
    General:
      type: object
      oneOf:
        - $ref: '#/components/schemas/Specific'
      properties:
        responseType:
          type: string
      required:
        - responseType
    Specific:
      type: object
      properties:
        question:
          type: string
```

### How we fixed it
Refactored `buildPayload` in `packages/core/src/generators/factory.ts` to no longer treat combiners and `properties` as mutually exclusive.
The logic now collects payloads from both the combiner and the parent properties into an array, and combines them using `Object.assign({}, ...payloads)`.
This ensures the generated factory correctly outputs an object satisfying the full intersection type, including all required parent and variant properties.

## Summary by CodeRabbit

* **Tests**
  * Added test coverage for handling combined schema payloads.

* **Refactor**
  * Improved payload generation logic for composed schemas to better support merging multiple property sources.

### DIFF
--- a/packages/core/src/generators/factory.test.ts
+++ b/packages/core/src/generators/factory.test.ts
@@ -584,4 +584,30 @@ describe('generateFactory', () => {
     expect(result?.model).not.toContain('user: createRefTarget()');
     expect(result?.model).toContain('token: createRefTarget()');
   });
+
+  it('combines combiner payloads with parent properties using Object.assign', () => {
+    const schema: OpenApiSchemaObject = {
+      type: 'object',
+      oneOf: [
+        {
+          type: 'object',
+          required: ['variantProp'],
+          properties: { variantProp: { type: 'string' } },
+        },
+      ],
+      required: ['parentProp'],
+      properties: {
+        parentProp: { type: 'string' },
+      },
+    };
+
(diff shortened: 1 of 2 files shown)


melloware (approved):

mironbalcerzak:
@melloware thanks once again 👍

PR #3528, opened 2026-06-02
fix(zod): escape newlines in $ref description with generateReusableSchemas

### LINKED ISSUE #3517: generateReusableSchemas (zod): $ref field with multi-line description emits unescaped newlines → TS1002 Unterminated string literal
## Summary

With `output.override.zod.generateReusableSchemas: true`, when an object property is
a **`$ref` to a reusable component** **and** carries a **multi-line `description`**,
orval emits `ReferencedSchema.describe("…")` using **double quotes with the raw,
unescaped description** — so the literal newlines break the string. The generated
file is invalid TypeScript.

Sibling **primitive** fields with the same multi-line description are emitted
correctly (single-quoted, `\n`-escaped) — so orval already has a working escaper;
the `$ref` + `.describe()` code path just doesn't use it.

## Environment

- orval **8.14.0** (first/only release with `generateReusableSchemas`)
- zod 4.4.3, Node v22.19.0
- `output.client: 'zod'`, `mode: 'tags-split'`

## Minimal reproduction

**`spec.json`** (OpenAPI 3.1 — sibling `$ref` + `description` is valid in 3.1):

```json
{
  "openapi": "3.1.0",
  "info": { "title": "repro", "version": "1.0.0" },
  "paths": {
    "/x": { "get": { "operationId": "getX", "tags": ["X"],
      "responses": { "200": { "description": "ok",
        "content": { "application/json": { "schema": { "$ref": "#/components/schemas/Claims" } } } } } } }
  },
  "components": { "schemas": {
    "UserId": { "type": "string", "description": "A user id." },
    "Claims": { "type": "object", "required": ["sub"], "properties": {
      "sub":  { "$ref": "#/components/schemas/UserId", "description": "Subject identifier.\n\nMust be normalized first." },
      "note": { "type": "string", "description": "Line one.\n\nLine two." }
    } }
  } }
}
```

**`orval.config.mjs`:**

```js
export default {
  repro: {
    input: { target: './spec.json' },
    output: {
      client: 'zod',
      mode: 'tags-split',
      target: './out/x.zod.ts',
      override: { zod: { generateReusableSchemas: true } },
    },
  },
}
```

**Run:** `orval --config ./orval.config.mjs`

## Actual output (`out/x.zod.schemas.ts`)

```ts
export const Claims = zod.object({
  "sub": UserId.describe("Subject identifier.

Must be normalized first."),
  "note": zod.string().optional().describe('Line one.\n\nLine two.')
});
```

- `sub` (a `$ref`): **double-quoted, raw newlines** → broken.
- `note` (a primitive): **single-quoted, `\n`-escaped** → correct.

`tsc --noEmit` on the file:

```
x.zod.schemas.ts(15,46): error TS1002: Unterminated string literal.
x.zod.schemas.ts(17,1): error TS1005: ',' expected.
…
```

## Expected output

The `$ref` field's description should be escaped the same way the primitive field's
is:

```ts
"sub": UserId.describe('Subject identifier.\n\nMust be normalized first.'),
```

## Root cause (suspected)

The branch that attaches `.describe()` to a **referenced reusable schema** passes the
raw description into a double-quoted literal without running it through the
newline-escaping/quoting that the primitive-field branch uses. It only surfaces with
`generateReusableSchemas`, because only that mode renders a `$ref` field as
`Name.describe("…")`; the default (inline) mode never produces that construct.

## Workaround

Pre-process the spec with `input.override.transformer` to flatten newlines in
`description` fields before generation.

z4o4z:
## Summary

Fixes #3517.

With `output.override.zod.generateReusableSchemas: true`, when an object property is a `$ref` to a reusable component **and** carries a multi-line `description`, orval emitted `ReferencedSchema.describe("…")` using double quotes around the **raw, unescaped** description. The literal newlines broke the string literal, producing invalid TypeScript:

```
x.zod.schemas.ts(15,46): error TS1002: Unterminated string literal.
```

Sibling primitive fields with the same multi-line description were emitted correctly (single-quoted, `\n`-escaped) — orval already had a working escaper; the `$ref` + `.describe()` path just wasn't using it.

## Root cause

`packages/zod/src/index.ts` — the chainable-sibling path for `$ref`s used:

```js
functions.push(['describe', `"${escape(siblingSchema.description)}"`]);
```

`escape()` only escapes quote characters, not newlines, and the result was wrapped in double quotes. This construct is only produced in `generateReusableSchemas` mode, which is why it didn't surface before.

## Fix

Use the same single-quoted, fully JS-escaped form as the primitive description path (`pushDescriptionOrMeta`):

```js
functions.push(['describe', `'${jsStringEscape(siblingSchema.description)}'`]);
```

Now `$ref` siblings and primitive fields are escaped identically:

```ts
"sub": UserId.describe('Subject identifier.\n\nMust be normalized first.'),
```

## Tests

- Updated the existing `$ref` description test to expect the single-quoted form.
- Added regression test `escapes newlines in a multi-line description sibling on a $ref (#3517)`, verified to fail before the fix and pass after.
- Full `@orval/zod` suite passes (204 tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Generated Zod validation schemas now emit description strings as single-quoted, fully JS-escaped literals, ensuring special characters and newlines are represented safely.

* **Tests**
  * Added a regression test verifying multi-line descriptions are emitted with escaped newline characters so generated code remains single-line and valid.

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -281,7 +281,14 @@ export const generateZodValidationSchemaDefinition = (
     }
 
     if (typeof siblingSchema.description === 'string') {
-      functions.push(['describe', `"${escape(siblingSchema.description)}"`]);
+      // Use the same single-quoted, fully JS-escaped form as the primitive
+      // description path (see `pushDescriptionOrMeta`). `escape` only escapes
+      // quote chars, so a multi-line description would emit raw newlines and
+      // break the generated string literal (TS1002).
+      functions.push([
+        'describe',
+        `'${jsStringEscape(siblingSchema.description)}'`,
+      ]);
     }
   };
 
--- a/packages/zod/src/zod.test.ts
+++ b/packages/zod/src/zod.test.ts
@@ -4101,7 +4101,10 @@ describe('generateZodValidationSchemaDefinition`', () => {
       // newlines that would break the generated string literal (TS1002).
       expect(result.functions).toEqual([
         ['namedRef', { name: 'Pet', sourceRef: '#/components/schemas/Pet' }],
-        ['describe', "'Subject identifier.\\n\\nMust be normalized first.'"],
+        [
+          'describe',
+          String.raw`'Subject identifier.\n\nMust be normalized first.'`,
+        ],
       ]);
     });
 
--- a/packages/zod/src/zod.test.ts
+++ b/packages/zod/src/zod.test.ts
@@ -4078,7 +4078,30 @@ describe('generateZodValidationSchemaDefinition`', () => {
       );
       expect(result.functions).toEqual([
         ['namedRef', { name: 'Pet', sourceRef: '#/components/schemas/Pet' }],
-        ['describe', '"optional pet"'],
+        ['describe', "'optional pet'"],
+      ]);
+    });
+
+    it('escapes newlines in a multi-line description sibling on a $ref (#3517)', () => {
+      const context = makeContextSpec({
+        spec: { components: { schemas: { Pet: { type: 'object' } } } },
+      });
+      const result = generateZodValidationSchemaDefinition(
+        {
+          $ref: '#/components/schemas/Pet',
+          description: 'Subject identifier.\n\nMust be normalized first.',
+        } as never,
+        context,
+        'someField',
+        false,
+        false,
+        { required: true, useReusableSchemas: true },
+      );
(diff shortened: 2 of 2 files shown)


melloware (changes requested):
Build is failing

melloware (approved):

PR #3529, opened 2026-06-02
feat(mock): emit strict Mock return types when required and nonNullable are both true

### LINKED ISSUE #3525: feat(mock): .faker types should reflect complete mocks when required and nonNullable are both true
## Goal

When `override.mock.required` and `override.mock.nonNullable` are both enabled, **the default faker mock should be typed as a complete object** — every property present, no optional/`null` on the **return type** of `getPetMock()`.

Callers must still be able to pass **`null` in `overrideResponse`** when that is a valid value for the real API (`Pet` allows `tag: string | null`, etc.). Strict typing applies to **what the generator produces by default**, not to forbidding legitimate override values.

```ts
export type PetMock = {
  [K in keyof Required<Pet>]: NonNullable<Required<Pet>[K]>;
};

export const getPetMock = (
  overrideResponse: Partial<Pet> = {},
): PetMock => ({
  id: faker.number.int(),
  tag: faker.string.alpha(),
  ...overrideResponse,
});
```

| Call | Types | Runtime |
|------|-------|---------|
| `getPetMock()` | `PetMock` — complete, non-null default | All fields filled, no random nulls |
| `getPetMock({ tag: null })` | Override accepts `null` (`Partial<Pet>`); return type still `PetMock` | `tag === null` is valid (#3476) |

TypeScript will **not** narrow the return type when you pass `null` in the override (still `PetMock`). That is acceptable: the **default mock** is what should “lend to the truth”; overrides are an explicit escape hatch for API-shaped values including `null`.

---

## Background

[PR #3476](https://github.com/orval-labs/orval/pull/3476) added `override.mock.nonNullable`. Together with `required`:

| Flag | Effect on **generated default** |
|------|----------------------------------|
| `required: true` | All properties emitted (no random omission). |
| `nonNullable: true` | No `arrayElement([value, null])` in generated faker code. |
| **Both** | Complete default object at runtime (see `object.test.ts`). |

This issue applies only when **both** are `true`. Docs: `docs/content/docs/reference/configuration/output.mdx`.

---

### 1. Default mock is complete at runtime, but typed as loose `Pet`

```ts
export const getPetMock = (
  overrideResponse: Partial<Pet> = {},
): Pet => ({ ... });
```

`getPetMock()` is typed as `Pet` (`birthDate?: string | null`, etc.) even when the generator always returns a full, non-null object. **We want `PetMock` on the return type**, not `Partial<Pet>` / nullable OpenAPI shape for the default path.

### 2. Operation mocks may not honor `nonNullable`

`getMockWithoutFunc` forwards `required` but not `nonNullable` today — operation/`get*ResponseMock` in `.faker.ts` may not get strict **runtime** defaults until fixed.

---

### Runtime

- Forward `nonNullable` through `getMockWithoutFunc` so all mock factories honor both flags.

### Types

When both flags are set:

1. Emit **`PetMock`** (name TBD): required keys + `NonNullable` on property types (top-level; nested can be follow-up).
2. **`getPetMock(): PetMock`** — calling without overrides returns a **complete** type.
3. **`overrideResponse: Partial<Pet>`** — keep the **OpenAPI model** so users can pass `null`, omit keys, or partial updates that match the real API. Do **not** use `Partial<PetMock>` here (that would block `{ tag: null }`).

Generated faker body still ends with `...overrideResponse` so runtime overrides work as today.

Apply to faker schema factories, faker operation factories, and MSW `get*ResponseMock`.

### When flags are unset

Unchanged `Partial<Pet>` / `Pet` (or `Partial<T>` / `T`) signatures.

---

## Scope boundaries (not anti-goals)

| In scope | Out of scope for this issue |
|----------|-----------------------------|
| `PetMock` return type for **default** complete mock | Changing signatures when flags are unset |
| `Partial<Pet>` on `overrideResponse` (includes `null`) | Replacing `Pet` in non-mock client/model files |
| Wire `nonNullable` on operation mock path | Per-call return inference (generics/overloads) |
| Same rules for `.faker.ts` and MSW | Deep recursive `PetMock` on nested objects (follow-up) |

---

## Motivation

`getPetMock()` in Storybook, seeds, and tests should be typed as **complete** without casts. Teams testing nullable API fields can still use `getPetMock({ tag: null })` with a parameter type that allows it.

---

## Related

- #3476 — `override.mock.nonNullable`
- #1853 — original non-null mock request
- #3484 — separate 3.1 double-wrap tracking

---

## Acceptance criteria

- [ ] `getMockWithoutFunc` forwards `nonNullable` for operation/response mocks.
- [ ] With both flags: `getPetMock(overrideResponse?: Partial<Pet>): PetMock`.
- [ ] `getPetMock()` typed as complete (`PetMock`); `getPetMock({ tag: null })` allowed on the parameter.
- [ ] With either flag unset: current signatures unchanged.
- [ ] MSW factories match when both flags are set.
- [ ] Tests/snapshots: optional + nullable schema (OAS 3.0 + 3.1).

Hypenate:
## Summary

- When `override.mock.required` and `override.mock.nonNullable` are both enabled, faker schema factories, operation response mocks, and MSW `get*ResponseMock` helpers emit a strict `{Schema}Mock` base type (`Required` keys + `NonNullable` property types).
- Overridable factories are generic so return types narrow when overrides include `null`:
  - `getPetMock()` → complete `PetMock` (`tag: string`)
  - `getPetMock({ tag: null })` → `tag: string | null`
  - `getPetMock({ tag: 'abc' })` → `tag: string`
- `overrideResponse` stays typed as `Partial<T>` (OpenAPI model) so all legitimate API-shaped overrides remain valid.
- `getMockWithoutFunc` now forwards `nonNullable` so operation/response mocks honor the flag at runtime.

## Generated signature (strict mode)

```ts
export type KeysWithNull<O> = { /* ... */ };
export type MockWithNullableOverrides<T, O extends Partial<T>, M> = { /* ... */ };
export type PetMock = { [K in keyof Required<Pet>]: NonNullable<Required<Pet>[K]>; };

export const getPetMock = <O extends Partial<Pet> = {}>(
  overrideResponse?: O,
): MockWithNullableOverrides<Pet, O, PetMock> => ({ ... }) as MockWithNullableOverrides<Pet, O, PetMock>;
```

## Test plan

- [x] Unit tests for `mock-types` helpers and strict mock generation in `faker/index.test.ts` and `msw/index.test.ts`
- [x] Snapshot fixtures for OpenAPI 3.0 and 3.1 nullable optional schemas (`issue-3525`, `issue-3525-oas31`)
- [x] `bun test` in `packages/mock` passes (209 tests)

Fixes #3525

## Summary by CodeRabbit

* **New Features**
  * Strict mock type generation for mock outputs (faker + MSW): non-nullable required fields in returned mock types while still supporting typed partial/nullable overrides; new nullable-override helper types and strict mock aliases are emitted in generated code.
* **Tests**
  * Added suites validating strict-mock behavior, factory signatures, and generated return types under strict flags.
* **Chores**
  * Added generator configs, specs and updated snapshots to cover strict-mock scenarios.


Hypenate:
@wadakatu I'll have to ask for your review skills :)
Should be last PR from our side (famous last words).

wadakatu:
Thanks for the ping @Hypenate, and nice work extracting the strict-mock logic into a testable `mock-types.ts` module. I pulled the branch (`eab1c2922`) and ran the generator end-to-end. The single-operation snapshots look right, but I hit two compile-breaking issues as soon as the spec has more than one operation or an array-of-`$ref` response — both are currently uncovered by the fixtures (which only exercise one object operation).

**Repro:** OAS 3.0, `override.mock: { required: true, nonNullable: true }`, `mock.generators: [{ type: 'msw' }, { type: 'faker', schemas: true, operationResponses: true }]`, three ops — `GET /pet` → `Pet`, `POST /pet` → `Pet`, `GET /pets` → `Pet[]`.

### 1. Duplicate type declarations across operations → TS2300

`generateMSW` prepends the helper block (`KeysWithNull` + `MockWithNullableOverrides`) once per operation, and each `generateDefinition` prepends the per-schema `PetMock` alias. The writer concatenates the per-operation strings verbatim (`core/src/writers/target.ts:73`, and `target-tags.ts:61` for tags mode) with no dedup, so a file with N operations emits N copies:

```
$ grep -c 'export type KeysWithNull'             out/endpoints.ts   # 3
$ grep -c 'export type MockWithNullableOverrides' out/endpoints.ts  # 3
$ grep -c 'export type PetMock'                  out/endpoints.ts   # 3
```

```
out/endpoints.ts(164,13): error TS2300: Duplicate identifier 'KeysWithNull'.
out/endpoints.ts(168,13): error TS2300: Duplicate identifier 'MockWithNullableOverrides'.
out/endpoints.ts(176,13): error TS2300: Duplicate identifier 'PetMock'.
... (9 total)
```

Any spec where ≥2 operations share a file (single mode, or one tag file) trips this — e.g. `GET /pet` + `POST /pet` both returning `Pet`. The faker *schemas* file is fine because `generateFakerForSchemas` dedups via a `Set`; the bug is limited to the per-operation MSW and faker `operationResponses` paths.

Suggestion: emit the helpers and each `{Schema}Mock` alias once per file (the pattern `generateFakerForSchemas` already uses) — hoist them into the file header after aggregation, or dedup by identifier. A multi-operation + shared-schema snapshot would lock this down.

### 2. Schema-factory imports treated as type names → TS2749

For the array endpoint the response delegates to the faker factory `getPetMock()`. `getSchemaTypeNamesFromResponses` adds *every* response import name (`imp.alias ?? imp.name`), including value imports like the `getPetMock` factory, so `getStrictMockTypeDeclarations` emits:

```ts
export type getPetMockMock = {
  [K in keyof Required<getPetMock>]: NonNullable<Required<getPetMock>[K]>;
};
```

`getPetMock` is a value, not a type:

```
out/endpoints.ts(199,24): error TS2749: 'getPetMock' refers to a value, but is being used as a type here.
```

(The alias is also dead — the array factory's return type already resolves to `PetMock[]`.) Reproduces with msw + `faker schemas:true` + an array-of-`$ref` response. Filtering `getSchemaTypeNamesFromResponses` to type-only imports (drop `values`/`schemaFactory` imports), and/or only emitting `{Name}Mock` for actual schema types, would fix it.

### Things that look good

- `getMockWithoutFunc` forwarding `nonNullable` — correct, matches the acceptance criteria.
- Single-object happy path checks out: `getPetMock()` → `PetMock`, `getPetMock({ tag: null })` → `tag: string | null`, `getPetMock({ tag: 'x' })` → `tag: string`.
- I briefly suspected the generic `<O extends Partial<T>>` param would stop catching typo'd override keys, but TS 5.9 still flags excess properties against the constraint (`getPetMock({ taag: null })` → TS2561), so no regression there.

### One design question

The issue listed per-call generic narrowing as out-of-scope ("TS will **not** narrow the return type when you pass `null`… that is acceptable"). The `MockWithNullableOverrides` machinery is exactly that narrowing, and it's also the part being duplicated in (1). If the simpler `getPetMock(overrideResponse?: Partial<Pet>): PetMock` from the issue is enough, it'd drop `KeysWithNull`/`MockWithNullableOverrides` entirely and shrink the dedup surface to just the `{Schema}Mock` aliases. Might be worth a quick word with @melloware on whether the narrowing is wanted — happy either way, just flagging the trade-off.

Repro spec available if useful. Thanks again!

Hypenate:
@wadakatu I gave it another go.

melloware:
in my opinion "shrinking the dedupe" surface is always ideal so I am ok with the narrowing if necessary!

packages/mock/src/mock-types.ts
```diff
@@ -0,0 +1,275 @@
...
+ * MSW/faker operation mocks are concatenated per file with no dedup. Hoist the
+ * shared strict-mock helper types and each `{Schema}Mock` alias once at the top.
+ */
+export function usesStrictMockInImplementation(
+  implementation: string,
+): boolean {
+  return (
+    implementation.includes('export type KeysWithNull') ||
+    implementation.includes('MockWithNullableOverrides<') ||
+    /\b[A-Z]\w*Mock\b/.test(implementation)
```
wadakatu: This `/\b[A-Z]\w*Mock\b/` gate matches **any** PascalCase identifier ending in `Mock`, including a user schema literally named `WidgetMock`. Because `dedupeStrictMockTypeDeclarations` runs for every mock file (see `finalizeMockImplementation` wiring), this flips to `true` for non-strict output too and triggers the rewrite. Gating on the actual strict flag (`isStrictMock(override.mock)`) rather than a name heuristic would avoid the false positive.

packages/mock/src/mock-types.ts
```diff
@@ -0,0 +1,275 @@
...
+): string[] {
+  const names = new Set<string>();
+
+  for (const match of implementation.matchAll(
+    /MockWithNullableOverrides<(\w+),/g,
+  )) {
+    names.add(match[1]);
+  }
+
+  for (const match of implementation.matchAll(/\b([A-Z]\w*)Mock\b/g)) {
```
wadakatu: This captures `Widget` from the model type `WidgetMock`, so `buildStrictMockTypeFileHeader` then emits:

```ts
export type WidgetMock = {
  [K in keyof Required<Widget>]: NonNullable<Required<Widget>[K]>;
};
```

`Widget` doesn't exist (→ `TS2304`) and the alias name `WidgetMock` collides with the imported model (→ `TS2440`). Re-deriving schema names from the rendered string is the fragile root cause here — threading the real type names through as structured data (you already compute them via `getStrictMockTypeName` / `getSchemaTypeNamesFromResponses`) would remove this class of bug.

packages/mock/src/mock-types.ts
```diff
@@ -0,0 +1,275 @@
...
+    implementation.includes('export type KeysWithNull') ||
+    implementation.includes('MockWithNullableOverrides<') ||
+    /\b[A-Z]\w*Mock\b/.test(implementation)
+  );
+}
+
+export function dedupeStrictMockTypeDeclarations(
+  implementation: string,
+): string {
+  let body = implementation.replaceAll(INVALID_STRICT_MOCK_DECL_PATTERN, '');
```
wadakatu: `dedupeStrictMockTypeDeclarations` runs unconditionally on every mock file (it's wired into `finalizeMockImplementation` plus both `*Imports` hooks), so non-strict generation gets rewritten too. An early `isStrictMock(override.mock)` guard — here or at the call sites — would scope this to the feature and remove most of the regression surface for non-strict users.

packages/orval/src/api.ts
```diff
@@ -145,6 +148,7 @@ export async function getApiBuilder({
     footer: generateClientFooter,
     imports: generateClientImports,
     importsMock: generateMockImports,
+    finalizeMockImplementation: dedupeStrictMockTypeDeclarations,
```
wadakatu, line 151: This wires the hoist for **all** mock generation, regardless of the strict flags. As a result a non-strict user whose schema name ends in `Mock` gets strict helper types + a broken `…Mock` alias injected into their output (`TS2440` / `TS2304`). Passing the strict flag (and ideally the structured strict type-name set) through here would let it stay a no-op for non-strict output.

wadakatu (commented):
Re-pulled `6b817246c` — both original issues are fixed (multi-op + array specs generate and `tsc` clean, and the `issue-3525-multi` fixture is a great addition 👍).

Following up on the dedup approach before it lands. After this change the strict-mock types are emitted **only** by `dedupeStrictMockTypeDeclarations` re-deriving schema names via regex from the rendered output, and it runs on **every** mock file regardless of the strict flags. That string-scraping regresses today for any schema whose **name ends in `Mock`** — including users who don't touch the strict flags at all. I left inline notes on the exact spots; full repro below.

**Repro** — plain **non-strict** mock (no `required`/`nonNullable`), one schema named `WidgetMock`:

```yaml
components:
  schemas:
    WidgetMock:
      type: object
      required: [id]
      properties: { id: { type: integer }, label: { type: string, nullable: true } }
```

```
endpoints.ts(8,3):   error TS2440: Import declaration conflicts with local declaration of 'WidgetMock'.
endpoints.ts(75,24): error TS2304: Cannot find name 'Widget'.
```

The dedup injects strict helpers + `export type WidgetMock = { [K in keyof Required<Widget>] … }` into the non-strict file — it collides with the imported model (`TS2440`) and references a non-existent `Widget` captured from `WidgetMock` (`TS2304`). The same breakage happens in strict mode for a `*Mock`-named schema. Normal non-strict specs without a `*Mock` schema are unaffected — the dedup is a clean no-op there, so the blast radius is "any schema whose name ends in `Mock`".

Two changes would make it robust:

1. Only run the hoist when strict mode is actually on (`isStrictMock(override.mock)`), so non-strict output is never rewritten.
2. Build the header from the real strict type names you already compute (`getStrictMockTypeName` / `getSchemaTypeNamesFromResponses`) instead of regex-scanning the rendered string — that also removes the whitespace coupling in `STRICT_MOCK_SCHEMA_DECL_PATTERN` / `replaceAll(helperBlock, …)`.

A fixture with a schema named `WidgetMock` (both strict and non-strict) would lock the regression down. Happy to help if useful — thanks again for iterating on this! 🙏

Hypenate:
@wadakatu I gave it another try, but maybe it's becoming too complex?
What do you think?

wadakatu:
First, the good news: I re-pulled `b2c64c5d3` and the regression is gone — both `issue-3525-widget-mock` (non-strict) and `…-strict` generate and `tsc` clean, non-strict output is untouched, and a `*Mock`-named schema now emits a correct `WidgetMockMock` with no collision. The flag gating + structured `strictMockSchemaTypeNames` are exactly the right move. 👍

On "is it too complex?" — honestly a little, but I don't think it's the amount of functionality; it's that the new structured-names path was added *alongside* the old regex-scraping path instead of replacing it. There are two parallel strategies right now:

- **MSW / operation mocks**: factories only *use* the types, declarations aren't emitted per-op, and `strictMockSchemaTypeNames` flows to `finalizeMockImplementation`, which emits them once. 👍 clean.
- **faker schemas file**: `generateFakerForSchemas` still emits the helper block + each `{Schema}Mock` *inline*, so `dedupeStrictMockTypeDeclarations` has to strip + rebuild them with regex (`INVALID_STRICT_MOCK_DECL_PATTERN`, `STRICT_MOCK_SCHEMA_DECL_PATTERN`, `collectStrictMockSchemaTypeNames`, `collectStrictMockSchemaNamesFromUsage`, `replaceAll(helperBlock, …)`).

So the complexity is basically the regex half, which only exists to clean up what the faker-schemas path emits inline.

You can collapse it to one strategy: have `generateFakerForSchemas` stop emitting the helpers/aliases inline and instead expose `strictMockSchemaTypeNames` the way the MSW builder does, then emit them once at the file level (like the operation path already does). At that point nothing needs de-duplicating and the whole regex layer can go — `INVALID_/STRICT_MOCK_DECL_PATTERN`, `collectStrictMockSchemaTypeNames`, `collectStrictMockSchemaNamesFromUsage`, and the strip/`replaceAll` logic. "Dedupe" becomes "emit once from the structured set," which also removes the whitespace coupling those patterns rely on.

That said, this simplification is an internal refactor — the PR as it stands fixes the regression and generates correct, compiling output (I verified locally). So there's no need to cram it all into this PR; I'd be totally fine doing the regex-layer removal / unification as a **follow-up PR** rather than piling more onto this one. Might be the easier path — what do you prefer?

Hypenate:
@wadakatu
Thanks again for your excellent review and testing! 
Indeed, the PR is getting quite big.

IF it's okay for @melloware to merge this, and then I'll do a follow-up PR to do the simplification, but atleast this PR stays under control.

melloware (approved):

PR #3541, opened 2026-06-04
fix(angular): url-encode path params when urlEncodeParameters is set

### LINKED ISSUE #3540: Path  parameters not encoded in angular clients
### Description

In the angular clients, path parameters are not encoded. Regardless of the 'urlEncodeParameters' configuration.
See related issue #679 (https://github.com/orval-labs/orval/issues/679)

All angular clients are affected
- angular
- angular-query
- httpresource

### Output client

angular

```ts
// paste your orval.config file here
export default defineConfig({
  nitroDomain: {
    input: {
      target: './api/openapi.json',
    },
    output: {
      target: './packages/nitro-domain-client-orval/src/services',
      schemas: './packages/nitro-domain-client-orval/src/model',
      client: 'angular',
      mode: 'tags-split',
      indexFiles: true,
      clean: true,
      prettier: false,
      urlEncodeParameters: true,
      // Bake a placeholder origin into every generated URL; the lib's
      // NitroDomainApiBaseUrlInterceptor swaps it for the runtime
      // DOMAIN_API_URI_TOKEN. See nitro-api-base-uri-sentinel.ts.
      baseUrl: {
        getBaseUrlFromSpecification: false,
        baseUrl: NITRO_API_BASE_URI_SENTINEL,
      },
      override: {
        useTypeOverInterfaces: false,
      },
    },
  },
});

```

### Environment

  System:
    OS: Linux 7.0 Arch Linux
    CPU: (24) x64 13th Gen Intel(R) Core(TM) i7-13700HX
    Memory: 35.57 GB / 62.46 GB
    Container: Yes
    Shell: 5.3.9 - /usr/bin/bash
  npmPackages:
    @angular/common: 21.2.15 => 21.2.15 
    @angular/core: 21.2.15 => 21.2.15 
    orval: 8.10.0 => 8.10.0 

(note: 8.15 also affected).

### Expected behavior

encodeUriParam called when building uris containing path parameters

### Actual behavior

encodeUriParam not called for path parameters

```yaml
openapi: 3.0.3
info:
  title: Demo
  version: 0.0.0
paths: {}
```

### Additional context

_No response_

cghislai:
Wire makeRouteSafe into the Angular httpClient and httpResource builders and the angular-query client, mirroring query/fetch. Previously the flag was a no-op for these clients, so path params were interpolated raw.

Fixes https://github.com/orval-labs/orval/issues/3540

## Summary by CodeRabbit

* **New Features**
  * Optional URL parameter encoding for Angular clients: path parameters and routes are URL-encoded when enabled to improve handling of special characters.

* **Tests**
  * Added end-to-end test coverage verifying encoding on client and resource flows with encoding toggled.

* **Chores**
  * Updated generator snapshots and test configurations to cover the new encoding behavior.


packages/query/src/client.ts
```diff
@@ -102,8 +102,13 @@ export const generateAngularHttpRequestFunction = (
     formUrlEncoded,
     override,
   }: GeneratorVerbOptions,
-  { route, context }: GeneratorOptions,
+  { route: _route, context }: GeneratorOptions,
 ) => {
+  let route = _route;
+  if (context.output.urlEncodeParameters) {
+    route = makeRouteSafe(route);
+  }
```
melloware, line 110: I don't think this belongs here as `makeRouteSafe` is used in `/fetch/index.ts` and `/query/client.ts` so moving this to here won't it double encode the routes now in Fetch and Tanstack Query?

melloware (changes requested):

melloware:
@soartec-lab do you remember why this is moved inside each client instead of at a higher level like it is in this PR?

@wadakatu any thoughts here

packages/query/src/client.ts
cghislai: This only affects the angular client (were in generateAngularHttpRequestFunction function).

Other clients are unaffected, and the generated code for them is identical with or without this change.

melloware:
Ahhh I see now @cghislai thanks!

melloware (approved):

soartec-lab:
@melloware 
I couldn't answer immediately. Thank you for contacting me.

PR #3543, opened 2026-06-04
refactor(mock): emit strict mock types once from structured names

### LINKED ISSUE #3542: refactor(mock): emit strict mock types once from structured names only
## Summary

Follow-up to #3525 strict mock types. The regression fix added `strictMockSchemaTypeNames` for MSW/operation mocks, but `generateFakerForSchemas` still emits strict helper/alias blocks inline. `dedupeStrictMockTypeDeclarations` then strips and rebuilds them using regex (`STRICT_MOCK_SCHEMA_DECL_PATTERN`, `collectStrictMockSchemaTypeNames`, etc.).

## Proposal

- Have `generateFakerForSchemas` return `strictMockSchemaTypeNames` and stop emitting helpers/aliases inline (same as MSW builders).
- Finalize once at file write time (`writeFakerSchemaMocks` + existing writers) using the structured name set.
- Remove the regex dedupe layer; dedupe becomes prepend header from structured set.

## Motivation

Single strategy for strict mock type emission, less fragile whitespace coupling, simpler maintenance.

Internal refactor only — generated output should stay the same.

Hypenate:
## Summary

Follow-up refactor from strict mock types (#3525). Aligns the faker `schemas: true` path with MSW/operation mocks: generators only emit factories, and `finalizeMockImplementation` prepends strict helpers and `{Schema}Mock` aliases once from `strictMockSchemaTypeNames`.

- `generateFakerForSchemas` returns `strictMockSchemaTypeNames` and no longer inlines strict type blocks
- `writeFakerSchemaMocks` calls `finalizeMockImplementation` (same as other writers)
- `dedupeStrictMockTypeDeclarations` only prepends from the structured set (removed regex scraping/strip logic)

Closes #3542

## Test plan

- [x] `@orval/mock` unit tests
- [x] `orval-tests build` (mock client typecheck)
- [x] CI green on PR

## Summary by CodeRabbit

* **Improvements**
  * Faker mock generation now exposes strict schema type names separately, producing leaner mock implementations and enabling predictable header insertion and ordering.

* **Bug Fixes**
  * Finalized mock output is used when computing imports and emitting files so generated files reflect finalized implementation.

* **New Features**
  * Generated mock types and factory functions added for additional schemas (e.g., Status, PetList, Score).

* **Tests**
  * Updated/added tests covering strict-type exposure, deduplication/header behavior, and finalized-output flow.

### DIFF
--- a/packages/mock/src/faker/index.test.ts
+++ b/packages/mock/src/faker/index.test.ts
@@ -11,6 +11,7 @@ import { isFakerMock, isMswMock, OutputMockType } from '@orval/core';
 import { describe, expect, expectTypeOf, it } from 'vitest';
 
 import { createTestContextSpec } from '../../../core/src/test-utils/context';
+import { dedupeStrictMockTypeDeclarations } from '../mock-types';
 import {
   generateFaker,
   generateFakerForSchemas,
@@ -203,5 +204,41 @@ describe('generateFakerForSchemas strict mock types (#3525)', () => {
       ') as MockWithNullableOverrides<Pet, O, PetMock>;',
     );
     expect(result.implementation).not.toContain(', null]');
+  });
+
+  it('includes non-overridable strict schemas in strictMockSchemaTypeNames', () => {
+    const result = generateFakerForSchemas(
+      [
+        {
+          name: 'Status',
+          model: 'Status',
+          imports: [],
+          schema: {
+            type: 'string',
+            enum: ['active', 'inactive'],
+          },
+        },
+      ],
+      context,
+      { type: OutputMockType.FAKER, schemas: true },
+    );
+
+    expect(result.strictMockSchemaTypeNames).toEqual(['Status']);
+    expect(result.implementation).not.toContain('overrideResponse');
+    expect(result.implementation).toContain(
+      'export const getStatusMock = (): StatusMock =>',
+    );
+    expect(result.implementation).not.toContain('export type StatusMock = {');
+
+    const finalized = dedupeStrictMockTypeDeclarations(result.implementation, {
+      mockOptions: { required: true, nonNullable: true },
+      strictSchemaTypeNames: result.strictMockSchemaTypeNames,
+    });
+
+    expect(finalized).toContain('export type StatusMock = {');
+    expect(finalized).toContain('export type KeysWithNull<O>');
+    expect(finalized.indexOf('export type StatusMock')).toBeLessThan(
+      finalized.indexOf('export const getStatusMock'),
+    );
   });
 });
--- a/packages/mock/src/faker/index.test.ts
+++ b/packages/mock/src/faker/index.test.ts
@@ -171,7 +171,7 @@ describe('generateFakerForSchemas strict mock types (#3525)', () => {
     },
   });
 
-  it('emits PetMock alias and return type for schema factories', () => {
+  it('exposes strict schema names and omits inline type declarations', () => {
     const result = generateFakerForSchemas(
       [
         {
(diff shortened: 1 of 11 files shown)


Hypenate:
@wadakatu We merged the previous PR.
This PR contains the simplification you wanted earlier (I hope)
If you have time, can you review it?

wadakatu:
## Review — LGTM ✅

Beyond being a clean refactor, this actually fixes a latent TS2304 bug for non-overridable strict faker schemas. I regenerated and typechecked locally against a spec with top-level enum/array/scalar schemas.

### The one real behavioral change: dropping `&& isOverridable` (= a latent bug fix)

`getMockFactorySignatureParts` returns `${Type}Mock` as the return type even for non-overridable schemas (top-level enum/array/scalar) under strict mode, but on `master` the guard means the `XMock` type declaration is never emitted:

- **master**: `error TS2304: Cannot find name 'StatusMock' / 'PetListMock' / 'ScoreMock'` → doesn't compile
- **this PR**: the same spec typechecks ✅

Note that keeping the guard and only doing the refactor would have left this just as broken as `master`, so commit 3 is an intentional, standalone bug fix rather than a side effect of the refactor. Dropping the guard makes "every `XMock` referenced by a strict factory" line up exactly with `strictMockSchemaTypeNames`, which is why removing the scraping safety nets (`collectStrictMockSchemaNamesFromUsage`, etc.) is safe here. The MSW path is sound for the same reason: `applyStrictMockReturnType` and `strictMockSchemaTypeNames` both derive from the same source (`getSchemaTypeNamesFromResponses`), so completeness is guaranteed by construction.

### Suggestions (non-blocking)

1. **[Recommended] Add an e2e fixture for the non-overridable strict faker case.** This bug fix isn't guarded by the current CI gates (snapshot / `typecheck-generated`): the existing strict fixtures (issue-3525 family) only contain `Pet` (an object = overridable). Adding a top-level enum + array schema to the issue-3525 spec would lock the regression down in CI.
2. `dedupeStrictMockTypeDeclarations` is no longer idempotent (it prepends unconditionally without stripping an existing header). A one-line comment noting the "called once per file" invariant would make that safe to rely on.
3. Nit: `.replace(/^\n+/, '').trimStart()` — `.trimStart()` alone already covers leading newlines. And the `header ? ... : trimmedBody` false branch is effectively dead since we early-return when there are no names.

Nice cleanup 👍

Hypenate:
@wadakatu I summon you again :D 
I've implemented the 3 suggestions you gave.

wadakatu (approved):
All three are addressed cleanly — thanks for the quick turnaround! 🙌

Verified locally on the latest commit:

- **Suggestion 1 (fixture):** regenerated `issue-3525` and the new `Status` / `PetList` / `Score` schemas now emit `StatusMock` / `PetListMock` / `ScoreMock` and **typecheck** (incl. the odd-but-valid `ScoreMock` over `number`). The non-overridable strict-faker path is now locked down by both the snapshot and `typecheck-generated` gates.
- **Suggestion 2 (invariant):** the "not idempotent / call once per file" note matches how the writers and `writeFakerSchemaMocks` use it. 👍
- **Suggestion 3 (cleanup):** `${header}\n\n${implementation.trimStart()}` is output-equivalent (`header` is always non-empty once we're past the empty-names early return, and `.trimStart()` subsumes the old `.replace(/^\n+/, '')`).

Full check: `@orval/mock` units (219) green, snapshot suite green after a fresh regen, and `tsc --noEmit` over all of `generated/mock` is clean — no output changes anywhere outside `issue-3525`, so no regressions.

LGTM 🚀

melloware (approved):

PR #3552, opened 2026-06-06
fix(zod): avoid useless escapes in generated RegExp pattern literals

### LINKED ISSUE #3337: Zod generation on string property with pattern generates a wrongly escaped wildcard in RegExp
### Description

When generating a string property with a pattern ending with a `*` in its capture group, the generated string litteral (RegExp constructor parameter) is escaping the `*`

### Output client

zod

```ts
    output: {
      clean: true,
      mode: "single",
      client: "zod",
      target: "./schema.api.ts",
      schemas: "./models",
      override: {
        header: false,
        useNullForOptional: true,
        zod: {
          coerce: {
            query: ["number", "boolean"],
          },
        },
      },
    },

```

### Environment

  npmPackages:
    orval: ^8.9.1 => 8.9.1
    zod: ^4.4.3 => 4.4.3

### Expected behavior

`= new RegExp('^(0|[1-9]\\d*)$')`

### Actual behavior

`= new RegExp('^(0|[1-9]\\d\*)$')`

```yaml
openapi: 3.2.0
info:
  title: Demo
  version: 0.0.0
paths: {}
components:
  schemas:
    SBU:
      type: string
      pattern: '^(0|[1-9]\d*)$'
    PTS:
      type: object
      required:
        - hit
      properties:
        hit:
          $ref: "#/components/schemas/SBU"

```

### Additional context

_No response_

wadakatu:
## What

Generated Zod (and Effect) `new RegExp('...')` pattern literals over-escape `*` (and `/`, `"`). For the schema in #3337:

```ts
// before
export const SbuRegExp = new RegExp('^(0|[1-9]\\d\*)$');
// after
export const SbuRegExp = new RegExp('^(0|[1-9]\\d*)$');
```

The `\*` round-trips to the same regex at runtime, but ESLint's [`no-useless-escape`](https://eslint.org/docs/latest/rules/no-useless-escape) (part of `eslint:recommended`) flags it as `Unnecessary escape character: \*`, so generated code fails linting in consumers' projects.

## Why

The pattern is escaped with `jsStringEscape`, which — besides the characters a string literal actually needs — also escapes `/` and `*`. That extra escaping was added to neutralize JS **comment** delimiters (CVE-2026-25141 / #2882), where enum descriptions are embedded in a `/** … */` comment. It is correct for comments, but unnecessary (and lint-breaking) for single-quoted string literals such as RegExp patterns.

## How

- Add `jsStringLiteralEscape` in `@orval/core`, which escapes only what a single-quoted JS string literal needs: single quotes, backslashes, and line terminators (`\n`, `\r`, `\u2028`, `\u2029`). It intentionally does **not** escape `/`, `*`, or `"`.
- Use it for the RegExp pattern literals in `@orval/zod` (2 sites) and `@orval/effect` (1 site, same bug).
- `jsStringEscape` and the comment-generation path are **left completely untouched**, so the CVE-2026-25141 mitigation is fully preserved (verified: malicious `*/` in an enum description is still neutralized to `\*\/` and cannot break out of the comment; adversarial RegExp patterns cannot break out of the single-quoted literal).

## Tests

- `@orval/core`: unit tests for `jsStringLiteralEscape` (escapes `'`/`\`/line terminators; does **not** escape `*`/`/`/`"`).
- `@orval/zod` and `@orval/effect`: regression tests asserting the emitted `new RegExp('…')` has no `\*`.
- Existing snapshots are unchanged.

## Note (out of scope, possible follow-up)

A cleaner long-term direction is to emit regex **literals** (`z.string().regex(/^(0|[1-9]\d*)$/)`) instead of `new RegExp('…')`, which sidesteps string-escaping entirely — this is the direction Kubb took. It is an architectural change (named-const reuse/dedup, plus edge cases like `/` in patterns and empty patterns) and is deliberately left out here; happy to follow up if desired.

Closes #3337

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved regex pattern generation across core, effect, and zod packages to produce cleaner literal output without unnecessary escape sequences.

* **Tests**
  * Added test coverage for regex pattern generation to verify correct handling of edge cases and expected output format.

### DIFF
--- a/packages/core/src/utils/string.test.ts
+++ b/packages/core/src/utils/string.test.ts
@@ -5,6 +5,7 @@ import {
   escape,
   escapeRegExp,
   jsStringEscape,
+  jsStringLiteralEscape,
   stringify,
 } from './string';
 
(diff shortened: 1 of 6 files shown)


melloware (approved):

melloware:
@wadakatu se this one too https://github.com/orval-labs/orval/issues/3505 where the user submitted a PR but it broke other users so I reverted it.

PR #3559, opened 2026-06-06
feat(mock): support separate mock output paths

### LINKED ISSUE #3537: feat: support separate output paths for mocks and fakers
## Problem

With `mode: 'tags-split'` and `mock: true`, Orval writes `.faker.ts` and `.msw.ts` files next to the generated client services.

That is useful for simple projects, but package builds often want mocks in a separate test/dev entrypoint.

Example package layout:

```ts
@acme/client/angular
@acme/client/mocks
```

This keeps production client imports away from `msw` and `@faker-js/faker`.

## Suggested API

Allow configuring a separate mock output path:

```ts
output: {
  mock: {
    type: 'msw',
    path: './libs/client/mocks/src/lib',
  },
}
```

Or support separate paths for MSW handlers and faker data:

```ts
output: {
  mock: {
    type: 'msw',
    mswPath: './libs/client/mocks/src/lib',
    fakerPath: './libs/client/fakers/src/lib',
  },
}
```

## Expected behavior

When a mock output path is configured:

- `.msw.ts` files are written under the configured mock path.
- `.faker.ts` files are written under the configured faker path, or the shared mock path if only one path is provided.
- Generated client files do not need to live next to mock files.
- Imports between mock/faker files and client/model files remain valid.

## Related issue

This is related to #3318, where generated MSW files in `split` mode can be accidentally re-exported through a production barrel.

In jsdom/Vitest environments, importing MSW transitively can fail at module evaluation time because MSW v2 references stream globals such as `WritableStream`.

Separate mock output roots would make it easier to publish or import mocks from a dedicated test/dev entrypoint and keep MSW out of production client barrels.

## Why this helps

- Allows production client entrypoints to avoid `msw` and `@faker-js/faker`.
- Makes mocks easier to publish as separate package subpath exports.
- Better supports secondary entrypoints and monorepo package layouts.

### LINKED ISSUE #3554: Function generators are handled inconsistently across write modes
## Description

Function-form mock generators (`ClientMockBuilder`) are handled differently depending on the write mode:

| Mode | Behavior |
|------|----------|
| `single` | Treated as MSW |
| `tags` | Treated as MSW |
| `split` | Silently skipped (no mock output generated) |
| `tags-split` | Previously threw `Function mock generators ... are not supported in tags-split mode`. The throw was removed during #3537 implementation — now treated as MSW like the other modes. |

Upstream, `client.ts:304` already maps function generators to MSW:

```ts
type: isFunction(entry) ? OutputMockType.MSW : entry.type
```

## The inconsistency

1. **`split` mode silently drops function generators.** The `rawEntry` lookup finds the function, but the mock output type (`OutputMockType.MSW`) may not match what the generator produces. No error, no output.

2. **The old `tags-split` throw was removed** in the #3537 implementation to allow function generators to use the shared `mock.path`. The removal was necessary because `getMockDir()` needs to handle function generators (falling back to `mockConfig.path` since functions cannot have their own `path` property).

3. **`single` and `tags` modes** treat function generators as MSW and generate mock output inline (or de-inlined to separate files when `path` is set).

## Proposed resolution

Decide on one consistent behavior and apply it to all four modes:

- **Option A**: Support function generators everywhere. Map them to MSW consistently. This is the current de facto behavior after the #3537 changes, minus the `split` mode silent skip.
- **Option B**: Reject function generators explicitly in all modes with a clear error message, directing users to use the object form `{ type: "msw" }` instead.

## Context

Discovered during implementation of #3537 (mock `path` option for custom output directories). The `split-tags-mode` throw was removed to unify behavior, but `split-mode` still silently skips function generators.

aqeelat:
Adds support for writing mock outputs to separate directories via a shared `mock.path` or per-generator `path` values. This lets MSW handlers and faker fixtures live outside the production client entrypoint while preserving imports back to generated schemas.

Also fixes function-form mock generators so they are consistently treated as MSW across write modes, matching the upstream normalization in `client.ts`.

Closes #3537
Fixes #3554

Validation:
- `node_modules/.bin/vitest run --config vitest.config.ts packages/core/src/writers/split-mode.test.ts packages/core/src/writers/tags-mode.test.ts packages/core/src/writers/split-tags-mode.test.ts packages/core/src/writers/mock-utils.test.ts`
- `node_modules/.bin/vp run -F './packages/*' typecheck`
- `node_modules/.bin/vp lint --type-aware --type-check packages/core packages/orval`

Snapshot note:
Running `bun run test:snapshots:update` generated 199 files. I did not include those generated snapshot/fixture updates in this PR yet; I will add them after the source changes are reviewed.

## Summary by CodeRabbit

* **New Features**
  * Shared mock output path with optional per-generator path overrides; conditional de-inlining to emit mocks as separate files and optional per-directory index barrels.

* **Bug Fixes**
  * Function-form mock generators are treated as MSW in split and tags-split modes; deterministic ordering of index exports ensured.

* **Documentation**
  * Docs updated with mock.path, per-generator path examples, and indexMockFiles guidance.

* **Tests**
  * Added unit and regression tests for mock-path resolution, schema import paths, de-inlining, and index generation.


melloware (changes requested):
Build is failing

aqeelat:
@melloware It is because I didn't include the snapshot updates. they're > 150 files and will cause rabbit to not review the code.
I'll add them when the logic is done.

melloware:
@wadakatu please review this when available

packages/core/src/writers/single-mode.ts
```diff
@@ -230,14 +247,115 @@ export async function writeSingleMode({
...
+
+        await writeGeneratedFile(mockFilePath, mockData);
+        extraPaths.push(mockFilePath);
+
+        const indexKey = `${mockExtension}::${mockDir}`;
+        if (!seenMockIndexKeys.has(indexKey)) {
+          seenMockIndexKeys.add(indexKey);
+          writtenMockEntries.push({ extension: mockExtension, mockDir });
+        }
+      }
```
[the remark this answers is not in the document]
aqeelat: The premise here is incorrect. `rawMockOutputs` at `single-mode.ts:53` comes from `generateTarget(builder, output)`, which already merges per-operation `mockOutputs` by type.

`packages/core/src/writers/target.ts:67-90`:
```ts
for (const opMock of operation.mockOutputs) {
  let acc = target.mockOutputs.find((m) => m.type === opMock.type);
  if (!acc) {
    acc = emptyMockOutputFull(opMock.type);
    target.mockOutputs.push(acc);
  }
  acc.imports.push(...opMock.imports);
  acc.implementation.function += opMock.implementation.function;
  acc.implementation.handler += opMock.implementation.handler;
  // ...
}
```

By the time `writeSingleMode` runs, `rawMockOutputs` has at most **one entry per `OutputMockType`** (one MSW, one FAKER). The de-inline loop at `single-mode.ts:275-334` iterates at most two entries, and they write to **different files** — `<filename>.msw.ts` and `<filename>.faker.ts` — so there is no same-type collision.

The inline branch needs `collapseInlineMockOutputs` (`packages/core/src/writers/mock-outputs.ts:15-21`) because both MSW and FAKER are concatenated into the same `implementation` string; dropping FAKER when MSW is present avoids duplicate `get<Op>ResponseMock` declarations. The de-inline branch does not have that problem, and applying the same function here would *cause* data loss: with `mock.generators = [{ msw }, { faker }]`, `collapseInlineMockOutputs` would drop the faker entry, so `<filename>.faker.ts` would never be written.

If the concern is that a future refactor could regress the per-type merge, the right defensive guard is an assertion in `writeSingleMode`, e.g. before the loop:

```ts
const seenMockTypes = new Set<string>();
for (const m of rawMockOutputs) {
  if (seenMockTypes.has(m.type)) {
    throw new Error(
      `writeSingleMode expected at most one mock output per type, got duplicate ${m.type}`,
    );
  }
  seenMockTypes.add(m.type);
}
```

Happy to add that as a separate commit if you want belt-and-suspenders, but the current behavior is correct as-is and the proposed fix would regress the de-inline faker file.

suzunn:
I noticed the implementation intentionally keeps the generated snapshot refresh out of this PR for reviewability, but that is also where a lot of the new path/import behavior will be exercised. Before merging, I would add at least one committed fixture that combines `mock.path`, per-generator `path`, `indexMockFiles`, and `output.schemas` using the `.schemas` suffix case covered by `resolveMockSchemasPath`.

That would lock down the highest-risk contract here: generated MSW/faker files living outside the client tree still import schemas from the right relative path, and the per-directory barrels stay deterministic after tags are sorted. The unit tests cover the helpers well, but a generated-output fixture would catch a future writer-level regression where one mode bypasses the shared helper or computes the mock directory before normalization.

aqeelat:
@suzunn already added in a previous commit.

packages/core/src/writers/tags-mode.ts
```diff
@@ -260,5 +367,25 @@ export async function writeTagsMode({
     }),
   );
 
+  if (shouldDeinlineMocks && output.mock.indexMockFiles) {
+    for (const { ext, mockDir, tags } of mockIndexEntries) {
+      const indexPath = path.join(mockDir, `index.${ext}${extension}`);
+      const indexContent = tags
```
wadakatu, line 375: **`tags` mode + `mock.path` + `indexMockFiles` produces a non-deterministic barrel.**

The ordering fix from `f46b20db2` was only applied to `split-tags-mode.ts`, not here. `mockIndexEntries[].tags` is populated inside the concurrent `Promise.all(tagEntries.map(async …))` loop via `indexEntry.tags.push(kebabTag)`, so insertion order depends on which async write resolves first — and this barrel emit maps over `tags` without sorting.

Generating the same spec 8× (petstore, tags `pets`/`health`) flips the order between runs:

```
run 1: export { getPetsMock }   … ; export { getHealthMock } …
run 2: export { getHealthMock } … ; export { getPetsMock }   …
run 5: export { getPetsMock }   … ; export { getHealthMock } …
(other runs: health → pets)
```

This causes regeneration diff noise for users and would make any future fixture for this combo a flaky snapshot. `split-tags-mode` is unaffected (it sorts) and `single` mode is unaffected (single mock file per type).

Fix — sort at emit time, mirroring `split-tags-mode.ts`:

```ts
const sortedTags = [...tags].sort();
const indexContent = sortedTags.map((kebabTag) => { … }).join('');
```

tests/configs/mock.config.ts
```diff
@@ -655,4 +716,39 @@ export default defineConfig({
       target: '../specifications/faker-array-items-tags-split.yaml',
     },
   },
+  splitMockPath: {
```
wadakatu, line 821: **Committed fixtures don't cover the most complex new path code.**

These configs only exercise a **shared** `mock.path` in `split` / `tags-split`. The following new behaviors have no committed fixture — I verified them manually and they generate correctly, but there's no regression net:

| Combination | Covered? |
|---|---|
| `single` mode + `path` (de-inline path) | ❌ |
| `tags` mode + `path` (+ `indexMockFiles`) | ❌ — this is exactly where the determinism bug lives |
| per-generator distinct `path` (e.g. `msw` → `./msw`, `faker` → `./faker`) | ❌ |

The de-inline branch in `single-mode.ts` / `tags-mode.ts` is the largest piece of new logic, so a fixture there is the highest-value addition (and a `tags`+`path`+`indexMockFiles` fixture would have caught the ordering bug). This is the part of @suzunn's earlier request that's still open — `single`/`tags` and per-generator paths aren't in the current fixtures.

wadakatu (commented):
@melloware @aqeelat — took a thorough pass through the source changes (the 7 non-snapshot files) and exercised the new code paths locally. The core design is solid: the de-inline logic preserves existing no-path behavior (only the `index-mock-file` sort diff changed among existing snapshots), import resolution is correct across all the path layouts I tried (`mocks/ → ../model`, `mocks/<tag>/ → ../../model`, and the inline-schema `../endpoints` / `../endpoints.schemas` cases), and the #3554 function-generator unification works as intended.

Two things I'd want addressed before merge, left as inline comments below: one confirmed determinism bug, and a test-coverage gap around the new path code.

melloware:
@aqeelat let @wadakatu know when this is ready for re-review.

aqeelat:
@coderabbitai review --dir packages
@coderabbitai review --dir docs

aqeelat:
@melloware @wadakatu the PR is ready

packages/core/src/writers/tags-mode.ts
wadakatu: Thanks for the fast turnaround — and the three new fixtures (`singleMockPath`, `tagsMockPath`, `tagsSplitPerGeneratorPath`) close the coverage gap nicely. 🙏

However, the determinism fix in `e9e2d838a` doesn't actually resolve it, and it regressed `split-tags-mode`. Sorting `Object.entries(target)` at the source doesn't constrain the barrel order, because `indexEntry.tags` is accumulated **after** `await writeGeneratedFile(...)` inside the concurrent `Promise.all(tagEntries.map(async …))` — so the push order follows I/O completion, not the sorted input order. Removing the emit-time `[...tags].sort()` therefore re-opens the race.

Verified locally by regenerating the new fixtures 10× each:

```
tagsMockPath        (tags + path + index):       still flips health↔pets across runs
tagsSplitMockPath   (split-tags, per-gen paths): now ALSO flips  ← regressed by removing the emit-time sort
```

So `tagsMockPath`, `tagsSplitMockPath`, and the existing `index-mock-file` snapshots are now flaky — CI passing here is just luck of the draw on the I/O completion order.

The robust fix is to keep the sort at **emit time** (where the order actually matters), in both writers:

```ts
// tags-mode.ts and split-tags-mode.ts — barrel emit loop
const indexContent = [...tags]
  .sort()
  .map((tag) => { … })
  .join('');
```

With that restored, all three fixtures are stable across 10/10 regenerations. The `.toSorted()` on `tagEntries` is harmless but isn't sufficient on its own.

wadakatu:
@aqeelat Left a follow-up on the determinism thread — the new fixtures look great, but the `toSorted()` approach in e9e2d838a doesn't actually fix the ordering (and removing the emit-time sort regressed `split-tags-mode`). I verified it still flips across regenerations. Details + the one-line fix here: https://github.com/orval-labs/orval/pull/3559#discussion_r3376938636 — could you take a look when you get a chance? 🙏

aqeelat:
@coderabbitai full review but ignore `tests/__snapshots__/*`

aqeelat:
@melloware is it a good idea to configure coderabbit to ignore `tests/__snapshots__/*` by default?

aqeelat:
@wadakatu I now understand the issue. I fixed it and added a test case for it. The test has a 10ms delay but I couldn't get around that.

melloware:
@aqeelat sorry I just merged another PR and you have merge conflicts!  Apologies!

melloware:
@aqeelat i just asked @anymaniax if we can ignore `tests/snapshots` i will let you know what he says!

aqeelat:
@melloware rebased

melloware (approved):

PR #3575, opened 2026-06-09
fix(mock): emit strict mock types in tags-split MSW+faker output

### LINKED ISSUE #3574: fix(mock): prepend strict mock type declarations in tags-split MSW/faker output
## Problem

Orval 8.16.0 generates mock factory signatures that reference strict mock helper types (`MockWithNullableOverrides`, `{Schema}Mock`, etc.) but does not emit the corresponding `export type` declarations when all of the following are combined:

- `output.mode: 'tags-split'`
- `override.mock.required: true` and `override.mock.nonNullable: true` (strict mock mode, #3529)
- `mock.generators: [{ type: 'msw' }, { type: 'faker', arrayItems: true }]`
- `mock.indexMockFiles: true`

This produces TypeScript errors such as:

```
error TS2304: Cannot find name 'MockWithNullableOverrides'.
error TS2304: Cannot find name 'PetMock'.
error TS2552: Cannot find name 'getPetMock'. Did you mean 'getGetPetsResponseMock'?
```

## Why existing tests do not catch it

Strict-mock fixtures (`issue-3525`, `issue-3525-multi`, etc.) use `client: 'fetch'` and a single-file / non-tags-split layout. They do not cover tags-split per-tag `.msw.ts` + `.faker.ts` with `arrayItems: true`.

In that layout, `strictMockSchemaTypeNames` was dropped when building tag targets, so `finalizeMockImplementation` / `dedupeStrictMockTypeDeclarations` never prepended strict types. Faker files also referenced schema factories emitted only in the MSW file because array-item factory dedup was scoped per tag but not per mock generator file.

## Expected behavior

Each generated mock file that references strict mock types should include (once per file):

- `export type KeysWithNull<O> = { ... };`
- `export type MockWithNullableOverrides<...> = ...;`
- `export type PetMock = { [K in keyof Required<Pet>]: NonNullable<Required<Pet>[K]>; };`

Faker files must define or import schema factories they call (e.g. `getPetMock()`).

## Related PRs

- #3529 — strict mock return types when `required` + `nonNullable` are both true
- #3543 — emit strict types once via `strictMockSchemaTypeNames` / `finalizeMockImplementation`

Hypenate:
## Summary

- Preserve `strictMockSchemaTypeNames` when building tag targets in `tags-split` mode so `finalizeMockImplementation` / `dedupeStrictMockTypeDeclarations` prepends `KeysWithNull`, `MockWithNullableOverrides`, and `{Schema}Mock` aliases in per-tag `.msw.ts` and `.faker.ts` files.
- Scope array-item factory dedup per mock generator file (`msw` vs `faker`) so faker output defines schema factories (e.g. `getPetMock`) it references instead of relying on the MSW file.
- Add regression fixtures for #3574: OpenAPI spec, angular + fetch mock configs, snapshots, and a focused assertion in `api-generation.spec.ts`.

Fixes #3574

## Test plan

- [x] `bun test` in `@orval/mock` (includes new array-item factory scope test)
- [x] `bun run generate:mock` for `issue3574StrictMockTagsSplitAngular` and `issue3574StrictMockTagsSplitFetch`
- [x] Snapshot tests cover new fixtures under `tests/__snapshots__/mock/issue-3574-*`
- [x] `node tests/scripts/typecheck-generated.mjs` passes (mock folder includes new generated output)
- [x] Existing `issue-3525` / `issue-3525-multi` / `issue-3525-widget-mock` snapshots unchanged

## Summary by CodeRabbit

* **New Features**
  * Mock generation now scopes outputs per mock type (e.g., MSW vs Faker) and consistently preserves strict mock type names across generated mock artifacts.
  * Generated client and mock modules for pets/store include faker-backed factories, MSW handlers, nullable-aware override typings, and new index entrypoints for simpler imports.

* **Tests**
  * Added tests, snapshots and configs validating output-type-specific mock emission and strict-mock type handling.


Hypenate:
@wadakatu 

Can you verify the fix?
We're having some troubles with the latest version:

```
✘ [ERROR] TS2304: Cannot find name 'MockWithNullableOverrides'. [plugin angular-compiler]

    libs/content-engine/shared/api/src/lib/generated/admin/admin.msw.ts:99:8:
      99 │   }) as MockWithNullableOverrides<BatchDto, O, BatchDtoMock>;
         ╵         ~~~~~~~~~~~~~~~~~~~~~~~~~


✘ [ERROR] TS2304: Cannot find name 'BatchDtoMock'. [plugin angular-compiler]

    libs/content-engine/shared/api/src/lib/generated/admin/admin.msw.ts:99:47:
      99 │   }) as MockWithNullableOverrides<BatchDto, O, BatchDtoMock>;
```

This PR should fix this.

wadakatu:
Thanks for the fix @Hypenate. I verified the per-generator file scoping (B) and the first-op / wrapped-output preservation (A) — both work. But there's a remaining drop point in (A), and I think it still leaves the reporter's real-world case broken.

### Residual bug: `mergeOperationMockOutputs` drops `strictMockSchemaTypeNames`

In `packages/core/src/writers/target-tags.ts`, `mergeOperationMockOutputs` is called for **every operation after the first one in a tag**. Its initial `.map()` copies `imports` but not `strictMockSchemaTypeNames`:

```ts
// target-tags.ts (~L50)
const result: GeneratorMockOutputFull[] = accMockOutputs.map((m) => ({
  type: m.type,
  implementation: { ...m.implementation },
  imports: [...m.imports],
  // strictMockSchemaTypeNames is NOT carried over here
}));
```

So the accumulator's strict names are reset to `undefined` on each merge, and the loop below only re-adds the *current* op's names. Net effect: **only the schema referenced by the last-merged operation in a tag keeps its strict type.** (The non-tags `target.ts` mutates the same `acc` object in place, so it accumulates correctly — only the tags/tags-split path has this asymmetry. Note this code also serves `mode: 'tags'`.)

### Repro (built this PR, then generated)

One tag (`Store`) with two operations returning different schemas — `getPets → Pet[]`, `getOwner → Owner` — under `tags-split` + strict + `generators: [msw, faker]`:

```ts
// generated store.msw.ts
export type OwnerMock = { ... };        // last op only — PetMock is never declared
export const getPetMock = <O extends Partial<Pet> = {}>(overrideResponse?: O):
  MockWithNullableOverrides<Pet, O, PetMock> => ...   // references undeclared PetMock
```

```
store.msw.ts(41):  error TS2304: Cannot find name 'PetMock'.
store.faker.ts(33): error TS2304: Cannot find name 'PetMock'.   (x3)
```

This is the same error class as #3574. A real tag with many operations/DTOs (e.g. the reporter's `admin` module with `BatchDto`) would still fail after this PR — every DTO except the last operation's would be missing its `Mock` type. The current regression fixture doesn't catch it because both ops reference the same `Pet` (so the last merge re-adds it) — I'll note the coverage side separately.

### Suggested fix (mirror `initialMockOutputsForOperation`)

```ts
const result: GeneratorMockOutputFull[] = accMockOutputs.map((m) => ({
  type: m.type,
  implementation: { ...m.implementation },
  imports: [...m.imports],
  strictMockSchemaTypeNames: m.strictMockSchemaTypeNames
    ? [...m.strictMockSchemaTypeNames]
    : undefined,
}));
```

With this, both `PetMock` and `OwnerMock` are emitted in the `.msw.ts` and `.faker.ts` files and `tsc` reports **0 TS2304**. I also re-generated the existing `issue-3574-strict-mock-tags-split-fetch` fixtures with this change and they are **byte-identical** to the committed snapshots, so it doesn't regress the single-schema case.

wadakatu:
### Test coverage gap (separate from the residual bug above)

The regression fixtures (`issue-3574-strict-mock-tags-split` / `-fetch`) put a single schema (`Pet`) behind both operations of the one tag. Because both ops reference the same schema, the `mergeOperationMockOutputs` path re-adds `Pet` on the final merge, so the suite stays green even when per-operation strict names are being dropped. In other words, the current fixture can't distinguish "correctly accumulated" from "only the last op survived".

To actually guard the merge path, I'd suggest a fixture where **one tag has ≥2 operations returning distinct schemas**, e.g. a `Store` tag with `getPets → Pet[]` and `getOwner → Owner`. Then assert that **each** generated `.msw.ts` / `.faker.ts` declares `export type PetMock` *and* `export type OwnerMock` (or run `tsc` over the generated output). That fixture fails on `master` + this PR today and passes with the one-line fix.

Hypenate:
@wadakatu Applied your comments

wadakatu:
LGTM — thanks for the quick turnaround, and nice catch going past my two notes to the actual nested-factory case. I rebuilt the PR at b490ff4e5 and verified:

- **Residual merge bug** (`mergeOperationMockOutputs`): fixed exactly as suggested, and the new multi-schema `Store` fixture/test guards it. 👍
- **Nested split factories** (`f2eb97327`): I reproduced the reporter-style case — an `Admin` tag with `getBatches → BatchDto[]` (arrayItems), `getTenant → TenantInfoDto`, and `getReport → ReportDto` where `ReportDto.lines` is a nested `$ref` array of `LineItemDto`. All of `BatchDtoMock` / `TenantInfoDtoMock` / `ReportDtoMock` / `LineItemDtoMock` are now declared in both `admin.msw.ts` and `admin.faker.ts`. `LineItemDto` only ever appears via the nested item factory, so it's a good proof that the implementation scan does its job.
- `tsc` over the generated output: **0 `TS2304`**.
- `@orval/mock` unit tests: **232 passing**.
- No churn in existing snapshots (only the new `issue-3574-...-multi-fetch` fixture), CI green.

🚢 from my side.

wadakatu:
Two **non-blocking** nits for a possible follow-up — feel free to ignore for this PR:

1. Recovering schema names by regex-scanning generated text (`collectStrictMockSchemaTypeNamesFromImplementation`) is pragmatic and well tested, but it's reverse-parsing the output, so it's coupled to the exact factory syntax. The structurally robust place to collect a nested item's schema name is where the `$ref` split factory is *generated* (the array-item / faker getters, where the ref name is known), then thread it into `strictMockSchemaTypeNames`. Worth at most a code comment noting the coupling.

2. The `WidgetMock` false-positive guard (`filter(name => !names.has(`${name}Mock`))`) has one residual edge: if a spec has *both* a `Widget` and a `WidgetMock` schema used as strict mocks in the same file, `Widget` gets dropped. Extremely contrived, just flagging it.

Hypenate:
I'll do a follow-up PR with your remarks when the PR from @kfrederix has been merged. 
And if that is ok with @melloware

melloware (approved):

PR #3578, opened 2026-06-09
refactor(mock): tighten strict schema name recovery (#3574 follow-up)

### LINKED ISSUE #3574: fix(mock): prepend strict mock type declarations in tags-split MSW/faker output

Hypenate:
Follow-up to #3575 (closes #3574).

#3575 landed the tags-split strict mock type fix; this PR addresses two review nits from that work:

## Changes

- **Document regex coupling** — `collectStrictMockSchemaTypeNamesFromImplementation` now notes it reverse-parses emitted factory syntax and that the robust long-term fix is threading schema names from array-item / faker getters at generation time.
- **Tighter alias recovery** — Replace the broad prefix scan and Widget/WidgetMock filter with targeted patterns:
  - `MockWithNullableOverrides` first type param (unchanged)
  - Third generic alias → schema via inverse of `getStrictMockTypeName`
  - Non-overridable array return types like `): TenantInfoDtoMock[]`

This fixes the residual edge case where both `Widget` and `WidgetMock` strict mocks in the same file would drop `Widget`, while keeping `issue-3525-widget-mock-strict` snapshot-stable.

## Test plan

- [x] `vitest run packages/mock/src/mock-types.test.ts`
- [x] New test: both `Widget` and `WidgetMock` collected when present
- [x] Existing WidgetMock-only regression test passes
- [x] `issue-3525-widget-mock-strict` / `issue-3574` fixtures unchanged

Relates to #3574

## Summary by CodeRabbit

* **Tests**
  * Added a unit test to verify collection of multiple strict-mock schema names across factory implementations.

* **Refactor**
  * Improved internal pattern-matching for extracting strict-mock schema type names to increase accuracy and maintainability.

* **Documentation**
  * Added clarifying inline comments describing handling of nested factories and name-threading considerations.

### DIFF
--- a/packages/mock/src/mock-types.test.ts
+++ b/packages/mock/src/mock-types.test.ts
@@ -288,6 +288,19 @@ describe('mock-types', () => {
         collectStrictMockSchemaTypeNamesFromImplementation(implementation),
       ).toEqual(['WidgetMock']);
     });
+
+    it('keeps both Widget and WidgetMock when both schemas are strict-mocked', () => {
+      const implementation = [
+        'export const getWidgetMock = <O extends Partial<Widget> = {}>(overrideResponse?: O): MockWithNullableOverrides<Widget, O, WidgetMock> => ({}) as MockWithNullableOverrides<Widget, O, WidgetMock>;',
+        'export const getWidgetTypeMock = <O extends Partial<WidgetMock> = {}>(overrideResponse?: O): MockWithNullableOverrides<WidgetMock, O, WidgetMockMock> => ({}) as MockWithNullableOverrides<WidgetMock, O, WidgetMockMock>;',
+      ].join('\n');
+
+      expect(
+        collectStrictMockSchemaTypeNamesFromImplementation(
+          implementation,
+        ).toSorted(),
+      ).toEqual(['Widget', 'WidgetMock']);
+    });
   });
 
   describe('buildStrictMockTypeFileHeader', () => {
--- a/packages/mock/src/mock-types.ts
+++ b/packages/mock/src/mock-types.ts
@@ -222,11 +222,25 @@ export function applyStrictMockReturnType(
 
 const STRICT_MOCK_SCHEMA_TYPE_FROM_OVERRIDES =
   /MockWithNullableOverrides<([A-Z]\w*),/g;
-const STRICT_MOCK_SCHEMA_TYPE_FROM_MOCK_ALIAS = /\b([A-Z]\w*)Mock\b/g;
+const STRICT_MOCK_SCHEMA_TYPE_FROM_OVERRIDE_ALIAS =
+  /MockWithNullableOverrides<[^,]+,\s*[^,]+,\s*([A-Z]\w*Mock)>/g;
+const STRICT_MOCK_SCHEMA_TYPE_FROM_MOCK_ALIAS_RETURN =
+  /\): ([A-Z]\w*Mock)(?:\[\]|;)/g;
+
+/** Inverse of {@link getStrictMockTypeName}: `PetMock` → `Pet`, `WidgetMockMock` → `WidgetMock`. */
+function getSchemaTypeNameFromStrictMockAlias(alias: string): string {
+  return alias.endsWith('Mock') ? alias.slice(0, -4) : alias;
+}
 
 /**
  * Collect schema type names referenced by strict mock factories in generated
  * implementation text (nested split factories, array item helpers, etc.).
+ *
+ * This reverse-parses emitted factory syntax and is therefore coupled to the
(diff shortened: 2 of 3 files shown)


Hypenate:
@wadakatu this PR addresses your comments on #3575

wadakatu:
**Verified locally — LGTM (non-blocking notes below)** ✅

**What I checked:**

- `mock-types.test.ts` 25/25, full `@orval/mock` suite 233/233
- Regenerated all 7 strict-mock fixture configs (`issue3525*`, `issue3574StrictMockTagsSplit*`) — **zero churn**, both on this branch and after merging current master (#3577) locally
- Regex coverage against actual emit shapes: nested split factories (`value.ts` / `array-item-factory.ts`) are always emitted overridable with the full `MockWithNullableOverrides<T, O, TMock>` signature, so they're covered by both the first-param and third-param patterns. The shapes the new patterns *don't* match (e.g. non-overridable scalar returns like `(): StatusMock =>` in `issue-3525/model/index.faker.ts`) all come from paths that thread names structurally (faker writer) or are already covered by response-derived `schemaTypeNames` — I couldn't construct a false negative.
- New combination enabled by #3577 (strict mock + MSW-only + `arrayItems: true`, tags-split, nested `$ref` array items): generated output recovers the nested item alias via this scan and compiles clean with `tsc --strict`.

**Non-blocking — the Widget/WidgetMock co-exist case still doesn't compile end-to-end:**

The collection fix works (both names are now kept), but when schemas `Widget` and `WidgetMock` actually co-exist in one strict-mocked file, the emitted file still fails `tsc`: the strict alias for `Widget` is *named* `WidgetMock`, which collides with the schema import:

```
error TS2440: Import declaration conflicts with local declaration of 'WidgetMock'.
error TS2344: Type 'WidgetMock' does not satisfy the constraint 'Record<keyof Widget, unknown>'.
```

To be clear, **this is not a regression** — I ran the same spec against master and it also fails (`TS2344` ×4, because the dropped `Widget` alias makes `WidgetMock` resolve to the schema import). Both before and after, this input produces non-compiling output; this PR just changes the failure mode. The remaining fix would be collision-aware alias naming (detect that `{name}Mock` is taken by a real schema and pick a different alias), which I'd suggest tracking as a follow-up issue rather than expanding this PR. Repro spec, strict + tags-split:

```yaml
Order:
  type: object
  properties:
    widgets:
      type: array
      items:
        $ref: '#/components/schemas/Widget'
    widgetMocks:
      type: array
      items:
        $ref: '#/components/schemas/WidgetMock'
```

**Nit:** the `;` alternative in `STRICT_MOCK_SCHEMA_TYPE_FROM_MOCK_ALIAS_RETURN` (`): XxxMock;`) never fires today — `formatMockFactoryDeclaration` always emits ` => ` right after the return annotation, and the scan runs before prettier. Harmless belt-and-suspenders, fine to keep.

melloware (approved):

PR #3591, opened 2026-06-12
fix(mock): strict faker schema mock types for enums, nested spreads, and binary

### LINKED ISSUE #3590: Strict faker mocks (required+nonNullable): enum *Mock types, nested spreads, and Blob/ArrayBuffer mismatch
## Summary

With Orval **8.17.0**, enabling both `override.mock.required` and `override.mock.nonNullable` (strict mock feature from #3525 / PR #3529) causes **compile-time TypeScript errors** in generated `schemas/index.faker.ts` when `mock.generators` includes `{ type: 'faker', schemas: true }`.

Runtime mock values look correct; the failures are type mismatches in generated code.

**Workaround:** remove `required: true` and `nonNullable: true` — all errors disappear, but we lose strict mock return types needed for Cypress e2e (`getXMock({ tag: null })` narrowing per #3525).

## Environment

- Orval: 8.17.0
- TypeScript: 5.9
- Output: `tags-split`, client: `angular`
- Schema suffix: `Dto`
- `mock.generators: [{ type: 'faker', schemas: true, arrayItems: true }]`

## Reproduction

A sanitized minimal repro lives in this repo at:

`tests/repro/strict-faker-mock-type-errors/`

```bash
bun run build:release
cd tests/repro/strict-faker-mock-type-errors
bun ../../../packages/orval/dist/bin/orval.mjs --config orval.config.ts
bunx tsc --noEmit --project tsconfig.json
```

### `orval.config.ts`

```ts

export default defineConfig({
  repro: {
    input: { target: './repro.yaml' },
    output: {
      mode: 'tags-split',
      target: './generated',
      schemas: './generated/schemas',
      client: 'angular',
      mock: {
        generators: [{ type: 'faker', schemas: true, arrayItems: true }],
      },
      override: {
        components: { schemas: { suffix: 'Dto' } },
        mock: {
          required: true,
          nonNullable: true,
          properties: {
            '/^([Ii]d)$/': () => faker.string.uuid(),
          },
        },
      },
    },
  },
});
```

### Minimal OpenAPI (`repro.yaml`)

See `tests/repro/strict-faker-mock-type-errors/repro.yaml` — covers string enums, nested object arrays, nullable properties, and `format: binary`.

## Observed errors (6 total in `generated/schemas/index.faker.ts`)

```
error TS2552: Cannot find name 'IntegrationDtoSettingsItemMock'. Did you mean 'IntegrationDtoSettingsItem'?
error TS2352: Conversion of type '{ ... channels: ChannelTypeDtoMock[] ... type: SettingValueTypeDto | SettingValueTypeDtoMock; }'
  to type 'MockWithNullableOverrides<SettingMetaDto, O, SettingMetaDtoMock>' may be a mistake ...
error TS2352: Conversion of type '{ channels: ChannelTypeDtoMock[]; settings: SettingMetaDto[]; ... }'
  to type 'MockWithNullableOverrides<ProviderMetaDto, O, ProviderMetaDtoMock>' may be a mistake ...
error TS2352: Conversion of type '{ ... activeChannels: ChannelTypeDtoMock[]; ... providerType: ProviderKindDto | ProviderKindDtoMock; }'
  to type 'MockWithNullableOverrides<IntegrationDto, O, IntegrationDtoMock>' may be a mistake ...
error TS2740: Type 'ArrayBuffer' is missing the following properties from type 'UploadFileDtoMock': size, type, arrayBuffer, bytes, and 2 more.
```

---

## Bug 1: String-union enum `*Mock` types collapse to `{}`

For const-object enums like:

```ts
export type ChannelTypeDto = (typeof ChannelTypeDto)[keyof typeof ChannelTypeDto];
```

Orval generates:

```ts
export type ChannelTypeDtoMock = {
  [K in keyof Required<ChannelTypeDto>]: NonNullable<Required<ChannelTypeDto>[K]>;
};
```

Because `ChannelTypeDto` is a string-literal union, `keyof Required<ChannelTypeDto>` is `never`, so `ChannelTypeDtoMock` becomes `{}`.

Mock factory:

```ts
export const getChannelTypeDtoMock = (): ChannelTypeDtoMock =>
  faker.helpers.arrayElement(['sms', 'rcs', 'chat', 'mms'] as const);
```

Parent object literals then infer `ChannelTypeDtoMock[]` and `ProviderKindDto | ProviderKindDtoMock`, which fail the `as MockWithNullableOverrides<...>` casts on parent factories (`getSettingMetaDtoMock`, `getProviderMetaDtoMock`, `getIntegrationDtoMock`).

**Expected fix:** For string-union / const-object enum schemas, `{Schema}Mock` should alias the enum union itself (e.g. `ChannelTypeDtoMock = ChannelTypeDto`), not use the `keyof Required<T>` mapped type.

This also affects the existing `issue-3525` fixture: `StatusMock` is emitted as the same broken mapped type in `tests/__snapshots__/mock/issue-3525/model/index.faker.ts`.

---

## Bug 2: Nested mock spread incompatible with strict cast

Nested arrays use spread of generic mock factories:

```ts
settings: Array.from({ length: n }, () => ({
  ...getSettingMetaDtoMock(),
})),
```

`getSettingMetaDtoMock()` returns `MockWithNullableOverrides<SettingMetaDto, O, SettingMetaDtoMock>`, but the parent `*DtoMock` expects `SettingMetaDto[]` (plain DTO element types). The `as MockWithNullableOverrides<...>` cast on the parent fails.

Additionally, inline nested schemas (e.g. `Integration.settings[]` items) reference `IntegrationDtoSettingsItemMock` in factory signatures but **do not emit** the `{Schema}Mock` type alias.

**Expected fix (pick one or combine):**

1. Emit nested values typed as the strict `{Nested}Mock` base type (not the full `MockWithNullableOverrides` return type), and emit missing inline-schema `*Mock` aliases.
2. Call nested factories in a way TypeScript can structurally verify.
3. Use `as unknown as MockWithNullableOverrides<...>` in generated output (less ideal).

---

## Bug 3: Binary schema DTO type (`Blob`) vs faker mock (`ArrayBuffer`) mismatch

OpenAPI:

```yaml
UploadFile:
  type: string
  format: binary
```

Generated DTO:

```ts
export type UploadFileDto = Blob;
```

Generated mock:

```ts
export type UploadFileDtoMock = {
  [K in keyof Required<UploadFileDto>]: NonNullable<Required<UploadFileDto>[K]>;
};
export const getUploadFileDtoMock = (): UploadFileDtoMock =>
  new ArrayBuffer(faker.number.int({ min: 1, max: 64 }));
```

TS error: `ArrayBuffer` is not assignable to `UploadFileDtoMock` (which resolves to `Blob` properties via the mapped type).

This is a generator inconsistency: DTO generator maps `format: binary` → `Blob`, faker generator maps binary → `ArrayBuffer` (intentional for MSW per #3065 / PR #3248). For faker-only schema mocks (no MSW handler), return type and value should align.

**Expected fix:** When emitting faker schema mocks for binary types, either:

1. Generate `UploadFileDtoMock = ArrayBuffer` and `getUploadFileDtoMock(): ArrayBuffer`, or
2. Generate `new Blob([...])` to match `Blob` DTO type, or
3. Use a dedicated binary mock type consistently across DTO + faker.

Related: #2934 (Blob typing).

---

## What works without errors

Disabling strict mode avoids all errors:

```ts
// Remove:
// required: true,
// nonNullable: true,
```

---

## Related

- #3525 — strict mock return types (closed, implemented in PR #3529)
- PR #3529 — `MockWithNullableOverrides` / `{Schema}Mock` generation
- #2934 — Blob typing in mocks
- #3065 / PR #3248 — `ArrayBuffer` for binary MSW mocks

## Note

Errors appear in **`schemas/index.faker.ts`** with `schemas: true`, not only in operation-level `.faker.ts` files.

Hypenate:
## Summary

- Emit schema-aware strict `{Schema}Mock` aliases: mapped types for objects, type aliases for enums/scalars/arrays, and `ArrayBuffer` for binary faker schema mocks
- Cast nested `get*Mock()` spreads to base mock types so `MockWithNullableOverrides` parent casts typecheck
- Collect inline split-factory schema names for missing mock type declarations in `schemas/index.faker.ts`
- Add petstore regression spec `tests/specifications/issue-3590.yaml` (follows `issue-3525.yaml` style)

Fixes #3590

## Test plan

- [x] `packages/mock` unit tests (247 pass)
- [x] `tests/generated/mock` typecheck passes (`node tests/scripts/typecheck-generated.mjs`)
- [x] Snapshot test `mock issue-3590 strict faker schema mocks typecheck and emit correct aliases`

## Summary by CodeRabbit

* **New Features**
  * Generated mocks now include per-type schema "kinds" (object | alias | binary) and propagate this metadata through mock and MSW outputs.
  * Improved import recovery and de-duplication so consolidated schema factory imports are restored when needed.

* **Bug Fixes**
  * Stricter mock typings and casts for delegated/array/binary schemas to produce correct aliasing and binary mappings.

* **Tests**
  * New fixtures and tests covering kind classification, import recovery, and regression cases for strict mock generation.


Hypenate:
@wadakatu We still have some issues wit the latest version of Orval.
But it's already in a much beter state, so hopefully this will be the nail in the coffin (famous last words).

Hypenate:
I'm going for a full AI approach here...

I'll let it make a local build of Orval and check again against all our yaml files.
I'll let it itterate over this processes until all our errors are gone.

I'm not sure how we'll do the review process here...
I'll wait a bit for what it spits out.

Hypenate:
@wadakatu & @melloware 

The build is green, but we have almost 100 changed files...
How do we proceed? 

Do we go yolo mode or do I split this up?

melloware:
@Hypenate a lot of the changes are because it added tests which I am OK with.  I usually don't look too much at the tests but at the core code changed. It doesn't seem "too bad" but i will let @wadakatu review.

PR #3600, opened 2026-06-14
fix(mcp): order handler body/query args and propagate optional body to handler + tool schema

### LINKED ISSUE #3599: [mcp] Generated handler swaps body/query arguments for operations with a required query param + optional body
### Description

For an `mcp`-client operation that has **both a query parameter and a request body**, the generated `handlers.ts` always calls the http-client with a fixed **body-first** argument order:

```ts
await createItem(args.bodyParams, args.queryParams, options);
```

But the generated `http-client.ts` orders the client function's parameters **required-before-optional**. So when the **query param is required and the body is optional**, the client signature becomes **query-first**:

```ts
export const createItem = async (params: CreateItemParams, item?: Item, options?: RequestInit) => { ... }
```

The handler then passes `bodyParams` into the query slot and `queryParams` into the body slot. The generated code **fails to type-check** (`TS2345`), and if compilation is bypassed (`@ts-nocheck`, JS output, or a permissive `override.mutator`), the request goes out with **the body serialized into the URL query string and the query object sent as the JSON body**.

This is **not** specific to `override.mutator` — it reproduces with the plain `mcp` fetch client. It only surfaces for the *required-query + optional-body* shape: when the body is required and the query optional, the client signature is body-first and the call happens to line up, which is why it's easy to miss.

### Output client

`mcp`

```ts

export default defineConfig({
  repro: {
    output: {
      client: 'mcp',
      target: './out/handlers.ts',
      schemas: './out/model',
      baseUrl: '',
    },
  },
});
```

### Environment

```
  System:
    OS: Linux 6.12 Ubuntu 26.04 LTS 26.04 LTS (Resolute Raccoon)
    CPU: (16) arm64 unknown
    Memory: 21.30 GB / 23.43 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  npmPackages:
    orval: ^8.17.0 => 8.17.0
```

### Expected behavior

The handler should pass arguments in the same order the http-client declares its parameters, so the generated handler compiles and routes body/query to the correct slots regardless of which one is required.

### Actual behavior

Running `npx orval --config ./orval.config.ts` on the spec below produces:

`out/http-client.ts` — **query-first** (because `params` is required and `item` optional):

```ts
export const createItem = async (params: CreateItemParams, item?: Item, options?: RequestInit) => { ... }
```

`out/handlers.ts` — still **body-first**:

```ts
export type createItemArgs = {
  queryParams: CreateItemParams;
  bodyParams: Item;
};

export const createItemHandler = async (args: createItemArgs, options?: RequestInit) => {
  const res = await createItem(args.bodyParams, args.queryParams, options); // <-- swapped
  ...
};
```

`npx tsc --noEmit --strict --skipLibCheck out/handlers.ts out/http-client.ts out/model/*.ts` fails:

```
out/handlers.ts(26,32): error TS2345: Argument of type 'Item' is not assignable to parameter of type 'CreateItemParams'.
  Property 'dryRun' is missing in type 'Item' but required in type 'CreateItemParams'.
```

Validated with `npx @scalar/cli document validate ./openapi.yaml` (passes).

```yaml
openapi: 3.0.3
info:
  title: MCP arg-order repro
  version: 0.0.0
paths:
  /items:
    post:
      operationId: createItem
      summary: Create an item (REQUIRED query param + OPTIONAL body)
      parameters:
        - name: dryRun
          in: query
          required: true
          schema:
            type: boolean
      requestBody:
        required: false
        content:
          application/json:
            schema:
      responses:
        '200':
          description: ok
components:
  schemas:
    Item:
      type: object
      properties:
        name:
          type: string
      required:
        - name
```

### Additional context

- Reproduces on `8.15.0` through `8.17.0`; no MCP codegen changes have landed between those releases.
- **Root cause:** the MCP handler template emits the client call with a hardcoded `(bodyParams, queryParams, ...)` order, while the fetch http-client builds its parameter list required-before-optional. The two orderings diverge whenever required/optional differs between the query param and the body. The handler call should be generated from the same ordered parameter model used to emit the client signature, rather than assuming body-then-query.
- Workaround we currently use: a post-generation script that swaps `(args.bodyParams, args.queryParams,` -> `(args.queryParams, args.bodyParams,` at the affected call sites.

yagop:
## Description

Fixes #3599.

Two related correctness issues in the MCP generator's handler/tool wiring, both
exercised by one new sample operation (required query param + optional body).

### 1. Argument order (the reported bug)

The generated MCP handler called the fetch client with a hardcoded
body-then-query argument order:

    await createItem(args.bodyParams, args.queryParams, options);

But the generated client sorts its parameters required-before-optional
(`getProps` -> `sortByPriority`). When an operation has a **required query param
and an optional body**, the signature is query-first:

    export const createItem = (params: CreateItemParams, item?: Item, options?: RequestInit) => ...

so the handler passed body into the query slot and vice-versa: the generated
`handlers.ts` failed to type-check (`TS2345`), and if compiled past that, sent the
body as the URL query string and the query object as the JSON body. (Operations
with a required body happened to line up, which is why it went unnoticed.)

**Fix:** build the body/query arguments by iterating `verbOptions.props` — the
same already-sorted list the client signature is generated from — so the call
order can never diverge from the signature.

### 2. Optional request body typed as required (raised in review)

When the OpenAPI `requestBody` is optional, the handler still typed `bodyParams`
as required and the tool `inputSchema` required it, even though the client accepts
`pet?: Pet` — making the tool impossible to call without a body.

**Fix:** consult `body.isOptional` so the handler emits `bodyParams?` and the tool
`inputSchema` emits `Body.optional()`, keeping the MCP contract consistent with
the OpenAPI spec and the client signature.

## Tests

Adds a `filterPetsByStatus` operation (required query param + optional body) to
`samples/mcp/petstore` and its snapshots — it exercises both fixes at once
(`await filterPetsByStatus(args.queryParams, args.bodyParams, options)` with
`bodyParams?: Pet` and `FilterPetsByStatusBody.optional()`). Reverting the
ordering fix makes **only** `samples/mcp/petstore/src/handlers.ts` fail the
snapshot test, confirming it guards exactly that regression. Existing mcp sample
snapshots are unchanged and the full package test suite passes.

## Summary by CodeRabbit

* **New Features**
  * Added a new “filter pets by status” MCP tool/API endpoint.
  * Supports a required `status` query parameter plus an optional JSON request body.
  * Returns filtered pet results with success and error responses (including a default unexpected-error case).
  * Uses OAuth2 security with scopes that require both write and read permissions.

* **Bug Fixes**
  * Improved MCP handler argument ordering to avoid mismatches between query and body parameters when optionality differs.
  * Ensures optional request bodies are treated as optional in the generated schemas.

### DIFF
--- a/packages/mcp/src/index.ts
+++ b/packages/mcp/src/index.ts
@@ -12,6 +12,7 @@ import {
   getFileInfo,
   getFullRoute,
   getParamsInPath,
+  GetterPropType,
   isObject,
   isString,
   jsDoc,
@@ -174,8 +175,18 @@ ${handlerArgsTypes.join('\n')}
 
     fetchParams.push(pathParamsArgs);
   }
-  if (verbOptions.body.definition) fetchParams.push(`args.bodyParams`);
-  if (verbOptions.queryParams) fetchParams.push(`args.queryParams`);
+  // Body and query args must follow the same order as the generated client
+  // function signature, which sorts required params before optional ones (see
+  // `getProps`/`sortByPriority`). Emitting a fixed body-then-query order would
+  // swap the two arguments whenever the query param is required and the body is
+  // optional, sending the body as the query string and vice versa.
+  for (const prop of verbOptions.props) {
+    if (prop.type === GetterPropType.BODY) {
+      fetchParams.push('args.bodyParams');
+    } else if (prop.type === GetterPropType.QUERY_PARAM) {
+      fetchParams.push('args.queryParams');
+    }
+  }
 
   const handlerName = `${verbOptions.operationName}Handler`;
   const handlerImplementation = `
--- a/packages/mcp/src/index.ts
+++ b/packages/mcp/src/index.ts
@@ -154,7 +154,9 @@ export const generateMcp: ClientBuilder = (verbOptions) => {
     );
   }
   if (verbOptions.body.definition) {
-    handlerArgsTypes.push(`  bodyParams: ${verbOptions.body.definition};`);
+    handlerArgsTypes.push(
+      `  bodyParams${verbOptions.body.isOptional ? '?' : ''}: ${verbOptions.body.definition};`,
+    );
   }
 
   const handlerArgsName = `${verbOptions.operationName}Args`;
@@ -251,7 +253,9 @@ export const generateServer = (
       if (verbOption.queryParams)
         inputSchemaTypes.push(`queryParams: ${pascalOperationName}QueryParams`);
       if (verbOption.body.definition)
-        inputSchemaTypes.push(`bodyParams: ${pascalOperationName}Body`);
+        inputSchemaTypes.push(
+          `bodyParams: ${pascalOperationName}Body${verbOption.body.isOptional ? '.optional()' : ''}`,
+        );
 
       const inputSchemaImplementation =
         inputSchemaTypes.length > 0
--- a/samples/mcp/petstore/__snapshots__/handlers.ts
+++ b/samples/mcp/petstore/__snapshots__/handlers.ts
@@ -13,6 +13,8 @@
  * OpenAPI spec version: 1.0.27-SNAPSHOT
  */
 import {
+  FilterPetsByStatusParams,
+  Pet,
   FindPetsByStatusParams,
   FindPetsByTagsParams,
   UpdatePetWithFormParams,
@@ -20,6 +22,7 @@ import {
 } from './http-schemas';
 
 import {
+  filterPetsByStatus,
   findPetsByStatus,
   findPetsByTags,
   getPetById,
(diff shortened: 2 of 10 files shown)


yagop:
Thanks for the reviews. Addressed the automated feedback:

- **Optional body required in handler args + tool `inputSchema`** (Copilot on `server.ts`/`handlers.ts`, CodeRabbit on `handlers.ts`): fixed at the **generator** rather than hand-editing the generated files — `packages/mcp/src/index.ts` now consults `body.isOptional`, emitting `bodyParams?` in the handler args type and `…Body.optional()` in the tool `inputSchema`. Regenerated the sample, so `filterPetsByStatus` now shows `bodyParams?: Pet` and `FilterPetsByStatusBody.optional()`. I verified across the matrix (required/optional body × with/without query/path × no body) that the handler + inputSchema optionality now matches the generated client signature in every case, and that the argument **order** fix is not regressed.

- **`data` set to `{}` for empty bodies** (Copilot on `http-client.ts`): this is pre-existing `@orval/fetch` behaviour shared by every JSON-body operation (e.g. `getPetById`, `findPetsByStatus` in this same file), unrelated to the MCP change here — left out of scope.

- Note: optional **query** params have the same latent gap as the body did (still emitted as required in the handler type / inputSchema). It's harmless (a required-typed object is assignable to the client's optional param, and `tsc` passes) and pre-existing, so I kept it out of this PR to avoid churning every optional-query operation across the samples — happy to follow up separately if you'd like it fixed too.

melloware (approved):

PR #3609, opened 2026-06-15
feat(zod): coerce single-value array query params via opt-in coerce: ['array']

### LINKED ISSUE #3608: fix(zod): array query params reject a single value — explode arrays don't round-trip on the server
## Describe the bug

Array query params declared with `type: array` (+ `explode: true`, the OpenAPI default for query arrays) **serialize** correctly on the client side — orval emits repeated keys `?tags=a&tags=b` 👍 — but they **fail validation on the server** as soon as a single value is sent.

Server frameworks deliver a single repeated-key value as a **scalar**, not a 1-element array. Hono, for example:
```ts
// hono/validator, query target
v.length === 1 ? [k, v[0]] : [k, v]   // 1 value → "a" ; 2+ → ["a","b"]
```
So `?tags=a` arrives as `"a"`, and the generated `zod.array(...)` rejects it:
```
✖ Expected array, received string → tags
```
Multi-value works, single-value (the common "filter on one item" case) doesn't — so an explode array param never round-trips reliably.

## To Reproduce

A `type: array` query param, hono client:
```ts
// generated
export const ListPetsQueryParams = zod.object({
  tags: zod.array(zod.string()).optional(),
})
```
- `GET /pets?tags=a&tags=b` → `["a","b"]` ✅
- `GET /pets?tags=a` → ❌ `Expected array, received string`

## Expected behavior

A `type: array` query param accepts both a single value and multiple values (the OpenAPI `explode` contract), parsing to an array in both cases.

## Root cause

orval generates a bare `zod.array(...)` for array query params. At codegen time it knows the param is `in: query` + `type: array`, but emits nothing to absorb the framework's single→scalar collapse.

(Hono's collapse is schema-blind and defensible — it optimizes the common scalar case; zod is correctly strict and even provides the fix primitive, `z.preprocess`. orval is the only layer with both the context and the timing to bridge it.)

Related: #3231 covers the **client** half of the same gap (fetch serialization not defaulting `explode: true`). This issue is the **server** half — validating those exploded arrays. (Until #3231 lands, `explode: true` must be set explicitly for the client to emit repeated keys.)

## Proposed solution (opt-in)

Extend `override.zod.coerce.<location>` to accept an **`'array'`** target. When present, array params in that location are wrapped in a single→array `preprocess`:

```ts
// override: { zod: { coerce: { query: ['array'] } } }   (combines with ['number','array'])

tags: zod.preprocess(
  (value) => (value === undefined || Array.isArray(value) ? value : [value]),
  zod.array(zod.string()).optional(),
)
```

- `?tags=a` → `["a"]`, `?tags=a&tags=b` → `["a","b"]`, omitted → `undefined` (passes through, so an outer `.optional()`/`.default()` still applies), invalid item → still rejected.
- It is **not** a real `z.coerce.array()` — it's a structural single→array normalization, expressed through the existing `coerce` option to reuse its per-location plumbing and "coerce wire → schema type" mental model. Open to a dedicated option instead if you prefer — happy to adjust.
- **Fully opt-in**: with `'array'` absent, output is byte-identical.

## Why the existing tests wouldn't catch it

Array query params are exercised in fixtures only with the default (no coercion), where the bare `zod.array(...)` is the expected output — nothing asserts the single-value round-trip.

## Notes

I have a working, tested branch ready to open as a PR:
- new regression tests (wrap emitted when `coerce` includes `'array'`; untouched otherwise);
- `@orval/zod` unit suite green (244 → 246);
- **all generation snapshots unchanged** (verified — opt-in, zero churn);
- typecheck + lint clean;
- validated end-to-end against a real Hono + zod backend: `?tags=a` and `?tags=a&tags=b` both parse.

Same area as #3231 (client-side explode), #3511/#3512 (fixed in #3515/#3516), and the import fix #3532 shipped in 8.17.

NathanFichefet:
## Summary

Fixes #3608.

Array query params (`type: array`, `explode: true`) serialize fine client-side but **fail server-side validation on a single value**: a server framework (e.g. Hono) delivers one repeated-key value as a scalar (`?tags=a` → `"a"`), which the generated `zod.array(...)` rejects.

## Change

Opt-in: `override.zod.coerce.<location>` accepts a new **`'array'`** target. When set, array params in that location are wrapped in a single→array `preprocess`:
```ts
// coerce.query: ['array']  (combines with ['number','array'])
tags: zod.preprocess(
  (value) => (value === undefined || Array.isArray(value) ? value : [value]),
  zod.array(zod.string()).optional(),
)
```
`?tags=a` → `["a"]`, `?tags=a&tags=b` → `["a","b"]`, omitted → `undefined` (so an outer `.optional()`/`.default()` still applies), invalid item → still rejected.

It's not a real `z.coerce.array()` — it's a structural single→array normalization, routed through `coerce` to reuse its per-location plumbing. The leaf `array` branch returns before the `.coerce.<fn>` path, so reading `'array'` from `coerce` doesn't affect primitive coercion. Happy to switch to a dedicated option if preferred.

## Tests

- Two regression tests in `packages/zod/src/zod.test.ts`: wrap emitted when `coerce.query` includes `'array'`; untouched when it doesn't (opt-in).
- `@orval/zod` unit suite green (246).
- All generation snapshots unchanged (opt-in → zero churn).
- Validated end-to-end against a Hono + zod backend (`?tags=a` and `?tags=a&tags=b` both parse).

Related: #3231 (client-side explode default).

## Summary by CodeRabbit

* **New Features**
  * Enhanced array parameter coercion: scalar values are now automatically converted to single-element arrays when array coercion is enabled for query parameters.

* **Tests**
  * Added test coverage for array query parameter validation behavior.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -833,7 +833,16 @@ export interface EffectOptions {
   useBrandedTypes?: boolean;
 }
 
-export type ZodCoerceType = 'string' | 'number' | 'boolean' | 'bigint' | 'date';
+export type ZodCoerceType =
+  | 'string'
+  | 'number'
+  | 'boolean'
+  | 'bigint'
+  | 'date'
+  // Not a real `z.coerce.array()` — opting 'array' into `coerce.<location>`
+  // wraps array params in a single→array preprocess so a single repeated-key
+  // query value (delivered as a scalar by the server framework) still parses.
+  | 'array';
 
 export interface NormalizedZodOptions {
   strict: {
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -1405,7 +1405,19 @@ ${Object.entries(objectArgs)
       .map((prop) => parseProperty(prop, [...fieldPath, key]))
       .join('');
     appendConstsChunk(schema.consts.join('\n'));
-    return `  "${key}": ${value.startsWith('.') ? 'zod' : ''}${value}`;
+    const fieldZod = `${value.startsWith('.') ? 'zod' : ''}${value}`;
+    // Opt-in via `coerce.<location>: ['array', ...]`: a server framework (e.g.
+    // Hono) delivers a single repeated-key value as a scalar, not a 1-element
+    // array, so `?t=a` would fail z.array(...). Wrap so both `?t=a` and
+    // `?t=a&t=b` parse. Undefined passes through so an outer `.optional()`/
+    // `.default()` still applies. Already-arrays are left untouched (no-op for
+    // JSON-body arrays).
+    const coerceArrays =
+      Array.isArray(coerceTypes) && coerceTypes.includes('array');
+    if (coerceArrays && schema.functions.some(([fn]) => fn === 'array')) {
+      return `  "${key}": zod.preprocess((value) => value === undefined || Array.isArray(value) ? value : [value], ${fieldZod})`;
+    }
+    return `  "${key}": ${fieldZod}`;
   })
   .join(',\n')}
 })`;
--- a/packages/zod/src/zod.test.ts
+++ b/packages/zod/src/zod.test.ts
@@ -4620,6 +4620,51 @@ const bodyOnlyApiSchema = {
     },
   },
 } as unknown as GeneratorOptions;
+
+// An operation with an array query param, to assert the single→array coercion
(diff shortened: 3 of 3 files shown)


melloware:
@Hypenate you have been involved in this and want your thoughts.

Hypenate:
Makes sense yes.

melloware (approved):

PR #3617, opened 2026-06-17
fix(mock): handle nullable object schemas and per-import strict kinds

Hypenate:
## Summary
- Fix strict faker mocks for schemas that are nullable objects (`type: object` + `nullable: true`, or `type: ['object', 'null']`): when `nonNullable` is enabled, factories no longer emit top-level `null` branches that conflict with `{Schema}Mock` return types.
- Classify strict mock schema kinds per import by resolving `$ref` targets and matching `oneOf`/`anyOf` branches, instead of applying the response wrapper schema to every imported type.
- Resolve `$ref` targets in `classifyStrictMockSchemaType` so enum refs emit alias mocks instead of object mapped types.

## Test plan
- [x] `packages/mock/src/mock-types.test.ts`
- [x] `packages/mock/src/faker/getters/object.test.ts`
- [x] `packages/mock/src/faker/index.test.ts`
- [x] `bun run lint`

EOF

## Summary by CodeRabbit

* **New Features**
  * Enhanced strict mock type generation to detect nullable-at-root object schemas and emit nullable type declarations, including context-aware handling for referenced and composed schemas.

* **Bug Fixes**
  * Improved OpenAPI 3.1 nullable object mocking: root null wrapping is applied consistently (including additional properties and empty object cases), `nonNullable` overrides are respected, and `nullWrapped` behavior is propagated correctly.

* **Tests**
  * Added and expanded Vitest coverage for nullable-at-root behavior, context-aware `$ref` resolution, and `oneOf` response classification, plus strict-mock expectations for `type: ['object', 'null']`.


Hypenate:
@wadakatu this follow up PR for #3607 with issue #3590 fixes your remarks.

melloware (approved):

PR #3636, opened 2026-06-22
feat(fetch): add paramsSerializer and arrayFormat support for query params

### LINKED ISSUE #3635: Allow different handling of array query parameters when using fetch
### Description

We need more control over the handling of array-based query parameters when using fetch, so that it works better with different API backends.

### Current behavior

As described in https://github.com/orval-labs/orval/discussions/2368, the current fetch implementation doesn't work well together with different approaches to handling query parameters with array values. The current implementation defaults to a comma-separated resolution of these values. However, many APIs prefer a repetition of the key for each value. 

### Proposed behavior

Provide a way to configure the handling of array-based query params.

### Proposed solution

I've already implemented the required changes in two ways:
1. an `overrides.fetch.arrayFormat` configuration parameter, which allows the user to choose between "repeat", "brackets" or "comma"
2. an `overrides.paramsSerializer`, which works in the same way as the axios and angular configuration.

Option 1 would be for basic use, option 2 would be the power user option.

ronderksen:
Adds two new override options for the fetch client:
- `override.paramsSerializer`: a custom mutator that receives the full params object and returns a serialized query string, replacing the built-in URLSearchParams logic entirely (power-user escape hatch).
- `override.fetch.arrayFormat`: controls how array query params are serialized when the OpenAPI spec does not set `explode` explicitly. Supports 'repeat' (key=a&key=b), 'brackets' (key[]=a&key[]=b), and 'comma' (key=a,b). The spec's own `explode` property always takes precedence.
- Updated docs to explain these new options.

Closes https://github.com/orval-labs/orval/issues/3635

## Summary by CodeRabbit

* **New Features**
  * Added `override.fetch.arrayFormat` (`repeat`, `brackets`, `comma`) to control array query serialization in the fetch client.
  * Enhanced fetch URL generation to support `override.paramsSerializer` for fully custom query-string handling.
* **Documentation**
  * Documented `override.fetch.arrayFormat` with examples and clarified that the OpenAPI spec’s `explode` setting takes precedence.
  * Clarified `override.paramsSerializer` support across Axios, Angular, and fetch (fetch expects a raw query string).
* **Tests**
  * Added coverage for all array-format variants and `paramsSerializer` delegation behavior.


melloware:
This seems like a great improvement to align fetch with the other clients

melloware (changes requested):
Build failing

ronderksen:
I had some trouble figuring out what issues were actually reported here on Github, but I think I now fixed all of them.

melloware (approved):

PR #3640, opened 2026-06-23
fix(angular): emit filterParams helper for acronym tags in tags-split

### LINKED ISSUE #3634: filterParams helper is missing for some tags (tags-split + Angular)
### Description

When generating an Angular service in tags-split mode, the generated service for certain tags calls the `filterParams(...)` helper but never defines or imports it, resulting in `TS2304: Cannot find name 'filterParams')`.

| Endpoint | Tag | Generated file | Result |
| --- | --- | --- | --- |
| `GET /widgets` | `Widget` | `widget/widget.service.ts` | defines `filterParams` |
| `GET /ab-widgets` | `AB Widget` | `ab-widget/ab-widget.service.ts` | calls `filterParams`, never defines it |

### Output client

angular

```ts
module.exports = {
  repro: {
    output: {
      mode: 'tags-split',
      client: 'angular',
      target: './src/generated/services',
      schemas: './src/generated/model',
    },
  },
};
```

### Environment

Node 24.11.0
Orval 8.18.0
Using Ubuntu

### Expected behavior

Every generated service that calls `filterParams` should also define (or import) it, regardless of how the tag is spelled.

### Actual behavior

`TS2304: Cannot find name 'filterParams')` errors for some generated files.

<details>
<summary>OpenAPI spec</summary>

```yaml
openapi: 3.0.0
info:
  title: Orval tags-split filterParams repro
  version: 1.0.0
paths:
  # Control: a single-word tag. Generates correctly.
  /widgets:
    get:
      tags: [Widget]
      operationId: listWidgets
      parameters:
        - name: search
          in: query
          schema:
            type: string
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: array
                items:
                  type: object

  # Trigger: a tag whose first word is an acronym ("AB") followed by a space.
  # Generates a broken service (calls filterParams but never defines it).
  /ab-widgets:
    get:
      tags: [AB Widget]
      operationId: listAbWidgets
      parameters:
        - name: search
          in: query
          schema:
            type: string
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: array
                items:
                  type: object

```

</details>

### Additional context

Minimal reproduction repo (generated with Claude Code): https://github.com/jeroentvb/orval-filterparams-repro

wadakatu:
## Summary

Fixes a `TS2304: Cannot find name 'filterParams'` error in Angular `tags-split` output for tags whose first word is an acronym (e.g. `AB Widget`). The generated service calls the shared `filterParams` helper but never defines or imports it. Single-word tags (`Widget`) and ordinary multi-word tags (`User Account`) were unaffected.

## Root cause

In `tags-split` mode each per-tag output file is keyed by `kebab(operation.tags[0])` (`packages/core/src/writers/target-tags.ts`), and that kebab-cased string is passed to the Angular header builder as `tag`. But `getRelevantVerbOptionsForTag` — which decides whether the shared `filterParams` helper is emitted — matched tags with `camel`:

- `camel('ab-widget')` → `'abWidget'` (the incoming kebab-cased tag)
- `camel('AB Widget')` → `'aBWidget'` (the operation's original tag — the acronym keeps its second capital)

`'abWidget' !== 'aBWidget'`, so every operation was filtered out, `hasBuiltInFilteredQueryParams` became `false`, and the helper was suppressed — while the operation body still called `filterParams`. `camel` is not idempotent through a `kebab` round-trip for acronym tags, which is why only those tags broke.

## Fix

Match on `kebab` instead of `camel` in `getRelevantVerbOptionsForTag` so the comparison mirrors the writer's grouping key (`kebab(tags[0])`). `kebab` is idempotent (`kebab('ab-widget') === 'ab-widget'`) and `DefaultTag` is unaffected (`kebab('default') === 'default'`). This is a single shared function, so the fix covers the `httpClient`, `httpResource`, and `both` Angular modes at once.

## Tests

- New regression spec `tests/specifications/issue-3634.yaml` with a single-word control tag (`Widget`) and an acronym trigger tag (`AB Widget`), wired through a new `tags-split` Angular config.
- Focused assertion in `tests/api-generation.spec.ts` that the acronym-tag service defines `filterParams` (fails before the fix, passes after).
- Generated output for all 16 clients type-checks (`orval-tests build`); full snapshot suite passes with no churn on existing fixtures.

Closes #3634

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tag-scoped output handling in Angular code generation to correctly process multi-word tags with acronym prefixes.

* **Tests**
  * Added regression test to verify proper code generation behavior for tag-scoped operations with acronym-based tags.


the-ult:
Heads up @wadakatu. I was working in it as well — we independently converged on the same fix… 😇

Your PR cleanly resolves #3634. I opened #3641 which builds on the same root cause and goes further: generalizes the rule into a single `@orval/core` source of truth (it's hand-copied across ~6 sites in 4 packages) and fixes the same bug in `@orval/hono` (raw AB Widget/ dir vs canonical import + empty handler block). Happy to rebase on top of yours once it lands and drop the duplicated Angular bits.

- https://github.com/orval-labs/orval/pull/3641

cc: @melloware

wadakatu:
@the-ult 

Thank you for your PR!
I think that your PR is much cleaner than mine, so I will leave it up to you.

PR #3645, opened 2026-06-23
feat(mock): add override.mock.schemas for per-schema property overrides

jakiestfu:
## Summary

`override.mock.properties` matches by property name, so an override for `color` applies to **every** schema that has a `color` property. There was no way to give the same property name different mock values depending on which schema it belongs to.

This adds **`override.mock.schemas`**, keyed by schema name, so the same property name can mock differently per schema:

```ts
override: {
  mock: {
    schemas: {
      Apple: { properties: { color: () => faker.helpers.arrayElement(['red', 'green']) } },
      Car:   { properties: { color: () => 'midnight black' } },
    },
  },
}
```

`getAppleMock()` now mocks `color` as a fruit color and `getCarMock()` as a car color, even though both declare `color: string`.

The keys under `properties` use the same matching rules as `override.mock.properties` — bare name, `/regex/`, or exact `#.path`.

**Precedence** (first match wins): `override.operations` → `override.tags` → `override.mock.schemas` → `override.mock.properties`.

## Generated output

Given two schemas that both have `color: string`:

```ts
// Apple-scoped override applied
export const getAppleMock = (overrideResponse: Partial<Apple> = {}): Apple => ({
  color: faker.helpers.arrayElement(['red', 'green']),
  ...overrideResponse,
});

// Car-scoped override applied
export const getCarMock = (overrideResponse: Partial<Car> = {}): Car => ({
  color: 'midnight black',
  ...overrideResponse,
});
```

With `schemas: true`, a referencing schema keeps delegating to the factory and the override rides along:

```ts
export const getBasketMock = (overrideResponse: Partial<Basket> = {}): Basket => ({
  apple: { ...getAppleMock() }, // getAppleMock() already bakes in the Apple-scoped override
  ...overrideResponse,
});
```

## Changes

- **`packages/core`** — add `schemas?: Record<string, { properties }>` to `OverrideMockOptions` (user-facing) and `MockOptions` (serialized form).
- **`packages/mock` (msw/mocks.ts)** — `getMockWithoutFunc` serializes function-valued schema overrides to IIFE strings, mirroring the existing `operations`/`tags` handling.
- **`packages/mock` (faker/getters/scalar.ts)** — `getMockScalar` resolves a schema-scoped tier keyed on the property's enclosing schema (`item.parentName`), between the tag and global-property tiers.
- **Docs** — new "Per-schema overrides" section in the Faker guide + an `override.mock.schemas` reference entry.

## Notes

- Schema-scoped overrides apply to a schema's **own** properties (matched by immediate `parentName`). No change to `hasOverrideTouchingSchema` was needed for `schemas: true`: each `get<Schema>Mock` factory is built with the same mock options, so delegation preserves the override (verified by test).
- Works with both the `faker` and `msw` generators.

## Tests

- Added unit tests in `scalar.test.ts` (override applied per schema; falls through when the schema name doesn't match; precedence vs operation/global).
- Added end-to-end tests in `index.test.ts` (different override per schema; preserved through factory delegation).
- Added serialization tests in `mocks.test.ts` (function → IIFE, value stringification, omitted when unset).
- Full suite: mock 293/293, core 2084/2084. Both packages typecheck clean.

## Summary by CodeRabbit

* **New Features**
  * Added support for schema-scoped mock property overrides, allowing users to customize mock values for specific OpenAPI schemas with clearly defined precedence rules.

* **Documentation**
  * Added "Per-schema overrides" guide explaining scoping rules and behavior.
  * Expanded configuration reference with examples and override precedence order.

### DIFF
--- a/docs/content/docs/guides/faker.mdx
+++ b/docs/content/docs/guides/faker.mdx
@@ -210,6 +210,44 @@ override: {
 
 You can also scope overrides per-operation or per-tag via `override.operations` and `override.tags`.
 
+### Per-schema overrides
+
+`override.mock.properties` matches by property name, so a `color` override applies to
+**every** schema that has a `color` property. To give the same property name different
+mock values depending on the schema it belongs to, scope the override to a schema by name
+under `override.mock.schemas`:
+
+```ts title="orval.config.ts"
+override: {
+  mock: {
+    schemas: {
+      Apple: {
+        properties: {
+          color: () => faker.helpers.arrayElement(['red', 'green']),
+        },
+      },
+      Car: {
+        properties: {
+          color: () => 'midnight black',
+        },
+      },
+    },
+  },
+}
+```
+
+`getAppleMock()` now mocks `color` as a fruit color and `getCarMock()` as a car color,
+even though both schemas declare `color: string`.
+
+The keys under `properties` use the same matching rules as `override.mock.properties` —
+bare property name, regex (`/.../`), or exact path (`#.foo.bar`) — but only apply to
+properties of the named schema (matched against the schema's own `parentName`). When
+`schemas: true` is enabled, each `get<SchemaName>Mock` factory bakes its schema-scoped
+override in, so references that delegate to the factory keep the override.
+
+**Precedence** (first match wins): `override.operations` → `override.tags` →
+`override.mock.schemas` → `override.mock.properties`.
+
 ## Usage
 
 ### Unit Tests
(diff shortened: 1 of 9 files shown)


jakiestfu:
Thanks for taking a peek @melloware. Do you have any concern or changes you'd like to request?

P.S. We've really been loving the faker mocks <-> Orval at Turo!

melloware:
@jakiestfu this seems like a slick improvement to me!  @wadakatu and @Hypenate any thoughts?

Hypenate:
Nice addition! Thanks

melloware (approved):

PR #3676, opened 2026-07-01
fix(core): fix issue when barrel file was re-exporting itself (#3675)

### LINKED ISSUE #3675: since v8.18.0 re-exports index file if `target` is `index.ts`
### Description

8.17.0 is fine, but in 8.18.0 `export * from './index.ts';`  get's added at the end of `index.ts` leading to circular dependency import (`index.ts` imports itself).

Same issue in 8.19.0.

### Output client

other

```ts
import { join } from 'path';

import builder from './src/API/orval/builder';

const SPEC_FILE_PATH = join(__dirname, 'openapi.json');

export default defineConfig({
  'backend': {
    input: {
      target: SPEC_FILE_PATH,
    },
    output: {
      clean: true,
      mode: 'split',
      workspace: join(__dirname, 'src', 'API', 'generated'),
      target: 'index.ts',
      client: () => builder,
      formatter: 'prettier',
      override: {
        aliasCombinedTypes: true,
        // Axios since v0.27.0 can convert our object to FormData
        // automatically, so we don't need to do it manually.
        // Also enabling it causes TS errors.
        formData: false,
        useNamedParameters: true,
        mutator: {
          path: join(__dirname, 'src', 'API', 'orval', 'custom-instance.ts'),
          name: 'customInstance',
        },
      },
    },
  },
});
```

Also, we use our own builder based on Axios, but the only thing it changes is the signature of the functions from parameters to one object.

```ts
// It's mainly a copy of https://github.com/orval-labs/orval/blob/master/packages/axios/src/index.ts
// (@orval/axios v8) with some modifications to fit our needs.
//
// The main difference is that instead of having function declaration like:
// const getApiEndpointMethod = (bodyTypeInCamelCase: BodyTypeInCamelCase, params: EndpointParams) => {
// it generates:
// const getApiEndpointMethod = (params: { body: BodyTypeInCamelCase, params: EndpointParams }) => {
import {
  type ClientBuilder,
  type ClientDependenciesBuilder,
  type ClientFooterBuilder,
  type ClientGeneratorsBuilder,
  type ClientHeaderBuilder,
  type ClientTitleBuilder,
  generateFormDataAndUrlEncodedFunction,
  generateMutatorConfig,
  generateMutatorRequestOptions,
  generateOptions,
  generateVerbImports,
  type GeneratorDependency,
  type GeneratorOptions,
  type GeneratorVerbOptions,
  type GetterBody,
  type GetterProp,
  isSyntheticDefaultImportsAllow,
  pascal,
  sanitize,
} from '@orval/core';

const replacePropName = (prop: GetterProp, replaceWith: string): GetterProp => {
  const prevName = prop.name;
  const nameRegex = new RegExp(prevName, 'g');

  return {
    ...prop,
    definition: prop.definition.replace(nameRegex, replaceWith),
    implementation: prop.implementation.replace(nameRegex, replaceWith),
    name: replaceWith,
  };
};

const replaceBodyName = (body: GetterBody, replaceWith: string): GetterBody => {
  if (!body.definition) {
    return body;
  }

  return {
    ...body,
    implementation: body.implementation.replace(/\w+/g, replaceWith),
  };
};

const AXIOS_DEPENDENCIES: GeneratorDependency[] = [
  {
    exports: [
      {
        name: 'axios',
        default: true,
        values: true,
        syntheticDefaultImport: true,
      },
      { name: 'AxiosRequestConfig' },
      { name: 'AxiosResponse' },
    ],
    dependency: 'axios',
  },
];

const PARAMS_SERIALIZER_DEPENDENCIES: GeneratorDependency[] = [
  {
    exports: [
      {
        name: 'qs',
        default: true,
        values: true,
        syntheticDefaultImport: true,
      },
    ],
    dependency: 'qs',
  },
];

const returnTypesToWrite = new Map<string, (title?: string) => string>();

export const getAxiosDependencies: ClientDependenciesBuilder = (
  hasGlobalMutator,
  hasParamsSerializerOptions: boolean
) => [
  ...(hasGlobalMutator ? [] : AXIOS_DEPENDENCIES),
  ...(hasParamsSerializerOptions ? PARAMS_SERIALIZER_DEPENDENCIES : []),
];

const getFunctionPropsImplementation = (props: GetterProp[]) => {
  const typeDefinitionLines = props.map((prop) => prop.definition);
  const atLeastOnePropRequired = props.some((prop) => prop.required);

  const type = `{\n${typeDefinitionLines}\n}`;

  const semicolon = atLeastOnePropRequired ? ':' : '?:';
  return `props${semicolon} ${type},`;
};

const getDestructedProps = (props: GetterProp[]) => {
  const atLeastOnePropRequired = props.some((prop) => prop.required);
  const accessPropertyDot = atLeastOnePropRequired ? '' : '?.';

  const destructuredPropsLines = props.map((prop) => {
    const propDeclaration = prop.implementation.slice(0, prop.implementation.search(/(\?:|:)/));
    const propDefinitionName = prop.definition.slice(0, prop.definition.search(/(\?:|:)/));

    return `const ${propDeclaration} = props${accessPropertyDot}["${propDefinitionName}"];`;
  });

  return destructuredPropsLines.join('\n');
};

const generateAxiosImplementation = (
  {
    headers,
    queryParams,
    operationName,
    response,
    mutator,
    body,
    props,
    verb,
    override,
    formData,
    formUrlEncoded,
    paramsSerializer,
  }: GeneratorVerbOptions,
  { route, context }: GeneratorOptions
) => {
  const isRequestOptions = override.requestOptions !== false;
  const isFormData = !override.formData.disabled;
  const isFormUrlEncoded = override.formUrlEncoded !== false;
  const isExactOptionalPropertyTypes =
    !!context.output.tsconfig?.compilerOptions?.exactOptionalPropertyTypes;

  const isSyntheticDefaultImportsAllowed = isSyntheticDefaultImportsAllow(context.output.tsconfig);

  const bodyForm = generateFormDataAndUrlEncodedFunction({
    formData,
    formUrlEncoded,
    body,
    isFormData,
    isFormUrlEncoded,
  });

  const functionPropsImplementation = getFunctionPropsImplementation(props);
  const destructedProps = getDestructedProps(props);

  if (mutator) {
    const mutatorConfig = generateMutatorConfig({
      route,
      body,
      headers,
      queryParams,
      response,
      verb,
      isFormData,
      isFormUrlEncoded,
      hasSignal: false,
      isExactOptionalPropertyTypes,
    });

    const requestOptions = isRequestOptions
      ? generateMutatorRequestOptions(override.requestOptions, mutator.hasSecondArg)
      : '';

    returnTypesToWrite.set(
      operationName,
      (title?: string) =>
        `export type ${pascal(operationName)}Result = NonNullable<Awaited<ReturnType<${
          title ? `ReturnType<typeof ${title}>['${operationName}']` : `typeof ${operationName}`
        }>>>`
    );

    const propsImplementation =
      mutator.bodyTypeName && body.definition
        ? functionPropsImplementation.replace(
            new RegExp(String.raw`(\w*):\s?${body.definition}`),
            `$1: ${mutator.bodyTypeName}<${body.definition}>`
          )
        : functionPropsImplementation;

    return `const ${operationName} = (\n    ${propsImplementation}\n ${
      isRequestOptions && mutator.hasSecondArg
        ? `options${context.output.optionsParamRequired ? '' : '?'}: SecondParameter<typeof ${mutator.name}<${response.definition.success || 'unknown'}>>,`
        : ''
    }) => {
      ${destructedProps}
      ${bodyForm}
      return ${mutator.name}<${response.definition.success || 'unknown'}>(
      ${mutatorConfig},
      ${requestOptions});
    }
  `;
  }

  // You can ignore this block of code
  // because we use `mutator` option in our config.
  // If we don't use it - this block will be used to generate the API client.
  // I left it here just for compatibility.
  const options = generateOptions({
    route,
    body,
    headers,
    queryParams,
    response,
    verb,
    requestOptions: override.requestOptions,
    isFormData,
    isFormUrlEncoded,
    paramsSerializer,
    paramsSerializerOptions: override.paramsSerializerOptions,
    isExactOptionalPropertyTypes,
    hasSignal: false,
  });

  returnTypesToWrite.set(
    operationName,
    () =>
      `export type ${pascal(operationName)}Result = AxiosResponse<${
        response.definition.success || 'unknown'
      }>`
  );

  return `const ${operationName} = <TData = AxiosResponse<${
    response.definition.success || 'unknown'
  }>>(\n    ${functionPropsImplementation} ${
    isRequestOptions ? `options?: AxiosRequestConfig\n` : ''
  } ): Promise<TData> => {
    ${destructedProps}
    ${bodyForm}
    return axios${isSyntheticDefaultImportsAllowed ? '' : '.default'}.${verb}(${options});
  }
`;
};

export const generateAxiosTitle: ClientTitleBuilder = (title) => {
  const sanTitle = sanitize(title);
  return `get${pascal(sanTitle)}`;
};

export const generateAxiosHeader: ClientHeaderBuilder = ({
  title,
  isRequestOptions,
  isMutator,
  noFunction,
}) => `
${
  isRequestOptions && isMutator
    ? `type SecondParameter<T extends (...args: never) => unknown> = Parameters<T>[1];\n\n`
    : ''
}
  ${noFunction ? '' : `export const ${title} = () => {\n`}`;

export const generateAxiosFooter: ClientFooterBuilder = ({
  operationNames,
  title,
  noFunction,
  hasMutator,
  hasAwaitedType,
}) => {
  let footer = '';

  if (!noFunction) {
    footer += `return {${operationNames.join(',')}}};\n`;
  }

  if (hasMutator && !hasAwaitedType) {
    footer += `\ntype AwaitedInput<T> = PromiseLike<T> | T;\n
    type Awaited<O> = O extends AwaitedInput<infer T> ? T : never;
\n`;
  }

  for (const operationName of operationNames) {
    if (returnTypesToWrite.has(operationName)) {
      const func = returnTypesToWrite.get(operationName)!;
      footer += `${func(noFunction ? undefined : title)}\n`;
    }
  }

  return footer;
};

// the only purpose of this function is to change the name of the body argument to 'body'
const patchVerbOptions = (verbOptions: GeneratorVerbOptions): GeneratorVerbOptions => {
  // this helper changes the name of the body argument to 'body'
  const propsWithCorrectBodyArgumentName = verbOptions.props.map((prop) =>
    prop.type === 'body' ? replacePropName(prop, 'body') : prop
  );

  // this helper changes the name of the body argument to 'body' in the generated function itself
  const bodyWithCorrectName = replaceBodyName(verbOptions.body, 'body');

  return {
    ...verbOptions,
    props: propsWithCorrectBodyArgumentName,
    body: bodyWithCorrectName,
  };
};

export const generateAxios = (verbOptions: GeneratorVerbOptions, options: GeneratorOptions) => {
  const imports = generateVerbImports(verbOptions);
  const implementation = generateAxiosImplementation(patchVerbOptions(verbOptions), options);

  return { implementation, imports };
};

export const generateAxiosFunctions: ClientBuilder = (verbOptions, options) => {
  const { implementation, imports } = generateAxios(verbOptions, options);

  return {
    implementation: `export ${implementation}`,
    imports,
  };
};

const axiosFunctionsClientBuilder: ClientGeneratorsBuilder = {
  client: generateAxiosFunctions,
  header: (options) => generateAxiosHeader({ ...options, noFunction: true }),
  dependencies: getAxiosDependencies,
  footer: (options) => generateAxiosFooter({ ...options, noFunction: true }),
  title: generateAxiosTitle,
};

export default axiosFunctionsClientBuilder;
```

### Environment

  System:
    OS: Linux 7.0 Arch Linux
    CPU: (16) x64 AMD Ryzen 7 9800X3D 8-Core Processor
    Memory: 39.60 GB / 62.40 GB
    Container: Yes
    Shell: 5.3.15 - /usr/bin/bash
  npmPackages:
    @tanstack/react-query: ^5.101.0 => 5.101.0
    axios: ^1.18.0 => 1.18.0
    orval: 8.18.0 => 8.18.0
    react: ^19.2.7 => 19.2.7
    zod: ^4.4.3 => 4.4.3

### Expected behavior

`index.ts` should be generated as it is in 8.17.0

### Actual behavior

`index.ts` contains re-export of itself.

```yaml
openapi: 3.0.3
info:
  title: Demo
  version: 0.0.0
paths: {}
```

### Additional context

Normally I would just set `indexFiles` to `false` and it would fix the problem, but then #2912 pops up. And basically, the only workaround for now is to use different name for the `target` and update imports across the project.

Opening the bug, because found it strange that issue didn't exist before 😄

PowerSlime:
Tried to fix the issue by adding additional filter if path resolves in the same path as index file. Also added couple of tests to check that behavior in future.

Hopefully it's helpful.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved generated workspace barrel files to avoid self-referencing re-exports when the output target is `index.ts`.
  * Refined the logic that appends missing re-exports to prevent incorrect or incomplete barrel contents.

* **Tests**
  * Added Axios regression coverage for both `index.ts` and non-`index.ts` workspace targets.
  * Added new test configurations and an OpenAPI fixture to support the updated barrel-generation scenarios.
  * Expanded unit tests to cover undeclared module specifier detection and workspace `index.ts` creation behavior.


PowerSlime:
Tests generated by AI are quite bad... Give me slightly more time, will get my hands dirty a little 🫠

packages/orval/src/write-specs.ts
```diff
@@ -98,6 +98,73 @@ export async function runFormatter(
...
+    filePath,
+    comparablePath: getComparableFilePath(filePath),
+  }));
+
+  return comparableFilePaths
+    .filter(({ comparablePath }) => comparablePath !== comparablePathToExclude)
+    .map(({ filePath }) => filePath);
+}
+
+const moduleSpecifierPatterns = [
```
aqeelat: This preserves the desired output, but through the wrong mechanism. In the `target: index.ts` case, the operation file already has `import type { Pet } from './index.schemas'`; this parser sees that import and suppresses the schemas re-export. The output matches the intended 8.17 behavior, but only because operation imports are being treated as already-declared barrel exports.

I think this should be explicit instead: delete the module-specifier parser, keep a targeted duplicate-export check at the append site (`!data.includes(`export * from '${imp}'`)`), and exclude the co-located target's own generated sibling schemas file from `implementationPaths` before building the barrel imports.

That keeps the same snapshots, but for the right reason: the co-located barrel excludes itself and its own sibling schemas file by path, while non-index-target barrels still re-export both implementation and schemas.

packages/orval/src/write-specs.ts
```diff
@@ -98,6 +98,73 @@ export async function runFormatter(
...
+  try {
+    comparablePath = fs.realpathSync(resolvedPath);
+  } catch (error) {
+    // The workspace index may not exist yet on a first generation run.
+    if ((error as NodeJS.ErrnoException).code !== 'ENOENT') {
+      throw error;
+    }
+  }
+
+  // paths in Linux can be named the same but have different casing,
```
aqeelat: This case handling is correct. Please add one short note that this is a platform heuristic: it intentionally does not lowercase on Linux because Linux paths are usually case-sensitive, even though a case-insensitive filesystem could theoretically be mounted there. That limitation is fine; documenting it avoids future confusion.

Also, the ENOENT fallback above is good. It would help to mention which not-yet-created path it protects, since `realpathSync` otherwise looks surprising in generation code.

packages/orval/src/write-specs.test.ts
```diff
@@ -83,3 +90,66 @@ describe('runFormatter', () => {
...
+    expect(
+      getUndeclaredModuleSpecifiers(
+        ['./index.schemas', './endpoints', './endpoints.msw'],
+        data,
+      ),
+    ).toEqual(['./endpoints.msw']);
+  });
+});
+
+describe('writeSpecs workspace index', () => {
```
aqeelat: Once the regex helper is removed, I think this file should go back to only testing `runFormatter`. The new `generateSpec` + `mkdtemp` case is an integration test inside a unit-test file that mocks `@orval/core`, and the same behavior is already covered by the new `tests/api-generation.spec.ts` regression tests. Reverting this file should be possible after removing `getUndeclaredModuleSpecifiers`.

aqeelat (changes requested):
Thanks for the quick turnaround. The structural fix (`excludeFilePath` + `getComparableFilePath`) is correct and well-reasoned, and the OS-gated lowercasing plus ENOENT fallback are both the right shape.

A few things to adjust before merge:

1. Drop the regex-based dedup. It currently preserves the desired output, but indirectly: it suppresses `./index.schemas` because it sees the operation file's own `import type { Pet } from './index.schemas'`. That is not really a barrel dedup rule. The explicit rule should be: when the workspace barrel is co-located with `output.target` (`target: index.ts`), exclude the barrel file itself and its own generated sibling schemas file from the barrel import list.
2. Add a short comment documenting the platform-based case-sensitivity heuristic in `getComparableFilePath`.
3. Revert the `write-specs.test.ts` additions once the regex helper is removed. The integration coverage already exists in `tests/api-generation.spec.ts`, and the helper-specific tests go away with the helper.

On your two questions from #3675:

1. Your Linux case-sensitivity read is correct. Do not lowercase on Linux; `Index.ts` and `index.ts` can be different files there.
2. The ENOENT fallback is also the right shape because `realpathSync` requires the path to exist. Please add a short comment explaining which not-yet-created path this fallback protects so future changes do not accidentally remove it.

<details>
<summary>Agent prompt</summary>

```text
Repo: orval-labs/orval, PR #3676, fixing issue #3675.

Context:
- Bug #3675: with output.target = "index.ts", output.workspace set, and mode = "split",
  the workspace barrel appended a self-reexport into index.ts:
  `export * from './index'` or `export * from './index.ts'`.
- #3558 established `index.ts` at the target root as an intentional barrel pattern.
- For this split-mode colocated case, preserve the 8.17 behavior:
  index.ts should contain the generated client code, should not reexport itself,
  and should not add a redundant reexport for its own generated sibling schemas file.
- Non-index-target barrels are separate files and must still reexport both the
  implementation file and schemas file.

Goal:
Keep the good structural fix (`excludeFilePath` + `getComparableFilePath`), but remove
the regex-based module-specifier parser. Replace it with explicit barrel self-reference
exclusion and a targeted duplicate-export check.

Required changes in packages/orval/src/write-specs.ts:
1. Delete `moduleSpecifierPatterns`, `getDeclaredModuleSpecifiers`, and
   `getUndeclaredModuleSpecifiers`.
2. At the append-to-existing-index call site, replace:
   `getUndeclaredModuleSpecifiers(imports, data)`
   with a targeted duplicate-export check:
   imports.filter((imp) => !data.includes(`export * from '${imp}'`))
   This checks for the exact statement shape this function writes, not any import
   or export that happens to reference the same module.
3. Keep excluding `indexFile` from `implementationPaths` before building the barrel imports.
4. If `output.target` resolves to the same file as `indexFile`, also exclude the generated
   sibling schemas file for that target from `implementationPaths` before building barrel
   imports. This should only apply to the default sibling schemas file generated from the
   target stem, e.g. `index.schemas.ts` for `index.ts`. Do not use this rule to exclude a
   configured `output.schemas` directory.
5. Keep `getComparableFilePath`'s ENOENT fallback. Add a short comment explaining the
   not-yet-created path scenario it protects.
6. Keep OS-gated lowercasing for `win32` and `darwin`; do not lowercase on Linux. Add a
   short comment noting that this is a platform heuristic and intentionally does not try
   to detect case-insensitive filesystems mounted on Linux.

Required changes in packages/orval/src/write-specs.test.ts:
- Revert this PR's additions to this file. After deleting the regex helper there is no
  helper to unit test, and the new `generateSpec`/`mkdtemp` integration test duplicates
  the existing regression coverage in `tests/api-generation.spec.ts`.

Expected output:
- `tests/__snapshots__/axios/issue-3675-index-target/index.ts`:
  - contains generated client code
  - does not contain `export * from './index'`
  - does not contain `export * from './index.ts'`
  - does not contain `export * from './index.schemas'`
- `tests/__snapshots__/axios/issue-3675-non-index-target/index.ts`:
  - still contains `export * from './endpoints'`
  - still contains `export * from './endpoints.schemas'`

Acceptance:
- Existing regression tests in `tests/api-generation.spec.ts` pass.
- No module-specifier regex parser remains in `write-specs.ts`.
- The fix is implemented through explicit path exclusion, not by treating operation imports
  as already-declared barrel exports.
```

</details>

PowerSlime:
Amazing! Thanks a lot! I will take a look at it tomorrow :) Since I'm already afk :)

PowerSlime:
@aqeelat should be better now :)

aqeelat:
@PowerSlime This is great work. All three points addressed cleanly, and the `getImplementationPathsForIndex` extraction is genuinely better than what I suggested. Scoping the own schemas exclusion to `SPLIT` is exactly the right call, since single mode inlines schemas and there is nothing to exclude there. CI is green and the snapshots are spot on. Really nicely done, and thank you for turning this around so fast.

Bravo

melloware (approved):

PowerSlime:
Thanks a lot! Well... I learned some things about paths resolving too :) Thank you too :) 
For the amazing tool as well 😉

melloware:
@PowerSlime thank you for the PR!

PR #3679, opened 2026-07-02
feat(vue-query): use MaybeRefOrGetter/toValue for reactive params on v5

### LINKED ISSUE #3412: vue-query: support MaybeRefOrGetter for generated hook parameters
## Description

TanStack Vue Query v5 [recommends using getter functions](https://tanstack.com/query/v5/docs/framework/vue/reactivity) (`() => value`) as the idiomatic way to pass reactive parameters to `useQuery`, and explicitly states: *"For trivial cases of simple property access, `computed` is an optimization with no real benefit."* Vue 3.3+ introduced `MaybeRefOrGetter<T>` (= `T | Ref<T> | (() => T)`) and `toValue()` to standardize this pattern. However, orval's vue-query template currently wraps all parameters with `MaybeRef<T>` via `vueWrapTypeWithMaybeRef`, which only accepts `T | Ref<T>` and leaves out getter functions entirely.

## Current behavior

Generated hook signature:

```ts
import { unref } from 'vue';
import type { MaybeRef } from 'vue';

export const getCustomerDetail = (
  customerId: MaybeRef<number>,  // ❌ getter () => number not accepted
  ...
) => {
  customerId = unref(customerId); // unref() ignores getter functions
  ...
};
```

Callers who follow TanStack's recommended getter pattern get a type error and must wrap manually:

```ts
// Recommended by TanStack Vue Query docs — but orval rejects this
useGetCustomerDetail(() => props.customerId)

// Must wrap in computed instead
const idRef = computed(() => props.customerId ?? 0)
useGetCustomerDetail(idRef, { query: { enabled: computed(() => !!props.customerId) } })
```

## Proposed behavior

```ts
import { toValue } from 'vue';
import type { MaybeRefOrGetter } from 'vue';

export const getCustomerDetail = (
  customerId: MaybeRefOrGetter<number>, // ✅ accepts T | Ref<T> | (() => T)
  ...
) => {
  customerId = toValue(customerId); // toValue() handles all three forms
  ...
};
```

Callers can use the idiomatic getter pattern directly:

```ts
useGetCustomerDetail(() => props.customerId) // ✅ no wrapper needed
```

## Proposed solution

The change is localized to `vueWrapTypeWithMaybeRef` in `packages/query/src/frameworks/vue.ts`:

1. Replace the generated type `MaybeRef<T>` → `MaybeRefOrGetter<T>`
2. Replace `import type { MaybeRef } from 'vue'` → `import type { MaybeRefOrGetter } from 'vue'`
3. Replace the call-site `unref(param)` → `toValue(param)` for path/query params

`MaybeRefOrGetter` is a superset of `MaybeRef`, so existing call sites passing `Ref` or plain values continue to work without changes. An opt-in config flag (e.g. `query.useMaybeRefOrGetter: true`) could be added to avoid breaking existing projects.

### LINKED ISSUE #3682: vue-query: TypeScript error in generated unref() reassignment when schema has a property named `value`
### Description

When using the vue-query client, orval generates parameter reassignments like:

```
param = unref(
    param,
);
```

If the schema type for `param` has a property named `value`, TypeScript's structural typing confuses it with Vue's `Ref<T>` interface (which also requires `value: T`). This causes `unref<T>()` to infer the wrong `T`, making the reassignment fail with:

`Type 'Schema | null | undefined' is not assignable to type 'MaybeRef<Schema> | undefined'. Type 'null' is not assignable to type 'MaybeRef<Schema> | undefined'.`

Renaming the `value` property to anything else makes the error disappear, confirming
the property name is the trigger.

### Output client

vue-query

```ts

export default defineConfig({
    api: {
        output: {
            allParamsOptional: true,
            client: 'vue-query',
            mode: 'single',
            httpClient: 'axios',
            override: {
                useTypeOverInterfaces: true,
            },
            target: 'src/api/client.ts',
        },
        input: {
            target: './openapi.yaml',
        },
    },
});
```

### Environment

```
  System:
    OS: macOS 26.5.1
    CPU: (8) arm64 Apple M3
    Memory: 83.89 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    @tanstack/vue-query: ^5.101.0 => 5.101.0 
    axios: ^1.18.0 => 1.18.0 
    orval: ^8.18.0 => 8.19.0 
    vue: ^3.5.38 => 3.5.38 
```

### Expected behavior

The generated unref() reassignment compiles without errors. TypeScript correctly infers T = UpdateItemRequest for unref(param) regardless of the schema's property names.

### Actual behavior

TypeScript reports:

```
Type 'UpdateItemRequest | null | undefined' is not assignable to type 'MaybeRef<UpdateItemRequest> | undefined'. Type 'null' is not assignable to type 'MaybeRef<UpdateItemRequest> | undefined'.

The error is on the generated line:
updateItemRequestSchema = unref(
    updateItemRequestSchema,
);
```

```yaml
openapi: 3.0.3
info:
  title: Demo
  version: 0.0.0
paths:
  /items/{id}:
    patch:
      operationId: items.update
      parameters:
        - in: path
          name: id
          required: true
          schema:
            type: string
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/UpdateItemRequest'
      responses:
        '200':
          description: OK
components:
  schemas:
    Money:
      type: object
      properties:
        amount:
          type: number
        currency:
          type: string
    UpdateItemRequest:
      type: object
      properties:
        name:
          type: string
        value:
          nullable: true
          allOf:
            - $ref: '#/components/schemas/Money'
```

### Additional context

_No response_

daugvinasr:
Closes https://github.com/orval-labs/orval/issues/3412. Zero changes for users as same code is still valid just now accepts getters and refs. Tested on our giant work dashboard and seems to work with zero issues.

Fix #3682

As `MaybeRefOrGetter` requires Vue 3.3+ this change is scoped only to query `v5` consumers. Query `v4` generated code is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Vue Query generated APIs now accept reactive parameters as refs, plain values, or getters (`MaybeRefOrGetter`).
  * Vue Query v5 resolves these values with `toValue` when generating request/query bindings.

* **Bug Fixes**
  * Fixed version-aware reactive handling so Vue Query v4 continues using `unref`/`MaybeRef`, while v5 uses the new resolver logic.
  * Improved `enabled`/request gating for queries with optional inputs.

* **Documentation**
  * Updated the Vue Query guide with a new “Reactive Parameters” section, including the updated helper example types.

* **Tests**
  * Added/updated coverage for v4 vs v5 reactive parameter behavior.


melloware (changes requested):
Build failing

melloware (approved):

PR #3687, opened 2026-07-07
fix(ci): trigger docker image publish on release

### LINKED ISSUE #3686: Docker image publish workflow never runs
### Description

#3654 added a Dockerfile and a "Publish Docker Image" workflow to publish docker images to `ghcr.io/orval-labs/orval` upon each release. This workflow has never run, and no image exists at that location. The installation docs instruct users to pull the image and reference version `8.19.0`, which causes confusion because users cannot successfully follow these instructions.

Cause: `docker-publish.yaml` uses `push: tags: 'v*'` as the trigger, which doesn't work because the release tags are created by the "Publish Release" workflow using the `GITHUB_TOKEN`. GitHub Actions prevents push events created by the `GITHUB_TOKEN` from creating new workflow runs, meaning that the "Publish Docker Image" action will never execute.

### Output client

other

N/A - this is a CI issue

### Environment

N/A - this is a CI issue

### Expected behavior

The "Publish Docker Image" workflow runs after every release, and docker images are pushed to `ghcr.io/orval-labs/orval` as described by the installation docs.

### Actual behavior

The "Publish Docker Image" workflow has never run, and no docker image exists at `ghcr.io/orval-labs/orval`. The command specified in the installation docs, `docker pull ghcr.io/orval-labs/orval`, results in a docker pull error.

N/A - this is a CI issue

### Additional context

The fix + some improvements to the trigger and version detection are available [on a branch](https://github.com/orval-labs/orval/compare/master...vondenstein:orval:fix/docker-publish-trigger), and I can open a PR immediately.

I tested this workflow by re-creating the `8.20.0` release on my forked repository, and the action successfully ran and publish an image for `v8.20.0`. The action can be viewed [here](https://github.com/vondenstein/orval/actions/runs/28893540555), and the resulting image can be tested with `docker run --rm ghcr.io/vondenstein/orval --version`.

Additionally, the installation docs mention version `8.19.0`, but no docker image exists for this release. I see a few ways to fix this:
- A maintainer manually builds the docker images for previous releases and pushes them
- The "Publish Docker Image" is updated to include the `workflow_dispatch` trigger and accept `VERSION` and `TAG` variables, so that images can be built and pushed for previous releases
- The docs are updated to reference the first version to have an accompanying docker image (say, `8.21.0`)

I wasn't sure which approach would be preferred, but I would be happy to implement options two or three in my PR.

vondenstein:
Addresses an issue in which the Docker publish workflow wasn't running on new releases. Switched to using the `workflow_run` trigger on "Publish Release", updated `if:` condition, and bumped the pinned version number in the docs to the first version expected to have a docker image: `8.21.0`.

Fixes #3686 

## Summary by CodeRabbit

* **New Features**
  * Docker images are now published via the completed release workflow, using versioned and `latest` tags that reflect the current repository.
  * Image versioning is derived from the corresponding release branch.
* **Bug Fixes**
  * Added safeguards so publishing runs only after a successful release workflow and the matching version is available (with a short wait window).
* **Documentation**
  * Updated the Docker installation instructions to pin the image to version `8.21.0` (from `8.19.0`).

### DIFF
--- a/.github/workflows/docker-publish.yaml
+++ b/.github/workflows/docker-publish.yaml
@@ -1,9 +1,9 @@
 name: Publish Docker Image
 
 on:
-  push:
-    tags:
-      - 'v*'
+  workflow_run:
+    workflows: ["Publish Release"]
+    types: [completed]
 
 permissions:
   contents: read
@@ -12,6 +12,10 @@ permissions:
 jobs:
   docker:
     runs-on: ubuntu-latest
+    if: |
+      github.event.workflow_run.head_repository.full_name == github.repository &&
+      github.event.workflow_run.conclusion == 'success' && 
+      startsWith(github.event.workflow_run.head_branch, 'release/v')
     steps:
       - name: Checkout repository
         uses: actions/checkout@v7
@@ -20,7 +24,28 @@ jobs:
 
       - name: Extract version
         id: version
-        run: echo "version=${GITHUB_REF#refs/tags/v}" >> "$GITHUB_OUTPUT"
+        env:
+          HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
+        run: |
+          VERSION="${HEAD_BRANCH#release/v}"
+          if ! [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "Could not extract a valid version from branch '$HEAD_BRANCH'" >&2
+            exit 1
+          fi
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+
+      - name: Wait for npm package
+        env:
+          VERSION: ${{ steps.version.outputs.version }}
+        run: |
+          for i in $(seq 1 30); do
+            if npm view "orval@${VERSION}" version > /dev/null 2>&1; then
+              exit 0
+            fi
+            echo "orval@${VERSION} not yet on npm, retrying..."
+            sleep 10
+          done
+          echo "Package never appeared on npm" && exit 1
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v4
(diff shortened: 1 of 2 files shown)


vondenstein:
@melloware I addressed the initial CodeRabbit comment to pass the version via `env`, but was going to leave the exponential backoff comment unaddressed for now since re-running the job would work in the unlikely event the timeout is reached. Let me know what you think.

melloware (approved):

melloware:
LGTM1

melloware:
@vondenstein can you investigate it looks like it failed when I released v8.24.0?  https://github.com/orval-labs/orval/actions/runs/31259416947/job/93107790970

melloware:
@vondenstein its still failing can you investigate?

vondenstein:
Whoops, yes taking a look at it today.

vondenstein:
@melloware could you do me a favor and try re-running one of the failed jobs? I suspect that it might have tried to build the container image before all of the orval packages were pushed to npm.

melloware:
@vondenstein that was it I just re=ran the build and it published 8.25.0.  So its a timing issue?

vondenstein:
@melloware yes, the job checks whether the Orval package exists on npm before builder the docker container, but it doesn't check that all of its dependencies do.

The main Orval package depends on @orval/core etc. and some of those weren't showing as available on npm yet.

The permanent fix is probably to check that all of the dependencies were also pushed to npm before building the image.

I can open a PR for that this weekend if that would help. Until then, just re-run the job if it fails. You may need to re-run the job for 8.24.0 too, that image is still missing.

melloware:
@vondenstein what if we just switched it to a daily job at midnight that checks the version and if new version publishes  the image else bails?

but yeah ANY PR to fix this would be great.  Right now I manually run it after deploys waiting 10 minutes.

PR #3697, opened 2026-07-10
fix(mock): generate tuple values for OpenAPI 3.1 prefixItems

### LINKED ISSUE #3691: Generated MSW mock is not assignable to generated TypeScript tuple type (OpenAPI 3.1 prefixItems)
### Description

When generating MSW mocks from an OpenAPI 3.1 schema containing a tuple (`prefixItems`), the generated MSW mock is not assignable to the generated TypeScript type.

I couldn't determine from the documentation or existing issues whether this behaviour is expected, so I'm opening this issue with a minimal reproduction.

### Output client

axios

```ts

export default defineConfig({
  api: {
    input: "./openapi.json",
    output: {
      target: "./src/api.ts",
      httpClient: "axios",
      client: "vue-query",
      mode: "tags",
      mock: {
        path: "./src/mocks",
        indexMockFiles: true,
        generators: [
          {
            type: "msw",
            delay: 250,
            useExamples: true,
          },
        ],
      },
      override: {
        header: false,
        mutator: {
          path: "./src/client.ts",
          name: "client",
        },
      },
    },
  },
});
```

### Environment

System:
    OS: macOS 26.4.1
    CPU: (10) arm64 Apple M4
    Memory: 87.89 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
npmPackages:
    @tanstack/vue-query: ^5.101.1 => 5.101.1 
    axios: ^1.18.1 => 1.18.1 
    msw: ^2.14.6 => 2.14.6 
    orval: ^8.19.0 => 8.19.0 
    vue: ^3.5.30 => 3.5.38 
    zod: ^3.25.76 => 3.25.76 

### Expected behavior

When generating MSW mocks for an OpenAPI 3.1 schema that defines a tuple using `prefixItems`, the generated mock should be assignable to the generated TypeScript type.

For example, given the generated type:

```ts
point: [number, number] | null
```

the generated mock should return either a valid tuple or `null`, e.g.:

```ts
point: faker.helpers.arrayElement([
  [faker.number.float(), faker.number.float()],
  null,
])
```

### Actual behavior

The generated TypeScript type is:

```ts
point: [number, number] | null
```

However, the generated MSW mock is:

```ts
point: faker.helpers.arrayElement([[], null])
```

This results in the following TypeScript error during compilation:

```text
Type '[] | null' is not assignable to type '[number, number] | null'.
Type '[]' is not assignable to type '[number, number]'.
Source has 0 element(s) but target requires 2.
```

The generated mock is therefore not assignable to the generated type.

```yaml
openapi: 3.1.0

info:
  title: Tuple Mock Reproduction
  version: 1.0.0

paths: {}

components:
  schemas:
    Example:
      type: object
      properties:
        point:
          anyOf:
            - type: array
              prefixItems:
                - type: number
                - type: number
              minItems: 2
              maxItems: 2
            - type: "null"
```

### Additional context

I searched the existing issues but could not find describing this specific behaviour.

wadakatu:
## Overview

MSW mocks for OpenAPI 3.1 tuples (`prefixItems`) were generated as an empty
array `[]`, which is not assignable to the generated `[T0, T1, ...]` tuple
type, so the generated mock file failed to type-check.

Given the minimal reproduction from #3691:

```yaml
point:
  anyOf:
    - type: array
      prefixItems: [ { type: number }, { type: number } ]
      minItems: 2
      maxItems: 2
    - type: "null"
```

orval emits the correct type `point?: [number, number] | null`, but the mock
was `faker.helpers.arrayElement([[], null])` — the inner `[]` triggers:

```
Type '[]' is not assignable to type '[number, number]'.
  Source has 0 element(s) but target requires 2.
```

## Root cause

`getMockScalar`'s `case 'array'` in `packages/mock/src/faker/getters/scalar.ts`
only looked at `item.items` and returned `'[]'` when it was absent. OpenAPI 3.1
tuples carry their positional element schemas in `prefixItems`, not `items`, so
tuples hit the empty-array early return. The **type** side already handled this
correctly in `packages/core/src/getters/array.ts` (`getArray`); only the mock
side was never taught about `prefixItems`.

## Change

Add a `prefixItems` branch to `case 'array'`, **before** the `!item.items`
guard, that mocks each positional element through the existing
`resolveMockValue` helper and emits a tuple literal `[el0, el1, ...]`.

Placing the branch before the `items` guard also fixes the tuple-with-rest case
(`prefixItems` + `items`, generated type `[...prefix, ...Additional[]]`), which
previously fell through to the normal `Array.from(...)` path and produced a
plain array that was likewise not assignable. Emitting only the fixed positions
stays type-safe because the rest element permits zero entries.

## Tests

Adds `tests/specifications/issue-3691.yaml` and an `issue-3691` MSW project in
`tests/configs/mock.config.ts`, covering the `anyOf: [tuple, null]` shape from
the issue plus plain, object-element, tuple-with-rest, and empty-schema (`{}`)
element variants, with a focused assertion in `tests/api-generation.spec.ts`.
Before this change, `typecheck-generated` fails on all of these; after, it
passes with no churn to existing snapshots.

## Follow-up (out of scope)

`packages/mock/src/faker/format-example-value.ts` gates array-example handling
on `resolved.items` in the same `items`-only way, so a tuple with an `example`
under `useExamples: true` likely hits a parallel bug. The #3691 spec has no
examples, so it is untouched here; noting it for a separate change.

Closes #3691

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mock data generation for OpenAPI 3.1 tuple schemas using `prefixItems`.
  * Tuple positions now receive concrete, schema-compatible values (including nullable, object, and tuple-with-rest cases).
  * Avoids generating invalid empty-slot tuple output when a fixed-position schema is empty.
  * Preserved existing array mocking behavior when tuple-specific definitions are unavailable.

* **Tests**
  * Added coverage to ensure generated mock values match the expected tuple types for multiple tuple variants.  
  * Added a new mock generation configuration and an OpenAPI fixture for the tuple `prefixItems` scenario.


wadakatu:
Addressed the CodeRabbit nitpick in 8ef00a2a9 — added a `not.toContain('point: [])` guard so the issue's exact nullable-tuple case (`anyOf: [tuple, null]`) is covered directly.

melloware (approved):

Issue #3706, opened 2026-07-10
bug(msw): Orval MSW unexported item aliases


the-ult:
### Acknowledgement

- [x] I acknowledge that issues using this template may be closed without further explanation at the maintainers' discretion.

## Summary

Orval's MSW output can import operation-specific `…Item` aliases that are not exported by the generated Zod schema barrel. This makes the generated MSW module fail TypeScript compilation as soon as it is exported from a library's public API.

Example generated import:

```ts
import type { GetCatalogItemsItemsItem, CatalogItems } from './catalog-api.schemas';
```

`GetCatalogItemsItemsItem` is not exported from `catalog-api.schemas`; only `CatalogItems` is available.

## Desired behavior

Generate MSW factories using an exported type, for example:

```ts
type CatalogItem = CatalogItems[number];
```

or ensure every imported operation-specific item alias is exported from the Zod schema barrel.

## Current workspace workaround

A host-project OpenAPI post-processing step rewrites these aliases to the corresponding exported collection element types, such as `CatalogItems[number]`, before publicly exporting generated MSW handlers.

Issue #3704, opened 2026-07-10
feat(core): artifact output groups with runtime-safe, per-group entry points


the-ult:
## What's missing

One OpenAPI document can produce artifacts for three different runtimes: schemas and types (universal), Angular `HttpClient` / `httpResource` clients (browser), and MSW handlers plus Faker factories (Node test runners, Storybook, browser mocks). Orval can already write these to separate directories, but it has no first-class notion of an **artifact group**: a named bundle of generated files with its own barrel, safe to import without dragging in the other groups' runtime code. Two concrete holes remain, and one config still cannot describe more than one group boundary.

This is not Angular-specific. Angular is just where it hurts most, because a generated service imports `@angular/core` and evaluates an `@Injectable` decorator at module load (`samples/angular-app/src/api/http-client/pets/pets.service.ts:7-15`).

## Why it matters

In a monorepo, an app package re-exports generated output through one barrel:

```ts
export * from './lib/__generated__/openapi';
```

A Node-only mock package then imports a single handler:

```ts
import { getGetArticleMockHandler } from '@acme/reference-models';
```

Node evaluates the whole barrel, which evaluates the Angular client modules too. Under Playwright or Node instrumentation this can fail while Angular initializes platform code — even though the consumer only wanted an MSW handler. The reverse also bites: an Angular app that imports the client should not pull `@faker-js/faker` and `msw` into its module graph.

Note this is a **value**-level problem, not a type-only-export problem. Type imports are already safe: under `verbatimModuleSyntax`, `import type { A } from 'a'` is erased entirely from the emitted JS. What leaks is real runtime code.

## Today's behavior (verified on master @ 70fc4fad4)

Already works, and covers a good part of the original request:

- `schemas` accepts a path plus `importPath`, so types/Zod schemas already live in their own directory and can be imported through a package specifier — `packages/core/src/types.ts:330-334`.
- `mock` accepts a shared `path`, and each generator (`msw`, `faker`) accepts its own `path` — `packages/core/src/types.ts:530,554,578`.
- `mock.indexMockFiles: true` emits `index.msw.ts` / `index.faker.ts` mock-only barrels in `split` and `tags-split` — `packages/core/src/types.ts:570-577`, `packages/core/src/writers/split-tags-mode.ts:461-479`.
- Faker schema factories can be exposed through a dedicated barrel via `mock.generators[].schemasImportPath` — `packages/core/src/types.ts:539-544`.
[...]

Issue #3702, opened 2026-07-10
feat(angular): resolve the base URL through Angular DI instead of baking it into the route


the-ult:
## What's missing

The Angular generator has no way to resolve an API's base URL through Angular's dependency injection. `output.baseUrl` is glued onto every route while the code is generated, so the value is frozen into the bundle. There is no token to override, no provider to swap, and no way for one build to serve two environments.

## Why it matters

An Angular app usually reaches an API through a gateway or proxy. The browser-visible route is deployment configuration, not part of the API contract:

- OpenAPI path: `/v1/resources/{id}`
- OpenAPI server: `https://example-api-service`
- Browser route: `/api/example-api-service/v1/resources/{id}`

The gateway prefix differs per environment (local proxy, staging, production) and per consuming app. Today the only way to change it is to re-run codegen with a different `baseUrl`, which means one build artifact per environment. It also breaks down when one app consumes several generated APIs with different gateway prefixes, and it gives tests and SSR no seam to substitute a URL.

## Today's behavior

The base URL is concatenated into the route string during generation:

- `packages/orval/src/api.ts:99` calls `getFullRoute(route, servers, output.baseUrl)`.
- `packages/core/src/getters/route.ts:108-170` returns `` `${base}${route}` `` as a plain string.
- `packages/angular/src/http-client.ts:730-731` emits that string inside a template literal: `` this.http.${verb}(`${route}`, …) ``.
- `packages/angular/src/http-resource.ts:1563-1567` does the same for `httpResource` helpers.

Generated output (`tests/generated/angular/split/endpoints.service.ts:152`):

```ts
return this.http.get<TData>(`/v${version}/pets`, {
  ...(options as Omit<NonNullable<typeof options>, 'observe'>),
  observe: 'body',
  params: filteredParams,
});
```

And for `httpResource` (`tests/generated/angular/http-resource-tags/pets.ts:114`):

```ts
return httpResource<Pets>(() => {
  const request = {
    url: `/pets`,
    params: filterParams(params?.() ?? {}, new Set<string>([])),
  };
  return request;
}, options);
```

With `baseUrl: 'https://example-api-service'` the literal simply becomes `` `https://example-api-service/v${version}/pets` ``.

Two escape hatches exist, and neither is DI:

[...]

Issue #3700, opened 2026-07-10
bug: Angular `httpResource` calls an optional request-body signal unconditionally


the-ult:
### Description

# Bug: Angular `httpResource` calls an optional request-body signal unconditionally

## Summary

When generating an Angular client with `override.angular.retrievalClient: 'both'`, an operation with an optional request body can generate an `httpResource` request factory that calls the optional body signal unconditionally. Calling the generated resource without a body signal throws at runtime instead of remaining idle until input is supplied.

## Environment

- Orval: 8.20.0
- Angular: 21.x
- TypeScript: 5.9.x
- Zod schemas: enabled
- Output mode: `split`
- Platform: Windows (the problem is not platform-specific)

## Minimal OpenAPI input

```yaml
openapi: 3.0.3
info:
  title: Optional body reproduction
  version: 1.0.0
paths:
  /widgets:
    post:
      operationId: createWidget
      requestBody:
        required: false
        content:
          application/json:
            schema:
              type: object
              properties:
                name:
                  type: string
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: object
                properties:
                  id:
                    type: string
```

## Configuration

```ts
{
  input: { target: './optional-body.openapi.yaml' },
  output: {
    mode: 'split',
    target: './generated/optional-body.client.ts',
    schemas: { path: './generated/optional-body.schemas', type: 'zod' },
    client: 'angular',
    override: {
      angular: {
        retrievalClient: 'both',
        runtimeValidation: true,
      },
    },
  },
}
```

## Actual behavior

The generated resource factory has the equivalent of the following pattern:

```ts
return httpResource<Response>(() => {
  return {
    url: '/widgets',
    method: 'POST',
    body: bodySignal(),
  };
});
```

`bodySignal` is optional for an optional OpenAPI request body. If the caller does not provide it, evaluating the resource invokes `undefined` and throws.

## Expected behavior

An absent optional body signal should prevent creation of a request, following Angular `httpResource`'s `undefined` request-factory contract:

```ts
return httpResource<Response>(() => {
  if (!bodySignal) return undefined;

  return {
    url: '/widgets',
    method: 'POST',
    body: bodySignal(),
  };
});
```

Alternatively, the generator can emit an overload that requires a body signal whenever it emits `body: bodySignal()`.

## Impact

[...]

PR #3708, opened 2026-07-11
fix(angular): serialize object query params per OpenAPI style/explode instead of dropping them

### LINKED ISSUE #3705: feat(angular):  Orval Angular query-object serialization
### Acknowledgement

## Summary

For an OpenAPI operation that models a query object as a single parameter, Orval's Angular client generated a nested `arg0` query parameter instead of flattening the documented query fields.

The generated type is:

```ts
type SearchCatalogParams = {
  arg0: {
    itemReferences?: string[];
    pageNumber?: number;
  };
};
```

The generated request serializes `arg0`, while the API accepts top-level query parameters such as:

```text
/catalog-items?itemReferences=ITEM_REFERENCE%3B...&pageNumber=1
```

## Impact

The generated service cannot replace a working handwritten search service without changing the outgoing request shape. The service must remain until generated query serialization matches the contract.

## Desired behavior

Generate a flat parameter type and serialization for object-style query parameters, or expose an explicit configurable flattening strategy for Angular `HttpClient` and `httpResource` output.

The generated parameter should be:

```ts
type SearchCatalogParams = {
  itemReferences?: string[];
  pageNumber?: number;
};
```

the-ult:
## Summary

Closes #3705.

Angular clients silently **dropped** plain-object query params: the generated `filterParams` helper discards any object value unless a `paramsSerializer` is configured, so an operation like the issue's `searchCatalog(arg0)` never sent its documented parameters at all (verified against the committed `issue-3326` snapshot).

OpenAPI 3.x mandates `style: form` + `explode: true` as the default for query params, which for object schemas means the fields serialize as separate top-level params (`?itemReferences=...&pageNumber=1`). This PR implements exactly that.

## What changed

- **Core** (`getters/query-params.ts`): each plain-object query param gets a serialization strategy derived from the spec — `form`+`explode:true` → **flatten**, `form`+`explode:false` → **comma-joined**, `deepObject` → **bracketed keys** — exposed as optional `GetterQueryParam.objectQueryParams` (groundwork other clients, e.g. #2589, can adopt).
- **Angular** (HttpClient services *and* httpResource functions — they share one helper): the generated `filterParams` helper gains a strategies branch and the call sites pass per-param strategies. The flattened/comma/deepObject outputs are all `HttpParams`-compatible primitives; nested non-primitives are filtered with the same primitive rules (never `[object Object]`).
- **Escape hatch**: `override.angular.queryObjectSerialization: 'spec' | 'legacy'` (default `'spec'`; `'legacy'` restores the old dropping behavior). Settable globally, per-tag, per-operation.

## Deliberately NOT changed (scope decision)

- **Parameter types stay nested** and keyed by the declared param name — `{ arg0: { ... } }` faithfully mirrors the spec, matching what OpenAPI Generator / NSwag / openapi-fetch do. The `arg0` name itself is a springdoc/Spring artifact; the docs now include the spec-side fix (`-parameters` / `@ParameterObject`).
- **`paramsFilter` / `paramsSerializer` precedence unchanged**: a configured filter still owns everything; a configured serializer still receives the raw nested object (the #3326 passthrough contract) — strategies are suppressed in both cases, so existing workarounds see zero change.
- The helper body is **byte-identical for files without object query params**, keeping the blast radius to files that actually have them.

## Why default-on is safe

There was no working behavior to preserve — these params were silently omitted from every request. Sending the parameters the server's own contract documents is a bug fix (same rationale as the fetch client honoring `explode: true` by default). `'legacy'` exists for anyone relying on the omission.

## Verification

- Unit: core 2127 ✓, angular 232 ✓, mock 297 ✓, solid-start 16 ✓, orval 191 ✓ (1 pre-existing macOS-only `/tmp` flake)
- Snapshots: 5327/5327 ✓; blast radius exactly as predicted — **only** `tests/__snapshots__/angular/issue-3326/endpoints.ts` modified (+69/−7), 4 new fixture dirs (`issue-3705`, `-http-resource`, `-legacy`, `-serializer`); `issue-3326-filter`, `issue-3326-serializer`, all petstore-based Angular dirs, `angular-query`, and `samples/**` are untouched
- Lint + typecheck clean; all 16 generated clients compile

## Follow-ups (out of scope, flagged in docs)

- Arrays of objects and `content: application/json` params keep current behavior (serializer passthrough / drop) — no primitive-safe form encoding exists for the former; the latter wants a `'json'` strategy.
- Other clients can adopt the core strategy metadata (#2589 deepObject, #3231/#1285 array styles).
- A pre-existing per-operation `override.angular` merge quirk (setting one key re-defaults siblings at the normalization boundary) is documented by a new test but not changed here.
- zh Angular guide not translated (was already out of sync).

## Docs

`docs/content/docs/guides/angular.mdx`: “Query parameter filtering” defaults rewritten + new “Object query parameters (style & explode)” subsection (spec example, wire result, strategy table, legacy escape hatch, springdoc tip). `docs/content/docs/reference/configuration/output.mdx`: `queryObjectSerialization` reference.

## Related issues

#2589 (deepObject, open), #3326 (object-param passthrough — contract preserved), #3231 / #1285 / #3053 (explode handling elsewhere), #2493 (array styles), #2998 / #3634 (Angular filterParams gaps), #3164 (react-query flattened args). Sibling cluster: #3702, #3704, #3706, #3700.

## Summary by CodeRabbit

* **New Features**
  * Angular clients now serialize object-valued query parameters using OpenAPI `style`/`explode` rules by default (`spec`), including `form`/`deepObject`-style formats.
  * Added `override.angular.queryObjectSerialization` (`spec`/`legacy`) to restore prior behavior that dropped object-typed query params (`legacy`).
  * When `paramsSerializer` or `paramsFilter` is set, object serialization strategies are suppressed; `paramsFilter` must handle object param strategies (or rely on a downstream serializer).

* **Documentation**
  * Updated Angular generator docs with the object query parameter mappings and override/precedence details.

* **Tests**
  * Added Issue `#3705` coverage for default, `legacy`, `paramsSerializer`, and `paramsFilter`, plus `httpResource` scenarios.


the-ult:
Copilot's three findings (comma strategy dropping array-valued object properties at both emission sites + missing runtime test) are fixed in 3a51f0c9b: arrays are filtered to primitives and appended after the property name (matching the documented `itemReferences,a,b,pageNumber,1` shape), with runtime-eval tests for the array and null/object-filtering cases. CodeRabbit has confirmed both of its findings addressed in the same commit.

melloware (approved):

Issue #3656, opened 2026-06-26
Missing enum value import in generated MSW mock for oneOf branch


artemijkurganov:
### Description

After upgrading from Orval `8.17.x` to `8.18.0`, generated MSW mocks started failing TypeScript compilation.

The generated mock helper uses a referenced enum as a runtime value:

```ts
Object.values(ReasonEnum)
```

but `ReasonEnum` is not imported as a value. It is only used/imported through the generated model types, so TypeScript reports:

```text
error TS2304: Cannot find name 'ReasonEnum'.
```

This appears to be a regression of the same general class as #1201 / #1208 / #1205, but triggered through a `oneOf` discriminator branch mock helper.

The issue is still reproducible with `orval@8.19.0`.

### Output client

fetch

### Configuration (orval.config)

```ts
import { defineConfig } from "orval";

export default defineConfig({
  repro: {
    input: {
      target: "./openapi.yaml",
    },
    output: {
      mode: "single",
      target: "./src/generated/client.ts",
      schemas: "./src/generated/model",
      client: "fetch",
      httpClient: "fetch",
      mock: {
        generators: [
          {
            type: "msw",
            delay: 0,
          },
        ],
      },
      override: {
        mock: {
          required: true,
        },
      },
    },
  },
});
```

### Environment

Run the command above and paste the output here.
  System:
    OS: macOS 26.5.1
    CPU: (8) arm64 Apple M2
    Memory: 125.34 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    msw: ^2.14.6 => 2.14.6 
    orval: 8.19.0 => 8.19.0 

### Expected behavior

The generated MSW mock should either import the enum as a runtime value:

```ts
import { ReasonEnum } from "./model";
```

or inline the enum values in the mock:

```ts
reason: faker.helpers.arrayElement(["queue_rebalance"] as const)
```

The generated code should pass `tsc --noEmit`.

### Actual behavior

Orval generates code like this:

```ts
export const getListEventsResponseUpdatedDetailsMock = (
  overrideResponse: Partial<UpdatedDetails> = {},
): UpdatedDetails => ({
  ...{
    event_type: faker.helpers.arrayElement(["updated"] as const),
    reason: faker.helpers.arrayElement(Object.values(ReasonEnum)),
  },
  ...overrideResponse,
});
```

but `ReasonEnum` is not imported.

TypeScript error:

```text
[...]

Issue #3612, opened 2026-06-15
Faker schema factory import path doesn't resolve when schemas.importPath is set


aqeelat:
## Description

When `schemas.importPath` is configured and faker mock generation is enabled with `schemas: true`, orval generates an import path of `<importPath>/index.faker` (e.g., `@fathom/data-layer/sdk/index.faker`). This path doesn't resolve because `importPath` is a **package barrel specifier**, not a directory prefix — you can't append sub-path segments to it.

**Output client:** `angular` (affects all clients using faker mocks with `importPath`)

## Configuration

```ts
output: {
  mode: 'tags-split',
  client: 'angular',
  target: './libs/client/sdk/generated',
  schemas: {
    path: './libs/data-layer/sdk/generated',
    importPath: '@fathom/data-layer/sdk',
  },
  mock: {
    indexMockFiles: true,
    path: './libs/client/sdk/mocks',
    generators: [{ type: 'faker', schemas: true }],
  },
}
```

## Expected behavior

The faker factory import should be resolvable. Consumers need a way to specify a separate import path for faker schema factories, since the production barrel (`importPath`) cannot (and should not) expose internal `index.faker` files as a subpath export.

## Actual behavior

Generated mock file (`libs/client/sdk/mocks/data-quality/data-quality.faker.ts`):

```ts
import type { DataQualityRuleInfo } from '@fathom/data-layer/sdk';                // ✅ resolves
import { getDataQualityRuleInfoMock } from '@fathom/data-layer/sdk/index.faker';  // ❌ doesn't resolve
```

With tsconfig path mappings like `"@fathom/data-layer/*": ["libs/data-layer/*/public-api.ts"]`, `@fathom/data-layer/sdk` resolves to `public-api.ts` (a barrel). You cannot append `/index.faker` to reach an internal file — the path mapping resolves the full specifier, not a directory prefix.

## Root cause

In `packages/core/src/writers/generate-imports-for-builder.ts:38-41`, the faker schema factory dependency is computed as:

```ts
dependency: upath.joinSafe(
  relativeSchemasPath,           // e.g. "@fathom/data-layer/sdk"
  `index.faker${schemaFactoryImportExtension}`,
),
```

This joins the package specifier with `index.faker`, producing an unresolvable sub-path. The type imports (line 56/62) work fine because they use `relativeSchemasPath` directly.

## Suggested fix

[...]

PR #3709, opened 2026-07-11
fix(mock): use exported element aliases for top-level array responses in arrayItems factories

### LINKED ISSUE #3706: bug(msw): Orval MSW unexported item aliases

the-ult:
## Summary

Closes #3706.

With `mock.generators: [{ type: 'msw', arrayItems: true }]`, mocks for **top-level array responses** imported type names orval never emits — e.g. `import type { GetCatalogItemsItemsItem } from './model'` when only `ItemsItem` exists. The generated module then fails `tsc` as soon as it's exported from a library's public API (reproduced live at HEAD with three phantom shapes).

## Root cause

`getArrayItemFactoryNames` (`packages/mock/src/faker/getters/array-item-factory.ts`) fabricates `${pascal(operationId)}${pascal(propertyName)}${itemSuffix}` whenever `parentName` is undefined — i.e. whenever the array *is* the top-level response schema — instead of reusing the element alias the schema generator actually emits (`<RefName><itemSuffix>` for `$ref`'d arrays, the pre-`[]` alias for inline ones). The nested variant of this class was already fixed via the `isAmbiguousInlineItemContext` guard, confirming the intended contract: mock type names must equal schema-generator names.

## Fix (pure bug fix, no config surface)

In the no-`parentName` branch only:

- inline top-level array (`Foo[]` definition) → reuse the pre-`[]` alias, **bailing out to the always-correct inline `.map()` body** when the base isn't a plain identifier (`(A | B)[]`, `readonly X[]`);
- `$ref`'d array schema → reuse `${pascal(refName)}${itemSuffix}` (same `itemSuffix` config core uses), bailing out for multi-ref `allOf` items with no direct properties (core emits no alias there);
- `factoryName` formulas unchanged (outputs hitting this branch never compiled, so they're not a compat surface);
- `itemsRef` branch, `parentName` branch, and ambiguity guard byte-identical.

Because every output that hit this branch failed `tsc`, no working project can observe a behavior change.

## Verification

- Unit: mock 302/302 ✓ (new failing-first tests for all branch shapes incl. bail-outs)
- Snapshots: 5317/5317 ✓; 18 new snapshot files under `tests/__snapshots__/mock/issue-3706-msw-array-item-aliases{,-strict}/**`, **zero modified existing snapshots** (canaries `faker-array-items`, `msw-array-items`, `issue-3590*`, `issue-3574*` untouched)
- New fixture reproduces the reporter's exact `Get…ItemsItem` phantom plus inline/nullable/parentName regression guards; the strict variant (`override.mock.required/nonNullable`) exercises `${typeName}Mock` naming over the corrected aliases
- All 16 generated clients typecheck (`tests` build gate) — the phantom import is exactly what this gate catches; lint clean

## Follow-up (out of scope, discovered during research)

`output.schemas: { type: 'zod' }` barrels export **no synthetic inline aliases at all** (e.g. `GetCatalogItemsInline200Item` is imported by the plain fetch client itself yet missing from the zod model dir). That's a separate defect in the zod schema writer / import-rerouting path (`packages/orval/src/write-specs.ts:294-330`) affecting non-mock output too — it deserves its own issue rather than being folded in here; this PR's fix is correct independent of it. I can file that issue with a minimal repro if wanted.

## Related issues

#3513 (reusable array-item mock factories — the feature this hardens), #3514 (arrayItems origin), #3269 / #3656 (same class: MSW missing imports), #3108 (zod + tags-split phantom exports), #3612 (faker import path), #3574 / #3590 (strict mock naming). Sibling cluster: #3702, #3704, #3705.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mock generation for top-level array responses by producing correct reusable item factory names and imports.
  * Added safeguards for ambiguous array schemas to prevent incorrect factory generation.
* **Documentation**
  * Updated the Faker guide to clarify how faker factories are derived for top-level array responses, including nullable and fallback/inlining behavior.
* **Tests**
  * Expanded test coverage for referenced and inline top-level arrays, ambiguous/allOf cases, nullable arrays, and stricter item requirements.


the-ult:
Copilot's nullable-array finding is fixed in fb4c14deb: the ' | null' suffix is stripped before branch selection (so 'CatalogItems | null' resolves to the real `CatalogItemsItem` alias and 'Foo[] | null' to `Foo`), and the bare-name branch now has an identifier guard bailing to the inline item body for any non-identifier shape. The fixture gained a `getNullableCatalogItems` nullable top-level array operation — its generated snapshot imports only real, exported model types; unit tests cover all three nullable shapes.

melloware (approved):

Issue #3593, opened 2026-06-12
TS6133: 'requiredNullableParamKeys' is declared but its value is never read.


jjfreez:
### Description

Since **v8.6.0**, the Angular client generator emits `requiredNullableParamKeys` 
unconditionally in the inline IIFE template, but the only branch that consumes it 
(`preserveNullableBranch`) is only injected when `preserveRequiredNullables === true`.

`preserveRequiredNullables` is only set to `true` when a `paramsSerializer` is 
configured. In the common case — Angular client **without** `paramsSerializer` — 
the branch is omitted from the template but the variable declaration remains, 
causing TypeScript to report:

> TS6133: 'requiredNullableParamKeys' is declared but its value is never read.

This error surfaces at compile time and breaks Angular test runners (Karma/Jest) 
that use `noUnusedLocals: true` in their `tsconfig`.

### Steps to reproduce

1. Use `client: 'angular'` with **no** `paramsSerializer` in the orval config
2. Have at least one endpoint with query params
3. Run `orval` to generate code
4. Run TypeScript compiler with `noUnusedLocals: true`

### Output client

angular

### Configuration (orval.config)

```ts
import { defineConfig } from 'orval';

export default defineConfig({
  myApi: {
    output: {
      client: 'angular',
      mode: 'tags-split',
      target: 'generated',
      schemas: 'generated/models',
      override: {
        requestOptions: false
      }
    },
    input: {
      target: './my-api.yaml'
    }
  }
});
```

### Environment

  System:
    OS: macOS 26.5
    CPU: (12) arm64 Apple M2 Pro
    Memory: 154.45 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    @angular/common: ^21.2.16 => 21.2.16 
    @angular/core: ^21.2.16 => 21.2.16 
    orval: ^8.17.0 => 8.17.0 

### Expected behavior

When preserveRequiredNullables is false (i.e. no paramsSerializer configured), requiredNullableParamKeys should not be emitted since it is never consumed. The generated code should be:

```ts
@Injectable({ providedIn: 'root' })
export class UserService {
  private readonly http = inject(HttpClient);
/**
 * log user into the system.
 * @summary Logs user into the system.
 */

  loginUser(
    params?: LoginUserParams, ): Observable<string> {
    return this.http.get<string>(
      `http://localhost:3000/user/login`,{
        params: (() => {
[...]

Issue #3326, opened 2026-05-06
Angular filteredParams object support


MichalK6677:
### Description

Orval filters out query params of object type. I have java API where query params are objects and it is generated as objects in ApiService. Hovewer there is filtering only for string, number and boolean so all my object params are gone. I have flattenParams method in mutator which makes flat params from object params but the new orval filters out my object query params before mutator is hit. How can I disable that filtering?

### Output client

angular

### Configuration (orval.config)

```ts
export default defineConfig({
  ezko: {
    input: {
      target: specFile,
      override: {
        transformer: inputTransformer,
      },
    },
    output: {
      client: 'angular',
      tsconfig: '../../../../tsconfig.json',
      formatter: 'prettier',
      mode: 'split',
      target: './api.ts',
      override: {
        mutator: {
          path: './api-mutator.ts',
          name: 'httpRequestAny',
        },
        title: () => 'Api',
        transformer: (schema) => outputTransformer(schema),
        header: ({ title, description, version }) => [
          `Generated by orval 🍺`,
          `Do not edit manually.`,
          ...(title ? [title] : []),
          ...(description ? [description] : []),
          ...(version ? [`OpenAPI spec version: ${version}`] : []),
          'eslint-disable',
        ],
      },
    },
  },
});

```

### Environment

 System:
    OS: Windows 11 10.0.26200
    CPU: (22) x64 Intel(R) Core(TM) Ultra 7 155H
    Memory: 3.99 GB / 31.48 GB
 npmPackages:
    @angular/common: ^21.2.11 => 21.2.11 
    @angular/core: ^21.2.11 => 21.2.11 
    orval: 8.9.1 => 8.9.1 

### OpenAPI document (minimal, if applicable)

```yaml
openapi: 3.0.3
info:
  title: Demo
  version: 0.0.0
paths: {}
```

the-ult:
Working on it
But it is dependent on: https://github.com/orval-labs/orval/pull/3359, finishing that first

PR #3168, opened 2026-03-30
fix(angular): use null literal in filterParams to fix TypeScript error

zeriong:
## Summary

Closes #3104

The generated `filterParams` function assigns `filteredParams[key] = value` in the null-check branch, where `value` is typed as `unknown` from `Object.entries()`. TypeScript may not reliably narrow `unknown` to `null` in compound `&&` conditions, causing:

> Type 'unknown' is not assignable to type 'string | number | boolean | Array<string | number | boolean>'

Since `value === null` is already asserted in the condition, this PR replaces `filteredParams[key] = value` with `filteredParams[key] = null` — semantically equivalent but avoids the narrowing dependency.

## Changes

- `getAngularFilteredParamsExpression()` (IIFE version)
- `getAngularFilteredParamsHelperBody()` (standalone function version)
- Updated all affected snapshots and samples (27 files)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of null values in required nullable query parameters to ensure they are properly preserved in HTTP requests.

PR #3716, opened 2026-07-11
fix(angular): send required nullable query params as empty string without a paramsSerializer

### LINKED ISSUE #3712: bug(angular): required+nullable query params are silently dropped without a paramsSerializer
## Summary

For Angular `HttpClient` services and `httpResource` functions, a query parameter that is both `required: true` and nullable per the OpenAPI schema is **silently dropped from the request** whenever its runtime value is `null` — unless a custom `paramsSerializer` mutator is configured. This can produce a request the server doesn't recognize as complete.

## Reproduction (on current master)

`samples/angular-app/src/api/http-client/pets/pets.service.ts` (generated from `tests/specifications/petstore.yaml`'s `searchPets` operation, whose `SearchPetsParams` declares `requirednullableString`/`requirednullableStringTwo` as required + nullable):

```ts
searchPets<TData = Pets>(
  params: SearchPetsParams,
  version: number = 1,
  options?: HttpClientObserveOptions,
): Observable<TData | HttpEvent<TData> | AngularHttpResponse<TData>> {
  const filteredParams = filterParams(
    { ...params, ...options?.params },
    new Set<string>(['requirednullableString', 'requirednullableStringTwo']),
  );
  ...
```

`filterParams` is called with only 2 arguments, so `preserveRequiredNullables` defaults to `false`. When `params.requirednullableString === null`, the key is dropped from `filteredParams` entirely — the request goes out without it, even though the parameter is documented as required.

## Root cause

`preserveRequiredNullables` is only ever set to `true` when a `paramsSerializer` mutator is configured:

- `packages/core/src/generators/options.ts`, `generateAxiosOptions` — all three `isAngular` param-building branches (`!isRequestOptions` IIFE, and the `isRequestOptions` `angularParamsRef`/`paramsSerializer`/plain branches) gate on `!!paramsSerializer` or unconditional `true` **only inside the `paramsSerializer` branch**; the plain (no-serializer) branch never passes it.
- `packages/angular/src/http-resource.ts:491` — same gating (`preserveRequiredNullables: !!paramsSerializer`), so `httpResource` functions have the identical gap.

This is not a regression — it has been the design since the feature was introduced in 667236784 (`feat(angular): preserve required nullable query params`). The reason: Angular's own `HttpClient`/`httpResource` `params` option type (`@angular/common/http`, e.g. `RequestOptions.params` and `HttpResourceRequest.params`) is:

```ts
params?: HttpParams | Record<string, string | number | boolean | ReadonlyArray<string | number | boolean>>;
```

— it structurally excludes `null`. Preserving a literal `null` in the filtered params object is only type-safe when a custom `paramsSerializer` consumes and converts it before the result reaches Angular's typed `params:` field; without a serializer, the value has nowhere valid to go.

**Note for reviewers**: this means the naive fix of adding `true` as a 3rd argument to the `filterParams(...)` call in the plain path (as suggested by an automated review comment on #3707) does not actually work — `filterParams(..., true)` returns `Record<string, AngularHttpParamValueWithNullable>` (includes `null`), which is not assignable to Angular's `params:` field and would fail to compile.

## Proposed fix

Don't try to preserve the literal `null` on the no-serializer path — instead, when a required-nullable key's value is `null`, emit an **empty string** (`''`) instead of dropping the key. This:

- needs no type widening (`''` is already a valid `AngularHttpParamValue`, no `| null` needed, no overload split for this case);
- is accepted by Angular's `HttpParams` natively, encoding as `?key=` in the request;
- preserves the key's presence (closer to the "required" contract) instead of silently omitting it, without misrepresenting `null` as some other value.

Concretely, in `packages/core/src/generators/options.ts`'s `getAngularFilteredParamsHelperBody()` (and the non-shared-helper IIFE twin, `getAngularFilteredParamsExpression`), add an unconditional branch (no `preserveRequiredNullables` gate needed, since `''` requires no type change):

```ts
} else if (value === null && requiredNullableParamKeys.has(key)) {
  filteredParams[key] = '';
}
```

placed before the existing `preserveRequiredNullables`-gated null-preserving branch, so:
- no `paramsSerializer` → required-nullable `null` becomes `''` (key preserved, always-on, no config needed);
- `paramsSerializer` configured (`preserveRequiredNullables: true`) → unchanged, literal `null` still passed through to the serializer as today, since the serializer explicitly opted into richer null-handling.

This should also thread through `packages/angular/src/http-resource.ts` for consistency, since it hits the identical gap.

Open question for maintainer input: whether `''` is the right universal default, or whether this should be configurable (e.g. an opt-out for users who'd prefer the current drop-silently behavior) — flagging as a design decision rather than presupposing the answer.

## Affected areas

- `packages/core/src/generators/options.ts` (`getAngularFilteredParamsHelperBody`, `getAngularFilteredParamsExpression`, `generateAxiosOptions`)
- `packages/angular/src/http-client.ts` (consumer)
- `packages/angular/src/http-resource.ts:491` (consumer, same gap)

## Related issues

- #3104 (closed, fixed by #3168) — the original TypeScript compile error from preserving a literal `null` in `filterParams`' return type unconditionally. This is *why* `preserveRequiredNullables` exists as a gated, opt-in flag at all rather than always-on: #3168 introduced the overloaded `filterParams` signature (`AngularHttpParamValue` vs `AngularHttpParamValueWithNullable`) specifically to keep the no-serializer path type-safe. This issue is the direct consequence of that constraint — the fix must respect it, not reintroduce #3104.
- #3593 (closed, fixed by #3597) — documents the exact `preserveRequiredNullables === !!paramsSerializer` asymmetry from the opposite angle: a TS6133 unused-variable error because `requiredNullableParamKeys` was emitted even when the branch that reads it was stripped out (i.e. the no-serializer path). Confirms this gating is long-standing, deliberate, and has already surfaced adjacent bugs.
- #3326 (closed) — the same *shape* of bug for a different value category: object-typed query params are dropped by `filterParams` entirely unless a `paramsSerializer`/`paramsFilter` is configured to consume the raw value. Established the precedent this issue follows (serializer-gated passthrough for values `HttpParams` can't natively represent) — the proposed empty-string fallback here is meant to compose with, not duplicate, that passthrough contract.
- #3192 (closed) — same root theme (null query param serialization) on the `fetch` client, with the opposite failure mode: `null` is serialized as the literal string `"null"` instead of being dropped. Cross-client evidence that `null`-valued query params are a recurring gap across generators, worth keeping in mind if a future fix aims for consistent semantics rather than an Angular-only patch.

Feature history for `preserveRequiredNullables`/`requiredNullableParamKeys`: 667236784, e2d3c2e92, 4acceae4e (original feature), 7b8084442 (#3168, the type-safety fix above), 5440a4c65 (#3597, the TS6133 fix above).

Surfaced via automated review on #3707 (`output.artifacts`, unrelated feature — `output.artifacts` only emits barrels from already-generated files and cannot fix generator-level behavior).

the-ult:
## Summary

Fixes #3712.

For Angular `HttpClient` services and `httpResource` functions, a query parameter that is both `required: true` and nullable per the OpenAPI schema was **silently dropped from the request** whenever its runtime value was `null` — unless a custom `paramsSerializer` mutator was configured. The request then went out without a parameter the spec documents as required.

## Root cause

`preserveRequiredNullables` was only ever passed as `true` in the `paramsSerializer` branches (`packages/core/src/generators/options.ts` and the `httpResource` builder). That gating exists for a real reason: Angular's `params` type (`HttpParams | Record<string, string | number | boolean | ReadonlyArray<...>>`) structurally excludes `null`, so a literal `null` is only type-safe when a serializer consumes it before it reaches Angular. The naive fix of passing `true` unconditionally does not compile (see the analysis in #3712).

## Fix

- The shared/emitted `filterParams` helper now emits `filteredParams[key] = preserveRequiredNullables ? null : '';` for required-nullable `null` values: **without** a serializer, `null` is sent as an empty string (`?key=`) so the key still reaches the wire; **with** a serializer, the literal `null` is preserved for the serializer to encode (unchanged behavior).
- The inline IIFE generator now emits the `requiredNullableParamKeys` `Set` + its reading branch whenever required-nullable keys exist (previously only when a serializer was configured), while preserving the #3593 guarantee that the `Set` is only declared when the branch reading it is emitted (no TS6133).
- The `''` assignment fits the existing non-nullable overload of `filterParams`, so the four overloads and `filteredParamValueType` are unchanged — no type widening.

## Tests

- 3 new unit tests in `packages/core/src/generators/options.test.ts` (empty-string fallback on the IIFE path, branch omission when no required-nullable keys, shared-helper body).
- New fixture spec `tests/specifications/issue-3712.yaml` + three `tests/configs/angular.config.ts` entries (`issue3712`, `issue3712HttpResource`, `issue3712Serializer`, modeled on the issue-3326 pattern) with `tests/api-generation.spec.ts` assertions for all three variants.
- New runtime proof: `samples/angular-app/src/app/required-nullable-params.spec.ts` uses `HttpTestingController` against the generated no-serializer `PetsService` and asserts `req.params.get('requirednullableString') === ''` — i.e. the key actually reaches the wire.
- All affected snapshots/samples regenerated (`tests/__snapshots__/angular/**`, `angular-query/**`, `default/issue-2998`, `samples/angular-app`, `samples/angular-query`) — every diff is the same expected helper-body hunk.

## Docs

`docs/content/docs/guides/angular.mdx` "Query parameter filtering": documents the empty-string fallback, notes that `override.paramsFilter` replaces this fallback (escape hatch back to drop-silently semantics), and the reference `default-filter-behaviour.ts` snippet now mirrors the emitted ternary exactly.

## Backwards compatibility

- Serializer path: unchanged (literal `null` still preserved).
- No new config surface. Users who want the old dropping behavior can use `override.paramsFilter`.
- Behavior change is scoped to a case that previously violated the OpenAPI contract (required param missing from the request).

## Validation

`bun run build:release` ✅ · `bun run lint` ✅ · core 2105/2105, angular 222/222, query 140/140 ✅ · snapshots 5313 tests, update + verify both green ✅ · angular-app `ng test` 17/17 incl. the new wire-level spec ✅ (one pre-existing macOS-only `/tmp` realpath flake in `packages/orval` options.test.ts, unrelated — package untouched).

## Related issues

- #3104 / #3168 — why `preserveRequiredNullables` is serializer-gated (literal `null` doesn't compile otherwise)
- #3593 / #3597 — TS6133 unused-`Set` guarantee this change preserves
- #3326 — object-param passthrough precedent; fixture pattern reused here
- #3192 — fetch client serialized `null` as literal `"null"` and it was treated as a bug — evidence against the literal-`"null"` alternative
- #3707 — the PR whose review surfaced this bug

## Summary by CodeRabbit

* **Bug Fixes**
  * Required+nullable Angular query parameters are now retained when explicitly set to `null`.
  * When no custom `paramsSerializer` is used, `null` is sent as an empty value (`?key=`); when a serializer is configured, the literal `null` is preserved.
  * Consistent behavior across generated clients, HTTP resources, and request modes.
* **Documentation**
  * Updated Angular docs to clarify the `required` + `nullable` `null` handling and how filtering overrides affect it.
* **Tests**
  * Added/updated coverage for request serialization and generated outputs for the relevant scenarios.


melloware (dismissed):

melloware:
Merge Conflicts  here too

melloware (approved):

PR #3721, opened 2026-07-13
fix(core): skip Required<Pick> for required keys missing from properties

### LINKED ISSUE #3722: Invalid Required<Pick> emitted when required property is missing from properties
## Description

Orval 8.21 can emit invalid TypeScript when a referenced schema lists a property in `required` but omits it from `properties`, and another schema references it via `nullable: true` + `allOf: [$ref]`.

### Example

```yaml
TagMetadataItem:
  type: object
  required:
    - tagId
    - label
    - color
  properties:
    id:
      type: integer
    label:
      type: string
    color:
      type: string

PetTagInfo:
  properties:
    tagMetadata:
      nullable: true
      allOf:
        - $ref: '#/components/schemas/TagMetadataItem'
```

### Generated (broken)

```ts
tagMetadata: TagMetadataItem & Required<Pick<TagMetadataItem, 'tagId'>> | null;
```

TypeScript fails with:

```
TS2344: Type '"tagId"' does not satisfy the constraint 'keyof TagMetadataItem'.
```

### Expected

```ts
tagMetadata: TagMetadataItem | null;
```

`Required<Pick>` should only be applied for keys that exist on subschema properties.

## Context

This was hit in production OpenAPI specs where `FieldId` appears in `required` but only `Id` is defined in `properties`.

Hypenate:
Fixes #3722

## Summary

- Fixes invalid TypeScript output in Orval 8.21 when a referenced schema lists a property in `required` but omits it from `properties`
- Orval was emitting types like `T & Required<Pick<T, 'tagId'>>`, which fails compilation with TS2344
- Only apply `Required<Pick>` for keys that exist on subschema properties
- Adds a petstore-based repro (`petstore-nullable-allof-required.yaml`) plus unit and integration regression tests

## Repro

Given a schema where `TagMetadataItem` requires `tagId` but only defines `id` in `properties`, and `PetTagInfo.tagMetadata` is `nullable: true` + `allOf: [$ref]`, Orval 8.21 generated:

```ts
tagMetadata: TagMetadataItem & Required<Pick<TagMetadataItem, 'tagId'>> | null;
```

After this fix:

```ts
tagMetadata: TagMetadataItem | null;
```

## Test plan

- [x] `vitest run src/getters/combine.test.ts` in `packages/core`
- [x] `bun run build`
- [x] `bun run generate:default -- --project petstore-nullable-allof-required`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed generated TypeScript types for composed schemas with required properties.
  * Prevented invalid required-property wrappers and phantom fields when keys are unavailable on the composed type.
  * Preserved correct handling of nullable `allOf` schema compositions.

* **Tests**
  * Added coverage for nested required properties and generated output involving tag metadata.
  * Added an OpenAPI fixture covering the corrected schema-generation scenario.


Hypenate:
@wadakatu We've tried to upgrade Orval from 8.18.0 to 8.21.0 but it breaks the build, this should fix the issue.

packages/core/src/getters/combine.ts
```diff
@@ -188,8 +188,11 @@ function combineValues({
...
         })(),
     );
-    if (overrideRequiredProperties.length > 0) {
-      return `${joined} & Required<Pick<${joined}, '${overrideRequiredProperties.join("' | '")}'>>`;
+    const pickableRequiredProperties = overrideRequiredProperties.filter(
+      (prop) => resolvedData.allProperties.includes(prop),
+    );
+    if (pickableRequiredProperties.length > 0) {
+      return `${joined} & Required<Pick<${joined}, '${pickableRequiredProperties.join("' | '")}'>>`;
     }
```
wadakatu, line 209: The new guard also drops keys that **do** exist on the member's type, when the member is itself a composition. `allProperties` only collects each member's top-level `properties` keys, so a required override pointing one `$ref` level deeper is silently lost:

```yaml
Base:        { type: object, properties: { baseProp: { type: string } } }
MidWrapper:  { allOf: [ { $ref: '#/components/schemas/Base' } ] }  # no top-level properties
WrappedItem: { type: object, required: [baseProp], allOf: [ { $ref: '#/components/schemas/MidWrapper' } ] }
```

- master: `export type WrappedItem = MidWrapper & Required<Pick<MidWrapper, 'baseProp'>>;` — valid TS (`keyof` resolves through the alias; compiles under `--strict`)
- this PR: `export type WrappedItem = MidWrapper;` — the override is silently gone, `{}` becomes assignable

One way to keep both fixes: emit the unresolved keys behind an `Extract` guard instead of dropping them. Ghost keys collapse to a no-op (`Extract<keyof T, 'tagId'>` = `never` → `Required<Pick<T, never>>` = `{}`, so #3722 still compiles), while real-but-nested keys stay enforced:

```ts
const pickable = overrideRequiredProperties.filter((p) =>
  resolvedData.allProperties.includes(p),
);
const unresolved = overrideRequiredProperties.filter(
  (p) => !resolvedData.allProperties.includes(p),
);
let result = joined;
if (pickable.length > 0) {
  result = `${result} & Required<Pick<${joined}, '${pickable.join("' | '")}'>>`;
}
if (unresolved.length > 0) {
  result = `${result} & Required<Pick<${joined}, Extract<keyof (${joined}), '${unresolved.join("' | '")}'>>>`;
}
return result;
```

Verified both behaviors with `tsc --strict`; resolvable keys keep the plain `Pick` form, so existing snapshots don't churn.

tests/specifications/issue-3722.yaml
```diff
@@ -0,0 +1,61 @@
+openapi: 3.1.0
```
wadakatu, line 1: nit: regression fixtures in `tests/specifications/` follow the `issue-NNNN.yaml` naming (50 existing) — `issue-3722.yaml` would match.

wadakatu (commented):
Confirmed the fix resolves the #3722 repro (built the branch, regenerated, `tsc` clean on the previously-broken output). Found one shape where the new guard over-filters — details inline.

melloware (approved):

PR #3723, opened 2026-07-14
fix(core): gate external $ref resolution behind explicit allow-list

aqeelat:
## Summary

Adds an explicit allow-list for external OpenAPI `$ref` resolution.

External refs are now blocked by default. Users can opt in with `input.parserOptions.externalRefs.allow`, including `['*']` for the previous allow-all behavior. Wildcard mode emits a warning for external documents referenced by the top-level spec.

## Changes

- Gate external file and URL `$ref` resolution behind `externalRefs.allow`
- Preserve existing header handling for protected remote specs
- Support relative allow-list entries for remote specs
- Update test configs that intentionally use external refs
- Document the new `input.parserOptions.externalRefs.allow` option

## Validation

- `bun vitest run packages/orval/src/import-specs.test.ts`
- `bun run lint`
- `bun run vp run -F orval-tests generate-api`

`bun run test` currently fails on an existing macOS `/tmp` vs `/private/tmp` path assertion in `packages/orval/src/utils/options.test.ts`.

## Summary by CodeRabbit

* **New Features**
  * Added `parserOptions.externalRefs.allow` to control which external OpenAPI `$ref` documents/URLs may be resolved (blocked by default).
  * When allowing all external refs, orval warns and documents the security implications; remote refs honor configured request headers.
  * Exported URL request header helper for public use.
* **Bug Fixes**
  * Improved Angular query-parameter filtering by using a safer property-existence check.
* **Documentation**
  * Documented external `$ref` allow-list behavior and warning/security callouts.

### DIFF
--- a/packages/orval/src/import-specs.test.ts
+++ b/packages/orval/src/import-specs.test.ts
@@ -522,6 +522,115 @@ describe('externalRefs', () => {
       await rm(workspace, { recursive: true, force: true });
     }
   });
+
+  it('should use configured headers when loading a remote top-level spec', async () => {
+    const remoteTarget = 'https://api.example.com/openapi.json';
+    const spec = {
+      openapi: '3.0.2',
+      info: { title: 'remote', version: '1.0' },
+      paths: {},
+    };
+    const fetchMock = vi.fn<typeof fetch>(async (_input, init) => {
+      expect(new Headers(init?.headers).get('Authorization')).toBe(
+        'Bearer token',
+      );
+      if (init?.method === 'HEAD') {
+        return new Response(undefined, { status: 200 });
+      }
+      return new Response(JSON.stringify(spec), { status: 200 });
+    });
+
+    vi.stubGlobal('fetch', fetchMock as typeof fetch);
+    try {
+      const normalizedOptions = await normalizeOptions(
+        {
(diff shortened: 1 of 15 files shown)


packages/orval/src/import-specs.test.ts
```diff
@@ -377,6 +377,7 @@ describe('validation', () => {
           output: { target: '' },
           input: {
             target: specPath,
+            parserOptions: { externalRefs: { allow: ['refs.yaml'] } },
```
melloware, line 380: ParserOptions is redundant should it just be "parser". Since you are in an options object it's implied

melloware (changes requested):

packages/orval/src/import-specs.test.ts
aqeelat: This is a pre-existing field in the config api. Renaming it to `parser` would be a breaking change. If you want, we can do it in a separate PR with a deprecation alias.

melloware: NO i didn't realize it existed!  All good!

melloware (approved):

melloware:
I have marked this ticket breaking change and security

Issue #1939, opened 2025-02-25
[Feature] Improve date handling with `useDates`


ghost:
When you enable the `useDates` option you yourself are responsible for finding dates in response objects and processing them accordingly.

The documentation recommends the use [of a Regexp.](https://orval.dev/reference/configuration/output#usedates)

This is incredibly error prone because users might insert strings that look like dates anywhere such as in the user name field upon registration.

You could also generate zod validations and use those to parse the dates. However that is a manual process. Orval does not produce information that allows me to automatically associate an API client function with a validation function at runtime.

**What Is to Be Done?**

Orval should provide a function for every request schema and a function for every response schema that will transform dates according to what is declared in the schema.

Later we could allow users to customize these functions in order to use a third party date library or to automatically process altogether custom types.

melloware:
I  think this is a duplicate of this request: https://github.com/orval-labs/orval/issues/1432

melloware:
OK marked as enhancement and will close that other ticket and point them here.

alario-tang:
+1 for this feature!

  Our use case: We use **Temporal API** (`temporal-polyfill`) instead of `Date`, and want to automatically parse `format: date-time` fields into `Temporal.Instant`.

  Currently we work around this with a post-processing script that replaces generated Zod schemas with our custom codec. It works but is fragile.

  This feature combined with #1383 (`customStringFormatResolver` / `formatOverride`) would solve this elegantly. For Zod client, something like:

  ```typescript
  override: {
    zod: {
      formatOverride: {
        'date-time': {
          schema: 'isoDatetimeCodec',
          import: { name: 'isoDatetimeCodec', path: './zod-codec' },
        },
      },
    },
  }

  Would love to see this land!

sami286:
We hit the same problem at scale (Axios + react-query + custom mutator): useDates is types-only, and the documented handleDates interceptor is O(entire JSON tree) + regex on every string.

**Proposal aligned with this issue:**
- Generate deserialize{Schema}() for each component schema (only format: date|date-time fields).
- Generate deserialize{Operation}Response() per operation.
- Append .then(deserialize…) in generated client functions (works with custom mutators).
- Mirror with serialize{Operation}Request() for bodies/params.
- Opt-in via override.useDatesDeserialize: true.

Perf: O(date fields in schema) vs O(all properties). No-op when a response has no dates.
Happy to contribute an MVP (responses + plain objects/arrays/$ref) if this direction works.

[...]

PR #3731, opened 2026-07-15
feat(core): useDatesTransform option and response date deserializer generator

sami286:
## What

Implements the first half of #1939: an opt-in `override.useDatesTransform` flag (implies `useDates`) plus a pure core module (`packages/core/src/generators/date-transform.ts`) that generates per-operation `deserialize<Op>Response` functions converting **only schema-declared** `format: date` / `format: date-time` fields in place.

Example of what the emitter produces (wired into clients by the follow-up PR):

```ts
const deserializeGetOrderDetailsResponse = (data: OrderDetails): OrderDetails => {
  if (data == null) return data;
  data.placedAt = new Date(data.placedAt);
  if (data.deliveredAt != null) data.deliveredAt = new Date(data.deliveredAt);
  for (let i0 = 0; i0 < data.events.length; i0++) {
    const item0 = data.events[i0];
    item0.createdAt = new Date(item0.createdAt);
  }
  return data;
};
```

## Covered schema shapes

- guarded assignments for optional and nullable properties
- arrays: index loops; object items hoisted into a `const` so TypeScript null-narrowing survives variable-indexed access; nullable elements guarded
- `$ref` with path-scoped cycle guards (repeated sibling refs each convert; genuine cycles are cut)
- `allOf` merged with sibling `properties`
- discriminated `oneOf`/`anyOf` (explicit `discriminator.mapping`): `switch` on the discriminator property with per-variant conversion

Everything else deliberately **emits nothing** rather than wrong code: unions without a discriminator mapping, `additionalProperties`-only date maps, multi-status success shapes, blob/non-JSON responses, date-free subtrees.

## Why

`useDates` is types-only; the documented runtime workaround is a regex interceptor that walks the entire response tree and can convert date-looking strings that aren't schema dates. This makes the schema the single source of truth at O(date fields) cost, with no interceptor and no runtime dependency.

## Notes for review

- This PR introduces the generator without wiring it into any client — the follow-up PR (query wiring + end-to-end fixture + docs) consumes it and serves as its integration test. Split kept small for reviewability.
- Zero behavior change for any existing config: the flag defaults to `false` and nothing consumes the module yet.
- Naming (`useDatesTransform` vs `useDatesDeserialize` vs a single `useDates: false | true | 'transform'` enum) — see the discussion on #1939; happy to rename.

Refs #1939

## Summary by CodeRabbit

* **New Features**
  * Added the `useDatesTransform` option to convert OpenAPI `date` and `date-time` response fields into runtime `Date` values.
  * Supports nested objects, arrays, nullable and optional fields, referenced schemas, and discriminator-based unions.
  * Enabling `useDatesTransform` automatically enables the related date-handling behavior.
* **Tests**
  * Added comprehensive coverage for date transformation, response deserialization, unsupported response scenarios, and option defaults.


melloware:
@wadakatu @aqeelat what do you guys think?

melloware:
pinging @wadakatu and @aqeelat again for opinions

wadakatu:
@melloware 

Sorry, I missed your comment. I’ll take a look at the PR once I get home.

melloware:
thank you @wadakatu much appreciated!

wadakatu:
@melloware Overall, I like the schema-driven approach. It should be more predictable and efficient than recursively scanning the whole response.

Before merging, though, I think it needs another correctness pass: some generated outputs do not type-check (`readOnly` properties and `allOf`-wrapped array items), and some valid schemas are mishandled, including component-name discriminator mappings, recursive schemas, and OAS 3.1 nullable dates. A compile-level integration test would catch most of these cases.

I do not see a major performance or security blocker. Runtime work is limited to schema-known paths, although large arrays still scale with their element count and per-operation inlining may increase generated code size. For hardening, schema-derived literals should be escaped consistently, ideally with `JSON.stringify`.

packages/orval/src/utils/options.ts
```diff
@@ -779,7 +779,10 @@ export async function normalizeOptions(
               }
             : {}),
         },
-        useDates: outputOptions.override?.useDates ?? false,
+        useDates:
+          (outputOptions.override?.useDates ?? false) ||
+          (outputOptions.override?.useDatesTransform ?? false),
+        useDatesTransform: outputOptions.override?.useDatesTransform ?? false,
```
melloware, line 785: Don't we need to add this to the docs in `/docs/output.mdx`?

melloware (changes requested):

packages/orval/src/utils/options.ts
sami286: These are added in #3732 

Moving the docs here means that if #3731 merges and releases before #3732, the docs would describe `.then(deserializeXxxResponse)` conversion that isn't wired yet.

melloware: OK

melloware (approved):

PR #3740, opened 2026-07-18
fix(fetch): infer request options type from mutator

kijk2869:
When a mutator is used with `fetch` client, the generated `options` parameter now infers its type from the mutator's second argument instead of always using `RequestInit`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TypeScript typings for generated request functions so the `options` parameter is inferred from the configured request handler rather than defaulting to `RequestInit`.
  * Updated sample request and hook factory signatures so `options` aligns with the second parameter type expected by the underlying fetch utilities (including hook-based mutators).

### DIFF
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -66,11 +66,11 @@ const getRequestOptionsType = (mutator?: GeneratorMutator) => {
   }
 
   if (mutator.hasSecondArg && !mutator.isHook) {
-    return `options?: SecondParameter<typeof ${mutator.name}>`;
+    return `options?: Parameters<typeof ${mutator.name}>[1]`;
   }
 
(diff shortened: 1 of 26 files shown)


melloware (changes requested):
Looks like the build is failing

kijk2869:
> Looks like the build is failing

I've checked the workflow, but it seems to be failing to compare snapshots.
This is the intended change point because it changes the generated fetch client.
Should I update snapshots?

melloware:
Yes the snapshots show you what has changed so your expected results can be verified or whether you have introduced a new issue!

melloware:
@soartec-lab can you please review this one?

melloware (approved):

PR #3741, opened 2026-07-18
fix(core): dotted path parameter names generate malformed request

### LINKED ISSUE #3735: Dotted path parameter names generate malformed request URLs
## Description

Orval generates malformed request URLs when an OpenAPI/Swagger path parameter name contains a dot, such as `{scope.id}`. This shape is produced by grpc-gateway when an HTTP binding references a nested protobuf request field.

Generation succeeds without warnings, but the generated URL omits both the opening brace and the function parameter. The malformed URL is only discovered at runtime.

## Output client

`fetch`. The same core route parser also affects `react-query` and other generated clients.

```js
export default {
  api: {
    input: {
      target: "./openapi.yaml",
    },
    output: {
      target: "./generated.ts",
      client: "fetch",
    },
  },
};
```

Run:

```sh
npx orval@8.22.0 --config ./orval.config.mjs --fail-on-warnings
```

## Environment

```text
System:
  OS: macOS 26.5.2
  CPU: arm64 Apple M5 Pro
  Shell: zsh 5.9
Binaries:
  Node: 22.23.0
  npm: 10.9.8
  pnpm: 11.7.0
Package:
  orval: 8.22.0
```

## Expected behavior

Orval should recognize `scope.id` as the required path parameter, sanitize it to a valid TypeScript identifier, and interpolate it into the generated URL. For example:

```ts
export const getResourceGetResourceUrl = (scopeId: string) => {
  return `/api/v1/resources/${scopeId}`;
};
```

## Actual behavior

Orval generates a zero-argument URL function containing malformed literal text:

```ts
export const getResourceGetResourceUrl = () => {
  return `/api/v1/resources/scope.id}`;
};
```

The CLI exits successfully, including with `--fail-on-warnings`.

## OpenAPI document (minimal)

The document validates successfully with `npx @scalar/cli document validate ./openapi.yaml`.

```yaml
swagger: "2.0"
info:
  title: Dotted path parameter reproduction
  version: 0.0.0
paths:
  /api/v1/resources/{scope.id}:
    get:
      operationId: Resource_GetResource
      parameters:
        - name: scope.id
          in: path
          required: true
          type: string
      responses:
        "200":
          description: OK
```

## Additional context

The current route parser only recognizes `\w`, `*`, `_`, and `-` inside braces in both `hasParam` and `getRoutePath`:

https://github.com/orval-labs/orval/blob/v8.22.0/packages/core/src/getters/route.ts#L34-L87

The later `sanitize(..., { dot: true })` call suggests dotted input was anticipated, but the parser never captures the complete dotted name.

PR #322 previously added `_` and `-` support to URL parameter names, but did not add `.`. Issue #3703 is related route-generation architecture work, but concerns template-literal escaping rather than dotted path parameter names.

A working local workaround is `input.override.transformer`: rewrite `{scope.id}` and its matching `in: path` parameter to a simple identifier only in Orval's in-memory specification. The source protobuf/OpenAPI contract can then remain unchanged.

daugvinasr:
Closes: #3735

Refactor replaces three duplicated recursive-regex route parsers with shared helpers, also fixes dotted path parameter names generating malformed requests #3735

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved generated client handling for dotted, underscored, dashed, wildcard, and nested path parameters.
  - Prevented malformed URLs and incorrect parameter references in generated clients.
  - Preserved template expressions and literal braces when converting routes.
  - Normalized routes that omit a leading slash.

- **New Features**
  - Added consistent route conversion for Hono and MSW integrations.
  - Added safe property access for parameter names requiring bracket notation.

- **Tests**
  - Expanded coverage for route formatting, parameter matching, escaping, collisions, and missing parameters.


melloware:
@daugvinasr just let us know when its ready for review

daugvinasr:
@melloware should be good to go

melloware (approved):

PR #3751, opened 2026-07-22
fix(core): safely collect nested allOf properties

wadakatu:
## Summary

- keep top-level properties on `anyOf`/`oneOf` schemas eligible for plain `Required<Pick>` when they are emitted into every union branch
- exclude non-object schema nodes and anyOf nullability that is propagated to a referenced wrapper from deep property-key collection
- parenthesize union-valued members before intersecting them with parent properties
- keep grouped nested unions eligible for plain `Required<Pick>` when nullability is not propagated to the referenced wrapper
- keep direct non-null scalar union members eligible for plain `Required<Pick>` after their intersection with parent properties
- keep nullable `$ref` anyOf members eligible for plain `Required<Pick>` because their nullability also stays inside the intersection
- keep standalone and sibling all-enum compositions on the compile-safe `Extract` path because parent properties are not guaranteed in every branch
- keep canonical nullable `oneOf: [object, null]` compositions on the `Extract` path because their object shortcut drops parent properties
- retain keys contributed by sibling `allOf` members when an all-enum or canonical nullable oneOf composition only makes the current node's own properties unsafe
- retain keys from inline allOf members with nullable anyOf because their null branch stays inside the intersection, while component `$ref` targets still use `Extract`
- retain nullable member keys when a non-null object sibling removes the null branch from the full allOf intersection
- retain nullable member keys when guaranteed properties on the parent itself remove the null branch
- add unit, generated snapshot, and compile-time regression coverage for all fourteen failure modes

## Root cause

The deep property-key collector added in #3749 returned before collecting a schema's own properties whenever `anyOf` or `oneOf` was present. Those top-level properties are intersected into every emitted branch, so treating them as unresolved could make `Extract<keyof T, K>` collapse to `never` under an `additionalProperties` index signature and silently make required keys optional.

The same guard only rejected null or multi-entry type arrays. A nested schema with `type: 'string'` or `type: ['string']` and a `properties` sibling was therefore treated as an object source, producing an invalid `Required<Pick<ScalarWrapper, 'id'>>` and TS2344.

A union-valued member also needs parentheses before intersecting it with parent properties. Without them, `A | null & Properties` leaves `A` outside the intersection because `&` binds more tightly than `|`. This change emits `(A | null) & Properties`.

The first guard for that precedence fix recursively treated every non-object value inside a nested union as if it propagated to the referenced wrapper. After grouping, however, a nested `null` is eliminated by its intersection with the parent properties and this depth of nullability is not propagated by reference resolution. Falling back to `Extract` there could again collapse the required key to `never`.

The narrowed guard still treated a direct non-null scalar union member as unsafe. Such a member also stays inside the grouped intersection (`string & Properties`); reference resolution only appends directly detected inline anyOf nullability to the imported wrapper name. The checks for a schema node's own output and for nullability propagated from an anyOf member are now separate, so scalar members and oneOf members use plain `Pick` while a directly nullable inline anyOf member on a component target retains the `Extract` fallback.

That propagation check must also match `resolveValue`'s exclusion of referenced anyOf members. Nullability on a `$ref` sibling is emitted inside the member value (`(Ref | null) & Properties`) but is not appended to the referenced wrapper. The collector now excludes `$ref` members from the propagated-nullability guard, preserving the required key after the intersection removes null.

All-enum compositions take a separate `combineValues` path: instead of intersecting parent properties into every enum branch, they emit the parent object as another union branch (or extract a combined enum under const generation). Those properties are therefore not guaranteed in `keyof` of the wrapper. The collector now uses the same resolved `isEnum` result as `combineSchemas` and keeps all-enum compositions on the `Extract` fallback, avoiding an invalid plain `Pick`.

An all-enum composition can also be a sibling of a non-enum composition. Checking `allOf ?? oneOf ?? anyOf` inspected only the first composition even though a sibling `anyOf`/`oneOf` is emitted through `resolvedValue` and has the same missing-key behavior. Each composition is now checked independently.

Canonical nullable oneOf objects take another special path in `getObject`: `oneOf: [inline object, null]` emits the inline object and null directly, without the parent properties. The collector now mirrors the shortcut's exact shape (one null member and one non-empty inline object member) before choosing `Extract`, while ordinary grouped oneOf members remain eligible for plain `Pick`.

That own-property fallback must not discard keys from sibling `allOf` members. For `ObjectBase & (InlineObject | null)`, the nullable branch is eliminated by the intersection and every surviving branch still contains `ObjectBase` keys. The same invariant holds when an all-enum sibling composition is intersected with `allOf`. The collector now separates unsafe own properties from conditions that invalidate the entire `allOf` intersection, so those descendant keys remain eligible for plain `Pick`.

Direct nullable anyOf only escapes that intersection when `resolveValue` imports the containing component through a `$ref` and appends `| null` to the alias. The same schema used as an inline allOf member emits `ObjectBase & (InlineObject | null)`, so the object-base keys remain guaranteed. The collector now tracks component-ref boundaries during recursive allOf traversal and applies the full descendant-key fallback only across those boundaries.

A nullable allOf member must also be evaluated in its parent intersection rather than in isolation. In `(Base | null) & { marker?: string }`, the object sibling makes the null branch `never`, so every surviving value still has the keys from `Base`. Parent traversal now recognizes siblings with guaranteed own properties and carries that proof into object-or-null members, without relaxing scalar, mixed-type, or enum branches.

The same null elimination occurs when `{ marker?: string }` comes from the parent schema's own properties rather than another allOf member. The collector now carries safely collected parent properties into every child as an additional proof that null cannot survive the full intersection.

## Validation

- `bunx vitest run` in `packages/core`: 54 files, 2,230 tests passed
- `bunx tsc --noEmit` in `packages/core`
- type-aware lint for the changed core files
- targeted generated snapshot/API tests: 53 passed
- generated TypeScript typecheck: all 16 clients passed
- targeted `regressions` and `regressions-oas31` project generation succeeded
- repository format check passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved generated TypeScript types for complex `allOf`, `anyOf`, and `oneOf` schemas.
  * Preserved required properties across nested compositions, references, unions, enums, and nullable schemas.
  * Improved handling of nullable values and non-object schema variants.
  * Prevented properties from being incorrectly inferred from union members.

* **Tests**
  * Added comprehensive regression coverage for schema composition and generated type behavior.


melloware (approved):

melloware:
Did this fix any open issues or just a continuation of the other fix?

PR #3785, opened 2026-07-30
feat(fetch): add serializeResponseHeaders option for serializable responses

### LINKED ISSUE #2363: Generated usePrefetch functions cause 'Only plain objects can be passed' error in Next.js Server Components
## Description

When using Orval-generated `usePrefetch` functions in Next.js Server Components, I encounter the following error:

This error occurs specifically when trying to use the generated `prefetchGetPostsQuery` function in a Server Component's `prefetchQuery` call.

## Steps to Reproduce

1. Configure Orval with the following settings:
   ```typescript
   // orval.config.ts

   export default defineConfig({
     client: {
       input: "../backend_app/openapi.json",
       output: {
         client: "react-query",
         target: "./src/client/index.ts",
         baseUrl: "${process.env.NEXT_PUBLIC_BACKEND_APP_URL}",
         httpClient: "fetch",
         mode: "split",
         override: {
           query: {
             useSuspenseQuery: true,
             usePrefetch: true,
           },
         },
       },
     },
   });
   ```

2. Generate client code using Orval
3. Use the generated `prefetchGetPostsQuery` function in a Next.js Server Component:

   ```typescript
   // app/post/page.tsx
   export default async function PostPage() {
     const queryClient = new QueryClient();
     await prefetchGetPostsQuery(queryClient);

     return (
       <HydrationBoundary state={dehydrate(queryClient)}>
         <Index />
       </HydrationBoundary>
     );
   }
   ```

## Expected Behavior

The `prefetchGetPostsQuery` function should work seamlessly in Next.js Server Components without causing serialization errors.

## Actual Behavior

The error occurs because the generated `prefetchGetPostsQuery` function returns a complex object that cannot be serialized and passed from Server Components to Client Components.

>Only plain objects can be passed to Client Components from Server Components. Headers objects are not supported.
  {data: ..., status: 200, headers: Headers}

Image

## Workaround

Currently, I'm using a manual workaround by creating a custom `queryFn`:

```typescript
await queryClient.prefetchQuery({
  queryKey: getGetPostsQueryKey(),
  queryFn: async () => {
    const { data, status } = await getPosts();
    return { data, status };
  },
});
```

ref https://github.com/mikan3rd/hono-next-example/pull/218

## Environment

- **Orval version**: 7.11.2
- **Next.js version**: 15.x
- **React Query version**: 5.x
- **Node.js version**: 18.x

## Additional Context

This issue is particularly problematic when using Next.js App Router with Server Components, as the generated prefetch functions are not compatible with the serialization requirements between Server and Client Components.

The issue affects the developer experience significantly, as developers need to manually create query functions instead of using the generated prefetch utilities.

## Related Issues

This might be related to the broader issue of function serialization in Next.js Server Components, but it specifically affects Orval-generated code.

## Proposed Solution

The generated `prefetchGetPostsQuery` function should return serializable data or provide a separate utility for Server Component usage that returns only the necessary data without complex object structures.

ostapondo:
Closes #2363.

## Problem

With `client: react-query` + `httpClient: fetch`, generated request functions return `{ data, status, headers: res.headers }`, where `headers` is a `Headers` instance. A generated `prefetch*` function puts that object in the query cache, and `dehydrate()` state containing it cannot cross the RSC serialization boundary:

> Only plain objects can be passed to Client Components from Server Components. Headers objects are not supported.

(The `useCallback` theory in the issue comments is not the cause — the non-hook prefetch helpers are already plain async functions. The failure comes from the cached response shape.)

## Change

New opt-in `override.fetch.serializeResponseHeaders` (default `false`). When enabled, generated clients type `headers` as `Record<string, string>` and build it from `res.headers.entries()` in all three response paths (json/text, blob, ndjson stream).

`set-cookie` is dropped rather than copied. Server-side `fetch` exposes it, and a dehydrated cache is embedded in the RSC payload, so copying it would ship an upstream session cookie to the browser. Header names are lowercased by `Headers` iteration, and repeated headers are joined with `, ` per the spec.

With the option off, generated output is unchanged — verified by regenerating every project in `tests/` and diffing the trees (zero differences outside the new test projects).

## Verification

Runtime tests in `tests/serialize-response-headers.spec.ts` drive the generated code against a stubbed `fetch`:

- `prefetchListPetsQuery(queryClient, ...)` → `dehydrate()` → `structuredClone(state)`. Without the option the cached entry is a live `Headers` instance and the assertion fails; with it the state clones cleanly.
- A response carrying `set-cookie` never surfaces it on the serialized object.

Generation tests pin the emitted source for the json, ndjson `stream`, blob and mutator configs. `tests/regressions/prefetch-serializable-headers.ts` pins the type and rejects the `Headers` API.

Every assertion was mutation-tested: removing the `set-cookie` filter and disabling serialization each fail a different, non-overlapping set of tests.

Full suite (5588 tests), generated-output typecheck for all 16 clients, type-aware lint and unit tests for the touched packages pass.

## Also in this PR

`normalizeOptions` spread the raw `override.fetch` after the normalized defaults, so an explicit `undefined` erased them. This affected all five fetch booleans, not just the new one; the spread now runs first, with a test covering it.

## Not covered

- The option is opt-in, so existing configs behave exactly as before until it is set.
- With a custom `mutator` it changes the declared type only — the mutator issues the request itself and must return headers in that shape, as it already must for `includeHttpResponseReturnType`.
- Only `headers` is converted: `blob` responses still return a `Blob`, and ndjson responses still return the raw `Response` under `stream`, so those remain non-serializable.
- Prefetch helpers generated for hook mutators (`use-prefetch-*`) remain hooks and still cannot run in Server Components.

## Summary by CodeRabbit

* **New Features**
  * Added an option to serialize fetch response headers as a plain, cloneable key-value object.
  * Header names are lowercased, repeated values are combined, and `set-cookie` headers are excluded.
  * Supports compatibility with Next.js Server Components and React Query data dehydration.
  * Added configuration guidance and examples for enabling the option.

* **Bug Fixes**
  * Preserved default fetch settings when optional overrides are unspecified.


ostapondo:
@coderabbitai resume

melloware (approved):

PR #3794, opened 2026-08-01
Fix/zod namespace import

### LINKED ISSUE #3792: hono: zod schemas use `import { z as zod }`, which defeats tree-shaking (+260K minified)
## Description

The `hono` client generates its zod schema file with a named import:

```ts
import { z as zod } from 'zod';
```

This pulls in zod's assembled `z` object, which transitively references every locale table. No bundler can eliminate them, so ~260K of translated error messages end up in the bundle even when only one schema is used.

The `zod` client, generating an identical schema body from the same spec in the same run, emits the namespace form instead:

```ts
import * as zod from 'zod';
```

That form tree-shakes correctly. Since orval already writes every call site as `zod.*`, the two are interchangeable — swapping the import in the hono output changes nothing else and produces identical runtime behaviour.

This matters most for hono, which is frequently deployed to serverless and edge runtimes where bundle size drives cold-start cost.

## Output client

hono

```js
export default {
  zodClient: {
    input: './spec.yaml',
    output: { mode: 'single', client: 'zod', target: './out/zod-client.ts' },
  },
  honoServer: {
    input: './spec.yaml',
    output: {
      mode: 'split',
      client: 'hono',
      target: './out/hono/api.ts',
      override: { hono: { handlers: './out/hono/handlers' } },
    },
  },
};
```

## Environment

```
  System:
    OS: macOS 15.7.7
    CPU: (12) arm64 Apple M3 Pro
    Shell: 5.9 - /bin/zsh
  npmPackages:
    orval: ^8.23.0 => 8.23.0
    zod: ^4.4.3 => 4.4.3
  Bundler:
    bun 1.3.14 (esbuild reproduces the same result)
```

## Expected behavior

`@orval/hono` emits `import * as zod from 'zod'`, matching what `@orval/zod` already does.

## Actual behavior

`out/hono/api.zod.ts`:

```ts
import { z as zod } from 'zod';
```

`out/zod-client.ts`, same run, same schema body:

```ts
import * as zod from 'zod';
```

Bundling a single `zod.object({ name: zod.string() })` from each, minified:

| import form | bundle | zod locale strings present |
|---|---|---|
| `import { z as zod } from 'zod'` | **328K** | 1503 |
| `import * as zod from 'zod'` | **68K** | 0 |

Both `safeParse` calls return the same result; validation and error messages are unaffected.

## OpenAPI document (minimal)

```yaml
openapi: 3.1.0
info:
  title: Repro
  version: 1.0.0
paths:
  /pets:
    get:
      operationId: listPets
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                type: object
                required: [name]
                properties:
                  name:
                    type: string
```

## Additional context

**Why the named form cannot be shaken.** `zod/v4/core/index.js` contains `export * as locales from '../locales/index.js'`. Importing the `z` binding forces that namespace object to be constructed, and neither bun nor esbuild can prove the locales are unused. Verified with both bundlers — 328K vs 319K, i.e. the same outcome, so this is not a bundler-specific quirk. Related upstream: colinhacks/zod#5561, colinhacks/zod#4369.

`override.zod.variant: 'mini'` does not help: it switches to `zod/mini`, but that entry point re-exports the same core and still bundles all 53 locale files (308K vs 328K).

**Real-world impact.** On a Hono-on-Bun service with three operations, applying the one-line swap as a post-generation step took the production bundle from **356K to 104K** with no behavioural change — validation still rejects bad input with 400 and error messages are intact.

**Workaround** for anyone hitting this before a fix, via `hooks.afterAllFilesWrite`:

```ts
const NAMED_ZOD_IMPORT = /import \{\s*z as zod\s*\} from (['"])zod\1;?/;

function zodNamespaceImport(paths = []) {
  for (const path of paths.filter((p) => p.endsWith('.ts'))) {
    const before = readFileSync(path, 'utf8');
    const after = before.replace(NAMED_ZOD_IMPORT, 'import * as zod from "zod";');
    if (after !== before) writeFileSync(path, after);
  }
}
```

Happy to open a PR if you'd like — the change looks like aligning the hono generator's zod import with whatever `@orval/zod` already emits.

Bielik20:
Fixes #3792.

Generated zod schema files used `import { z as zod } from 'zod'`, which pulls in zod's
assembled `z` object. That transitively references every locale table, and neither bun nor
esbuild can tree-shake it — so ~194K of translated error messages ship in every bundle.

`import * as zod` is a drop-in swap, since orval already writes every call site as `zod.*`.

### Changes

- Always emit the namespace form, dropping the `variant === 'mini'` special case in:
  - `packages/hono/src/index.ts`
  - `packages/mcp/src/index.ts`
  - `packages/orval/src/write-zod-specs.ts`
- Regenerated samples and snapshots (one import line each)
- Removed two comments describing the old form

### Impact

Bundling a single `zod.object({ name: zod.string() })`, minified:

| import | bundle | locale strings |
|---|---|---|
| `{ z as zod }` | 328K | 1503 |
| `* as zod` | 68K | 0 |

Same runtime behaviour — validation and error messages are unaffected. On a real
Hono-on-Bun service this took the production bundle from 356K to 104K.

### Notes

- This aligns the three generators with `@orval/zod` and `@orval/effect`, which already emit
  the namespace form via `namespaceImport`.
- It also removes the reason for the workaround noted at `write-zod-specs.ts:221`, where the
  two differing forms had to be reconciled when concatenated.
- The issue is titled `hono:`, but the same helper is duplicated in `mcp` and
  `write-zod-specs`. I fixed all three since it is one bug.
- I have split it into 2 commits: fix and test adjust for easier CR.

`build`, `test` (220), `test:snapshots` (5588), `typecheck`, `lint` and `format:check` all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized generated Zod schemas to use consistent namespace imports.
  * Preserved existing schema behavior and exported types across supported integrations.
  * Prevented unnecessary Zod imports when a path parameter is named `z`.

* **Tests**
  * Clarified coverage for import generation in this edge case.

### DIFF
--- a/packages/hono/src/index.ts
+++ b/packages/hono/src/index.ts
@@ -43,12 +43,12 @@ import {
 } from './handler-merge';
 import { getRoute } from './route';
 
+// Always a namespace import: `import { z as zod }` pulls in zod's assembled `z` object,
+// which transitively references every locale table and cannot be tree-shaken. Matches
+// what `@orval/zod` and `@orval/effect` already emit via `namespaceImport`.
 const getZodSchemaImportStatement = (
   variant: NormalizedOutputOptions['override']['zod']['variant'],
-) =>
-  variant === 'mini'
-    ? `import * as zod from '${getZodImportSource(variant)}';`
-    : `import { z as zod } from '${getZodImportSource(variant)}';`;
+) => `import * as zod from '${getZodImportSource(variant)}';`;
 
 // Warn at most once per run when the optional `typescript` peer is missing and a
 // non-`skip` strategy was requested, so the degraded behavior is never silent.
--- a/packages/mcp/src/index.ts
+++ b/packages/mcp/src/index.ts
@@ -28,12 +28,12 @@ import {
 import { generateClient, generateFetchHeader } from '@orval/fetch';
 import { generateZod, getZodImportSource } from '@orval/zod';
 
+// Always a namespace import: `import { z as zod }` pulls in zod's assembled `z` object,
+// which transitively references every locale table and cannot be tree-shaken. Matches
+// what `@orval/zod` and `@orval/effect` already emit via `namespaceImport`.
 const getZodSchemaImportStatement = (
   variant: NormalizedOutputOptions['override']['zod']['variant'],
-) =>
-  variant === 'mini'
-    ? `import * as zod from '${getZodImportSource(variant)}';`
-    : `import { z as zod } from '${getZodImportSource(variant)}';`;
+) => `import * as zod from '${getZodImportSource(variant)}';`;
 
 const getHeader = (
   option: false | ((info: OpenApiInfoObject) => string | string[]),
--- a/packages/orval/src/write-zod-specs.ts
+++ b/packages/orval/src/write-zod-specs.ts
@@ -58,10 +58,11 @@ type ZodSchemaFileToWrite = ZodSchemaFileEntry & {
   filePath: string;
 };
 
+// Always a namespace import: `import { z as zod }` pulls in zod's assembled `z` object,
+// which transitively references every locale table and cannot be tree-shaken. Matches
+// what `@orval/zod` and `@orval/effect` already emit via `namespaceImport`.
 const getZodSchemaImportStatement = (variant: ZodVariantOption) =>
-  variant === 'mini'
-    ? `import * as zod from '${getZodImportSource(variant)}';`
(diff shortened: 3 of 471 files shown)


melloware:
Running the tests now.

melloware (approved):

PR #3797, opened 2026-08-01
fix(orval): drop version from default generated header (#3788)

### LINKED ISSUE #3788: Default generated header includes the orval version and creates upgrade churn
### Description

The default generated file header includes the installed orval version. This causes every generated file to change whenever orval is upgraded, even when the generated API code is identical.

### Current behavior

The default header is:

```ts
Generated by orval v<version> 🍺
```

Upgrading orval therefore creates a repo-wide generated-file diff containing only a new version number.

Users can replace the full header with `output.override.header`, but that is an undiscoverable workaround for what is default behavior.

### Discussion

Possible directions:

- Remove the version from the default header while keeping the generated-file marker.
- Add an opt-in option such as `removeVersionFromHeader`.
- Document the existing custom-header callback as the supported way to omit it.

I think removing the version from the default header is preferable. It avoids unnecessary diffs without adding another option, and the installed generator version is already available from the lockfile or dependency manifest.

aqeelat:
## Summary

Removes the orval package version from the default generated-file header so upgrading orval no longer produces a repo-wide diff of identical generated code. The header changes from `Generated by orval v<x.y.z> 🍺` to `Generated by orval 🍺`; the generated-file marker is preserved and the OpenAPI spec version is still emitted on its own line.

## Changes

- `packages/orval/src/utils/options.ts` — drop `v${pkg.version}` from `getDefaultFilesHeader`.
- Remove `snapshot-version-check.yaml` and `snapshot-version-rerun.yaml`. These enforced the version string in snapshot headers and existed solely to manage the upgrade churn this PR eliminates. With the version gone they would false-fail on every PR, and the per-package snapshot tests already gate every byte of generated output.

## Context

Per discussion in #3788, @melloware endorsed both the default-header change ("the simplest fix") and removing the version-check workflow ("no real value ... relic of past times"). The installed orval version remains available from the lockfile/dependency manifest; users who want it in the header can use the existing `output.override.header` escape hatch.

## Test plan

- Regenerated all test snapshots and sample output (mechanical; one header line per file — `6442 insertions / 6442 deletions`).
- `orval` unit tests: 219 passed.
- `orval-tests` snapshot suite: 5525 passed.
- Samples snapshot suites: all passed.

Closes #3788


melloware:
I totally agree with this change the version just creates noise!

melloware (approved):

Issue #607, opened 2022-10-03
Bug: operationId "options" clashes with parameter name `options`, causing typescript errors


thraizz:
## What are the steps to reproduce this issue?

1. Setup a sample project with react + react-query/tanstack query + typescript where one of the endpoints has `operationId: "options"` or use the provided sample repo or the sample configuration
2. Generate the api 

## What happens?
If we check the generated api.ts, we see a variable name clash of the generated `export const options = ...` for the endpoint and the first parameter for`useOptions`, `options?:UseMutateOptions...`.

## What were you expecting to happen?

I expected orval to generate a valid name for endpoints with `operationId: "options"`.

## Any logs, error output, etc?

There are multiple typescript errors because of the name-clash, e.g. in `return options(petId, axiosOptions` it gives me:
```
TS2349: This expression is not callable.   Type '{ mutation?: UseMutationOptions<any, TError, { petId: number; }, TContext> | undefined; axios?: AxiosRequestConfig<any> | undefined; }' has no call signatures.
TS2722: Cannot invoke an object which is possibly 'undefined'.
```

…

## Any other comments?

Currently, i patch this by renaming the generated `export const options` and all references to `options1`, which fixes the clash between it and the parameter name.

I've created a sample repo for this: https://github.com/thraizz/orval-bug-operationId

The essence is the following setup:

openApiCatalog.json
```json
{
  "swagger": "2.0",
  "info": {
    "description": "This is a sample server Petstore server.  You can find out more about Swagger at [http://swagger.io](http://swagger.io) or on [irc.freenode.net, #swagger](http://swagger.io/irc/).  For this sample, you can use the api key `special-key` to test the authorization filters.",
    "version": "1.0.6",
    "title": "Swagger Petstore",
    "termsOfService": "http://swagger.io/terms/",
    "contact": { "email": "apiteam@swagger.io" },
    "license": {
      "name": "Apache 2.0",
      "url": "http://www.apache.org/licenses/LICENSE-2.0.html"
    }
  },
  "host": "petstore.swagger.io",
  "basePath": "/v2",
  "schemes": ["https", "http"],
  "paths": {
    "/pet/{petId}/options": {
      "post": {
        "tags": ["pet"],
        "summary": "get options",
        "operationId": "options",
        "consumes": ["multipart/form-data"],
        "produces": ["application/json"],
        "parameters": [
          {
            "name": "petId",
            "in": "path",
            "description": "ID of pet to update",
            "required": true,
            "type": "integer",
[...]

PR #3798, opened 2026-08-02
fix(core): avoid request body parameter shadowing the generated operation

### LINKED ISSUE #3784: Generated parameter shadows function name when request body schema matches operationId
### Description

When an endpoint's request body schema name, camelCased, equals the operationId, orval generates a function parameter that shadows the function itself. This causes `ReturnType<typeof fn>` to resolve against the parameter (a type) instead of the outer function, producing TS2344 and TS2349 errors in the generated output.

This is the same class of naming collision that was fixed for options in #607 and signal in #2786 — the body parameter variable just needs the same collision-avoidance treatment.

### Output client

react-query

```ts

export default defineConfig({
  api: {
    input: {
      target: './spec.yaml',
    },
    output: {
      target: './src/generated',
      client: 'react-query',
      mode: 'tags-split',
      override: {
        query: {
          version: 5,
          useQuery: true,
          signal: true,
        },
      },
    },
  },
});
```

### Environment

System:
  OS: macOS 26.3
  CPU: (14) arm64 Apple M4 Max
  Memory: 758.13 MB / 36.00 GB
  Shell: 5.9 - /bin/zsh
  Node: v24.14.1
npmPackages:
  orval: 8.23.0
  @tanstack/react-query: 5.101.4
  react: 18.3.1

### Expected behavior

When the request body schema name camelCases to the same string as the operationId (e.g., CreateItem → createItem === operationId createItem), orval should detect the collision and rename the body parameter variable to avoid shadowing (e.g., createItemBody or createItemData), producing valid TypeScript that compiles under strict: true.

### Actual behavior

Orval generates a parameter with the same name as the function, shadowing it. `ReturnType<typeof createItem>` resolves against the parameter type instead of the function, producing:

```
TS2344: Type 'CreateItem' does not satisfy the constraint '(...args: any) => any'.
TS2349: This expression is not callable. Type 'CreateItem' has no call signatures.
```

```yaml
openapi: 3.0.0
info:
  title: Minimal Repro
  version: 1.0.0
paths:
  /api/v1/items:
    post:
      operationId: createItem
      tags:
        - Items
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/CreateItem'
      responses:
        '200':
          description: Success
          content:
            application/json:
              schema:
components:
  schemas:
    CreateItem:
      type: object
      properties:
        name:
          type: string
      required:
        - name
    Item:
      type: object
      properties:
        id:
          type: string
        name:
          type: string

```

### Additional context

_No response_

ostapondo:
Fixes #3784.

When a request body schema's camel-cased name equals the operationId (schema `CreateItem`, operationId `createItem`), the generated parameter shadowed the operation function. `ReturnType<typeof createItem>` then resolved to the parameter type (TS2344/TS2349 under `strict`), and the react-query `queryFn` called the parameter instead of the function.

The fix renames the body parameter when it matches the operation name (`createItem` -> `createItemBody`), same collision-avoidance treatment as #607/#2786. Inline `FormData`/`URLSearchParams` code is rewritten to reference the renamed parameter, so multipart/url-encoded bodies with the same collision stay correct. The rename lives in core, so all clients are covered.

Verified with a regression spec (JSON body + multipart body, both colliding) that fails to compile and mistargets the call without the fix; regenerating all test targets changes no other snapshot, so non-colliding output is untouched.

Not covered: a schema named after some other generated helper (e.g. `GetCreateItemQueryKey`) could shadow that helper the same way; helper names are client-specific, so this change only handles the operation-name collision reported in the issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed generated API clients when a request-body name matches an operation name.
  * Prevented naming conflicts with parameters, headers, and other request fields.
  * Updated regular, multipart, and URL-encoded requests to use the corrected body name.

* **Tests**
  * Added regression coverage for standard and multipart request generation.


ostapondo:
Both points the check couldn't verify are covered:

- Strict compilation: the `Verify generated code can build` step of pr-checks type-checks every generated test output against `tests/tsconfig.json` (`strict: true`); it's green on both OS runners here. Without the fix that step fails on this output with the exact TS2344/TS2349 errors from the issue.
- The generated snapshots are committed under `tests/__snapshots__/react-query/body-schema-name-matches-operation-id/` and are part of the diff — the automated review just excludes generated files.

ostapondo:
@coderabbitai review

melloware (approved):

melloware (approved):

Issue #3750, opened 2026-07-21
`required` keys declared on the parent schema's own `properties` fall into the `Extract<keyof T, ...>` guard and lose requiredness with `additionalProperties` (regression in 8.22.0)


wadakatu:
### Description

Since v8.22.0 (#3721), `required` keys whose properties are declared on the composing schema's **own top-level `properties`**, while the `required` marker lives in a constraint-only `allOf` member (an overlay with `required` but no `properties`), are treated as unresolvable and wrapped in the `Extract<keyof T, ...>` guard:

```ts
Required<Pick<T, Extract<keyof T, 'id' | 'name'>>>
```

As in #3748, with `additionalProperties: true` the emitted intersection carries an index signature, `keyof T` collapses to `string | number`, `Extract<...>` evaluates to `never`, and the required override becomes a no-op — **properties the OpenAPI document declares as `required` silently become optional**. The generated code still compiles, so nothing fails in CI.

This is the sibling of #3748 (fixed by #3749): that fix resolves keys reached through `allOf` **members** (including nested `$ref` compositions), but the parent schema's own `properties` are still never fed into the resolvability check, so this shape remains misclassified on master with #3749 applied.

### Output client

fetch (the model emission is shared, so any client is affected)

### Configuration (orval.config)

```ts
import { defineConfig } from 'orval';

export default defineConfig({
  repro: {
    input: {
      target: './api.yaml',
    },
    output: {
      target: './generated/repro.ts',
      client: 'fetch',
    },
  },
});
```

### Environment

```
  System:
    OS: macOS 26.5.1
    CPU: (11) arm64 Apple M3 Pro
    Shell: 5.9 - /bin/zsh
  npmPackages:
    orval: 8.22.0 (also reproduced on current master with #3749 applied; typescript 6.0.3)
```

### Expected behavior

Same output as v8.20.0 — a plain `Required<Pick>`:

```ts
export type ItemDetail = { [key: string]: unknown } & {
  id?: string;
  name?: string;
  [key: string]: unknown;
} & Required<Pick<{ [key: string]: unknown } & {
  id?: string;
  name?: string;
  [key: string]: unknown;
}, 'id' | 'name'>>;
```

This is valid TS (`'id' | 'name'` satisfies `keyof T` because `keyof T` widens to `string | number`) and enforces requiredness:

```ts
declare const item: ItemDetail;
const id: string = item.id; // OK on 8.20.0
```

### Actual behavior

[...]

Issue #3663, opened 2026-06-28
Composite types do not combine required fields


FWest98:
### Description

In our API, we use the "JSON Merge Patch" standard for our `PATCH` endpoints. In this setup, the endpoint expects a "partial" model representation, where every (writable) model field is optional: only the provided fields will be updated. Of course, for `GET`/`POST`/`PUT`, there can still be normal required fields, as those still expect a complete model instance.

To prevent writing duplicate schemas in our OpenAPI specification, we split up our definition in two parts: a "partial" schema with all fields optional, and then a "required fields" overlay. This looks as follows:
```yaml
components:
  schemas:
    fooPartial:
      type: object
      properties:
        id:
          type: string
          readOnly: true
        name:
          type: string
        detail:
          type: string

    fooRequiredFields:
      required: [ "id", "name" ]

    foo:
      allOf:
        - $ref: '#/components/schemas/fooPartial'
        - $ref: '#/components/schemas/fooRequiredFields'
```

The `PATCH` endpoint will use the `fooPartial` schema, while all other endpoints use the `foo` schema. This works perfectly fine in the documentation generators we tried, where the `id` and `name` fields are correctly marked as required in the non-`PATCH` endpoints.

In Orval, however, this fails: Orval tries to make a distinct type for `fooRequiredFields`, but that has no properties defined so is empty, causing the `foo` type to still have all fields be optional:
```ts
export interface FooPartial {
    readonly id?: string;
    name ?: string;
    detail ?: string;
}
export interface FooRequiredFields {}
export type Foo = FooPartial & FooRequiredFields;
```

I have seen similar issues in the code generation elsewhere in our spec, where schema configuration affecting the same property in multiple parts of an `allOf` did not properly combine (e.g. `readOnly`). What would be the best approach to resolve this, without having to define two copies of all our schemas?

### Output client

fetch

### Configuration (orval.config)

```ts
    output: {
      mode: 'tags-split',
      target: './src/generated/',
      client: 'fetch',
      clean: true,
      override: {
        enumGenerationType: 'enum',
        useDates: true,
        useBigInt: true,
      },
      formatter: 'prettier',
      tsconfig: './tsconfig.json',
    },
```

### Environment

Orval 8.19.0

### OpenAPI document (minimal, if applicable)

See above

Ciantic:
[...]

Issue #910, opened 2023-08-08
MSW: nested objects with the same attribute names.


Cohey0727:
## What are the steps to reproduce this issue?

1. **Set Up the Environment**:
    - Install `orval` v6.17.0 on your machine.
    
2. **Provide the OpenAPI Definition**:
    Create an OpenAPI document in which the `parent` attribute appears multiple times in a nested structure.
    ```yaml
    openapi: "3.0.0"
    info:
      version: 1.0.0
      title: Swagger Petstore
      license:
        name: MIT
    servers:
      - url: http://petstore.swagger.io/v1
    paths:
      /pets/{petId}:
        get:
          summary: Info for a specific pet
          operationId: showPetById
          parameters:
            - name: petId
              in: path
              required: true
              description: The id of the pet to retrieve
              schema:
                type: string
          responses:
            "200":
              description: Expected response to a valid request
              content:
                application/json:
                  schema:
                    $ref: "#/components/schemas/Pet"
    components:
      schemas:
        Pet:
          type: object
          required:
            - id
            - name
          properties:
            id:
              type: integer
              format: int64
            name:
              type: string
            tag:
              type: string
            parent:
              $ref: "#/components/schemas/Parent"
        Parent:
          type: object
          required:
            - id
            - name
          properties:
            id:
              type: integer
              format: int64
            name:
              type: string
            tag:
              type: string
            parent:
              $ref: "#/components/schemas/GrandParent"
        GrandParent:
          type: object
          required:
            - id
            - name
            - parent
          properties:
            id:
              type: integer
              format: int64
            name:
              type: string
            tag:
              type: string
            parent:
              $ref: "#/components/schemas/GreatParent"
        GreatParent:
          type: object
          required:
            - id
            - name
            - parent
          properties:
            id:
              type: integer
              format: int64
            name:
              type: string
            tag:
              type: string
    ```

4. **Run Orval Code Generator**:
    run command with config `mock:true`.
    ```bash
    orval 
    ```

5. **Inspect the Generated MSW Mock**:
    ```typescript
    /**
     * Generated by orval v6.17.0 🍺
     * Do not edit manually.
     * Swagger Petstore
     * OpenAPI spec version: 1.0.0
     */
    import { rest } from "msw";
    import { faker } from "@faker-js/faker";
    
    export const getShowPetByIdMock = () => ({
      id: faker.datatype.number({ min: undefined, max: undefined }),
      name: faker.random.word(),
[...]

PR #3801, opened 2026-08-03
fix(mock): propagate required from $ref allOf members to overriding branches

federicobartoli:
When an allOf member narrows an inherited enum property to a single literal without re-listing it in its own required, the mock generator emitted faker.helpers.arrayElement([value, undefined]) for that property. The property stays required through the $ref'd base schema, so the generated mock fails to typecheck (Type 'undefined' is not assignable to the enum union).

Root cause: combineSchemasMock collected the allOf required union only from inline members and never dereferenced $ref members, so required fields declared on a referenced base were invisible to sibling override branches.

The fix dereferences $ref members when collecting the union, following $ref alias chains and nested allOf compositions with a cycle guard. Names collected from a nested composition cross its boundary only when that composition also declares the property, keeping mock requiredness aligned with the generated types (#3663) and preventing the #910 cyclic reference guard from flipping property omission to null. The union now also applies to sibling properties declared next to allOf, and constraint-only members ({ required: [...] }) count as well (#3750 semantics). With faker schemas: true, factory delegation is skipped when it would drop the composition-wide required union; the schema body is inlined instead.

Covered by 7 new unit tests in packages/mock and an e2e regression spec (tests/specifications/allof-narrowed-enum.yaml) exercising both the msw inline path and the faker schema-factory path, with the generated output part of the typechecked corpus. All pre-existing snapshots are unchanged.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved mock generation for OpenAPI schemas using composed definitions and references.
  - Required properties are now preserved across nested compositions, aliases, sibling schemas, and cyclic references.
  - Prevented undeclared nested fields from being incorrectly marked as required.
  - Preserved optional properties when no composition branch requires them.
  - Improved handling of narrowed enum properties in generated mock responses.

- **Tests**
  - Added comprehensive regression coverage for composed schemas, cyclic references, and validation-error responses.


melloware (approved):

melloware (changes requested):
Can you address all the AI feedback?

federicobartoli:
> Can you address all the AI feedback?

All AI feedback addressed. Copilot's point about the shallow delegation guard was fixed in 2e53962 (recursive target resolution), and CodeRabbit's shared-seen finding is fixed in 533d552fe (path-scoped cycle guard, with regression tests for the reported shape). All checks pass locally including the generated-output typecheck.

melloware (approved):

Issue #3411, opened 2026-05-21
isBlob is derived from content type only and disregards `format: binary`


jmorel:
### Description

The isBlob property of a response is derived from its content type only. It disregards the `format: binary` that could be declared on the response schema.

This means that Orval has to maintain a list of ALL content types that are binary. I'd argue that a shortlist of well known binary content types (such as zip, octet-stream, pdf etc.) does help with openapi specs that may not be perfectly formed, but we should also check for `{"type": "string", "format": "binary"}` in the response schema to decide whether to enable the isBlob property.

I ran into this issue while trying to serve a `.docx` export of a document (see the response object below)

If you agree, I'd be happy to make a PR. 

### Output client

axios

### Configuration (orval.config)

```ts
"responses": {
  "200": {
    "description": "Successful Response",
    "content": {
      "application/vnd.openxmlformats-officedocument.wordprocessingml.document": {
        "schema": {
          "type": "string",
          "format": "binary"
        }
      }
    }
  }
}
```

jmorel:
My bad, it looks like it's been fixed in https://github.com/orval-labs/orval/commit/0ef60b46c2842a93c0789cb26aa60d5eb313cb8b

melloware:
Yep should be in 8.11.0

mauriceackel:
@jmorel, @melloware: It seems like this change only landed for the response side but not for the request side. In particular, the generated fetch command still generates a JSON.stringify(), even if the format is binary.

I'm referring to: https://github.com/orval-labs/orval/blob/992a525f7bf0e070c6a521cd17dfa1806b06b690/packages/fetch/src/index.ts#L523

```ts
const fetchBodyOption = requestBodyParams
    ? (isFormData && body.formData) ||
      (isFormUrlEncoded && body.formUrlEncoded) ||
      isRawRequestBodyContentType(body.contentType)
      ? `body: ${requestBodyParams}`
      : `body: JSON.stringify(${requestBodyParams})`
    : '';
```

PR #3397, opened 2026-05-19
fix(core): raw binary request bodies such as image/png must be passed to fetch as the Blob itself

daugvinasr:
Closes: #3396

Adds additional checks so this:

```ts
export const replaceLotteryLogo = async (
  id: number,
  replaceLotteryLogoBody: Blob,
  options?: RequestInit,
): Promise<replaceLotteryLogoResponse> => {
  const res = await fetch(getReplaceLotteryLogoUrl(id), {
    ...options,
    method: 'PUT',
    headers: { 'Content-Type': 'image/png', ...options?.headers },
    body: JSON.stringify(replaceLotteryLogoBody),
  });
```
is now

```ts
export const replaceLotteryLogo = async (
  id: number,
  replaceLotteryLogoBody: Blob,
  options?: RequestInit,
): Promise<replaceLotteryLogoResponse> => {
  const res = await fetch(getReplaceLotteryLogoUrl(id), {
    ...options,
    method: 'PUT',
    headers: { 'Content-Type': 'image/png', ...options?.headers },
    body: replaceLotteryLogoBody,
  });
```

## Summary by CodeRabbit

* **New Features**
  * Fetch client now properly handles binary request bodies (such as image uploads) without JSON serialization. Requests with raw content types are transmitted as-is.

* **Tests**
  * Added test coverage for binary request body handling in the fetch client.

[Review Change Stack](https://app.coderabbit.ai/change-stack/orval-labs/orval/pull/3397?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

Issue #3396, opened 2026-05-19
raw binary request bodies must be passed to fetch as the Blob itself


adv-daugvinas:
### Description

When an endpoint defines a raw binary request body, for example `image/png`, Orval correctly generates the request body parameter as Blob, but the generated fetch client sends it using JSON.stringify.

### Output client

fetch

### Configuration (orval.config)

```ts
 binaryRequestBody: {
    output: {
      target: '../generated/fetch/binary-request-body/endpoints.ts',
      schemas: '../generated/fetch/binary-request-body/model',
      client: 'fetch',
      clean: true,
      formatter: 'prettier',
    },
    input: {
      target: '../specifications/binary-request-body.yaml',
    },
  },
```

### Environment

System:
    OS: macOS 26.3.1
    CPU: (12) arm64 Apple M4 Pro
    Memory: 1.18 GB / 24.00 GB
    Shell: 5.9 - /bin/zsh
npmPackages:
    @tanstack/vue-query: ^5.96.1 => 5.96.1 
    orval: ^8.6.2 => 8.6.2 
    vue: ^3.5.31 => 3.5.31

### Expected behavior

Blob should not be wrapped in `JSON.stringify`

### Actual behavior

Blob is wrapped in `JSON.stringify`

### OpenAPI document (minimal, if applicable)

```yaml
openapi: 3.0.0
info:
  title: Binary Request Body
  description: Raw binary request bodies
  version: 1.0.0
servers:
  - url: http://localhost
tags:
  - name: Lottery
    description: Lottery operations
  - name: Logo
    description: Logo operations
paths:
  /lottery/{id}/logo:
    put:
      operationId: replaceLotteryLogo
      summary: Replace lottery logo
      description: This endpoint sets or replaces lottery logo.
      tags:
        - Lottery
        - Logo
      parameters:
        - name: id
          in: path
          description: Lottery ID
          required: true
          schema:
            type: integer
      requestBody:
        description: New logo as binary PNG image
        required: true
        content:
          image/png:
            schema:
              type: string
              format: byte
              maxLength: 1048576
      responses:
        '204':
          description: Success
```

PR #3804, opened 2026-08-03
fix: made fetchBodyOption handle binary format

mauriceackel:
Adds a check to the body serialization to check if the message body has binary type

## Summary by CodeRabbit

* **Bug Fixes**
  * Binary request bodies are now sent in their native format instead of being converted to JSON, improving compatibility with binary uploads and APIs.
  * Improved detection of binary payloads across common media types, schemas, and encodings.
  * Ensured JSON and encoded string request bodies continue to use the appropriate encoding.


mauriceackel:
Relates to: #3396 & #3397
Relates to: #3411

melloware (approved):

melloware:
Running the build now!

melloware:
Looks like build failed

melloware (changes requested):
typescript issue

mauriceackel:
On it!

mauriceackel:
@melloware, I think I fixed the issue but I'm having issues setting up the dev environment properly. Sorry for the inconvenience but is there a chance you can run the checks again?

mauriceackel:
Actually, I think this is a bit more involved as I originally anticipated as we are running this on a body schema. I will have to work on this a little more.

mauriceackel:
After a bit of debugging, I found the `body.definition` field, which could be used like this:


 ```ts
 const isBinaryRequestBodySchema = (body: GetterBody) => body.definition === 'Blob';
 ```
 
 Is there anything inherently wrong with this approach/ are we missing any other binary cases this way?
 
 The main reason the other flow is not that easy is that the body can also be a reference, which we would need to resolve somehow. The definition seems to already be a resolved type.

melloware:
some deeper analysis would need to be done i guess before I could say whether its comprehensive enough...

melloware:
I assigned this PR review to @soartec-lab who wrote most of the Fetch code!

mauriceackel:
Hey @soartec-lab, @melloware

Thanks for sticking with me. Should have done my research before suggesting the change! However, now that I looked deeper into the code, I have majorly refactored this PR. The gist is:

Previously, a "response" already computed an `isBlob` flag and stored it. This was then used for other checks later. The "body" did not have this flag. I now added the `isBlob` flag to the "body" using the same logic as was already used for the "response". This flag can then be used to perform checks on whether the body type is a blob.

Happy for your review! 💜

melloware:
This actually looks much cleaner!

melloware:
OK now looks like some Lint issues

soartec-lab (commented):
There is the issue with Lintner, but this LGTM 👍

mauriceackel:
I think I fixed the remaining linter errors now after getting my dev setup actually properly configured.

melloware (approved):

mauriceackel:
🚀

PR #3806, opened 2026-08-04
fix(fetch): serialize path-item-level query parameters

matamatanot:
## Description

The fetch client's generated URL builder (`getXUrl`) collected only **operation-level** parameters (`spec[verb].parameters`) when building the `URLSearchParams` serialization loop. Query parameters declared at the OpenAPI **path-item level** (shared by all operations under a path) were dropped from the generated request URL — the loop was emitted with an empty body:

```ts
// spec: query param `date` declared at the path-item level
export const getGetItemUrl = (itemId: string, params: GetItemParams) => {
  const normalizedParams = new URLSearchParams();
  Object.entries(params || {}).forEach(([key, value]) => {}); // empty → query string never built
  const stringifiedParams = normalizedParams.toString();      // ""
  return stringifiedParams.length > 0 ? `/items/${itemId}?${stringifiedParams}` : `/items/${itemId}`;
};
```

Path parameters were unaffected (they are injected via the URL template, e.g. `/items/${itemId}`); only query params went through the empty loop. There is no type error or runtime exception — the params type, hook signature and query key are all generated correctly — so the request just silently omits the query string. This also affects `react-query`/`vue-query`/`svelte-query` with `httpClient: 'fetch'`, since they reuse the fetch `generateRequestFunction`.

## Root cause

`packages/fetch/src/index.ts` read:

```ts
const parameters = spec?.[verb]?.parameters ?? []; // operation-level only
```

Path-item-level `parameters` are part of the OpenAPI [Path Item Object](https://spec.openapis.org/oas/v3.1.0#path-item-object) and apply to every operation under the path (overridable per operation), but they were never merged in here. The `zod` and `effect` generators (and core's `verbs-options`) already merge both levels — `fetch` was the outlier.

## Fix

Merge path-item-level and operation-level parameters when collecting the parameter objects, mirroring the other generators:

```ts
const parameters = [
  ...(spec?.parameters ?? []),
  ...(spec?.[verb]?.parameters ?? []),
];
```

## Tests

- Adds `tests/specifications/path-item-level-query-params/spec.yaml` — a path whose only query parameter (`date`) is declared at the path-item level.
- Adds a `fetch.config.ts` entry + api-generation snapshot.
- Adds an assertion that the generated URL builder serializes the query parameter (non-empty `normalizedParams.append(...)` loop).

Related to #2440 (same class of "path-level parameters ignored", previously fixed for the zod generator).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed fetch URL generation to include query parameters declared at the OpenAPI path level alongside operation-level parameters.
  * Ensured combined path and query parameters are correctly normalized and serialized.
  * Fixed parameter overrides so operation-level settings take precedence.

* **Tests**
  * Added regression coverage for path-level query parameter handling in generated fetch requests.


melloware:
See AI comments

melloware (approved):

melloware (changes requested):
See AI comment

soartec-lab (approved):
Thanks!

soartec-lab:
Hi @melloware , could you review this again?

melloware (approved):

Issue #3789, opened 2026-07-30
Clarify the two unrelated output header options and the file-header type


aqeelat:
### Description

Orval has two unrelated output options named around "header", and one has a misleading type.

### Current behavior

`output.headers` controls whether generated client functions include HTTP request-header parameters.

`output.override.header` controls the comment block at the top of generated files.

The current `output.headers` documentation says only:

> Enable generation of headers.

That does not explain that it refers to HTTP request-header parameters, and can be confused with the generated file header.

Additionally, `output.override.header` is typed as `boolean | function`, but `true` has no effect:

- `false` disables the file header.
- A callback customizes it.
- `true` produces the same result as omitting the option.

### Discussion

Possible directions:

- Clarify both documentation entries so they explicitly distinguish HTTP request-header parameters from generated file comments.
- Narrow `output.override.header` from `boolean | function` to `false | function` in a future major release.
- Rename or move the file-header option, with a migration path.

I would avoid adding an alias immediately. Improving the docs is low-risk; changing the public config shape should be weighed against the migration cost.

luantaraschi:
I would like to take the documentation half of this, leaving the type change for a major as you suggested.

I checked both claims against the code first. `output.headers` gates `getQueryParams({ queryParams: parameters.header, suffix: 'headers' })` in `packages/core/src/generators/verbs-options.ts`, so it is the operation's HTTP header parameters. Its default is `false` in `packages/orval/src/utils/options.ts`, which the reference does not mention at all.

For `output.override.header`, the normalizer is:

```ts
header:
  outputOptions.override?.header === false
    ? false
    : isFunction(outputOptions.override?.header)
      ? outputOptions.override.header
      : getDefaultFilesHeader,
```

So `true` lands in the same branch as `undefined`. That confirms what you wrote, and it also means the reference is wrong where it says **Default:** `true`. The effective default is the built-in `getDefaultFilesHeader`, and `true` is just an alias for omitting the option.

Scope I have in mind, docs only, no config or type changes:

- Say that `output.headers` controls HTTP request header parameters, and give its `false` default.
- Say that `output.override.header` controls the comment block at the top of generated files, correct the stated default, and note that `true` has no effect.
- Cross-reference the two so the names stop reading as related.

Let me know if you would rather scope it differently.

PR #3821, opened 2026-08-07
docs: distinguish output.headers from output.override.header

luantaraschi:
Documentation half of #3789. No config or type changes.

`output.headers` and `output.override.header` are unrelated but read as if they were the same setting. The reference said only "Enable generation of headers" for the first and gave no default, and described the second as the "generated file header" without distinguishing it. Both entries now say which one they are and link to each other.

Two facts I checked against the code rather than taking from the issue:

`output.headers` gates `getQueryParams({ queryParams: parameters.header, suffix: 'headers' })` in `packages/core/src/generators/verbs-options.ts`, so it is the operation's HTTP header parameters. Its default is `false` in `packages/orval/src/utils/options.ts`, which the reference did not mention.

`output.override.header` was documented with a default of `true`. The normalizer is:

```ts
header:
  outputOptions.override?.header === false
    ? false
    : isFunction(outputOptions.override?.header)
      ? outputOptions.override.header
      : getDefaultFilesHeader,
```

Anything that is neither `false` nor a function falls through to `getDefaultFilesHeader`, so `true` behaves exactly like omitting the option, and the effective default is that function rather than `true`. The page now says so.

I left the type narrowing from `boolean | function` to `false | function` out, since #3789 puts it in a major.

## Summary by CodeRabbit

* **Documentation**
  * Expanded configuration guidance for output headers.
  * Clarified default behavior, customization options, disabling headers, and fallback behavior.
  * Added configuration examples and explained the distinction between related header settings.

### DIFF
--- a/docs/content/docs/reference/configuration/output.mdx
+++ b/docs/content/docs/reference/configuration/output.mdx
@@ -839,8 +839,24 @@ export default defineConfig({
 ## headers
 
 **Type:** `Boolean`
+**Default:** `false`
 
-Enable generation of headers.
+Generate typed parameters for the HTTP request headers an operation declares in
+the specification. When disabled, header parameters are omitted from the
+generated function signatures.
+
+This is unrelated to [`override.header`](#header), which controls the comment
+block written at the top of each generated file.
+
+```ts title="orval.config.ts"
+export default defineConfig({
+  petstore: {
+    output: {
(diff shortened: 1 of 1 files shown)


melloware:
@aqeelat please review!

melloware (approved):

PR #3822, opened 2026-08-07
fix(zod): report a misplaced boolean `required` instead of crashing

### LINKED ISSUE #3719: bug(zod): crash on `required: true` sibling of a schema `$ref` — `(schema.required ?? []) is not iterable`
### What's wrong

A schema node that carries `required: true` next to a `$ref` slips past spec validation and then crashes the zod generator with `(schema.required ?? []) is not iterable`. Some generators emit this shape. It is invalid either way: in a Schema Object `required` must be an array of property names; only Request Body Objects and Parameter Objects take a boolean. In OpenAPI 3.0 a node with `$ref` is a Reference Object, and the spec says sibling properties "shall be ignored".

Two things make this worth fixing:

1. The failure is an internal `TypeError`, not a diagnostic. It names no file, path, or spec rule.
2. The equivalent mistake **without** a `$ref` is already handled well. `@scalar/openapi-parser` rejects it with `type must be array` and a JSON pointer. Only the `$ref`-sibling form escapes, because validators treat that node as a Reference Object and skip its siblings.

### Repro

`spec.yaml`:

```yaml
openapi: 3.0.3
info: { title: Example API, version: 1.0.0 }
paths:
  /items:
    post:
      operationId: createItem
      requestBody:
        content:
          application/json:
            schema:
              required: true          # <- invalid: Schema Object `required` is an array
      responses:
        '204': { description: Created }
components:
  schemas:
    Item:
      type: object
      required: [name]
      properties:
        name: { type: string }
```

```sh
orval --input ./spec.yaml --output ./out.ts   # add `client: zod` in a config
```

Actual, with `client: zod` or `client: hono`:

```text
(schema.required ?? []) is not iterable
```

Expected: a validation error naming the offending path and the expected shape, e.g.

```text
OpenAPI spec validation failed:
  /paths/~1items/post/requestBody/content/application~1json/schema/required
  `required` in a Schema Object must be an array of property names, got boolean.
  A boolean `required` belongs on the Request Body Object or a Parameter Object.
  See: https://spec.openapis.org/oas/v3.0.3.html#schema-object
```

Verified on master `70fc4fad4`:

| input | client | result |
| --- | --- | --- |
| `required: true` beside `$ref` (3.0 or 3.1) | fetch / axios / mock | generates fine, boolean ignored |
| `required: true` beside `$ref` (3.0 or 3.1) | zod, hono | **crash** |
| `required: true` beside `$ref` in a property of a component schema | zod | **crash** |
| `required: true` beside `$ref` in a response schema | zod | **crash** |
| `required: true` on an inline schema (no `$ref`) | any | already rejected with a clear message |

### Where

- Crash site: `packages/zod/src/index.ts:1233` — `...(schema.required ?? [])` inside the `requiredKeys` set.
- Same unguarded spread: `packages/zod/src/index.ts:763` (the `allOf` path).
- Validation runs at `packages/orval/src/import-specs.ts:123` (`validateSpec`), which does not flag `$ref` siblings.
- Existing orval-side pre-validation to extend: `validateComponentKeys` at `packages/orval/src/import-specs.ts:427`, called at `packages/orval/src/import-specs.ts:121`.
- Existing defensive precedent: `packages/core/src/getters/object.ts:385-388` already does `Array.isArray(itemRequired) ? itemRequired : []`.
- Other unguarded `required` spreads worth the same treatment: `packages/mock/src/faker/resolvers/value.ts:275-276` and `:380`, `packages/mock/src/faker/getters/all-of-required.ts:86`.

### Proposed fix

Report, do not normalize. Both maintainers who commented agreed: a boolean on a schema node is ambiguous, and per OAS 3.0 a `$ref` sibling must be ignored rather than promoted, so moving it to `requestBody.required` would be a guess.

1. Add a walk next to `validateComponentKeys` in `packages/orval/src/import-specs.ts` that flags any Schema Object whose `required` is not an array. Throw with the JSON pointer to the offending node, the expected shape, and a spec link — matching the existing `validateComponentKeys` message style. Skip it under `unsafeDisableValidation`, like the rest of validation.
2. Guard `packages/zod/src/index.ts:1233` and `:763` with `Array.isArray(...)`, mirroring `packages/core/src/getters/object.ts:385`. Belt and braces, so a bad schema reaching the generator from any other path degrades instead of throwing a `TypeError`.
3. Open an upstream issue on `@scalar/openapi-parser` so the `$ref`-sibling case is caught by the parser itself, and drop the orval-side walk once it is.

### Acceptance

- [ ] A boolean `required` on a Schema Object — with or without a sibling `$ref`, in a request body, response, or nested property — fails with a message naming the JSON pointer and the expected shape, not `is not iterable`.
- [ ] A boolean `required` on a Request Body Object or Parameter Object still works unchanged.
- [ ] An array `required` on a Schema Object still works unchanged.
- [ ] `unsafeDisableValidation: true` skips the new check.
- [ ] Regression fixture plus a generation test covering both the `$ref`-sibling and inline forms, for `zod` and one TS client.

luantaraschi:
Implements option 2 from #3719, the one @snebjorn voted for and @melloware agreed with: report the invalid document rather than normalize it.

Reproduced first. A schema carrying `required: true` reaches the two spread sites in `packages/zod/src/index.ts` and fails with:

```
TypeError: (schema.required ?? []) is not iterable
```

That message names neither the schema nor what the keyword should hold, and generation of the whole client stops.

Both sites now read `required` through a small helper that returns the array when the document is valid and otherwise throws:

```
Invalid OpenAPI document: schema "Item" has `required: true`, but a schema
object's `required` must be an array of property names. A boolean `required`
belongs on the request body object or on a parameter, not on the schema it
references.
```

I did not normalize the value. A boolean carries no property names, so there is nothing to recover from it, and picking a set of required keys would change validation without the user knowing. That is the ambiguity @snebjorn raised.

Five tests, in `packages/zod/src/zod.test.ts`: the plain object path, the `allOf` path, the `allOf` member one level down, and two controls asserting a valid `required` array still generates, one on a schema and one on a member. Run against the commit before the fix, the first three fail and both controls pass on either side, so they are not asserting something that was already true. The three fail in two different ways, which is the point of the third one:

```
names the schema and the expected shape          got '(schema.required ?? []) is not iterable'
reports it on the allOf path too                got '(schema.required ?? []) is not iterable'
names the offending allOf member                expected [Function] to throw an error
```

The member one level down never threw at all. It rendered as `zod.unknown()` and said nothing.

`packages/zod` goes from 320 to 325 tests, all passing. `tsc --noEmit` and `vp fmt --check` are clean.

Scope note: `packages/mock/src/faker/resolvers/value.ts` spreads `schemaReference.required` the same way, so the same document shape would fail there too. I left it out to keep this focused on the reported path, and can send it separately if you want it covered.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for schema `required` values.
  * Invalid non-array values now produce clear, descriptive errors identifying the affected schema.
  * Validation now works consistently across objects, combined schemas, and referenced schemas.
  * Valid required-field arrays continue to work correctly.

* **Tests**
  * Added regression coverage for invalid and valid `required` configurations, including combined and referenced schemas.

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -784,18 +784,23 @@ export const generateZodValidationSchemaDefinition = (
       ? [
           ...new Set([
             ...getRequiredKeys(schema, name),
-            ...schemas.flatMap((member) => {
+            ...schemas.flatMap((member, index) => {
               // Only the member's top-level `required` is needed. For `$ref`
               // members resolve shallowly (no deep property dereference) and
               // tolerate unresolvable refs — they simply contribute no keys.
-              const resolved =
-                '$ref' in member && typeof member.$ref === 'string'
-                  ? tryResolveRefSchema(member.$ref, context)
-                  : (member as OpenApiSchemaObject);
-              const memberRequired = resolved?.required;
-              return Array.isArray(memberRequired)
-                ? (memberRequired as string[])
-                : [];
+              const isRef = '$ref' in member && typeof member.$ref === 'string';
+              const resolved = isRef
+                ? tryResolveRefSchema(member.$ref as string, context)
+                : (member as OpenApiSchemaObject);
+              if (!resolved) return [];
+              // A constraint-only member never reaches the object path below,
+              // so a misplaced boolean here would otherwise pass unreported.
+              // Name the member, not the composing schema, or the message
+              // points at the wrong place in the document.
+              return getRequiredKeys(
+                resolved,
+                isRef ? (member.$ref as string) : `${name}.allOf[${index}]`,
+              );
             }),
           ]),
         ]
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -416,6 +416,29 @@ const isDiscriminatableMember = (
   return hasLiteralDiscriminator(resolved, property);
 };
 
+/**
+ * Read a schema object's `required` keyword.
+ *
+ * Some generators emit `required: true` on the schema referenced by a request
+ * body, borrowing the boolean that belongs on the request body object. Spreading
+ * that boolean fails with `(schema.required ?? []) is not iterable`, which says
+ * nothing about the document. Report the offending schema and the expected shape
+ * instead. The document is not rewritten: a boolean carries no property names,
+ * so there is nothing to recover from it. (#3719)
+ */
+const getRequiredKeys = (schema: OpenApiSchemaObject, name: string) => {
+  const required = schema.required as unknown;
+
+  if (required === undefined) return [];
+  if (Array.isArray(required)) return required;
+
+  throw new Error(
+    `Invalid OpenAPI document: schema "${name}" has \`required: ${JSON.stringify(
+      required,
+    )}\`, but a schema object's \`required\` must be an array of property names. A boolean \`required\` belongs on the request body object or on a parameter, not on the schema it references.`,
+  );
+};
+
 export const generateZodValidationSchemaDefinition = (
   schema: OpenApiSchemaObject | OpenApiReferenceObject | undefined,
   context: ContextSpec,
@@ -760,7 +783,7 @@ export const generateZodValidationSchemaDefinition = (
     const allOfRequired = schema.allOf
       ? [
           ...new Set([
-            ...(schema.required ?? []),
+            ...getRequiredKeys(schema, name),
             ...schemas.flatMap((member) => {
               // Only the member's top-level `required` is needed. For `$ref`
               // members resolve shallowly (no deep property dereference) and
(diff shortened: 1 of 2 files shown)


melloware:
@snebjorn  @wadakatu could use your eyes on this.

luantaraschi:
The finding holds. I checked it before changing anything: with

```ts
{ allOf: [{ type: 'object', properties: { a: { type: 'string' } } }, { required: true }] }
```

generation completes today and the member renders as `zod.unknown()`. Nothing is lost, because a boolean carries no property names either way, but the document is malformed and the user hears nothing about it. Reporting it on the composing schema and ignoring it one level down was inconsistent, so I took the fix in `cb11cbc`.

One change from the proposed diff. It passes `name`, which is the composing schema's name, so a bad member would report as `schema "Composed" has required: true` and send the reader to a line that is fine. The member is labelled instead: an inline member reports as `Composed.allOf[1]`, and a `$ref` member reports as its ref, since that is where the reader has to go to fix it.

Added two tests: the bad inline member, and a valid `required: ['a']` on a member so the check does not swallow the legitimate case. The 289 tests in `zod.test.ts` pass.

melloware (approved):

Issue #3790, opened 2026-07-31
Prefix override for query keys/options factories


farsil:
### Description

When used with the `react-query` client, I would like an option in the orval configuration file to configure explicitly what prefix to use for factories.

### Current behavior

If I use orval with `react-query` client, it internally determines whether to use a `use` or `get` prefix based on the presence of a custom query options factory:

```
  const queryOptionsFnName = camel(
    queryKeyMutator || queryOptionsMutator || mutator?.isHook
      ? `use-${name}-queryOptions`
      : `get-${name}-queryOptions`,
  );
```

However, the `use` prefix is sometimes undesirable for query options, since the function may have to be used outside react (router loaders for example, with `ensureQueryData` or `prefetchQuery`). You can use it anyway of course but:
1. It would be confusing in any react codebase
2. If you have certain eslint rules configured, you may have to disable the rule about [React functions](https://react.dev/reference/rules/rules-of-hooks#only-call-hooks-from-react-functions)

### Proposed behavior

Orval should let the user decide what prefix to use (if any) for query options factories, and possibly for query keys factories as well.

### Proposed solution

The default automatic prefix could detect whether the mutators are hooks or not:

```
  const queryOptionsFnName = camel(
    queryKeyMutator?.isHook || queryOptionsMutator?.isHook || mutator?.isHook
      ? `use-${name}-queryOptions`
      : `get-${name}-queryOptions`,
  );
```

Moreover, we could introduce two new settings (`queryKeyFactoryPrefix` and `queryOptionsFactoryPrefix`, for example) that would live inside [`query`](https://orval.dev/docs/reference/configuration/output#overridequery). An empty prefix is sometimes desirable: my OpenAPI yaml files already prefix get operations with `get`, so with the default `get` prefix (assuming no mutators) it would result in a silly `getGetActivitiesQueryOptions`. 

I am willing to put the work myself if you like the idea.

melloware:
I feel like this is possible with one of the congfigs that is an override function and you write your own function to determine how you want it.

melloware:
A PR is definitely welcome but I would research first that what you want is not already possible.

melloware:
I think this is what you want: 

https://orval.dev/docs/reference/configuration/output#querykey--queryoptions--mutationoptions

might have to look for tickets for examples. We should updaete the docs to show an example usage

melloware:
@wadakatu do we have examples of this and mutationOptions overrides?  It seems like the docs are sorely lacking here.

[...]

PR #3825, opened 2026-08-07
docs: add examples for queryKey, queryOptions and mutationOptions

luantaraschi:
Picks up @melloware's note on #3790 that the docs are lacking here, scoped to documentation only. No config or behaviour change.

The `queryKey / queryOptions / mutationOptions` section described the three overrides in prose and named the operation identity each mutator receives, but had no code. So the answer to "how do I actually write one" lived in the test fixtures and in old tickets, which is what the issue thread ended up doing manually.

Adds the config snippet that wires a mutator up, then one example per option. They are taken from the fixtures @wadakatu pointed at rather than written from scratch:

- `queryKey` from `tests/mutators/custom-query-key.ts`
- `queryOptions` from `tests/mutators/custom-query-options-with-operation.ts`
- `mutationOptions` from `tests/mutators/custom-mutation.ts`

I trimmed them for the page: the `queryOptions` example inlines the key prefix instead of hoisting a constant, and the `mutationOptions` one drops the `Required<Pick<..., 'mutationFn'>>` intersection, which is noise in a doc example. The parameter lists and the operation-identity shapes are unchanged from the fixtures.

The closing line points at what the issue was actually asking for: since the mutator receives `operationId`, one file can serve every operation and branch where behaviour differs. That is the alternative to the per-operation prefix option requested in the issue, and it is what the thread converged on.

`vp fmt` skips `.mdx`, so there is no formatter to run against this file. The anchors and heading levels are unchanged.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for configuring custom query and mutation behavior.
  * Documented custom key generation, callback signatures, operation context, and operation-specific examples.
  * Included instructions for connecting custom configuration files and names.
  * Explained option wrapping and recommended practices for handling callback side effects.

### DIFF
--- a/docs/content/docs/reference/configuration/output.mdx
+++ b/docs/content/docs/reference/configuration/output.mdx
@@ -1251,6 +1251,81 @@ parameter, orval passes operation identity so the mutator can branch on it
 - `mutationOptions` mutator — `{ operationId, operationName }` (the `url`
   is supplied in the second parameter)
 
+Each option takes a mutator, so point it at a file and an exported name:
+
+```ts title="orval.config.ts"
+export default defineConfig({
+  petstore: {
+    output: {
+      client: 'react-query',
+      override: {
+        query: {
+          queryKey: {
+            path: './src/mutators/custom-query-key.ts',
+            name: 'customQueryKey',
+          },
+        },
+      },
+    },
+  },
+});
+```
+
+`queryKey` replaces the generated key factory. It receives the operation's
+query properties and a context carrying the `url`:
+
+```ts title="src/mutators/custom-query-key.ts"
+export function customQueryKey(
+  // The properties dictionary varies per operation (params, petId, ...), so
+  // type it broadly if one mutator serves every endpoint.
+  properties: Record<string, unknown>,
+  context: { url: string },
+) {
+  return ['tenant-abc', context.url, properties] as const;
+}
+```
+
+`queryOptions` wraps the options object passed to the generated hook. The
+third parameter is the operation identity described above:
+
+```ts title="src/mutators/custom-query-options.ts"
+import type { QueryKey } from '@tanstack/react-query';
+
+export function customQueryOptions<T extends { queryKey: QueryKey }>(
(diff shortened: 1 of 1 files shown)


melloware (dismissed):

melloware:
@luantaraschi  these are great can you review all the AI comments?

luantaraschi:
Went through all three. Two hold up, one had the right instinct but the wrong reason. Pushed `40a49a9`.

**Configure all three mutators** — fair, the snippet only wired `queryKey` while the page goes on to show two more. All three are in the config block now.

**`customQueryOptions` return type** — the finding is right, the stated reason is not. It claims the annotation breaks callers passing a fixed `queryKey` tuple, which would be a compile error. It is not: the function typechecks under `--strict` as written, because the spread's inferred type is assignable to `T` in return position. I checked before changing anything.

What is genuinely wrong is that `: T` promises the exact input type while the body replaces `queryKey`, so the declared type and the runtime value disagree for any caller with a narrower key. Widened to `T & { queryKey: QueryKey }`, which says what the function actually returns.

**Invalidate only after `deletePetById` succeeds** — this one is real, and worse than the comment suggests. `mutation-generator.ts` emits

```
const customOptions = useCustomMutation({...mutationOptions, mutationFn}, ...);
```

inside the generated options function, so the mutator body runs on every render of the hook, not when the mutation settles. `queryClient.invalidateQueries()` sitting there fires on each render, and since invalidating refetches `/pets` and a refetch re-renders, a component holding both hooks can loop. The example moves it into `onSuccess` and calls through to any `onSuccess` the caller passed, so it still composes.

Both revised examples typecheck under `--strict`.

One thing that follows from the third point and is outside this PR: `tests/mutators/custom-mutation.ts` has the same shape, invalidating in the mutator body. That is where I took the example from. It works as a fixture because nothing there renders twice, but it is the thing people copy. Happy to send a separate PR moving that one into `onSuccess` too if you want it changed.

melloware (approved):

PR #3829, opened 2026-08-07
fix(tests): invalidate in onSuccess in the custom mutation fixture

luantaraschi:
Follow-up to #3825, where I raised this at the end and offered to send it separately.

The `useCustomMutation` fixture invalidates in the hook body:

```ts
if (operation.operationId === 'deletePetById') {
  queryClient.invalidateQueries({ queryKey: ['/pets'] });
}
return options;
```

`mutation-generator.ts` emits the mutator call inside the generated options function:

```ts
const customOptions = useCustomMutation({...mutationOptions, mutationFn}, ...);
```

That function runs on every render of the hook, not when the mutation settles. So the invalidation fires on each render, and since invalidating `/pets` triggers a refetch and a refetch re-renders, a component holding both hooks can loop.

The fixture works as a fixture because nothing in the test suite renders it twice. It matters because it is what people copy: it is the only `mutationOptions` example in the repo, and it is where I took the example in #3825 from. That PR is now merged, so the documented example shows the `onSuccess` shape while the fixture still shows the other one.

The invalidation moves into `onSuccess` and calls through to any `onSuccess` the caller passed, so it still composes.

## Verification

The generator emits an import and a call, never the mutator body, so no snapshot changes. I checked: `tests/__snapshots__/react-query/custom-mutator-options/endpoints.ts` and `tests/__snapshots__/react-query/issue-2039/endpoints.ts` contain only `import { useCustomMutation } from '../../../mutators/custom-mutation'` and two call sites.

The return type does change, from the options object to a union of it and the spread with `onSuccess`, so I typechecked the consumer side rather than assume. Both generated snapshot directories plus `tests/mutators` under `strict`: 122 repo files, no errors. `tests/tsconfig.json` is clean too.

`vp fmt` skips `tests/mutators` by its ignore rules, so there is nothing to format there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Pet list data now refreshes after a successful pet deletion.
  * Mutation success callbacks continue to run as expected.

### DIFF
--- a/tests/mutators/custom-mutation.ts
+++ b/tests/mutators/custom-mutation.ts
@@ -20,8 +20,11 @@ export const useCustomMutation = <T, TError, TData, TContext>(
   // re-renders.
   return {
     ...options,
-    onSuccess: (data: T, variables: TData, context: TContext) => {
-      queryClient.invalidateQueries({ queryKey: ['/pets'] });
+    onSuccess: async (data: T, variables: TData, context: TContext) => {
+      // `onSuccess` may return a promise, and the mutation waits for it, so
+      // awaiting the invalidation keeps the mutation pending until `/pets` is
+      // refreshed instead of settling while the refetch is still in flight.
+      await queryClient.invalidateQueries({ queryKey: ['/pets'] });
       return options.onSuccess?.(data, variables, context);
     },
   };
--- a/tests/mutators/custom-mutation.ts
+++ b/tests/mutators/custom-mutation.ts
@@ -9,8 +9,20 @@ export const useCustomMutation = <T, TError, TData, TContext>(
   operation: { operationId: string; operationName: string },
 ) => {
   const queryClient = useQueryClient();
-  if (operation.operationId === 'deletePetById') {
-    queryClient.invalidateQueries({ queryKey: ['/pets'] });
+
+  if (operation.operationId !== 'deletePetById') {
+    return options;
   }
-  return options;
+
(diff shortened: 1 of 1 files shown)


melloware (changes requested):
See AI comments

luantaraschi:
One half holds, the other does not apply to the pinned version.

**Awaiting the invalidation** is right and I took it. `onSuccess` is typed `=> Promise<unknown> | unknown` and the mutation waits for what it returns, so firing `invalidateQueries` and dropping the promise let the mutation settle while the refetch was still in flight. It is now `async` and awaits before calling through to the caller's `onSuccess`. That is also the better shape for a fixture people copy.

**Forwarding "all" the arguments** does not apply here. The comment says v5 adds a mutation context parameter, but on `5.62.16`, which is what `tests/` resolves, the user-facing callback takes three:

```ts
// @tanstack/query-core@5.62.16, hydration-ClXcjjG9.d.ts:1130
onSuccess?: (data: TData, variables: TVariables, context: TContext) => Promise<unknown> | unknown;
```

The four-parameter form on line 361 is the `MutationCache` config callback, which takes the `Mutation` itself as the fourth argument. That is a different hook and not what this mutator implements. The three arguments were already being forwarded, so I left the signature alone.

`tests/tsconfig.json` is clean, and so is a typecheck of both generated snapshot directories against the new fixture, since the return type changed to a promise.

melloware (approved):

PR #3830, opened 2026-08-07
fix(core): derive the schema file name from the file name

luantaraschi:
`getSchemaFileName` computes an offset on one string and applies it to another:

```ts
return path
  .replace(`.${getExtension(path)}`, '')
  .slice(path.lastIndexOf('/') + 1);
```

`replace` with a string pattern removes the **first** occurrence, which may sit in a directory name, and `path.lastIndexOf('/')` is measured on the original `path` while `slice` runs on the already-shortened result. Once `replace` removes anything before the last `/`, the two disagree by exactly the number of characters removed and the name is cut in the wrong place:

```ts
getSchemaFileName('v1.yaml/petstore.yaml')  // 'ore.yaml', expected 'petstore'
getSchemaFileName('a.json/b/c.json')        // 'n',        expected 'c'
```

`ref.ts` feeds this the pathname of an external `$ref`, and the result becomes the generated type name through `pascal(originalName)`, so a schema split across a directory whose name carries `.json` or `.yaml` generates a type named after a fragment of its own path.

The function now takes the file name first and strips the extension only when it really is a suffix of that name. Everything that worked before still produces the same string: when the extension appears once and at the end, `replace` and `endsWith` remove the same characters, and the offset problem never arose because nothing before the last `/` was touched.

`getExtension` decides between yaml and json by scanning the whole path with `includes`, so `'v1.yaml/petstore.json'` still reports yaml and the `.json` suffix is left on the name. That is a separate question about `getExtension` and it is used in other places to pick output formats, so I left it alone rather than widen this.

## Tests

Three cases in `packages/core/src/utils/path.test.ts`: the plain extension drop, a name with no extension, and the directory-name case above.

I ran them against the code before the change. Only the third fails there; the first two pass on both sides, so they are not asserting something that was already broken.

`packages/core`: 2271 passing. The 3 failures in `src/utils/resolve-version.test.ts` reproduce on a clean `master` with this change stashed, so they are not from here. `vp fmt --check` and `tsc --noEmit` are clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved schema filename handling when paths include directories with periods.
  - File extensions are now removed only when they appear at the end of the filename.
  - Extensionless filenames are preserved correctly, providing more reliable schema name detection across varied paths.

- **Tests**
  - Added coverage for filenames with extensions, extensionless names, and directories containing periods.

### DIFF
--- a/packages/core/src/utils/path.test.ts
+++ b/packages/core/src/utils/path.test.ts
@@ -2,7 +2,7 @@ import path from 'node:path';
 
 import { describe, expect, it } from 'vitest';
 
-import { getRelativeImportPath } from './path';
+import { getRelativeImportPath, getSchemaFileName } from './path';
 
 // Build absolute paths that are valid on both Windows and Linux.
 // path.parse(process.cwd()).root gives 'C:\' on Windows and '/' on Linux.
@@ -148,5 +148,22 @@ describe('getRelativeImportPath', () => {
         getRelativeImportPath('bad/path.ts', abs('src', 'exporter.ts')),
       ).toThrow('"bad/path.ts"');
     });
+  });
+});
+
+describe('getSchemaFileName', () => {
+  it('drops the extension from the file name', () => {
+    expect(getSchemaFileName('petstore.yaml')).toBe('petstore');
+    expect(getSchemaFileName('specs/petstore.yaml')).toBe('petstore');
+    expect(getSchemaFileName('./petstore.json')).toBe('petstore');
+  });
+
+  it('keeps a file name that has no extension', () => {
+    expect(getSchemaFileName('specs/no-ext')).toBe('no-ext');
+  });
+
+  it('ignores an extension that appears in a directory name', () => {
+    expect(getSchemaFileName('v1.yaml/petstore.yaml')).toBe('petstore');
+    expect(getSchemaFileName('a.json/b/c.json')).toBe('c');
   });
 });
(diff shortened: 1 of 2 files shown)


melloware (changes requested):
See AI comments

luantaraschi:
Good catch, and it is a better fix than the one I talked myself out of. In the description I wrote that `getExtension` scanning the whole path was a separate question because it is used elsewhere. That was wrong on the facts: `getSchemaFileName` is its only caller, so there was nothing else to weigh, and the fix is not to change `getExtension` at all, just to call it on the file name I had already extracted one line above.

Changed to `getExtension(fileName)` and added `'v1.yaml/petstore.json'` to the directory-name test.

Only that case changes. The other four in the test go through the same path either way:

```
v1.yaml/petstore.json    getExtension(path)= "petstore.json"   getExtension(fileName)= "petstore"
v1.yaml/petstore.yaml    getExtension(path)= "petstore"        getExtension(fileName)= "petstore"
a.json/b/c.json          getExtension(path)= "c"               getExtension(fileName)= "c"
petstore.yaml            getExtension(path)= "petstore"        getExtension(fileName)= "petstore"
specs/no-ext             getExtension(path)= "no-ext"          getExtension(fileName)= "no-ext"
```

`vp fmt --check` and the `path.test.ts` suite are clean.

melloware (approved):

PR #3837, opened 2026-08-08
fix(core): don't read the pascal cache off Object.prototype

luantaraschi:
﻿`pascal` memoizes into an object literal keyed by the name being converted:

```ts
const pascalMemory: Record<string, string> = {};

export function pascal(s = '') {
  if (pascalMemory[s]) {
    return pascalMemory[s];
  }
```

Those keys come from the OpenAPI document, and a schema is free to name a property `toString` or `constructor`. For those the lookup finds the member inherited from `Object.prototype`, which is truthy, so the cache "hits" and hands back a function:

```js
pascal('toString')   // function toString() { [native code] }
camel('toString')    // TypeError: s.charAt is not a function
```

`camel` passes the result to `decap`, which calls `.charAt` on it, so generation stops on a `TypeError` that names neither the schema nor the property. Seven ordinary names reach it: `toString`, `valueOf`, `constructor`, `hasOwnProperty`, `isPrototypeOf`, `toLocaleString`, `propertyIsEnumerable`.

The write side is broken in the same place: `pascalMemory['__proto__'] = ...` goes through the `Object.prototype` setter and stores nothing, so that key never caches.

## The change

A `Map`. Reading with `get` and comparing against `undefined` rather than truthiness also lets an empty result stay cached, which the old check could never do.

## Tests

Added a block to `case.test.ts`: for each of the seven names, `pascal` returns a string with the expected shape and `camel` round-trips it. Plus `__proto__`, which the word rules shape to `_Proto` / `_proto` like any other name, and a second-call check so the cache is still exercised.

Fifteen of the twenty-nine cases in `case.test.ts` fail on `master` and all twenty-nine pass here.

`bunx vitest run packages/core`: 2285 passed. The three failures in `resolve-version.test.ts` are already there on a clean checkout of `master` on this machine and are untouched by this change. `bunx vp fmt --check` is clean on both files.

## How I got here

Reading the file rather than from an issue, so there is nothing to link. Happy to open one first if you would rather have it tracked that way.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved name conversion reliability for special names, including `__proto__`.
  * Ensured repeated PascalCase conversions return consistent results.
  * Prevented conflicts with inherited object property names during conversion.

### DIFF
--- a/packages/core/src/utils/case.test.ts
+++ b/packages/core/src/utils/case.test.ts
@@ -71,3 +71,39 @@ describe('kebab-case a few examples', () => {
     });
   }
 });
+
+describe('property names inherited from Object.prototype', () => {
+  // A schema is free to name a property `toString` or `constructor`. The
+  // pascal cache is keyed by that name, so on an object literal the lookup
+  // read the inherited member back instead of a cached string.
+  for (const name of [
+    'toString',
+    'valueOf',
+    'constructor',
+    'hasOwnProperty',
+    'isPrototypeOf',
+    'toLocaleString',
+    'propertyIsEnumerable',
+  ]) {
+    it(`pascal('${name}') returns a string`, () => {
+      const result = pascal(name);
+      expect(typeof result).toBe('string');
+      expect(result).toBe(name.charAt(0).toUpperCase() + name.slice(1));
+    });
+
+    it(`camel('${name}') returns a string`, () => {
+      expect(camel(name)).toBe(name);
+    });
+  }
+
+  it('shapes __proto__ by the same word rules as any other name', () => {
+    expect(pascal('__proto__')).toBe('_Proto');
+    expect(camel('__proto__')).toBe('_proto');
+  });
+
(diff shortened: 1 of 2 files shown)


luantaraschi:
﻿The red `pr-checks (windows-latest, 22.x)` is not this change.

The `external-ref` project in the snapshot suite resolves `https://geojson.org/schema/Point.json`, and that host answered 503 on the Windows runner:

```
[WARN] Fetch failed with status 503 Service Unavailable for URL: https://geojson.org/schema/Point.json
Failed to resolve external reference "https://geojson.org/schema/Point.json".
```

`pr-checks (ubuntu-latest, 22.x)` in the same run, on the same commit, passed. My other five open PRs here are green on the same Windows job, so it is the remote host rather than the platform.

Happy to push an empty commit whenever you want it re-run.

melloware (changes requested):
Build failing

melloware (approved):

Issue #3786, opened 2026-07-30
Keep configuration reference aligned with public types and document callback parameters


aqeelat:
### Description

The configuration reference is hand-written MDX and can drift from the public TypeScript configuration types. Several callback options expose parameters whose shape is undocumented.

For example, `output.override.header` accepts:

```ts
(info: OpenApiInfoObject) => string[] | string
```

The docs use `info.title` in an example but do not explain what `info` contains.

### Current behavior

`OpenApiInfoObject` is the OpenAPI Info Object and includes optional fields such as:

- `title`
- `description`
- `termsOfService`
- `contact`
- `license`
- `version`

Neither the callback option nor the aliased type documents this shape locally.

This is not limited to `header`: other public configuration callbacks and options have incomplete or missing JSDoc.

The configuration pages are hand-written MDX. Only roughly 10% of user-facing configuration fields currently have JSDoc, so fully generating the reference from types would currently produce mostly incomplete pages.

### Discussion

TypeDoc and its Markdown plugin are already used by this repository, so generating type-derived documentation or validation is technically possible without a new dependency.

Possible directions:

- Document public configuration types and callback parameters in JSDoc, starting with `header` and `OpenApiInfoObject`.
- Generate reference fragments from those annotations and embed them in curated MDX pages.
- Add CI validation that detects stale generated fragments or undocumented public configuration fields.
- Keep rich explanatory material, examples, callouts, and guides hand-written.

A GitHub Action that automatically commits documentation changes seems less useful than a deterministic local generation command plus a CI drift check.

aqeelat:
Working on this — draft PR: #3842.

**Approach.** The reference is generated from `packages/core/src/types.ts` into a committed, drift-checked registry consumed by the docs site. The generator uses the TypeScript compiler API — necessary specifically to expand callback parameters whose types are external aliases (e.g. `override.header`'s `info: OpenApiInfoObject` → the concrete OpenAPI Info Object fields). On the issue's TypeDoc note: TypeDoc resolves the alias name but does not flatten `@scalar`'s `Modify`/`Omit`/intersection utility types, so it can't drive the resolution; the TS checker can.

**Coverage.** JSDoc authored for all 175 public `defineConfig` fields (was ~10%). Callback params that are OpenAPI aliases auto-resolve to their fields with a spec link; internal types (e.g. `GeneratorVerbOptions`) are intentionally not expanded.

[...]

PR #3842, opened 2026-08-09
feat(docs): type-derived configuration reference [WIP]

aqeelat:
## Goals

Address #3786: keep the configuration reference aligned with the public
TypeScript types and document callback parameters, with no drift between
docs and types.

## Approach

- A manifest-driven generator (`scripts/generate-config-reference.ts`)
  reads the public `defineConfig` surface from `packages/core/src/types.ts`
  and emits a committed registry under `docs/src/generated/config-reference/`.
- The generator uses the **TypeScript compiler API** to resolve callback
  parameters whose types are external aliases (e.g. `override.header`'s
  `info: OpenApiInfoObject`) into their concrete OpenAPI fields, linked to
  the spec. Note on the issue's TypeDoc premise: TypeDoc resolves alias
  names but does not flatten `@scalar`'s `Modify`/`Omit`/intersection
  utility types, so it can't drive resolution; the TS checker does. No new
  dependency.
- Only approved public OpenAPI aliases expand (one level); internal types
  (e.g. `GeneratorVerbOptions`) are not expanded.
- Curated MDX stays for examples/callouts; structured field/type/default
  data is generated and embedded via a global `<ConfigReference path="…">`.
  The reference also renders into `llms.txt`/`llms-full.txt`.

## Status (draft)

- ✅ Generator + registry, resolver, `<ConfigReference>` (en/zh chrome),
  CI drift + coverage gate + resolver self-check.
- ✅ JSDoc for all 175 public config fields (was ~10%).
- 🚧 Migrating curated MDX descriptions into JSDoc as the single source
  (so prose can't drift). First section (`override.swr`, en + zh) done;
  remaining sections follow the same pattern. zh translations preserved
  via a `hideDescription` render mode.
- ⏳ Not yet manifest-covered: `schemas`, `baseUrl`, `factoryMethods`,
  `filters`, `parserOptions`, top-level `mock`.

## Verification

`gen:config-reference --check` (175/175), `git diff --exit-code` (drift),
`test:config-reference` (resolver), docs typecheck + build (en + zh), format.

Refs #3786.


aqeelat:
@coderabbitai review

PR #3858, opened 2026-08-15
fix(zod): add as const to Zod Mini enum object literals

### LINKED ISSUE #3852: Zod Mini enum output is missing `as const`, so literal types widen to `string`
### What happens

When the Zod client generates an enum with member names or descriptions, it emits an object literal passed to `zod.enum({ ... })`. For the Zod Mini variant, that object literal has no `as const` assertion. The literal types widen to `string`.

`z.infer` returns `string` instead of the expected literal union. Reading a member off `.def.entries` also returns `string`.

### Why Mini differs from Classic

The two variants have different signatures in zod 4.4.2:

- Classic: `_enum<const T extends util.EnumLike>(entries: T, ...)`
- Mini: `_enum<T extends util.EnumLike>(entries: T, ...)`

Classic has a `const` type parameter, so it preserves literal types on its own. Mini does not. Zod 3's `nativeEnum` has the same gap, which is why #3840 added `as const` to the v3 path during review. The Mini path was not updated at the same time.

### Reproduction

Verified with `tsc --noEmit --strict` against zod 4.4.2:

```ts
import * as zm from 'zod/mini';

const widened = zm.enum({ Black: 'BLACK' });
type W = zm.infer<typeof widened>;             // string — accepts any string
const a: 'BLACK' = widened.def.entries.Black;  // TS2322

const correct = zm.enum({ Black: 'BLACK' } as const);
type C = zm.infer<typeof correct>;             // 'BLACK'
const b: 'BLACK' = correct.def.entries.Black;  // ok
```

### Why CI does not catch it

The repository generates no Zod Mini output. No test config sets `variant: 'mini'`, so no Mini files exist under `tests/generated`. The unit test for Mini enum metadata compares generated strings only. A string comparison cannot detect a type-level regression. The compile-time regression fixture covers Zod 4 Classic and Zod 3 Classic, but not Mini.

### Proposed fix

Append ` as const` in the Mini `enumObject` handler, matching the Zod 3 Classic path.

### Acceptance criteria

- [ ] Zod Mini enum output includes `as const`.
- [ ] A Zod Mini test target exists, so Mini output is generated and typechecked in CI.
- [ ] The compile-time regression fixture asserts a Mini enum member has its literal type. The assertion fails if `as const` is removed.

### Context

The affected code path was added in #3840.

aryansk:
## What

Zod Mini's `zod.enum({ ... })` renders enum object literals without an
`as const` assertion, so every member type widens to `string`/`number` and
the generated schema silently loses its literal types. `z.infer` returns
`string` instead of the member union, and `.def.entries` members read as
`string`.

Zod 4 Classic's `_enum<const T extends util.EnumLike>` preserves literals
on its own; Mini's `_enum<T extends util.EnumLike>` does not. Classic v3
already appends `as const` for exactly this reason (`#3840`); the Mini
path was missed in that PR.

## Fix

- Append ` as const` in the Mini `enumObject` handler, matching the
  Zod 3 Classic path.
- Add an `enums-mini` generation target (Zod 4 Mini output from
  `enums.yaml`), so Mini output is actually generated and typechecked in
  CI — previously no config set `variant: 'mini'`.
- Add a compile-time regression fixture
  (`tests/regressions/zod-enums-mini.ts`) asserting Mini enum members keep
  their literal types. Verified that removing the `as const` makes the
  fixture fail to compile:
  `Type 'number' is not assignable to type '4'`.

## Validation

- `bunx vitest run src/zod.test.ts` — 304 passed
- `node tests/scripts/typecheck-generated.mjs` — all 18 generated clients pass
- `bunx vitest run --config vitest.snapshots.ts` — 6095 passed

Closes #3852

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated generated Zod Mini enums to preserve enum member literal types.
  * Enum values now retain their numeric literal types in generated schemas.

* **Tests**
  * Added regression coverage confirming numeric enum members remain correctly typed.
  * Updated metadata expectations and added validation for Mini-compatible enum output.


melloware:
@the-ult please review

the-ult:
LGTM 👍🏼

melloware (approved):

PR #3859, opened 2026-08-15
fix(zod): add u flag to RegExp with Unicode property escapes

### LINKED ISSUE #3841: zod: generated RegExp drops the unicode flag, silently breaking \p{...} patterns
## What happened

The zod client emits OpenAPI `pattern` constraints as:

```ts
export const updateContactBodyFirstNameRegExp = new RegExp(
  "^[\\p{L}][\\p{L}\\p{M}'’. -]{0,99}$",
);
```

Without the `u` flag, `\p{L}` / `\p{M}` are not Unicode property escapes — they parse as the literal characters `p{L}`. The constructed RegExp is syntactically valid, so nothing throws; it just matches the wrong language. In our case every real name (`"Anna"`, `"Müller"`) failed client-side validation while the server (which anchors the same pattern with Unicode semantics) accepted them. We only caught it because a contract test suite runs the generated zod schemas against the deployed API.

## Why it matters

OpenAPI declares `pattern` to be an ECMA 262 regular expression, and Unicode property escapes are the standard way to express "letters in any script" — increasingly common for name/address fields published by .NET / Java backends. A silently-wrong validator is worse than a crash: forms reject valid user input with no error anywhere in the toolchain.

Reproduced with orval 8.19.0, `client: 'zod'`, `mode: 'tags-split'`.

## Repro

```yaml
components:
  schemas:
    Person:
      type: object
      properties:
        firstName:
          type: string
          pattern: "^[\\p{L}]+$"
```

Generated: `new RegExp("^[\\p{L}]+$")` → `.regex(...)` — then `"Anna"` fails the zod schema even though it matches the declared pattern under ECMA semantics.

## Suggested fix

Emit the flag: `new RegExp("...", "u")` (or a regex literal with `/u`). Two considerations:

- `u` tightens escaping rules, so a handful of sloppy-but-working legacy patterns (e.g. unnecessary escapes like `\-` outside classes) would start throwing at module load. If that's a concern, an opt-in (`override.zod.unicodeRegExp: true` or similar) keeps existing outputs stable while giving spec-conformant behavior to those who need it — though arguably a loud error at load beats silently matching the wrong strings, so flipping the default with a changelog note would be defensible too.
- The `v` flag would be even stricter; `u` is the conservative choice that fixes property escapes.

## Workaround for anyone hitting this

We wrap every generated pattern at the composition boundary:

```ts
export function asUnicodePattern(pattern: RegExp): RegExp {
  return pattern.flags.includes('u')
    ? pattern
    : new RegExp(pattern.source, pattern.flags + 'u');
}
```

Happy to send a PR if maintainers indicate which direction (default vs opt-in) they'd take.

aryansk:
## What

OpenAPI declares `pattern` to be an ECMA 262 regular expression, and
Unicode property escapes (`\p{...}`) are the standard way to express
"letters in any script" — common for name/address fields from .NET/Java
backends. Without the `u` flag, `\p{L}` parses as the literal characters
`p{L}`: the generated `RegExp` is syntactically valid, nothing throws, and
the zod validator silently rejects valid input (`"Anna"`, `"Müller"`).

## Fix

Emit `new RegExp('<pattern>', 'u')` when the pattern uses a Unicode
property escape (`\p{` or `\P{`), matching ECMA-262 semantics. Patterns
without property escapes are unchanged (no new flags, no risk from `u`'s
stricter escaping rules). Both RegExp emission sites (plain `pattern` and
v4 custom `format`+`pattern`) use the same helper.

## Tests

- Unit tests: flagged emission for plain patterns and for custom
  format+pattern in v4, plus a flagless negative case.
- `tests/specifications/pattern.yaml` gains a `name` property with a
  `\p{L}` pattern; the generated output now shows
  `new RegExp("^[\\p{L}][\\p{L}\\p{M}'’. -]{0,99}$", 'u')`.

## Validation

- `bunx vitest run src/zod.test.ts` — 307 passed
- `bunx vitest run --config vitest.snapshots.ts` — 6094 passed
- `node tests/scripts/typecheck-generated.mjs` — all 18 clients pass

Closes #3841

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved regular-expression validation for Unicode property escapes by correctly applying Unicode support when generating patterns.
  * Preserved existing behavior for standard regular expressions without Unicode property escapes.

* **Validation**
  * Updated pattern validation to support names beginning with a Unicode letter and allowing letters, marks, apostrophes, periods, spaces, and hyphens.

* **Documentation**
  * Updated the OpenAPI specification to require the `Node` name field and document its Unicode-aware validation rules.

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -177,7 +177,8 @@ const zodMiniCoerceCall = (fn: string, args = '') =>
 // match the literal characters `p{...}`, so the generated validator silently
 // rejects valid input (#3841). Only patterns that use them get the flag,
 // leaving existing outputs unchanged.
-const hasUnicodePropertyEscape = (pattern: string) => /\\[pP]\{/.test(pattern);
+const hasUnicodePropertyEscape = (pattern: string) =>
+  /(^|[^\\])(?:\\\\)*\\[pP]\{/.test(pattern);
 
 const buildRegExpLiteral = (pattern: string) => {
   const innerPattern = pattern.slice(
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -173,6 +173,21 @@ const zodMiniCall = (fn: string, args = '') =>
 const zodMiniCoerceCall = (fn: string, args = '') =>
   `${PURE_COMMENT}zod.coerce.${fn}(${args})`;
 
+// Unicode property escapes (`\p{...}`) require the `u` flag; without it they
+// match the literal characters `p{...}`, so the generated validator silently
+// rejects valid input (#3841). Only patterns that use them get the flag,
+// leaving existing outputs unchanged.
+const hasUnicodePropertyEscape = (pattern: string) => /\\[pP]\{/.test(pattern);
+
+const buildRegExpLiteral = (pattern: string) => {
+  const innerPattern = pattern.slice(
+    pattern.startsWith('/') ? 1 : 0,
+    pattern.endsWith('/') ? -1 : undefined,
+  );
+  const unicodeFlag = hasUnicodePropertyEscape(innerPattern) ? ", 'u'" : '';
+  return `new RegExp('${jsStringLiteralEscape(innerPattern)}'${unicodeFlag})`;
+};
+
 /** Escapes string defaults for safe embedding in template literals. */
 function formatDefaultValue(value: unknown): string {
   if (isString(value)) {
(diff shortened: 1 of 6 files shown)


melloware (changes requested):
See AI feedback

melloware:
@aryansk bump. Looks like Ai caught something are you fixing it?

melloware (approved):

PR #3860, opened 2026-08-15
fix(angular): keep Date query params in filterParams as ISO strings

### LINKED ISSUE #3856: fix(angular): filterParams silently strips Date parameters when useDates is enabled
### Description

When using the Angular client target (`client: 'angular'`) with `override.useDates: true`, query parameters typed as `Date` (e.g., `fromDate?: Date`) are silently stripped out by the generated runtime helper function `filterParams()`.

### To Reproduce

1. Define an OpenAPI spec with date query parameters:
```yaml
paths:
  /items:
    get:
      parameters:
        - name: fromDate
          in: query
          schema:
            type: string
            format: date

```

2. Configure `orval.config.ts`:

```typescript

export default defineConfig({
  api: {
    input: './openapi.yml',
    output: {
      target: './src/lib/generated/client.ts',
      client: 'angular',
      override: {
        useDates: true,
      },
    },
  },
});

```

3. Call the generated service passing a `Date` object:

```typescript
this.defaultService.getItems({ fromDate: new Date() }).subscribe();

```

4. **Result:** The query parameter `fromDate` is missing from the outbound HTTP request.

---

### Expected Behavior

`filterParams()` should preserve `Date` objects (or convert them to ISO strings using `.toISOString()`) so they can be properly passed to Angular's `HttpClient` / `HttpParams`.

---

### Actual Behavior

The generated `filterParams` runtime function checks parameters against primitive types:

```typescript
} else if (
  value != null &&
  (typeof value === 'string' ||
    typeof value === 'number' ||
    typeof value === 'boolean')
) {
  filteredParams[key] = value;
}

```

Since `typeof new Date()` returns `'object'`, any parameter typed as `Date` fails this check and is excluded from `filteredParams`.

---

### Suggested Fix

Update the Angular generator template for `filterParams()` to explicitly account for `Date` instances or convert them to ISO strings:

```typescript
} else if (value instanceof Date) {
  filteredParams[key] = value.toISOString();
} else if (
  value != null &&
  (typeof value === 'string' ||
    typeof value === 'number' ||
    typeof value === 'boolean')
) {
  filteredParams[key] = value;
}

```

---

### Environment

* **Orval version:** latest
* **Client:** `angular`

### Output client

angular

```ts

export default defineConfig({
  api: {
    input: './openapi.yml',
    output: {
      target: './src/lib/generated/client.ts',
      client: 'angular',
      override: {
        useDates: true,
      },
    },
  },
});
```

### Environment

  System:
    OS: Windows 11 10.0.26200
    CPU: (24) x64 12th Gen Intel(R) Core(TM) i7-12850HX
    Memory: 11.30 GB / 31.61 GB
  npmPackages:
    @angular/common: 22.0.6 => 22.0.6 
    @angular/core: 22.0.6 => 22.0.6 
    orval: ^8.24.0 => 8.24.0 
    react: ^18.2.0 => 18.3.1 
    zod: ^4.3.6 => 4.4.3 

```yaml
openapi: 3.0.3
info:
  title: Items API
  version: 1.0.0
  description: Simple API spec for experimenting with Orval code generation.

paths:
  /items:
    get:
      summary: Retrieve a list of items filtered by date range
      operationId: getItems
      parameters:
        - name: fromDate
          in: query
          description: Start date filter (ISO format)
          required: false
          schema:
            type: string
            format: date
            example: '2026-01-01'
        - name: endDate
          in: query
          description: End date filter (ISO format)
          required: false
          schema:
            type: string
            format: date
            example: '2026-12-31'
      responses:
        '200':
          description: A list of items matching the query parameters
          content:
            application/json:
              schema:
                type: array
                items:
        '400':
          description: Invalid query parameters provided

components:
  schemas:
    Item:
      type: object
      required:
        - id
        - name
        - createdAt
      properties:
        id:
          type: string
          format: uuid
          example: '123e4567-e89b-12d3-a456-426614174000'
        name:
          type: string
          example: 'Sample Item'
        description:
          type: string
          example: 'A brief description of the item'
        createdAt:
          type: string
          format: date-time
          example: '2026-08-14T10:00:00Z'
```

### Additional context

_No response_

aryansk:
## Fix: keep `Date` query params in `filterParams` as ISO strings

Closes #3856

### Problem

`filterParams` drops `Date` objects from query params. When a query param is
declared with `format: date` / `format: date-time` and the caller passes a JS
`Date`, the generated Angular client silently omits it from the request —
param values must be serialized server-side from an ISO string.

### Root cause

The generated filter predicate only keeps `string` / `number` / `boolean`
values (and `preserveRequiredNullables` nulls), so `Date` instances fall
through and the key is dropped.

### Fix

Added a `value instanceof Date` branch to the filter predicate in all three
emission sites (the two inline IIFE bodies and the shared helper body):

```ts
} else if (value instanceof Date) {
  filteredParams[key] = value.toISOString();
}
```

`instanceof` is a no-op for non-Date values, so the change is safe for every
generated client, not just ones with `useDates`.

### Validation

- Core unit suite: 2308 pass, including a new byte-identity regression test
  for the generated helper and an emission test that asserts Date params are
  kept.
- Full snapshot suite regenerated: 6094 snapshots green (34 updated).
- Committed sample projects (`angular-app`, `angular-query`) regenerated;
  their snapshot suites and typechecks pass.
- Angular package suite: 146 pass.

The generated-code diff is mechanical: the single added `else if` branch in
each emitted filter predicate.

## Summary by CodeRabbit

- **Bug Fixes**
  - Angular query parameters containing `Date` values are now preserved and serialized as ISO-formatted strings.
  - Date handling is consistent across supported request configurations, preventing valid parameters from being discarded.

- **New Features**
  - Added support for optional date-based query filters, including “since” filters.

- **Tests**
  - Added coverage for ISO conversion, date-filter behavior, and omission of undefined parameters.

### DIFF
--- a/docs/content/docs/guides/angular.mdx
+++ b/docs/content/docs/guides/angular.mdx
@@ -1383,6 +1383,10 @@ export const defaultFilter = (
       if (filtered.length) out[key] = filtered;
     } else if (value === null && requiredNullableKeys.has(key)) {
       out[key] = preserveRequiredNullables ? null : '';
+    } else if (value instanceof Date) {
+      // Date params are objects; convert them to ISO strings so they survive
+      // the primitive-type filter instead of being silently dropped (#3856).
+      out[key] = value.toISOString();
     } else if (
       value != null &&
       (typeof value === 'string' ||
--- a/packages/core/src/generators/options.test.ts
+++ b/packages/core/src/generators/options.test.ts
@@ -679,10 +679,23 @@ function filterParams(
       expect(body).toContain('} else if (value instanceof Date) {');
     });
 
-    it('inline IIFE converts Date values to ISO strings', () => {
-      const expression = getAngularFilteredParamsExpression('params ?? {}');
+    it('inline IIFE converts Date values to ISO strings when date params exist', () => {
+      const expression = getAngularFilteredParamsExpression(
+        'params ?? {}',
+        [],
+        false,
+        [],
+        {},
+        ['since'],
+      );
       expect(expression).toContain('value instanceof Date');
       expect(expression).toContain('value.toISOString()');
+    });
+
+    it('inline IIFE omits the Date branch for primitive-only params', () => {
+      const expression = getAngularFilteredParamsExpression('params ?? {}');
+      expect(expression).not.toContain('instanceof Date');
+      expect(expression).not.toContain('toISOString()');
     });
   });
 
@@ -977,7 +990,14 @@ function filterParams(
       // The inline IIFE is what the non-helper emission sites use; the shared
       // helper covers the hasObjectParams path above, so exercise both at
       // runtime instead of only asserting on the generated source text.
-      const expression = getAngularFilteredParamsExpression('params ?? {}');
+      const expression = getAngularFilteredParamsExpression(
+        'params ?? {}',
+        [],
+        false,
+        [],
+        {},
+        ['from'],
+      );
       const { outputText } = ts.transpileModule(expression, {
         compilerOptions: {
           module: ts.ModuleKind.CommonJS,
@@ -990,6 +1010,7 @@ function filterParams(
         .replace(/^["']use strict["'];\s*/, '')
         .replace(/;\s*$/, '');
       const runInline = (params: Record<string, unknown>) =>
+        // eslint-disable-next-line @typescript-eslint/no-implied-eval, no-new-func
         new Function('params', `return (${iife});`)(params);
 
       const inlineResult = runInline({
--- a/packages/core/src/generators/options.test.ts
+++ b/packages/core/src/generators/options.test.ts
@@ -634,6 +634,10 @@ function filterParams(
       // string so the required key still reaches the wire as \`?key=\`
       // instead of being silently dropped. See #3712.
       filteredParams[key] = preserveRequiredNullables ? null : '';
+    } else if (value instanceof Date) {
+      // Date params are objects; convert them to ISO strings so they survive
+      // the primitive-type filter instead of being silently dropped (gh #3856).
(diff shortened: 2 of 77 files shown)


the-ult:
Might be an idea to add dates to the (test) OpenAPI spec and angular sample app. And add tests to test if they appear correctly as well?

aryansk:
Thanks @the-ult — good catch, this exposed a real gap. Done in ab78ea7:

1. **Dates in the test spec + sample app**: the angular sample's petstore spec now declares a `since` query param with `format: date-time`, and the sample configs enable `useDates` so it types as `Date` in the generated clients (`SearchPetsParams.since?: Date`).
2. **Tests that they appear correctly**: added `date-query-params.spec.ts` in the sample app — it calls `searchPets` with a JS `Date` and asserts the outgoing request carries the ISO string (HTTP-testing-controller, same pattern as the required-nullable spec).
3. **Fix surfaced by your suggestion**: the inline-IIFE Date branch was emitted unconditionally, but its `value` is typed from the operation's params union, so any operation with primitive-only params (e.g. the sample's `listPets`) failed to type-check (TS2358) — the sample app build was broken. The branch is now gated on the operation having date-format query params (computed only under `useDates`), and the sample regenerates cleanly. The shared helper keeps the unconditional branch since its `value` is `unknown`.
4. Also fixed the lint failure on this PR (the runtime IIFE test was missing the eslint-disable for the Function constructor).

Validation: core suite 2312 tests pass, sample `ng test` 33/33 incl. the new spec, vitest snapshots 78/78, `vp lint` and format check clean.

melloware:
looks like SNAPSHOTS are failing

PR #3861, opened 2026-08-15
fix(orval): keep items on Swagger 2.0 formData array parameters (#3857)

### LINKED ISSUE #3857: Regression in Orval 8:  Swagger 2.0 formData array parameters type is unknown[]
### Description

I've noticed a regression between orval 7 and orval 8 where swagger 2.0 documents that use formData parameters with an 'array' type lose their type information in the generated output.  I have so far been able to debug a bit on my end and it appears that the `items: string` portion of the `schema` is lost somewhere before the `getObject` function is called.  I'm going to keep investigating and will try to find exactly where the problem is.

I have created a minimal reproducible example of the issue here: https://github.com/redsky17/orval-barebones

The example repo uses orval 8.  You will see in the generated oas2-form-data.ts, the type gets generated as:

```typescript
export type SubmitFormBody = {
  tags: unknown[];
};
```

In the generated oas3-form-data.ts file, the type is generated as (the expected): 

```typescript
export type SubmitFormBody = {
  tags: string[];
};
```

Unfortunately in my case, we are consuming another service's OAS documents and they are producing Swagger 2.0 files instead of newer 3.0 or 3.1, etc, documents.  So I'm not in a position to just update the specs to a newer version.

### Output client

fetch

```ts

// from the minimal reproduction repo

export default defineConfig({
  'oas2-form-data': {
    input: './openapi/oas2-form-data.yaml',
    output: {
      target: './src/generated/oas2-form-data.ts',
      client: 'fetch',
      mode: 'single'
    }
  },
  'oas3-form-data': {
    input: './openapi/oas3-form-data.yaml',
    output: {
      target: './src/generated/oas3-form-data.ts',
      client: 'fetch',
      mode: 'single'
    }
  }
});

```

### Environment

  System:
    OS: macOS 26.5
    CPU: (18) arm64 Apple M5 Pro
    Memory: 633.23 MB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    orval: ^8.24.0 => 8.24.0 

I was able to reproduce this on Red Hat Enterprise Linux 8 as well.

### Expected behavior

The type in this example should be `string[]` rather than `unknown[]`

### Actual behavior

The type generated is `unknown[]`

```yaml
swagger: '2.0'
info:
  title: Form Data API (OAS 2.0)
  version: '1.0.0'
basePath: /
paths:
  /submit:
    post:
      operationId: submitForm
      consumes:
        - application/x-www-form-urlencoded
      produces:
        - application/json
      parameters:
        - name: tags
          in: formData
          required: true
          type: array
          items:
            type: string
          collectionFormat: csv
      responses:
        '200':
          description: Success
          schema:
            type: object
            properties:
              ok:
                type: boolean

```

### Additional context

```yaml

# working OAS 3 document to compare with

openapi: 3.0.3
info:
  title: Form Data API (OAS 3.0)
  version: '1.0.0'
paths:
  /submit:
    post:
      operationId: submitForm
      requestBody:
        required: true
        content:
          application/x-www-form-urlencoded:
            schema:
              type: object
              required:
                - tags
              properties:
                tags:
                  type: array
                  items:
                    type: string
      responses:
        '200':
          description: Success
          content:
            application/json:
              schema:
                type: object
                properties:
                  ok:
                    type: boolean

```

aryansk:
## What

Fixes #3857 — Swagger 2.0 `formData` array parameters lost their item type during the OAS 3.1 upgrade, so generated body types degraded from `string[]` to `unknown[]`.

## Root cause

`@scalar/openapi-parser`'s `upgrade()` converts Swagger 2.0 `formData` parameters into a `requestBody` schema, but drops the parameter's `items` — `tags` arrives as `{ type: 'array' }` with no `items`, and orval falls back to `unknown[]`.

## Fix

In `import-specs.ts`, capture the item schemas of Swagger 2.0 `formData` array parameters **before** calling `upgrade()` (the upgrader also mutates its input, so the `swagger: '2.0'` marker is gone afterwards), then re-apply them to the upgraded document. Handles:

- Inline `formData` array parameters (urlencoded and multipart) — `tags: string[]`, `ids: number[]`
- Reusable parameters referenced via `$ref: '#/parameters/...'` — the upgrader promotes these to `components.requestBodies` and leaves a `$ref` in the operation's `parameters` array, which the restore resolves
- Non-string item types (`integer` → `number[]`)

Only patches array-typed properties that the upgrader left without `items`, so OAS 3 documents and any items the upgrader already preserved are untouched (guarded by a regression test).

## Validation

- 4 new unit tests in `import-specs.test.ts` covering inline, reusable-ref, non-string items, and an OAS 3 no-op guard — all proven to fail before the fix
- Full `orval` package suite: 224 tests pass; typecheck, lint, and format clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Swagger 2.0 form-data arrays so their item types are preserved correctly.
  * Added support for inline and reusable array parameters, including integer-to-number conversions.
  * Ensured path-level and operation-level parameter handling follows Swagger semantics without altering shared request definitions.
  * Preserved existing OpenAPI 3 form-data behavior.

* **Tests**
  * Added regression coverage for inline, reusable, path-level, and overridden Swagger 2.0 form-data arrays.

### DIFF
--- a/packages/orval/src/import-specs.test.ts
+++ b/packages/orval/src/import-specs.test.ts
@@ -525,6 +525,122 @@ describe('swagger2FormData', () => {
     expect(body?.model).not.toContain('unknown[]');
   });
 
+  it('keeps items on path-level inline formData array parameters (#3857)', async () => {
+    // Swagger 2.0 allows parameters on a Path Item Object; they apply to every
+    // operation under the path. The capture must merge them in.
+    const spec = {
+      swagger: '2.0',
+      info: { title: 'Path-level form data', version: '1.0.0' },
+      paths: {
+        '/submit': {
+          parameters: [
+            {
+              name: 'tags',
+              in: 'formData',
+              required: true,
+              type: 'array',
+              items: { type: 'string' },
+            },
+          ],
+          post: {
+            operationId: 'submitForm',
+            consumes: ['application/x-www-form-urlencoded'],
+            responses: { '200': { description: 'Success' } },
+          },
+        },
+      },
+    };
+    const normalizedOptions = await normalizeOptions(
+      { output: { target: '' }, input: { target: spec } },
+      'test',
+      {},
+    );
+    const result = await importSpecs('test', normalizedOptions);
+    const body = result.schemas.find((s) => s.name === 'SubmitFormBody');
+    expect(body?.model).toContain('tags: string[]');
+    expect(body?.model).not.toContain('unknown[]');
+  });
+
+  it('keeps items on path-level reusable formData parameters (#3857)', async () => {
+    const spec = {
+      swagger: '2.0',
+      info: { title: 'Path-level reusable form data', version: '1.0.0' },
+      parameters: {
+        tagList: {
+          name: 'tags',
+          in: 'formData',
+          required: true,
+          type: 'array',
+          items: { type: 'string' },
+        },
+      },
+      paths: {
+        '/submit': {
+          parameters: [{ $ref: '#/parameters/tagList' }],
+          post: {
+            operationId: 'submitForm',
+            consumes: ['multipart/form-data'],
+            responses: { '200': { description: 'Success' } },
+          },
+        },
+      },
+    };
+    const normalizedOptions = await normalizeOptions(
+      { output: { target: '' }, input: { target: spec } },
+      'test',
+      {},
+    );
+    const result = await importSpecs('test', normalizedOptions);
+    // Reusable formData parameters are promoted to a requestBody component, so
+    // the generated body type takes its name from the reusable parameter key.
+    const body = result.schemas.find((s) => s.name === 'TagListBody');
+    expect(body?.model).toContain('tags: string[]');
+    expect(body?.model).not.toContain('unknown[]');
+  });
+
+  it('lets operation-level parameters override path-level ones (#3857)', async () => {
+    const spec = {
+      swagger: '2.0',
+      info: { title: 'Path-level override', version: '1.0.0' },
+      paths: {
+        '/submit': {
+          parameters: [
+            {
+              name: 'tags',
+              in: 'formData',
+              required: true,
+              type: 'array',
+              items: { type: 'string' },
+            },
+          ],
+          post: {
+            operationId: 'submitForm',
+            consumes: ['application/x-www-form-urlencoded'],
+            parameters: [
+              {
+                name: 'tags',
+                in: 'formData',
+                required: true,
+                type: 'array',
+                items: { type: 'integer', format: 'int64' },
+              },
+            ],
+            responses: { '200': { description: 'Success' } },
+          },
+        },
+      },
+    };
+    const normalizedOptions = await normalizeOptions(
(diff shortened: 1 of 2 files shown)


aryansk:
Review feedback addressed in 7d4c5ed: the capture now merges path-item parameters into every operation (operation-level overrides win on the same name/in pair, refs resolved so a ref and its inline equivalent dedupe), and the restore resolves requestBody refs from the path item's parameters array too (the upgrader promotes path-level reusable formData params to components.requestBodies referenced there). Three new regression tests cover path-level inline, path-level reusable, and the operation-level override — the two path-level tests fail against the pre-change code.

aryansk:
Both findings are addressed in `6c3d50c`:

**1. `keyOf` stringification (Maintainability):** the merge key now requires string `in` and `name` values (`paramKey` returns `undefined` for non-strings instead of stringifying them). This fixes the two `typescript(no-base-to-string)` errors — `vp lint --type-aware --type-check packages/orval` is now 0 warnings / 0 errors (previously 2 errors at line 262).

**2. Shared path-level body corruption (Data Integrity):** the restore loop now patches Path-Item-referenced request bodies (the shared `components.requestBodies` entries promoted from reusable `#/parameters`) with only the names the operation does **not** override. An operation-level override previously wrote its items into the shared body, and the result depended on operation iteration order: whichever operation restored first claimed the shared body, so a non-overriding operation could get the override's item type.

New regression test `keeps shared path-level bodies intact when an operation overrides a parameter (#3857)` declares the **overriding** operation first (so the bug cannot hide behind ordering) and asserts the shared `TagListBody` keeps `tags: string[]` while the overriding operation's own `ReplaceFormBody` gets `tags: number[]`. It fails on `7d4c5ed` and passes on `6c3d50c`.

Validation: orval package 228 tests pass (227 → 228), `tsc --noEmit` clean, lint 0/0, formatter clean.

melloware:
@aryansk shouldn't we just wait until Scalar fixes the upstream bug?  https://github.com/scalar/scalar/issues/9908

rather than code around it?

aryansk:
Good question. Two reasons we kept the consumer-side fix:

1. **It works today regardless of upstream.** scalar/scala#9908 is still open with no fix or timeline, and orval cannot wait for a parser release to stop emitting `unknown[]` for valid Swagger 2.0 specs. This PR fixes orval at the import layer so the generated types are correct for every version of the parser currently published.

2. **It's a no-op once the upgrader is fixed.** The restore only patches array-typed properties that the upgrader left without `items`. If/when scalar's `upgrade()` starts preserving `items`, the guard sees the items already present and touches nothing — so the two fixes compose rather than conflict.

I'll keep watching scalar/scala#9908 and can drop the workaround once the upstream fix ships in a released parser version.

melloware:
Thanks for the answer!

melloware (approved):

aryansk:
Thanks for the review and merge! Glad the workaround + regression coverage made sense while the upstream Scalar fix lands.

PR #3866, opened 2026-08-15
fix(angular): type multi-content text responses as Observable<string>

the-ult:
Follow-up to #3711, which merged with two unaddressed review threads on the Angular guide. Fixing those surfaced a real typing defect in the generated multi-content-type dispatch, which is the bulk of this diff.

## Generated services no longer cast to `any`

Operations with multiple response content types generate a three-way dispatch on the `Accept` header. The `text/`/`xml` branch cast its result to `Observable<any>`, which silently disabled type checking for every consumer of those operations:

```ts
// before
return this.http.get(`/v${version}/pets/${petId}`, {
  ...options,
  responseType: 'text',
  headers,
}) as Observable<any>;
```

The adjacent blob branch already casts precisely (`as Observable<Blob>`), so this was an inconsistency rather than a constraint — `responseType: 'text'` always resolves to `Observable<string>`, which is assignable to the declared union return type. The cast is kept (it documents where `HttpClient` overload resolution is ambiguous) but narrowed to `Observable<string>`.

This removes 23 `any` casts from the Angular sample output and 26 more across the committed snapshots. Consumers gain type checking they should already have had; anyone who was relying on the `any` to assign the result somewhere unrelated will now see an error, which is the point.

## Angular guide snippets now match generated output

Three snippets in the `override.angular.baseUrl` section could not be used as written:

- The `petstore.base-url.ts` snippet used `InjectionToken`, `Provider`, and `inject` with no import statement, so copying it verbatim fails to compile.
- The TestBed example imported `PetsService` from `./api/petstore.service`, but the `tags-split` config shown directly above it emits the service at `./api/pets/pets.service`.
- The `httpResource` snippet was titled `petstore.resource.ts` for the same reason; under `tags-split` it is `pets/pets.resource.ts`.

The first two were raised on #3711; the third is the same bug in a snippet nobody flagged.

## Where to start reviewing

`packages/angular/src/http-client.ts` is the only behavioral change — one line. Everything under `samples/` and `tests/__snapshots__/` is regenerated output.

Two things worth knowing about that regenerated diff:

**It is wider than the `any` fix, and that part is currently breaking master.** The committed `base-url-token` sample predates daa07fb06 (#3828) and was never regenerated when that landed, so this branch also picks up the `@experimental` → `@remarks` note and the `hasValue(): this is ResolvedResourceState<T>` type guard.

That drift is unrelated in origin but cannot be separated from a regeneration — and it is not cosmetic. Because `pr-checks` runs `on: pull_request` only, master was never re-verified after #3828 merged, so **every PR branched off current master fails `Verify snapshot`** on `samples/angular-app/src/api/base-url-token/pets/pets.resource.ts`, whatever it changes. Demonstrated on #3867, a branch whose only change is a workflow file. This PR already contains the regeneration, which is why its own CI is green; merging it restores master. (#3867 adds a `push` trigger so this class of drift is caught at merge time instead.)

**Every changed line falls into one of those two buckets.** The full generated diff reduces to exactly four distinct line changes — the `any` → `string` cast, the doc-comment note, and the two `hasValue` guard lines — so a spot check of one service and one resource file covers the rest.

`samples/angular-app/tsconfig.spec.json` is unrelated housekeeping: the setup file registered in `angular.json` lives at the project root and was outside the spec `include`, so every `ng test` run warned that it was bundled without being type-checked.

Refs #3711

## Summary by CodeRabbit

* **Improvements**
  * Improved type safety for text and XML HTTP responses by using string-specific observable types.
  * Enhanced resource state handling with clearer type narrowing and guaranteed values after resolution.
  * Updated Angular resource guidance to reflect availability in Angular 19.2 and later.

* **Documentation**
  * Updated Angular examples and generated service paths for consistency.

* **Tests**
  * Refined test configuration and response type expectations.

### DIFF
--- a/docs/content/docs/guides/angular.mdx
+++ b/docs/content/docs/guides/angular.mdx
@@ -203,6 +203,8 @@ With `override.angular.baseUrl` set, Orval emits a sibling `<target>.base-url.ts
 file alongside the generated client:
 
 ```ts title="petstore.base-url.ts"
+import { InjectionToken, inject, type Provider } from '@angular/core';
+
 export const PETSTORE_SERVER_URL: string = 'http://petstore.swagger.io/v1';
 
 export function normalizeBaseUrl(baseUrl: string): string {
@@ -332,7 +334,7 @@ import {
   providePetstoreBaseUrl,
   providePetstoreBaseUrlResolver,
 } from './api/petstore.base-url';
-import { PetsService } from './api/petstore.service';
+import { PetsService } from './api/pets/pets.service';
 
 describe('PetsService', () => {
   let httpMock: HttpTestingController;
@@ -373,7 +375,7 @@ inside an injection context, and fall back to `options.injector.get(...)` when
 an explicit `injector` is passed — the same rule that already applies to every
 other injected dependency in generated `httpResource` functions:
 
-```ts title="petstore.resource.ts"
+```ts title="pets/pets.resource.ts"
 export function showPetByIdResource(
   petId: Signal<string>,
   accept?: ShowPetByIdAccept,
--- a/packages/angular/src/http-client.test.ts
+++ b/packages/angular/src/http-client.test.ts
@@ -1304,7 +1304,7 @@ describe('angular HttpClient generator', () => {
       expect(impl).toContain('accept?: GetPetFileAccept');
       expect(impl).toContain('Observable<Pet | string>');
       expect(impl).toContain('this.http.get<Pet>');
-      expect(impl).toContain('as Observable<any>');
+      expect(impl).toContain('as Observable<string>');
       // Content-type dispatch logic
       expect(impl).toContain("responseType: 'json'");
       expect(impl).toContain("responseType: 'text'");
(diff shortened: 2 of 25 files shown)


the-ult:
Additional context that raises the priority of this PR: the regenerated `base-url-token` fixtures here are not just drift cleanup — **they are currently failing CI on master**.

`samples/angular-app/src/api/base-url-token/pets/pets.resource.ts` has been out of sync with the generator since daa07fb06, and `pr-checks` runs `on: pull_request` only, so nothing re-verified snapshots after that merged. The result is that any PR branched off current master fails the `Verify snapshot` step regardless of what it touches.

Demonstrated on #3867, a branch whose only change is adding a CI step — it fails with:

```
FAIL angular-app api-generation.spec.ts > API Generation Snapshots >
     samples/angular-app/src/api/base-url-token/pets/pets.resource.ts
```

This PR already contains the regeneration that fixes it, which is why its own CI is green. Merging it should restore master for everyone else.

the-ult:
@melloware this PR should have priority

melloware (approved):

PR #3872, opened 2026-08-16
perf(core): avoid repeated dependency scans

### LINKED ISSUE #3805: Optimize performance
### Description

I think we should improve performance. Config like this:

```typescript

import { execSync } from 'child_process';

const header = ({ title, description, version }) => [
  'Generated by orval 🍺',
  'Do not edit manually.',
  ...(title ? [title] : []),
  ...(description ? [description] : []),
  ...(version ? [`OpenAPI spec version: ${version}`] : []),
];

export default defineConfig({
  main: {
    input: 'https://.../schema/openapi.json',
    output: {
      schemas: 'projects/api/src/lib/main/schemas',
      target: 'projects/api/src/lib/main/target',
      client: 'angular-query',
      httpClient: 'angular',
      mode: 'tags',
      clean: true,
      override: {
        header,
        useTypeOverInterfaces: false,
      },
    },
    hooks: {
      afterAllFilesWrite: (paths) => {
        console.log('Running lint fix and patching generated files...');
        execSync(`bun scripts/fix-orval-params.ts ${paths.join(' ')}`);
        execSync(`bun run prettier --write ${paths.join(' ')}`);
        execSync(`bun run eslint --fix ${paths.join(' ')}`);
      },
    },
  },
  allauth1: {
    input: 'https://.../api/openapi.json',
    output: {
      schemas: 'projects/api/src/lib/allauth1/schemas',
      target: 'projects/api/src/lib/allauth1/target',
      client: 'angular-query',
      httpClient: 'angular',
      mode: 'tags',
      clean: true,
      override: {
        header,
        useTypeOverInterfaces: false,
      },
    },
    hooks: {
      afterAllFilesWrite: (paths) => {
        console.log('Running lint fix and patching generated files...');
        execSync(`bun scripts/fix-orval-params.ts ${paths.join(' ')}`);
        execSync(`bun run prettier --write ${paths.join(' ')}`);
        execSync(`bun run eslint --fix ${paths.join(' ')}`);
      },
    },
  },
  allauth2: {
    input: 'https://.../api/v2/openapi.json',
    output: {
      schemas: 'projects/api/src/lib/allauth2/schemas',
      target: 'projects/api/src/lib/allauth2/target',
      client: 'angular-query',
      httpClient: 'angular',
      mode: 'tags',
      clean: true,
      override: {
        header,
        useTypeOverInterfaces: false,
      },
    },
    hooks: {
      afterAllFilesWrite: (paths) => {
        console.log('Running lint fix and patching generated files...');
        execSync(`bun scripts/fix-orval-params.ts ${paths.join(' ')}`);
        execSync(`bun run prettier --write ${paths.join(' ')}`);
        execSync(`bun run eslint --fix ${paths.join(' ')}`);
      },
    },
  },
});

```

usually takes omething like 20 seconds. I expect something like 500ms in those cases. Shemas are not that big to take so much time ...

I double checked it with ai and there is a slow part in codegen ...

### Current behavior

Generation usually takes around 20 seconds which seems to me too much. My schemas are not that large...

### Proposed behavior

I expect time in milliseconds not in seconds ...

### Proposed solution

Focus on bottlenecks in the codebase asn measure it and add performance becnhmarks in readme etc...

jaydaVis04:
Avoids repeatedly constructing regular expressions and scanning the full generated implementation once per possible dependency export.

`addDependency` now tokenizes referenced identifiers once and checks each candidate import with `Set.has`, preserving the existing alias matching behavior and avoiding partial identifier matches.

Refs #3805

### Validation

- `vp exec vitest run packages/core/src/generators/imports.test.ts --config packages/core/vitest.config.ts --pool forks --reporter verbose` - passed
- `vp exec tsc --noEmit -p packages/core/tsconfig.json` - passed
- `vp run -F @orval/core test` - passed outside sandbox after sandbox IPC shared-memory failure
- `vp fmt --check packages/core/src/generators/imports.ts packages/core/src/generators/imports.test.ts` - passed
- `git diff --check` - passed

### Performance

Quick local synthetic check over 1000 candidate exports and a 2000-reference implementation:

- repeated regex scan median: 18.149ms
- single token Set scan median: 0.771ms
- matched dependency count: 1000 before and after

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed dependency imports being added incorrectly when an export name appeared only within a longer identifier.
  * Improved recognition of referenced symbols, including identifiers containing Unicode characters and aliases.
  * Ensured generated imports are more accurate and avoid bringing in unrelated symbols.

* **Tests**
  * Added regression coverage for exact symbol matching.
  * Added coverage confirming referenced symbols are tracked consistently across dependency groups.


packages/core/src/generators/imports.ts
```diff
@@ -234,13 +233,21 @@ interface AddDependencyOptions {
...
+}
+
 export function addDependency({
   implementation,
   exports,
   dependency,
   projectName,
   isAllowSyntheticDefaultImports,
 }: AddDependencyOptions) {
+  const referencedIdentifiers = getReferencedIdentifiers(implementation);
```
melloware: Maybe I am not seeing this right but doesn't this just create a SET with 1 value in it each time this is called? Its not storing them in a Set that is persistent?

melloware (changes requested):

packages/core/src/generators/imports.ts
```diff
@@ -234,13 +233,21 @@ interface AddDependencyOptions {
   isAllowSyntheticDefaultImports: boolean;
 }
 
+function getReferencedIdentifiers(implementation: string): Set<string> {
+  return new Set(
```
melloware, line 244: Isn't this just creating a new Set with 1 value every time not a persistent collection over time?

melloware (changes requested):

melloware:
@coderabbitai review

melloware (approved):

PR #3873, opened 2026-08-16
fix(fetch): spread Headers instance correctly when merging custom headers

### LINKED ISSUE #2375: Headers not correctly passed to fetch client
I noticed that my headers were missing when I call myFun({ headers }). The myFun was generated with orval. In my case, headers has type Headers. I looked into the generated code and found this line of code which causes the problem:

headers: { 'Content-Type': 'application/json', ...options?.headers }

The problem is that {...headers} is always an empty object.

### LINKED ISSUE #3880: Generated endpoint headers are not passed to `fetch` for a request with a body if `headers` is a `Headers` object or an array
### Description

Here is the generated code:

```
export const createConsumerSearch = async (
  createSearchRequest: CreateSearchRequest,
  options?: RequestInit,
): Promise<createConsumerSearchResponseSuccess> => {
  return customFetch<createConsumerSearchResponseSuccess>(getCreateConsumerSearchUrl(), {
    ...options,
    method: "POST",
    headers: { "Content-Type": "application/json", ...options?.headers },
    body: JSON.stringify(createSearchRequest),
  });
};
```

The spread of `{ "Content-Type": "application/json", ...options?.headers }` does not work if `headers` is anything other than a bare object. However, the `RequestInit` type additionally allows for an array of tuples or a Headers object:

```
interface RequestInit {
    /** A Headers object, an object literal, or an array of two-item arrays to set request's headers. */
    headers?: HeadersInit;
...
}

type HeadersInit = [string, string][] | Record<string, string> | Headers;
```

With a `Headers` object, all values are dropped. If an array is used, the literal headers `0`, `1`, etc are added instead of the correct key/value pairs. This is trivially seen in the console:

```
let testArr = [["foo", "one"], ["bar", "two"]];
console.log({ "Content-Type": "application/json", ...testArr } );
--> {"0": ["foo", "one"], "1": ["bar", "two"], "Content-Type": "application/json"}

let testHeaders = new Headers();
testHeaders.set("foo", "one");
testHeaders.set("bar", "two");
console.log({ "Content-Type": "application/json", ...testHeaders } );
--> { "Content-Type": "application/json" }
```
### Output client

fetch

```ts
export default defineConfig({
  consumer: {
    input: {
      target: process.env.OPENAPI_URL ?? "http://localhost:8000/apidoc/consumer/openapi.json",
    },
    output: {
      indexFiles: true,
      target: "./src/lib/api/generated/consumer/",
      client: "react-query",
      httpClient: "fetch",
      mode: "tags-split",
      clean: true,
      baseUrl: {
        runtime: "getClientEnv().VITE_API_URL",
        imports: [{ name: "getClientEnv", importPath: "@/lib/env/client" }],
      },
      override: {
        fetch: {
          includeHttpResponseReturnType: true,
          forceSuccessResponse: true,
        },
        mutator: {
          path: "./src/lib/api/customFetch.ts",
          name: "customFetch",
        },
      },
    },
  }
});
```

### Environment

  System:
    OS: macOS 26.6.1
    CPU: (15) arm64 Apple M5 Pro
    Memory: 70.08 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    @tanstack/react-query: ^5.101.2 => 5.101.2 
    orval: ^8.20.0 => 8.20.0 
    react: ^19.2.7 => 19.2.7 

### Expected behavior

Headers supplied as an array or a Headers object should be used in the fetch call.

### Actual behavior

Headers are either ignored (for the Headers object) or are sent over as literal index strings "0", "1", etc

```yaml
openapi: 3.0.3
info:
  title: Demo
  version: 0.0.0
paths: {}
```

### Additional context

_No response_

ErfanBagheri404:
## Summary

When a generated **fetch** client has default headers (e.g. a hardcoded Content-Type), user-supplied headers are merged in via object spread:

```ts
```

Object spread on a `Headers` instance yields `{}`, because `Headers` has no own enumerable properties — its entries are only accessible via `forEach` / `entries()`. When a caller passes a real `Headers` object (the standard way to call `fetch`), all their custom headers are silently dropped.

## Fix

Emit a per-operation helper that normalizes all three valid `HeadersInit` forms into a plain object before spreading:

```ts
const getHeaders = (h: HeadersInit | Headers): Record<string, string> => {
  if (!h) return {};
  if (h instanceof Headers) return Object.fromEntries(h.entries());
  if (Array.isArray(h)) return Object.fromEntries(h);
  return h;
};
```

and use `...getHeaders(options?.headers)` instead of `...options?.headers`.

## Verification

The full snapshot suite (6179 tests) passes with 137 snapshots updated to reflect the new output.

Fixes orval-labs/orval#2375

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request header handling across generated fetch clients and sample applications.
  * Caller-provided headers now work consistently when supplied as `Headers`, tuple arrays, or plain objects.
  * Existing custom headers are preserved while default JSON content headers are applied.

### DIFF
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -107,7 +107,7 @@ export const generateRequestFunction = (
   const isFormData = !override.formData.disabled;
   const isFormUrlEncoded = override.formUrlEncoded !== false;
 
-  const GET_HEADERS_HELPER = `  const getHeaders = (h: HeadersInit | Headers): Record<string, string> => {
+  const GET_HEADERS_HELPER = `  const getHeaders = (h?: HeadersInit | Headers): Record<string, string> => {
     if (!h) return {};
     if (h instanceof Headers) return Object.fromEntries(h.entries());
     if (Array.isArray(h)) return Object.fromEntries(h);
@@ -691,7 +691,7 @@ ${override.fetch.forceSuccessResponse && hasSuccess ? '' : `export type ${respon
   `;
   if (mutator?.isHook) {
     fetchImplementation = `export const use${pascal(operationName)}Hook = (): (${args}) => ${returnType} => {
-    ${fetchImplementationBody}}
+    ${fetchHeadersOption ? GET_HEADERS_HELPER : ''}${fetchImplementationBody}}
   `;
   }
 
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -107,6 +107,14 @@ export const generateRequestFunction = (
   const isFormData = !override.formData.disabled;
   const isFormUrlEncoded = override.formUrlEncoded !== false;
 
+  const GET_HEADERS_HELPER = `  const getHeaders = (h: HeadersInit | Headers): Record<string, string> => {
+    if (!h) return {};
+    if (h instanceof Headers) return Object.fromEntries(h.entries());
+    if (Array.isArray(h)) return Object.fromEntries(h);
+    return h;
+  };
+`;
+
   const getUrlFnName = camel(`get-${operationName}-url`);
   const getUrlFnProps = toObjectString(
     props.filter(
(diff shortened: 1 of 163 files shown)


melloware (commented):
Please see AI feedback

melloware:
SNAPSHOT failing as expected now

image

melloware (approved):

PR #3875, opened 2026-08-17
fix(zod): emit zod.never() for keys forbidden by not in oneOf/anyOf

### LINKED ISSUE #3780: Zod: constraint-only oneOf branches with sibling properties generate zod.unknown() and lose validation
## Description

A valid OpenAPI 3.1 object schema can define property types in a common `properties` block and use constraint-only `oneOf` branches to define conditional required or forbidden fields.

Orval generates these branches as `zod.unknown()`, causing the generated validator to accept values rejected by the OpenAPI schema.

## Current configuration

**Output client:** `zod`

### `orval.config.ts`

```ts

export default defineConfig({
  reproduction: {
    input: './openapi.yaml',
    output: {
      client: 'zod',
      target: './generated.ts',
      override: {
        zod: {
          version: 3,
          generate: {
            body: true,
            response: false,
            query: false,
            param: false,
            header: false,
          },
        },
      },
    },
  },
});
```

Generation command:

```sh
pnpm exec orval --config ./orval.config.ts
```

### Environment

```text
System:
  OS: Windows 11 10.0.26200

npmPackages:
  @tanstack/vue-query: ^5.83.1 => 5.85.6
  axios: ^1.18.0 => 1.18.0
  msw: ^2.11.5 => 2.11.5
  orval: ^8.23.0 => 8.23.0
  vue: ^3.5.35 => 3.5.35
  zod: ^3.25.76 => 3.25.76
```

## Swagger

```yaml
openapi: 3.1.0

info:
  title: Constraint-only oneOf reproduction
  version: 1.0.0

paths:
  /examples:
    post:
      operationId: createExample
      responses:
        '204':
          description: No Content
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/CreateExampleRequest'

components:
  schemas:
    CreateExampleRequest:
      type: object

      oneOf:
        - title: AB
          required:
            - A
            - B
          not:
            anyOf:
              - required:
                  - X
              - required:
                  - Y

        - title: XY
          required:
            - X
            - Y

      properties:
        A:
          type: string
        B:
          type: integer
        X:
          type: string
        Y:
          type: integer
```

The document was validated with the command requested by the issue template:

```sh
npx --yes @scalar/cli document validate ./openapi.yaml
```

Result:

```text
Matches the OpenAPI specification (OpenAPI 3.1)
File validated in 80 ms
```

The schema describes two mutually exclusive cases:

- `AB`: `A` and `B` are required; `X` and `Y` are forbidden.
- `XY`: `X` and `Y` are required; `A` and `B` remain optional.

## Generated Zod

Orval 8.23.0 generates:

```ts
export const CreateExampleBody = zod.union([zod.unknown(),zod.unknown()]).and(zod.object({
  "A": zod.string().optional(),
  "B": zod.number().int().optional(),
  "X": zod.string().optional(),
  "Y": zod.number().int().optional()
}))
```

Runtime reproduction:

```ts
const cases = {
  empty: {},
  incompleteXY: {
    X: 'x',
  },
  validAB: {
    A: 'a',
    B: 1,
  },
  validXY: {
    X: 'x',
    Y: 1,
  },
  invalidMixed: {
    A: 'a',
    B: 1,
    X: 'x',
  },
};

console.log(
  Object.fromEntries(
    Object.entries(cases).map(([name, value]) => [
      name,
      CreateExampleBody.safeParse(value).success,
    ]),
  ),
);
```

Actual result:

```json
{
  "empty": true,
  "incompleteXY": true,
  "validAB": true,
  "validXY": true,
  "invalidMixed": true
}
```

The generated schema incorrectly accepts:

- `empty`, which matches neither branch.
- `incompleteXY`, because `Y` is missing.
- `invalidMixed`, because the `AB` branch forbids `X`, while the `XY` branch requires `Y`.

## Expected Zod

The generated schema should preserve the common property types and the branch constraints.

A semantically equivalent Zod schema for JSON payloads would be:

```ts
const AB = zod.object({
  A: zod.string(),
  B: zod.number().int(),
  X: zod.never().optional(),
  Y: zod.never().optional(),
});

const XY = zod.object({
  A: zod.string().optional(),
  B: zod.number().int().optional(),
  X: zod.string(),
  Y: zod.number().int(),
});

export const CreateExampleBody = zod.union([AB, XY]);
```

Expected result:

```json
{
  "empty": false,
  "incompleteXY": false,
  "validAB": true,
  "validXY": true,
  "invalidMixed": false
}
```

The expected results were verified both with the Zod schema above and with Ajv 8.20.0 using its JSON Schema 2020-12 validator.

If Orval cannot preserve a supported OpenAPI constraint, generation should report an unsupported constraint instead of silently replacing the branch with `zod.unknown()`.

ErfanBagheri404:
## Summary

A oneOf/anyOf member that carries `not: { anyOf: [{ required: [X] }, { required: [Y] }] }` forbids X and Y on that branch. When such a member is rendered against the composing schema's sibling properties (#3780's constraint-only member fix), the forbidden keys were emitted as ordinary optional properties - silently dropping the constraint.

This PR emits `zod.never().optional()` for forbidden keys so any supplied value fails validation while allowing the key's absence (the correct semantics).

## Changes

- `getForbiddenKeys()`: new helper that extracts forbidden key names from `not.anyOf[].required` (and the bare `not.required` form).
- `withSiblingProperties()`: threads forbidden keys as pre-generated property overrides rendered as `zod.never().optional()`.
- `case 'never':` added to the zod type dispatch for the property override path.

## Generated output (after)

```ts
zod.union([
  zod.object({
    A: zod.string(),
    B: zod.number().int(),
    X: zod.never().optional(),
    Y: zod.never().optional(),
  }),
  zod.object({
    A: zod.string().optional(),
    B: zod.number().int().optional(),
    X: zod.string(),
    Y: zod.number().int(),
  }),
])
```

Closes #3780

## Summary by CodeRabbit

* **New Features**
  * Improved generated schema handling for union constraints, including constraint-only union members.
  * Forbidden properties now reject supplied values while remaining optional when omitted.
  * Added support for rendering `never` constraints in generated Zod definitions.
  * Generated branch schemas now consistently enforce these property restrictions.

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -344,7 +344,7 @@ const getForbiddenKeys = (
   // common OpenAPI tooling. A bare `not: { required: [X] }` is accepted too.
   const branches = Array.isArray(not.anyOf)
     ? not.anyOf
-    : isObject(not.required)
+    : Array.isArray(not.required)
       ? [not]
       : [];
 
@@ -852,12 +852,23 @@ export const generateZodValidationSchemaDefinition = (
       // sibling properties, forbidden keys must be emitted as `zod.never()`
       // (optional, so absence passes and any value fails). These are injected
       // as pre-generated property overrides into the object schema below. (#3780)
-      let propertyOverrides: Record<string, ZodValidationSchemaDefinition> | undefined;
+      let propertyOverrides:
+        | Record<string, ZodValidationSchemaDefinition>
+        | undefined;
       if (forbidden) {
         propertyOverrides = {};
         for (const key of forbidden) {
+          // A key that is both forbidden and required contradicts itself; emit
+          // the required form (no optional) so the `never` still rejects any
+          // value while the key is expected to be absent.
+          const isForbiddenAndRequired = required.includes(key);
           propertyOverrides[key] = {
-            functions: [['never', undefined], ['optional', undefined]],
+            functions: [
+              ['never', undefined],
+              ...(isForbiddenAndRequired
+                ? []
+                : ([['optional', undefined]] as const)),
+            ],
             consts: [],
           } as ZodValidationSchemaDefinition;
         }
@@ -885,9 +896,11 @@ export const generateZodValidationSchemaDefinition = (
         {
           required: true,
           additionalRequired: allOfRequired,
-          propertyOverrides: (withSiblingProperties(schema) as OpenApiSchemaObject)[
-            'x-orval-property-overrides'
-          ] as Record<string, ZodValidationSchemaDefinition> | undefined,
+          propertyOverrides: (
+            withSiblingProperties(schema) as OpenApiSchemaObject
+          )['x-orval-property-overrides'] as
+            | Record<string, ZodValidationSchemaDefinition>
+            | undefined,
           constNameRegistry,
           useReusableSchemas,
           urlEncoded,
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -330,6 +330,35 @@ const isConstraintOnlyMember = (
   return Object.keys(schema).every((key) => SHAPELESS_MEMBER_KEYS.has(key));
 };
 
+// Extracts the property names a member forbids via `not`. JSON Schema forbids a
+// key when `not` lists any subschema that `required`s it — i.e.
+// `not: { anyOf: [{ required: [X] }, { required: [Y] }] }` means X and Y must
+// be absent. Returns `undefined` when the member has no such constraint. (#3780)
+const getForbiddenKeys = (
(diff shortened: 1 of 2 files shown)


melloware (changes requested):
Please see AI feedback and build is failing

melloware:
Looks like tests are failing?

ErfanBagheri404:
> Looks like tests are failing?

looking into it

melloware:
@ErfanBagheri404 weird but it looks like no files have changed in this PR?

melloware (approved):

PR #3877, opened 2026-08-18
fix(orval): skip symlinked-mock test on EPERM (Windows)

### LINKED ISSUE #3874: Symlink regression test fails with EPERM on Windows without Developer Mode
### Summary

The symlink regression test added in #3814 hard-fails on Windows for contributors who do not have the symlink-creation privilege:

```
FAIL  src/generate-spec.test.ts > generateSpec - clean skips a symlinked mock directory > does not touch files in the target of a symlinked mock.path
Error: EPERM: operation not permitted, symlink 'C:\Users\<omitted>\AppData\Local\Temp\orval-gen-spec-p92IIl' -> 'C:\Users\<omitted>\AppData\Local\Temp\orval-gen-spec-TB49Q0\src\mocks'
```

Reported by @snebjorn in https://github.com/orval-labs/orval/pull/3814#discussion_r3792860163.

### Cause

`packages/orval/src/generate-spec.test.ts` calls `fs.symlink(outsideTarget, mockLink, 'dir')`. On Windows, creating a symlink requires `SeCreateSymbolicLinkPrivilege`, granted only to elevated processes or when Developer Mode is enabled. Without either, Node throws `EPERM`.

This is why CI stays green: GitHub's `windows-latest` runner executes as an administrator, so the `pr-checks (windows-latest, 22.x)` leg passes (e.g. master run [31954991545](https://github.com/orval-labs/orval/actions/runs/31954991545)). Only unprivileged local Windows machines hit it.

### Proposed fix

Make the test skip itself when the platform refuses the symlink, instead of failing:

- Wrap the `fs.symlink` call in a `try`/`catch` and `ctx.skip()` when the error code is `EPERM`.
- Prefer this over a blanket `it.skipIf(process.platform === 'win32')`, which would silently drop Windows coverage of a Windows-relevant code path — the test should still run on CI and for devs with Developer Mode enabled.

### Workaround until then

Enable Developer Mode (Settings → System → For developers), or run the test suite from an elevated terminal.

ErfanBagheri404:
## Summary
Fixes #3874 — the symlinked-mock regression test hard-fails on unprivileged Windows machines with `EPERM: operation not permitted`.

## Change
`packages/orval/src/generate-spec.test.ts` — wrap the `fs.symlink(...)` call in a `try/catch`. When the error code is `EPERM` (no symlink privilege / Developer Mode off on Windows), the test now calls `ctx.skip()` instead of failing. Other errors still propagate.

This is the approach the issue itself recommends (targeted skip on EPERM, rather than a blanket `it.skipIf(process.platform === 'win32')` which would silently drop Windows coverage of a Windows-specific code path).

## Test
`bunx vitest run packages/orval/src/generate-spec.test.ts` — 52/52 pass.

## Summary by CodeRabbit

* **Tests**
  * Improved compatibility for symlink-related tests on systems with restricted permissions.
  * Tests now skip setup gracefully when supported link creation is unavailable while continuing to report unexpected errors.
  * Retained checks confirming that files in linked targets remain unchanged.

### DIFF
--- a/packages/orval/src/generate-spec.test.ts
+++ b/packages/orval/src/generate-spec.test.ts
@@ -2473,7 +2473,9 @@ describe('generateSpec - clean skips a symlinked mock directory', () => {
   // it straight through to the link's target, which can sit outside the
   // workspace entirely. The fix lstats each mock root first and skips
   // cleaning it (with a warning) when it is a symbolic link.
-  it('does not touch files in the target of a symlinked mock.path', async () => {
+  it('does not touch files in the target of a symlinked mock.path', async ({
+    skip,
+  }) => {
     const workspace = await createTempWorkspace();
     const outsideTarget = await createTempWorkspace();
 
@@ -2483,7 +2485,19 @@ describe('generateSpec - clean skips a symlinked mock directory', () => {
 
       const mockLink = path.join(workspace, 'src/mocks');
       await fs.ensureDir(path.dirname(mockLink));
-      await fs.symlink(outsideTarget, mockLink, 'dir');
+      try {
+        // A junction, not a true symbolic link: Windows grants
+        // `SeCreateSymbolicLinkPrivilege` only to elevated tokens or under
+        // Developer Mode, but a junction needs no privilege. libuv reports any
+        // reparse point as a symlink under lstat, so the guard under test sees
+        // the same thing either way. Ignored on POSIX.
+        await fs.symlink(outsideTarget, mockLink, 'junction');
+      } catch (error: unknown) {
+        // Filesystems without reparse-point support (FAT32, some network mounts).
+        const code = (error as NodeJS.ErrnoException)?.code;
+        if (code === 'EPERM' || code === 'ENOTSUP') skip();
+        else throw error;
+      }
 
       // A stale-looking file already sitting in the symlink target, matching
       // the mock cleanup patterns, so a naive fix (just widening the glob


the-ult:
> *This was generated by AI during triage.*

Thanks for picking this up, @ErfanBagheri404 — and for reading #3874 closely enough to avoid the blanket `skipIf`. Two things to address before this can go in: one is a bug in the diff, the other is a better fix that makes the skip unnecessary.

### 1. The catch block does not skip — it passes silently

The PR description says the test "calls `ctx.skip()`", but the diff does a bare `return`:

```ts
if ((err as { code?: string })?.code === 'EPERM') return;
```

A `return` in a vitest test body ends the test with **zero assertions**, and the runner reports it as **passed**. That is worse than the current failure: a red test tells a Windows contributor that something needs attention, whereas a silent green tells them the code path is covered when it is not. `ctx.skip()` is what makes the runner print the test as skipped.

That is also why the arrow function was converted to `function () {}` — but `ctx` is never taken as a parameter, so the conversion currently buys nothing. Vitest's test context is the **first argument** to the callback, not `this`; the arrow function can stay:

```ts
it('does not touch files in the target of a symlinked mock.path', async ({ skip }) => {
```

### 2. A junction removes the need to skip at all

The root cause is narrower than "Windows cannot make links". Windows has two kinds of directory link, and only one of them needs a privilege:

| | privilege needed | reported by `lstat` as |
|---|---|---|
| symbolic link (`'dir'`) | `SeCreateSymbolicLinkPrivilege` — elevated token or Developer Mode | symlink |
| **junction** (`'junction'`) | **none** | symlink |

libuv reports **any** reparse point as `S_IFLNK` under `lstat`. From `fs__stat_assign_statbuf` in [`libuv/src/win/fs.c`](https://github.com/libuv/libuv/blob/v1.x/src/win/fs.c):

```c
if (do_lstat && (stat_info.FileAttributes & FILE_ATTRIBUTE_REPARSE_POINT)) {
  statbuf->st_mode |= S_IFLNK;
```

So the production guard this test protects — the `lstat(...).isSymbolicLink()` check on each mock root — behaves identically for a junction. The change is one word:

```ts
await fs.symlink(outsideTarget, mockLink, 'junction');
```

`'junction'` is ignored on POSIX (Node creates an ordinary symbolic link), so no `process.platform` branch is needed. Verified on macOS: `lstat(...).isSymbolicLink()` is `true` for a junction-typed link, and the test passes unchanged, still emitting `Skipped cleaning "…": the configured mock path is a symbolic link.`

Two reasons this beats the skip:

- **Windows contributors keep the coverage.** Nobody has to enable Developer Mode or run an elevated terminal to exercise a Windows-relevant code path.
- **It tests the more realistic case.** Junctions are what Windows users actually end up with — `npm`/`pnpm` create them for linked packages, and Link Shell Extension (which @snebjorn mentioned on #3874) creates them too.

Worth noting for context: @snebjorn is an Administrators-group member and *still* hit `EPERM`. That is expected — UAC strips `SeCreateSymbolicLinkPrivilege` from the unelevated token, so group membership alone never grants it. Developer Mode works only because libuv passes `SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE` in `fs__symlink`, with a retry-without-the-flag fallback on `ERROR_INVALID_PARAMETER`.

### Suggested shape

Junction as the fix, `ctx.skip()` as a genuine fallback for filesystems with no reparse-point support at all (FAT32, some network mounts):

```ts
it('does not touch files in the target of a symlinked mock.path', async ({ skip }) => {
  // ...
  try {
    // A junction, not a true symbolic link: Windows grants
    // `SeCreateSymbolicLinkPrivilege` only to elevated tokens or under
    // Developer Mode, but a junction needs no privilege. libuv reports any
    // reparse point as a symlink under lstat, so the guard under test sees
    // the same thing either way. Ignored on POSIX.
    await fs.symlink(outsideTarget, mockLink, 'junction');
  } catch (error) {
    // Filesystems without reparse-point support (FAT32, some network mounts).
    const code = (error as NodeJS.ErrnoException).code;
    if (code === 'EPERM' || code === 'ENOTSUP') skip();
    else throw error;
  }
  // ...
});
```

Please also confirm the suite passes on your Windows machine after the change — that is the one thing none of us can verify from macOS or from CI (the `windows-latest` runner is elevated, which is why this never went red upstream).

I hope this helps 👍🏼 

---

**References**
- Issue: #3874 (an agent brief with full acceptance criteria is now posted there)
- Introduced by: #3814, commit `6391b420a`
- Original report: https://github.com/orval-labs/orval/pull/3814#discussion_r3792860163
- Prior art for a platform-specific test fix in this same suite: #3742
- libuv `fs__symlink` / `fs__stat_assign_statbuf`: https://github.com/libuv/libuv/blob/v1.x/src/win/fs.c

snebjorn:
Can confirm the test now pass on my machine

packages/orval/src/generate-spec.test.ts
```diff
@@ -2473,7 +2473,7 @@ describe('generateSpec - clean skips a symlinked mock directory', () => {
   // it straight through to the link's target, which can sit outside the
   // workspace entirely. The fix lstats each mock root first and skips
   // cleaning it (with a warning) when it is a symbolic link.
-  it('does not touch files in the target of a symlinked mock.path', async () => {
+  it('does not touch files in the target of a symlinked mock.path', async function () {
```
snebjorn: why change to function?

melloware:
@the-ult  is this one ready to go?

the-ult:
I think so, but I can't test it, since I don't have windows.

@snebjorn I think you might be a better judge 👍

melloware (approved):

PR #3879, opened 2026-08-18
Vite plus upgrade

snebjorn:
Upgraded vite+ and changed vitest references to the [vite+ recommend way](https://viteplus.dev/guide/migrate#vitest)

Also:
- fixed .svelte-kit folder in mono root
- removed extra lint config that is on by default
- correctly prefix eslint rule names
- removed direct vite pkg where not needed

## Summary by CodeRabbit

* **Documentation**
  * Added Vite Plus usage guidance and formatter settings for JavaScript and TypeScript projects.
* **Refactor**
  * Standardized development, build, and test workflows around Vite Plus commands.
  * Updated sample projects to use the standardized workflow.
* **Tests**
  * Migrated test utilities and snapshot configurations while preserving existing coverage and behavior.
* **Chores**
  * Removed redundant test configuration files and updated tooling versions and package configuration.


packages/core/src/utils/resolve-version.test.ts
snebjorn, line 1: @ScriptType can you verify that the changes I made in here still does the correct thing

melloware (approved):

PR #1085, opened 2023-12-01
fix: do not use TVariables

Maxim-Mazurok:
## Status

**READY**

## Description

Fixes #857

Looks like something changed in Vue Query typings. Also `void` is known to be problematic in TS, and the general recommendation is to only use it for functions. Clearly `TVariables` wasn't a function, perhaps this is the source of error.

Also adds a route for reproduction. Here's an error without the actual `index.ts` change:
image

## Todos

- [x] Tests
- [x] Documentation
- [ ] Changelog Entry (unreleased)

## Steps to Test or Reproduce

1. clone
2. revert changes to `index.ts`
3. run all tests `cd ~/orval && yarn && yarn build && yarn update-samples && yarn format && yarn test:ci && cd tests && yarn && yarn generate && yarn build && cd ..`

PR #3883, opened 2026-08-22
fix(samples): unbreak six excluded react-query sample tests

### LINKED ISSUE #3868: react-query samples excluded from test:samples are broken: TS5112 under TypeScript 6, plus a mutator/httpClient mismatch in form-data-mutator
Six `react-query` samples are excluded from the root `test:samples` script, and their `test` scripts do not currently run anywhere. Checking why turned up a chain of problems: the scripts fail outright under TypeScript 6, and fixing that surfaces further errors in four of them — including one that looks like a genuine mutator/`httpClient` mismatch.

Surfaced while wiring sample tests into CI in #3867; that PR deliberately leaves these exclusions in place so it can land without dragging this in.

## The exclusions

```jsonc
// package.json
"test:samples": "vp run -F './samples/**' -F '!react-query-form-data' -F '!react-query-form-url-encoded' -F '!react-query-form-url-encoded-mutator' -F '!react-query-form-data-mutator' -F '!react-query-hook-mutator-axios' -F '!react-query-hook-mutator-fetch' test"
```

The filters date back to #1085 (Dec 2023), long before the failures below existed, so nothing has re-checked them since.

## 1. All six fail under TypeScript 6 (`TS5112`)

All six share the same script, `tsc --noEmit endpoints.ts`:

```
$ cd samples/react-query/form-data && tsc --noEmit endpoints.ts
error TS5112: tsconfig.json is present but will not be loaded if files are specified on commandline.
Use '--ignoreConfig' to skip this error.
```

Reproduced on all six with the repo's pinned `typescript@6.0.3`. None of these directories has its own `tsconfig.json` — TypeScript resolves the **repo-root** one, and TS6 now treats "config found + filenames on the command line" as a hard error.

The sibling samples `react-query-basic` and `react-query-custom-fetch` are unaffected because they use plain `tsc --noEmit` with no filename argument.

Note that plain `tsc --noEmit` is *not* the fix here: with no local `tsconfig.json`, it inherits the root config and type-checks the entire monorepo, including `docs/`. Adding `--ignoreConfig` as the error suggests does resolve this specific error.

## 2. With that resolved, four of the six still fail

```
$ tsc --noEmit --ignoreConfig endpoints.ts
```

| Sample | Result |
|---|---|
| `form-data` | ✅ passes |
| `form-url-encoded` | ✅ passes |
| `form-data-mutator` | ❌ `TS2554` |
| `form-url-encoded-mutator` | ❌ `TS2769` |
| `hook-mutator-axios` | ❌ `TS7016` |
| `hook-mutator-fetch` | ❌ `TS7016` |

### `form-data-mutator` — looks like a real bug

```
endpoints.ts(133,67): error TS2554: Expected 1 arguments, but got 2.
```

The generated code calls the mutator fetch-style with two arguments:

```ts
// endpoints.ts (generated)
return customInstance<listPetsResponse>(getListPetsUrl(params), {
  ...options,
  method: 'GET',
});
```

but the sample's mutator is Axios-style and takes a single config object:

```ts
// custom-instance.ts
export const customInstance = <T>(config: AxiosRequestConfig): Promise<T> => { ... };
```

The sample's `orval.config.ts` sets `client: 'react-query'` with no `httpClient`, and the default resolves to `fetch` for every client except `angular-query` (`packages/orval/src/utils/options.ts:536-542`). So the generated output is fetch-shaped while the committed mutator is still Axios-shaped — the mutator looks like it was never updated when that default changed. Either the config needs `httpClient: 'axios'` or the mutator needs to be rewritten fetch-style; I don't know which matches the intent of the sample.

### `form-url-encoded-mutator` — real, but minor

```
custom-form-url-encoded.ts(4,18): error TS2769: No overload matches this call.
  Argument of type 'Body' is not assignable to parameter of type '{ [s: string]: unknown; } | ArrayLike<unknown>'.
```

An unconstrained generic reaching `Object.entries`:

```ts
export const customFormUrlEncoded = <Body>(body: Body): URLSearchParams => {
  Object.entries(body).forEach(...)
```

Constraining it (`<Body extends Record<string, unknown>>`) should be enough.

### `hook-mutator-axios` / `hook-mutator-fetch` — likely a config artifact, not a code bug

```
endpoints.ts(7,29): error TS7016: Could not find a declaration file for module 'react'.
```

I suspect this is a side effect of `--ignoreConfig` discarding the root config's module-resolution and `types` settings rather than a genuine defect, but I have not confirmed that. Giving these samples their own small `tsconfig.json` would likely address both this and the `TS5112` problem more cleanly than `--ignoreConfig`.

## Suggested direction

Rather than patching the script flags, giving each of these six a minimal local `tsconfig.json` (as most other samples have) and reverting the script to plain `tsc --noEmit` would fix the root cause, remove the need for `--ignoreConfig`, and let the six exclusions be dropped from `test:samples` so they are actually covered.

I'm happy to open a PR for this if the direction sounds right — I mainly wanted the `form-data-mutator` mismatch on someone's radar, since that one may indicate a real generated-vs-mutator drift rather than just sample rot.

## Environment

- orval `master` @ 9e3194af5
- `typescript@6.0.3` (repo catalog pin)

cc @anymaniax — flagging you as you've authored most of the `samples/react-query` history; no urgency, and happy to be redirected if someone else owns this area.

ErfanBagheri404:
Fixes #3868.

Six react-query samples are excluded from the root test:samples script and their test scripts fail under TypeScript 6. This PR addresses the root cause and re-enables them in CI.

Changes:
- Add a minimal local tsconfig.json (moduleResolution: bundler, skipLibCheck, noEmit) to each of the six samples so tsc --noEmit uses the local config instead of resolving the repo-root TS6 config (TS5112).
- Revert each sample test script from 'tsc --noEmit endpoints.ts' to plain 'tsc --noEmit'.
- form-data-mutator: add httpClient: axios to match the committed Axios-style custom-instance (default changed to fetch in #1085, TS2554).
- form-url-encoded-mutator: constrain generic to Record<string,unknown> so Object.entries is valid (TS2769).
- Drop the six !react-query-* exclusions from the root test:samples script so they are covered again.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fetch request header handling to preserve readonly array values across generated clients and samples.
  * Improved type safety for URL-encoded form request bodies.

* **Tests**
  * Expanded sample validation to cover complete TypeScript projects, including form-data, URL-encoded, and mutator examples.
  * Added consistent strict TypeScript checking across React Query samples.
  * Ensured Axios-based form-data examples are validated correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "typecheck": "vp run -F './packages/*' typecheck",
     "test": "vp run -F './packages/*' test",
     "test:cli": "vp run -F orval-tests clean && vp run -F orval-tests generate-api && vp run -F orval-tests build",
-    "test:samples": "vp run -F './samples/**' -F '!react-query-form-data' -F '!react-query-form-url-encoded' -F '!react-query-form-url-encoded-mutator' -F '!react-query-form-data-mutator' -F '!react-query-hook-mutator-axios' -F '!react-query-hook-mutator-fetch' test",
+    "test:samples": "vp run -F './samples/**' test",
     "test:snapshots": "vp run -F './packages/*' build:release && vp run -F './samples/**' -F orval-tests generate-api && vp run -F './samples/**' -F orval-tests test:snapshots",
     "test:snapshots:update": "vp run -F './packages/*' build:release && vp run -F './samples/**' -F orval-tests generate-api && vp run -F './samples/**' -F orval-tests test:snapshots:update",
     "lint": "vp lint --type-aware --type-check packages",
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -107,7 +107,7 @@ export const generateRequestFunction = (
   const isFormData = !override.formData.disabled;
   const isFormUrlEncoded = override.formUrlEncoded !== false;
 
-  const GET_HEADERS_HELPER = `  const getHeaders = (h?: HeadersInit | Headers): Record<string, string> => {
+  const GET_HEADERS_HELPER = `  const getHeaders = (h?: NonNullable<RequestInit['headers']>): Record<string, string | readonly string[]> => {
     if (!h) return {};
     if (h instanceof Headers) return Object.fromEntries(h.entries());
     if (Array.isArray(h)) return Object.fromEntries(h);
--- a/samples/hono/hono-with-fetch-client/__snapshots__/next-app/pets/pets.ts
+++ b/samples/hono/hono-with-fetch-client/__snapshots__/next-app/pets/pets.ts
@@ -130,7 +130,9 @@ export const createPets = async (
   createPetsBodyItem: CreatePetsBodyItem[],
   options?: RequestInit,
 ): Promise<createPetsResponse> => {
-  const getHeaders = (h?: HeadersInit | Headers): Record<string, string> => {
+  const getHeaders = (
+    h?: NonNullable<RequestInit['headers']>,
+  ): Record<string, string | readonly string[]> => {
     if (!h) return {};
     if (h instanceof Headers) return Object.fromEntries(h.entries());
     if (Array.isArray(h)) return Object.fromEntries(h);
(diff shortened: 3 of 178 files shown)


melloware (changes requested):
build failing

PR #3889, opened 2026-08-23
fix(core): treat a falsy parameter default as a default

luantaraschi:
A parameter whose schema default is `0`, `false` or `""` is generated as if it had no default at all.

Take the repository's own `tests/transformers/add-version.js`, which injects a required path parameter `version` with `schema: { type: number, default: 1 }`, and change nothing but the default:

```
default: 1   ->  showPetById(petId: string, version: number = 1)
default: 0   ->  showPetById(version: number, petId: string)
```

Three things go wrong at once. The `= 0` is gone, the parameter turns mandatory, and it moves to the front of the signature.

The last one is what breaks the build rather than just the behaviour. `sortByPriority` keeps a defaulted parameter last precisely so it does not precede a required one, and the emitted type is a parameter list:

```
default=1 definitions: petId: string, version?: number
default=0 definitions: version?: number, petId: string     // TS1016
```

With `useNamedParameters` the two halves disagree instead. `props.ts` already asks `param.default !== undefined`, so it appends the `= {}` initializer, while `params.ts` asks for truthiness and marks the property required:

```ts
export type ListPetsPathParameters = { version: number };
listPets({ version = 0 }: ListPetsPathParameters = {}, ...)
// TS2322: Property 'version' is missing in type '{}'
```

Five places read the default for its truthiness: two in `getters/params.ts`, three in `getters/query-params.ts`, and `utils/sort.ts`. `stringify()` already renders `0`, `false` and `''` correctly, so the truthiness test is the only thing in the way. They now ask whether the default is `undefined`, which is what `getters/props.ts` has always asked.

`sort.ts` reads `default` off the prop objects, and `getters/props.ts` was using the literal `false` as its "no default" sentinel for the body, query and header props. Under the strict check that sentinel would read as a real default and move those props to the end, so it is now `undefined`, which is what it always meant. That is also what the workaround in `angular/src/http-resource.ts:358-365` is written against, and its guard keeps behaving the same way, so I left it alone.

## What I ran

Six new cases in `packages/core/src/getters/params.test.ts`, one per default value across number, boolean and string. The three falsy ones fail on `master` and the three truthy ones pass on both, so they pin the behaviour that was already right.

`vp run -F './packages/*' test` is green: 3885 tests across 11 packages, no failures. `vp run -F './packages/*' typecheck` exits 0 and `vp fmt --check` is clean on every touched file.

Four assertions in `packages/orval/src/import-specs.test.ts` pinned the old `false` sentinel and now expect `undefined`.

No snapshot moved. I rebuilt the packages and regenerated all 17 configs under `tests/configs` with the built CLI, and `git status tests/` came back empty. That matches what the corpus contains: the only two falsy defaults anywhere in it are `tests/specifications/issue-708.yaml`, where `default: 0` sits on a query parameter that is already `required: false` so the `!required` half short-circuits either way, and `zod-required-default-values.yaml`, where the `default: false` is on a component schema property rather than a parameter. The sentinel change is ordering-neutral too, since neither `false` nor `undefined` sorted a prop to the end under the old predicate and `undefined` does not under the new one.

I could not run `vp run -F orval-tests generate-api` itself on this machine: it goes through `npm run`, whose `devEngines.runtime` gate wants Node `>=22.22.3` and I am on 22.20.0. Driving `packages/orval/dist/bin/orval.mjs` per config is the same work without that gate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved path, query, body, and header parameter defaults when values are `0`, `false`, or empty strings.
  * Correctly marked parameters with defined defaults as optional across generated definitions and implementations.
  * Improved parameter ordering so defaulted and required parameters are ordered consistently.
  * Removed unintended `false` defaults from generated parameter properties when no default is provided.
  * Updated generated parameter metadata to accurately reflect missing defaults as undefined.

### DIFF
--- a/packages/core/src/getters/params.test.ts
+++ b/packages/core/src/getters/params.test.ts
@@ -11,6 +11,20 @@ const pathParam = (name: string): GetterParameters['path'][number] => ({
   imports: [],
 });
 
+const pathParamWithDefault = (
+  name: string,
+  type: string,
+  defaultValue: unknown,
+): GetterParameters['path'][number] => ({
+  parameter: {
+    name,
+    in: 'path',
+    required: true,
+    schema: { type, default: defaultValue } as never,
+  },
+  imports: [],
+});
+
 describe('getParams getter', () => {
   it('matches a dotted spec name to its generated identifier in the route', () => {
     const params = getParams({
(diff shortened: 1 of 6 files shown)


melloware:
@luantaraschi i updated the AI comments and now the snapshots are failing. Can you update the PR please?

melloware (approved):

PR #3895, opened 2026-08-24
fix(angular): serialize Date query params to ISO string in filterParams

### LINKED ISSUE #3856: fix(angular): filterParams silently strips Date parameters when useDates is enabled

ErfanBagheri404:
Fixes #3856

When `useDates: true`, query params typed as `Date` were silently dropped by the generated `filterParams` helper since it only kept `string | number | boolean`. The helper now serializes `Date` values with `toISOString()` before they reach HttpParams.

Added a test covering both helper variants (with and without object params). All 2361 core tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Angular query-parameter filtering now correctly includes `Date` values by converting them to ISO-formatted strings.
  * This behavior is supported whether or not object parameters are enabled.

### DIFF
--- a/packages/core/src/generators/options.test.ts
+++ b/packages/core/src/generators/options.test.ts
@@ -634,6 +634,8 @@ function filterParams(
       // string so the required key still reaches the wire as \`?key=\`
       // instead of being silently dropped. See #3712.
       filteredParams[key] = preserveRequiredNullables ? null : '';
+    } else if (value instanceof Date) {
+      filteredParams[key] = value.toISOString();
     } else if (
       value != null &&
       (typeof value === 'string' ||
@@ -665,6 +667,24 @@ function filterParams(
       expect(body).not.toBe(PRE_3705_HELPER_BODY);
       expect(body).toContain('objectParamStrategies');
       expect(body).toContain("'flatten' | 'comma' | 'deepObject'");
+    });
+  });
+
+  describe('getAngularFilteredParamsHelperBody date handling (#3856)', () => {
+    it('serializes Date values to ISO strings in both helper variants', () => {
+      for (const hasObjectParams of [false, true]) {
+        const body = getAngularFilteredParamsHelperBody({ hasObjectParams });
+
+        expect(body).toContain('value instanceof Date');
+        expect(body).toContain('value.toISOString()');
+        // The Date branch must run before the primitive check so Date values
+        // are never silently dropped by the typeof filter.
+        const dateBranch = body.indexOf('value instanceof Date');
+        const primitiveBranch = body.indexOf("typeof value === 'string'");
+        expect(dateBranch).toBeGreaterThan(-1);
+        expect(primitiveBranch).toBeGreaterThan(-1);
+        expect(dateBranch).toBeLessThan(primitiveBranch);
+      }
     });
   });
 
--- a/packages/core/src/generators/options.ts
+++ b/packages/core/src/generators/options.ts
@@ -253,6 +253,8 @@ function filterParams(
       // string so the required key still reaches the wire as \`?key=\`
       // instead of being silently dropped. See #3712.
       filteredParams[key] = preserveRequiredNullables ? null : '';
+    } else if (value instanceof Date) {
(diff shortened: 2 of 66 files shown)


ErfanBagheri404:
closing this, #3860 already covers the same fix

PR #3904, opened 2026-08-26
fix(core): handle unevaluatedProperties like additionalProperties (#2156)

### LINKED ISSUE #2156: Support unevaluatedProperties OpenAPI 3.1
**Info**

`unevaluatedProperties` supersedes `additionalProperties` in complex/composed schemas.

`unevaluatedProperties` is in the 3.1 spec. You won't find direct references to `unevealuatedProperties` in the OAS spec because it links to [the JSON Schema spec](https://json-schema.org/draft/2020-12/json-schema-core#name-unevaluatedproperties) instead.

For example `TypeSpec` (by Microsoft) uses it: https://github.com/microsoft/typespec/blob/d7629a4eb6414964220d2dbb63b3088dfe15707a/packages/openapi3/src/schema-emitter-3-1.ts#L155

**Example**
```yml
openapi: 3.1.0
components:
  schemas:
    Base:
      type: object
      properties:
        id:
          type: string
      required:
        - id

    WithExtras:
      allOf:
        - $ref: '#/components/schemas/Base'
        - type: object
          unevaluatedProperties:
            type: string
            maxLength: 10
```

**What this does**
* Base defines a required id of type string.
* WithExtras allows additional properties beyond id, but only if:
  * They are strings, and
  * Have a maximum length of 10 characters.
* Properties validated in Base are not re-evaluated, so id is not subject to the maxLength rule.

✅ **Valid object**
```json
{
  "id": "abc123",
  "nickname": "shortname",
  "color": "blue"
}
```

❌ **Invalid object**
```json
{
  "id": "abc123",
  "nickname": "thisnameiswaytoolong"  // >10 characters
}
```

ErfanBagheri404:
## Problem
Fixes #2156 
`unevaluatedProperties` from OpenAPI 3.1 / JSON Schema 2020-12 is silently ignored. Schemas that use it (e.g. from TypeSpec output) produce `unknown` types instead of index signatures or Record types.

## Fix

`unevaluatedProperties` has the same shape (`boolean | SchemaObject | ReferenceObject`) as `additionalProperties` and should produce the same TypeScript output. The fix falls back to `unevaluatedProperties` when `additionalProperties` is absent, in three spots:

- `packages/core/src/getters/object.ts` — properties-iteration path (after last property)
- `packages/core/src/getters/object.ts` — no-properties standalone object path
- `packages/core/src/generators/factory.ts` — circular reference detection

Two unit tests added covering `unevaluatedProperties: true` and `unevaluatedProperties: { type: 'string' }`.

## Test plan

- `vp run -F './packages/core' test` — all 2363 tests pass
- `vp run -F './packages/orval' -F './packages/fetch' -F './packages/query' test` — 274 tests pass

## Summary by CodeRabbit

* **New Features**
  * Added support for OpenAPI 3.1 `unevaluatedProperties` when generating object types.
  * Generates appropriate index signatures and `Record` types for schemas using this setting.

* **Tests**
  * Added coverage for schema-valued and boolean `unevaluatedProperties` configurations.
  * Updated test configuration for outputs that omit embedded Zod schemas.

### DIFF
--- a/packages/core/src/generators/factory.ts
+++ b/packages/core/src/generators/factory.ts
@@ -103,7 +103,9 @@ function getItems(
 function getAdditionalProperties(
   schema: OpenApiSchemaObject,
 ): OpenApiSchemaObject | OpenApiReferenceObject | boolean | undefined {
-  return schema.additionalProperties as
+  // `unevaluatedProperties` (OAS 3.1) has the same shape and circularity
+  // implications as `additionalProperties`. See issue #2156.
+  return (schema.additionalProperties ?? schema.unevaluatedProperties) as
     | OpenApiSchemaObject
     | OpenApiReferenceObject
     | boolean
--- a/packages/core/src/getters/object.test.ts
+++ b/packages/core/src/getters/object.test.ts
@@ -114,4 +114,46 @@ describe('getObject', () => {
       ]),
     );
   });
+
+  it('renders unevaluatedProperties as an index signature like additionalProperties', () => {
+    const context = createTestContextSpec({ spec: {} });
+
+    const result = getObject({
+      item: {
+        type: 'object',
+        properties: {
+          id: { type: 'string' },
+        },
+        required: ['id'],
+        unevaluatedProperties: { type: 'string', maxLength: 10 },
+      },
+      name: 'WithExtras',
+      context,
+      nullable: '',
+    });
+
+    // Falls back to unknown because named key types can't be proven equal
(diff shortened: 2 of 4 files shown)


melloware (changes requested):
See AI Feedback and can you check this ticket https://github.com/orval-labs/orval/issues/2156  Link it if this fixes that issue

ErfanBagheri404:
Yes it fixes #2156

melloware (approved):

Issue #2353, opened 2025-09-04
MSW mock generation produces multipleOf: undefined for integer types, causing TypeScript errors with exactOptionalPropertyTypes: true


suzukisan22:
## Title
`multipleOf: undefined` generated in MSW mocks causes TypeScript error with `exactOptionalPropertyTypes: true`

## Description

**Affected versions:** v7.10.0 and later (confirmed in v7.11.2)

**Current behavior:**
When generating MSW mocks from OpenAPI schemas containing integer fields without a `multipleOf` constraint, Orval generates:
```javascript
faker.number.int({min: 0, max: 9999, multipleOf: undefined})
```

This causes TypeScript compilation errors when `exactOptionalPropertyTypes: true` is enabled:
```
error TS2345: Argument of type '{ min: number; max: number; multipleOf: undefined; }' is not assignable to parameter of type 'FakerNumberIntOptions<number>'
```

**Expected behavior:**
The `multipleOf` parameter should be omitted when not defined in the OpenAPI schema:
```javascript
faker.number.int({min: 0, max: 9999})
```

**Root cause:**
In [`packages/mock/src/faker/getters/scalar.ts#L23`](https://github.com/orval-labs/orval/blob/e70256718431c40a323e28ac4e56675fa4033190/packages/mock/src/faker/getters/scalar.ts#L23):
```typescript
${isFakerV9 ? `, multipleOf: ${item.multipleOf}` : ''}
```
The code doesn't check if `item.multipleOf` is undefined before interpolating it.

**Proposed solution:**
```typescript
${isFakerV9 && item.multipleOf !== undefined ? `, multipleOf: ${item.multipleOf}` : ''}
```

## Steps to reproduce
1. Create an OpenAPI schema with integer fields (without `multipleOf` property)
   ```yaml
   type: integer
   minimum: 0
   maximum: 9999
   ```
2. Set `exactOptionalPropertyTypes: true` in tsconfig.json
3. Run Orval to generate MSW mocks
4. Run TypeScript compiler

## Environment
- **Orval:** 7.11.2
- **TypeScript:** 5.7.3
- **@faker-js/faker:** 10.0.0
- **Node.js:** 20.x

## Impact
This prevents projects using strict TypeScript settings from upgrading to Orval v7.10.0 or later.

## Workaround
Currently, there's no clean workaround without either:
- Disabling `exactOptionalPropertyTypes` (impacts entire project)
- Downgrading to Orval v7.9.0 (missing new features)
[...]

PR #3911, opened 2026-08-27
fix(angular): compile httpResource output under exactOptionalPropertyTypes

### LINKED ISSUE #3909: bug(angular): httpResource output does not compile under exactOptionalPropertyTypes
### Description

Every generated `*.resource.ts` fails to compile when the consuming project has `"exactOptionalPropertyTypes": true`, with one error per file:

```
src/app/core/api/system/system.resource.ts(95,5): error TS2375: Type '{ headers: HttpHeaders | Record<string, string | readonly string[]> | undefined; url: string; ... }'
is not assignable to type 'HttpResourceRequest' with 'exactOptionalPropertyTypes: true'.
  Types of property 'headers' are incompatible.
    Type 'undefined' is not assignable to type 'HttpHeaders | Record<string, string | readonly string[]>'.
```

The offending line is in `applyOrvalRequestExtension`, emitted by the httpResource header template (introduced in #3717):

```ts
let next: HttpResourceRequest = { ...base };
const extraHeaders =
  typeof options.headers === 'function' ? options.headers() : options.headers;
if (extraHeaders !== undefined) {
  next = { ...next, headers: mergeOrvalResourceHeaders(next.headers, extraHeaders) }; // TS2375
}
```

`mergeOrvalResourceHeaders` is declared as returning `HttpResourceRequest['headers']`, which includes `undefined` because the property is optional. Under `exactOptionalPropertyTypes` an optional property may be *absent*, but not *present and `undefined`*, so the spread is rejected.

It is a typing issue only: the call is already inside `if (extraHeaders !== undefined)`, and `mergeOrvalResourceHeaders` returns `extra` when `base` is falsy, so the value is never actually `undefined` there. The neighbouring `context` branch is written in the compatible shape already.

`retrievalClient: 'httpResource'` and `'both'` are affected; `'httpClient'` is clean.

### Output client

angular

```ts

export default defineConfig({
  api: {
    input: '../api/openapi.json',
    output: {
      clean: true,
      client: 'angular',
      mode: 'tags-split',
      override: {
        angular: { retrievalClient: 'both' },
      },
      schemas: 'src/app/core/api/model',
      target: 'src/app/core/api',
    },
  },
});
```

### Environment

```
System:
  OS: Linux 7.0 Ubuntu 26.04.1 LTS
  Node: 24.19.0
  pnpm: 11.22.0
npmPackages:
  @angular/common: 22.1.3 => 22.1.3
  @angular/core: 22.1.3 => 22.1.3
  orval: 8.25.0 => 8.25.0
  typescript: 6.0.3 => 6.0.3
```

Also reproduced on 8.26.0: the emitted template is identical (`@orval/angular@8.26.0`, `dist/index.mjs:1527`).

The only relevant tsconfig setting is:

```jsonc
{"compilerOptions": {"strict": true, "exactOptionalPropertyTypes": true}}
```

### Expected behavior

The generated client compiles in a project that enables `exactOptionalPropertyTypes`.

### Actual behavior

One `TS2375` per generated `*.resource.ts`, at the `headers` assignment in `applyOrvalRequestExtension`.

```json
{
  "openapi": "3.1.0",
  "info": { "title": "demo", "version": "0.0.0" },
  "paths": {
    "/health": {
      "get": {
        "tags": ["system"],
        "responses": {
          "200": {
            "description": "ok",
            "content": {
              "application/json": {
                "schema": {
                  "type": "object",
                  "required": ["status"],
                  "properties": { "status": { "type": "string" } }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

### Additional context

The operation shape does not matter: I reproduced it with the parameterless `GET` above, and with path params, optional and array query params, JSON request bodies and multipart uploads. In every case it is this one error per `.resource.ts` file and nothing else, so a fix here should unblock the whole `httpResource` output.

Suggested fix — set the key only when the merge produced a value, no behaviour change:

```ts
if (extraHeaders !== undefined) {
  const mergedHeaders = mergeOrvalResourceHeaders(next.headers, extraHeaders);
  next = mergedHeaders === undefined ? next : { ...next, headers: mergedHeaders };
}
```

Alternatively, overload `mergeOrvalResourceHeaders` so that a defined `extra` yields `NonNullable<HttpResourceRequest['headers']>`.

Happy to send the PR if you would like.

the-ult:
## Summary

Fixes #3909. Every generated `*.resource.ts` failed to compile in a project with `exactOptionalPropertyTypes: true`, one `TS2375` per file, always on the same statement.

`mergeOrvalResourceHeaders` was declared to return `HttpResourceRequest['headers']`. That is an indexed access on an *optional* property, so it resolves to `HttpHeaders | Record<…> | undefined`. The call site then writes that value as a *present* `headers:` key, and under `exactOptionalPropertyTypes` a present key may not hold `undefined`:

```ts
next = { ...next, headers: mergeOrvalResourceHeaders(next.headers, extraHeaders) }; // TS2375
```

The value is never `undefined` there. The call sits behind a guard, and the helper returns `extra` whenever `base` is falsy. So the type was wrong, not the logic. It now carries the invariant:

```diff
 function mergeOrvalResourceHeaders(
   base: HttpResourceRequest['headers'],
-  extra: HttpResourceRequest['headers'],
-): HttpResourceRequest['headers'] {
+  extra: NonNullable<HttpResourceRequest['headers']>,
+): NonNullable<HttpResourceRequest['headers']> {
   if (!base) return extra;
-  if (!extra) return base;
```

This is the second of the two fixes the issue suggests. I picked it over guarding the merged value at the call site because it states the real invariant instead of adding a branch that can never run, and because it deletes an already-unreachable line rather than adding a second one.

The call-site guard is now `if (extraHeaders)` instead of `if (extraHeaders !== undefined)`. `HttpHeaders` instances and header records are always truthy, so it is the same guard for every value the public type allows, and it narrows `extraHeaders` to the non-nullable parameter type without a cast.

This affects `retrievalClient: 'httpResource'` and `'both'`. `'httpClient'` was already clean: it always computes a header value (it must set `Accept`), so it never writes an `undefined` key. #3717 introduced the helper.

## Tests

Unit test in `packages/angular/src/http-resource.test.ts` pins the emitted signature and the removal of the dead guard. It fails before the fix.

Two runtime tests in `samples/angular-app/src/app/http-resource-request-extension.spec.ts` cover the `undefined` paths that nothing exercised before: the `() => undefined` function form over a record base, and an `HttpHeaders` base with `headers: undefined`. Record and `HttpHeaders` bases take different branches inside the helper, so both are pinned.

`tests/scripts/typecheck-generated.mjs` gained a second pass that compiles the 13 Angular fixtures emitting the request-extension helper with `exactOptionalPropertyTypes` enabled. It goes red when the fix is reverted while the existing `angular` pass stays green, so it covers something no existing check did. The per-folder typecheck body is now one `typecheck()` function shared by both passes.

## Validation

- `@orval/angular`: 275 tests pass
- Snapshots: 6248 assertions pass, no regeneration drift
- `typecheck-generated`: 19 of 19 runs, including the new gate
- `samples/angular-app`: `ng test` 45 of 45, `ng build` clean, `ng lint` clean
- `vp fmt --check` clean

Generated output differs from `master` by exactly the three type lines and the guard, across all 24 snapshots and 4 sample sources. Nothing else moved.

## Out of scope

Compiling the whole `tests/generated/angular` corpus under the flag turns up 72 further `TS2375`/`TS2322` errors, all in mock output (`petstore`, `tags`, `split`, `tags-split`, `custom-client`, `url-encode-parameters`). They come from `faker.helpers.arrayElement([value, undefined])` assigned to optional properties, which is a separate defect in the mock generator, close to #2353. It deserves its own issue rather than riding along here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Angular HTTP resource header handling when optional or undefined headers are provided.
  * Prevented generated resources from causing TypeScript errors with `exactOptionalPropertyTypes`.
  * Preserved base request headers when no additional headers are supplied.

* **Tests**
  * Added regression coverage for header merging and generated Angular resources.
  * Expanded type checking for Angular HTTP resource fixtures.


the-ult:
@andsouto could you review it?

melloware (approved):

Issue #3664, opened 2026-06-29
@orval/mcp: no Zod input schema generated for application/x-www-form-urlencoded request bodies


yagop:
### What are the steps to reproduce this issue?

With the **`mcp` client**, Orval generates a per-operation Zod input schema (`<Op>Body`) for **JSON** request bodies, but **not** for `application/x-www-form-urlencoded` request bodies. Form-body operations end up with no generated input schema at all.

**1. Minimal spec** (`spec.yaml`):

```yaml
openapi: 3.0.3
info: { title: repro, version: 1.0.0 }
paths:
  /token:
    post:
      operationId: issueToken
      requestBody:
        required: true
        content:
          application/x-www-form-urlencoded:   # <-- the trigger
            schema:
              type: object
              required: [grant_type]
              properties:
                grant_type: { type: string, enum: [client_credentials] }
                client_secret: { type: string }
      responses:
        '200': { description: ok }
```

**2. Config** (`orval.config.ts`):

```ts
export default {
  repro: {
    input: './spec.yaml',
    output: { client: 'mcp', target: './gen', mode: 'split' },
  },
};
```

**3. Run** `orval` and inspect `gen/tool-schemas.zod.ts`.

If you change the content type to `application/json`, Orval emits `IssueTokenBody = zod.object({ ... })` as expected. With `application/x-www-form-urlencoded`, **no `IssueTokenBody` schema is generated**, even though `handlers.ts` still types the handler arg with the body model.

### What is the expected behavior?

The `mcp` client emits a `<Op>Body` Zod schema for `application/x-www-form-urlencoded` request bodies too (same as it does for `application/json`), so the operation's body can be registered as the tool's `inputSchema` and validated/typed at the MCP boundary.

### What is the actual behavior?

No Zod schema is generated for the form-urlencoded body. There is nothing to pass as the tool's `inputSchema`, so downstream code must fall back to a permissive validator and cast to the handler's real param type, e.g.:

```ts
server.registerTool(
  name,
  { inputSchema: { bodyParams: z.record(z.string(), z.string()) } },
  async (args) =>
    handler(args as unknown as Parameters<typeof handler>[0]), // <-- forced cast
);
```

That single operation loses both per-op input validation and type safety — the rest of the MCP server's whole point.

### Notes / impact

- JSON bodies are handled: an op with an `application/json` body produces `<Op>Body` Zod; only the form-urlencoded content type is skipped.
[...]

PR #3915, opened 2026-08-28
fix(types): use File instead of Blob for multipart form-data binary fields (#3662)

### LINKED ISSUE #3662: File uploads using `multipart/form-data` do not include filename
### Acknowledgement

### Comment

In our API spec, we have an endpoint to which an array of files can be uploaded. Orval generates code for this using the `Blob` type, which does not support including the original file names. In practice, browsers will include the filename in the `Content-Disposition` header, in line with the [RFC](https://www.rfc-editor.org/info/rfc7578/#section-4.2). The `FormData` object also has an [overload](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append) accepting a filename in addition to the blob.

Ideally, the generated upload schema would include a field of `File` type (or something similar), but I did not find any way to configure this in Orval. How could I best approach this? Is this a shortcoming in Orval?

------

OpenAPI specification:
```yaml
openapi: 3.1.0
info:
  title: Demo
  version: 0.0.0
paths:
  /foo:
    post:
      description: This is an upload endpoint
      requestBody:
        required: true
        content:
          multipart/form-data:
            schema:
              type: object
              properties:
                bar:
                  type: string
                files:
                  type: array
                  items:
                    type: string
                    contentMediaType: application/octet-stream
      responses:
        '204':
          description: Success
```

ErfanBagheri404:
Fixes #3662

Multipart form-data binary fields were typed as `Blob`, which loses the original filename when appended to `FormData`. Per RFC 7578, the filename belongs in the part's `Content-Disposition` header, and `FormData.append` only preserves it when the value is a `File`.

This changes the emitted type for multipart binary fields from `Blob` to `File` (and `Blob | string` to `File | string` for text files). The generated runtime already handles `File` correctly, so no runtime changes were needed.

Non-multipart binary fields (response bodies, url-encoded bodies) stay `Blob`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Multipart form-data binary fields now accept both `Blob` and `File` values.
  - Text-compatible file fields support `Blob`, `File`, and string values.
  - Arrays of uploaded files now accept combinations of `Blob` and `File`.
  - Generated upload models consistently support `Blob` and `File` inputs.
  - Zod validation for binary uploads now recognizes `Blob` values, including multipart form-data fields.

### DIFF
--- a/packages/core/src/getters/res-req-types.test.ts
+++ b/packages/core/src/getters/res-req-types.test.ts
@@ -366,16 +366,18 @@ describe('getResReqTypes (content type handling)', () => {
       expect(bodySchema).toBeDefined();
       expect(bodySchema?.model).toContain('export type BodyRequestBody = {');
 
-      // Binary files → File (filename preserved in multipart, #3662)
-      expect(bodySchema?.model).toContain('encBinary: File;');
-      expect(bodySchema?.model).toContain('cmtBinary: File;');
-      expect(bodySchema?.model).toContain('formatBinary: File;');
-      expect(bodySchema?.model).toContain('wildcardFile: File | string;');
+      // Binary files → Blob | File (filename preserved in multipart, #3662)
+      expect(bodySchema?.model).toContain('encBinary: Blob | File;');
+      expect(bodySchema?.model).toContain('cmtBinary: Blob | File;');
+      expect(bodySchema?.model).toContain('formatBinary: Blob | File;');
+      expect(bodySchema?.model).toContain(
+        'wildcardFile: Blob | File | string;',
+      );
 
-      // Text files → File | string
-      expect(bodySchema?.model).toContain('encText: File | string;');
-      expect(bodySchema?.model).toContain('cmtText: File | string;');
-      expect(bodySchema?.model).toContain('encOverride: File | string;'); // encoding precedence
+      // Text files → Blob | File | string
+      expect(bodySchema?.model).toContain('encText: Blob | File | string;');
+      expect(bodySchema?.model).toContain('cmtText: Blob | File | string;');
+      expect(bodySchema?.model).toContain('encOverride: Blob | File | string;'); // encoding precedence
 
       // base64 encoded → string (not file)
       expect(bodySchema?.model).toContain('base64Field: string;');
@@ -383,8 +385,8 @@ describe('getResReqTypes (content type handling)', () => {
       // Object field → named type
       expect(bodySchema?.model).toContain('metadata: BodyRequestBodyMetadata;');
 
-      // Array of files → File[]
-      expect(bodySchema?.model).toContain('photos: File[];');
+      // Array of files → (Blob | File)[]
+      expect(bodySchema?.model).toContain('photos: (Blob | File)[];');
 
       // Nested contentMediaType should be ignored (JSON serialization)
       const metadataSchema = result.schemas.find(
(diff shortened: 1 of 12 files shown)


melloware:
@ErfanBagheri404 i marked this breaking change since it changes the output

luantaraschi (commented):
I came to this because of the `breaking change` label and wanted to see how much work it was doing. It is doing quite a lot.

The narrowing lands in an input position. The multipart body model is the only surface that moves, and for every client except hono that is the thing the caller passes in:

```ts
declare const b: Blob;                        // canvas.toBlob(), (await fetch(...)).blob(), new Blob([...])
interface UploadBody { photos?: File[] }
const x: UploadBody = { photos: [b] };
// error TS2739: Type 'Blob' is missing the following properties from type 'File': lastModified, name, webkitRelativePath
```

With `photos?: Blob[]` the same code compiles. And `new File()` wants a filename, so a caller who genuinely does not have one has no cheap way across.

What I keep circling back to is that `File extends Blob` already, so anyone who wanted the filename preserved could always pass a `File` into the old type and `FormData.append` kept it. The capability was there; this makes it compulsory. In #3662 the ask was for `Blob | File`, which widens instead and leaves existing callers compiling.

There is an argument in favour that the description does not make: the zod output has emitted `zod.instanceof(File)` for multipart since #3664, so the TS type was the odd one out. If that is the reasoning, it is worth saying, because it turns this from a break into an alignment.

One question rather than a claim. Does this reach a multipart body written as a `$ref` to a shared component? The form context looks like it only comes from the inline body path, and if that is right then the shared model keeps `Blob` while the zod validator for the same endpoint asks for a `File`.

melloware:
I marked it breaking change because it felt like one.  @luantaraschi is there a safer way to do it?

luantaraschi:
It depends which of the two you want to be the contract, and orval has half answered that already.

On this PR's base the zod generator emits `instanceof File` for multipart binary in five places, and the PR does not touch `packages/zod`. So today the generated validator rejects a `Blob` while the generated TS type accepts one. This change removes that disagreement by moving the type to where zod already stands. That is the strongest argument for it and the description does not make it.

The non-breaking version is the one asked for in #3662: `Blob | File`. It widens instead of narrowing, so every existing caller keeps compiling, and anyone who wants the filename preserved passes a `File`, which already worked before this PR because `File extends Blob`. The cost is that the TS type stays looser than the validator, so the disagreement survives.

So the safer path is not really a different diff, it is a different decision. If `File` is the contract, this PR is right, the label is right, and it wants a major. If you would rather nobody's build breaks now, `Blob | File` here plus letting zod accept either gets the filename support without the compile error, and you can narrow both together later.

Either way there is one shape worth checking first, which I could not settle from outside: whether the form context reaches a multipart body written as a `$ref` to a shared component. If it only comes from the inline body path then that shape keeps `Blob` and its own zod validator still asks for `File`, so it would carry both the bug and the disagreement past this PR.

melloware:
maybe it should be `Blob | File`

luantaraschi:
That works, and the zod half is a swap rather than a union. `File` extends `Blob` at runtime, so `zod.instanceof(Blob)` already accepts a `File`:

```
> new File(['x'], 'a.txt') instanceof Blob   // true
> new Blob(['x']) instanceof File            // false
```

The four `instanceof File` sites are all in `packages/zod/src/index.ts`: the `format === 'binary'` branch, the `contentMediaType === 'application/octet-stream'` one, `fileOrString`, and the form-data branch. Swapping `File` for `Blob` in those leaves the validator exactly as permissive as `Blob | File`, so the two stop disagreeing, and a caller who passes a `File` still gets the filename through `FormData.append` the way they do today.

melloware:
@luantaraschi did you want to submit a separate PR showing your idea?

ErfanBagheri404:
> That works, and the zod half is a swap rather than a union. `File` extends `Blob` at runtime, so `zod.instanceof(Blob)` already accepts a `File`:
> 
> ```
> > new File(['x'], 'a.txt') instanceof Blob   // true
> > new Blob(['x']) instanceof File            // false
> ```
> 
> The four `instanceof File` sites are all in `packages/zod/src/index.ts`: the `format === 'binary'` branch, the `contentMediaType === 'application/octet-stream'` one, `fileOrString`, and the form-data branch. Swapping `File` for `Blob` in those leaves the validator exactly as permissive as `Blob | File`, so the two stop disagreeing, and a caller who passes a `File` still gets the filename through `FormData.append` the way they do today.

what you say @melloware 
should i apply this?

melloware:
yes please @ErfanBagheri404 and let us know when its ready for review

melloware:
@luantaraschi please re-review!

melloware (approved):

PR #3916, opened 2026-08-28
feat(core): support runtime expressions for server URL variables

### LINKED ISSUE #3734: Dynamic variables for getBaseUrlFromSpecification: true
### Description

We want to generate type specifications for the Tebex headless API, which as the following URL format
https://headless.tebex.io/api/accounts/{publicApiToken}/...

Its servers field is described like this:
```
  - url: https://headless.tebex.io/api/accounts/{token}
    description: Main API scoped to public token
    variables:
      token:
        default: ""
        description: The store's public token
```

Unfortunately orval doesnt let me use an environment variable for this.

### Current behavior

```
            baseUrl: {
                getBaseUrlFromSpecification: true,
                variables: {
                    token: 'process.env.TEBEX_TOKEN',
                }
            }
```
or:
`token: '${process.env.TEBEX_TOKEN}'`

Both unfortunately don't work. The first is inserted as a string, the latter is escaped. 

### Proposed behavior

There should be a way to describe runtime expressions for specification variables, not just for the baseUrl itself. 

### Proposed solution

My idea would be that all variables could be considered a string or an object. 
Given a string, it behaves as right now.
Given an object like e.g. 
```
              variables: {
                    token: {
                        runtime: "process.env.TEBEX_TOKEN"
                    }
                }
```
it would instead insert `token` as a JavaScript expression. Shouldn't introduce any backwards compatibility issues.

ErfanBagheri404:
Closes #3734

Server URL variables now accept a `{ runtime: string }` value in addition to a plain string. The expression is embedded as a `${...}` interpolation in the generated URL template literal, so the value resolves at request time instead of being baked in at generation time.

```ts
baseUrl: {
  getBaseUrlFromSpecification: true,
  variables: {
    token: { runtime: 'process.env.TEBEX_TOKEN' },
  },
}
```

generates:

```ts
return `https://headless.tebex.io/api/accounts/${process.env.TEBEX_TOKEN}/packages`
```

Runtime variables mix freely with static string variables. The expression is spliced in after jsesc escaping, so it is never escaped.

## Summary by CodeRabbit

* **New Features**
  * Server URL variables now support runtime JavaScript expressions alongside static values.
  * Base URL configuration supports fixed and runtime-resolved values.
  * Runtime expressions are preserved when constructing API routes.

* **Bug Fixes**
  * Improved route construction for overlapping or placeholder-like server-variable values.
  * Literal placeholder text in static values is preserved correctly.

* **Tests**
  * Added coverage for runtime server variables, mixed static values, and tokenized API endpoints.


melloware (changes requested):
AI has a good suggestion

melloware (approved):

PR #3917, opened 2026-08-28
feat(core): add namingConvention.properties for schema property name conversion

### LINKED ISSUE #2381: Add naming convention conversion for schema property names
Hi everyone, this is a follow-up for the following discussion https://github.com/orval-labs/orval/discussions/218

### Summary
Add a native support for converting schema field names from diffent conventions like snake_case to camelCase in generated TypeScript interfaces and types, similar to what openapi-generator provides.

### Problem Description
Many APIs (especially those built with Python/FastAPI, Ruby on Rails, etc.) use snake_case naming conventions for their schema properties, but TypeScript/JavaScript codebases typically use camelCase. Currently, Orval users need to implement complex workarounds using mutators and transformers to achieve this conversion, which:

### Proposed Solution
Building on the recent implementation of `output.overrides.namingConvention.enum` (added in #2125) we could have something like :

```js
module.exports = {
  petstore: {
    output: {
      override: {
        namingConvention: {
          enum: 'camelCase', 
          properties: 'camelCase'      // New property
        }
      }
    }
  }
}
```

ErfanBagheri404:
Closes #2381

Adds `output.override.namingConvention.properties` — applies camelCase, snake_case, pascal_case, or kebab-case to schema property names in generated TypeScript types.

Minimal change: one line in `object.ts` applies `conventionName(key, convention)` before `getKey()` in the property emission. The config option flows through the existing normalization pipeline (the `?? {}` spread already handles the new field).

- `widget_id` → `widgetId` with `properties: 'camelCase'`
- `nested_data` → `nestedData` (recursively through nested objects)
- `inner_value` → `innerValue`
- `required` array matching against the original spec key is unaffected
- Runtime serialization (formData, query params) uses the original spec key unchanged

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring naming conventions for generated schema property names.
  * Property names can be transformed to camelCase or other supported conventions while preserving original names for required-field matching and runtime serialization.
  * Naming collisions are handled safely by retaining original schema keys when converted names would conflict.
* **Tests**
  * Added coverage for property conversion, collision handling, and form-data serialization in generated fetch client output.


melloware (changes requested):
see AI comments

melloware (approved):

PR #3919, opened 2026-08-28
feat(query): export a mutation key getter per operation

### LINKED ISSUE #3918: Export a mutation key getter per operation, mirroring `getXQueryKey`
### Description

Query keys get an exported getter. Mutation keys are computed the same way but stay local to the options factory. 

### Current behavior

No getter is created, meaning we have to get the mutation key from the options:

```ts
export const getAssetsStoreMutationOptions = <...>(options?) => {
  const mutationKey = ['assetsStore'];
  ...
};
```

this is kind of annoying especially in vue where you cant just do:
`getAssetsStoreMutationOptions().mutationKey`
Meaning you will need to create a helper function or typecast if you want to get the mutationKey from a `mutationOptions` function.

### Proposed behavior

Create a getter function for mutation keys, just like we already have for query keys:

```
export const getAssetsStoreMutationKey = () =>
  ['assetsStore'] as const;
```

### Proposed solution

The exact same as is currently used for query keys

spaansba:
Fixes #3918

Mutation keys were computed inside the options factory, so getting one meant calling `getXxxMutationOptions()` and reading `.mutationKey` off the result — awkward in Vue, and it couples callers to the options shape instead of a stable key API.

This hoists the key into an exported getter next to the factory, same shape as `getXxxQueryKey`:

```ts
export const getCreatePetsMutationKey = () => ['createPets'] as const;

export const getCreatePetsMutationOptions = <...>(options?) => {
  const mutationKey = getCreatePetsMutationKey();
  ...
};
```

Main use case is matching a key outside a hook, e.g. a global `MutationCache` callback that receives `mutation.options.mutationKey`.

Notes:
- Gated on the existing `shouldExportQueryKey` option, no new config. Emission is skipped only when the getter would be both unexported and unused (`shouldExportQueryKey: false` + `requestOptions: false`).
- Unchanged behaviour: with `requestOptions: false` the factory still sets no `mutationKey` at all, so nothing new is wired there.
- Snapshots and samples regenerated; the only diff outside the two source files is the new getter.

## Summary by CodeRabbit

* **New Features**
  * Mutation keys can now be exported through dedicated getter helpers alongside mutation options.
  * Request-option configurations automatically use the generated mutation-key getters.
* **Documentation**
  * Updated English and Chinese configuration guides to document mutation-key exports and include a usage example.
* **Tests**
  * Added coverage confirming mutation-key getters are generated when enabled and omitted when request options are disabled.

### DIFF
--- a/docs/content/docs/reference/configuration/output.mdx
+++ b/docs/content/docs/reference/configuration/output.mdx
@@ -1465,7 +1465,11 @@ Export mutator hooks.
 **Type:** `Boolean`
 **Default:** `true`
 
-Export query keys.
+Export query keys. Also applies to mutation keys, which are emitted as a getter next to the mutation options factory:
+
+```ts title="endpoints.ts"
+export const getCreatePetsMutationKey = () => ['createPets'] as const;
+```
 
 ### shouldFilterQueryKey
 
--- a/docs/content/docs/zh/reference/configuration/output.mdx
+++ b/docs/content/docs/zh/reference/configuration/output.mdx
@@ -497,7 +497,7 @@ export default defineConfig({
 
 ### shouldExportQueryKey
 
-控制是否导出 query key helper。
+控制是否导出 query key helper。该选项同样作用于 mutation key，会在 mutation options 工厂函数旁生成 `getXxxMutationKey()` 导出。
 
 ### shouldSplitQueryKey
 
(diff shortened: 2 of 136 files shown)


melloware:
running build now

melloware (approved):

PR #3929, opened 2026-08-30
feat(query): rename shouldExportQueryKey to shouldExportKeys

spaansba:
Missed in #3919.

`shouldExportQueryKey` doesn't apply to query keys alone. It also gates the `getXxxMutationKey` getters, so the name misdescribes what the flag does:

**Alternitive:**
**I was also thinking maybe it shoudn't be a `boolean` but a `'query' | 'mutation' | 'both' | 'false'`**

## New Features
- Added `shouldExportKeys` to control generated query and mutation cache key helpers.
- Cache key exports are enabled by default.

## Bug Fixes
- Ensured query and mutation key exports consistently follow the same configuration.

## Documentation
- Deprecated `shouldExportQueryKey` in favor of `shouldExportKeys`.
- Existing configurations using the old option remain supported, with the new option taking precedence when both are provided.

### DIFF
--- a/docs/content/docs/reference/configuration/output.mdx
+++ b/docs/content/docs/reference/configuration/output.mdx
@@ -1460,14 +1460,18 @@ export async function loader(queryClient: QueryClient) {
 
 Export mutator hooks.
 
-### shouldExportQueryKey
+### shouldExportKeys
 
 **Type:** `Boolean`
 **Default:** `true`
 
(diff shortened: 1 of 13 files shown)


melloware (approved):

melloware (approved):
@spaansba great job i love the warning to users so they can update before we remove

melloware (approved):

PR #3945, opened 2026-08-31
fix: honor allowReserved on path params when urlEncodeParameters is true (#3910)

### LINKED ISSUE #3910: urlEncodeParameters is all-or-nothing: no way to exclude a wildcard {path=**} parameter
### Description

`urlEncodeParameters` is all-or-nothing across every path parameter in a target, and there is no way to opt one out. For specs that mix ordinary single-segment path params with **wildcard / multi-segment** ones, neither setting is correct:

- **off** (the default): a single-segment param containing a reserved character silently addresses the wrong resource
- **on**: every wildcard path param has its `/` encoded to `%2F` and the route stops matching

Both failure modes are quiet — no exception, no type error, and in the first case an HTTP **200**.

Specs generated from **`google.api.http` annotations** hit this by construction: gRPC-transcoder routes use `{path=**}` for file paths and namespaces alongside plain `{name}` / `{id}` segments in the same document. That is how ours is shaped, and I expect it is common for anything fronted by an Envoy JSON transcoder or grpc-gateway.

### Output client

fetch (also generates the same shape for the other TS clients)

```ts

export default defineConfig({
  repro: {
    input: { target: './api.yaml' },
    output: { target: './gen.ts', client: 'fetch', urlEncodeParameters: true },
  },
})
```

### Environment

```
  System:
    OS: Linux 6.18 (WSL2)
  npmPackages:
    orval: 8.26.0   (repro run on latest; originally hit on 8.19.0)
  Runtime: bun 1.3.14
```

### Expected behavior

A way to keep encoding on for ordinary path params while leaving a wildcard/multi-segment param unencoded. Any of these would do:

- honour `allowReserved: true` on a path parameter as "do not encode" (see the caveat below)
- a per-parameter override in `output.override.operations[].parameters` or similar
- an `x-` extension orval reads (e.g. `x-orval-url-encode: false`)

### Actual behavior

`urlEncodeParameters: true` wraps **every** path param in `encodeURIComponent(String(...))`, including one documented as carrying a path, and `allowReserved: true` on the parameter changes nothing.

I want to flag honestly that OpenAPI says `allowReserved` "only applies to parameters with an `in` value of `query`", so ignoring it on a path param is defensible and this is a feature request rather than a spec violation. But it is currently the only thing in the document that could express the distinction, and nothing else does.

```yaml
openapi: 3.0.3
info: { title: Repro, version: 0.0.0 }
paths:
  /things/{name}:
    delete:
      operationId: deleteThing
      parameters:
        - { name: name, in: path, required: true, schema: { type: string } }
      responses: { '200': { description: ok } }
  /files/{path}:
    get:
      operationId: getFile
      parameters:
        - { name: path, in: path, required: true, allowReserved: true, schema: { type: string } }
      responses: { '200': { description: ok } }
```

Generated with `urlEncodeParameters: true` — both encoded, `allowReserved` ignored:

```ts
return `/things/${encodeURIComponent(String(name))}`
return `/files/${encodeURIComponent(String(path))}`   // breaks a {path=**} route
```

Generated with the option off (default) — neither encoded:

```ts
return `/things/${name}`   // silently addresses the wrong resource, see below
return `/files/${path}`
```

### Additional context

Both halves are observed against a real server, not inferred.

**Option off — wrong resource, HTTP 200.** Two resources exist, `zz-probe` and `zz-probe#danger`. Deleting the second through the generated builder:

```
generated URL   : "/v2/alert-rule-sets/zz-probe#danger"
what fetch sends: "/v2/alert-rule-sets/zz-probe"     <- the fragment never leaves the client
DELETE status   : 200
```

`zz-probe` was deleted; the intended target survived. The `#` is the sharp case because `new URL()` strips it as a fragment before the request is sent, so nothing reaches the server to reject — but `?` truncates the same way, and any reserved character is a candidate for a mismatch.

**Option on — wildcard route stops matching.** Same server, a `{path=**}` route, raw vs. percent-encoded:

```
raw  status=200 { "ns": "myrmex/system/collector/error/task", ... }
enc  status=404 { "code": 5, "message": "metric container \"myrmex%2Fsystem%2Fcollector%2Ferror%2Ftask\" not found" }
```

The proxy does not decode `%2F` back into separators — the literal `%2F`s reach the backend, as the error text shows. That is standard proxy behaviour for encoded slashes rather than anything unusual about this server.

So a single-target project containing both route shapes has no correct setting today. Happy to attempt a PR if you can say which of the three escape hatches above you would accept.

ErfanBagheri404:
fixes #3910

## Summary by CodeRabbit

* **Bug Fixes**
  * Generated HTTP clients now preserve reserved characters in path parameters explicitly configured to allow them.
  * URL encoding remains applied to other path parameters, improving request URL correctness across Fetch, Angular, Axios, and query clients.
  * Supports multi-segment and signal-based path parameters without unintended encoding.
  * Handles optional and named path-parameter expressions consistently when determining which values to encode.
  * Applies consistent behavior across generated request and resource helpers.


melloware (changes requested):
AI feedback

melloware (approved):

PR #3952, opened 2026-09-01
fix(core): collect enum members from nested anyOf/oneOf branches

### LINKED ISSUE #3951: If there are duplicate `type: "null"` in a nested `anyOf`, the enum values are not outputted since v8.25.0
### Description

If there are duplicate `type: "null"` in a nested `anyOf`, the enum values are not outputted since v8.25.0.

#### Minimal Reproduction
```yaml
# openapi.yaml

openapi: 3.1.0
info:
  title: "api"
  version: 0.0.0
paths:
  /:
    post:
      operationId: postRoot
      requestBody:
        required: true
        content:
          application/json:
            schema:
              anyOf:
                - anyOf:
                    - type: string
                      enum:
                        - a
                        - b
                    - type: "null"
                - type: "null"
      responses:
        "200":
          description: Default Response
```

#### Output (v8.27.0)

```ts
// output/api.ts

import {
  useMutation
} from '@tanstack/react-query';
import type {
  MutationFunction,
  UseMutationOptions,
  UseMutationResult
} from '@tanstack/react-query';

export type PostRootBody = typeof PostRootBody[keyof typeof PostRootBody] | null;

// * PostRootBody is an empty object.
export const PostRootBody = {
} as const;

...
```

#### Output (v8.24.0)

```ts
// output/api.ts

import {
  useMutation
} from '@tanstack/react-query';
import type {
  MutationFunction,
  UseMutationOptions,
  UseMutationResult
} from '@tanstack/react-query';

export type PostRootBody = typeof PostRootBody[keyof typeof PostRootBody] | null;

// * PostRootBody is an enum map object.
export const PostRootBody = {
  a: 'a',
  b: 'b',
} as const;

...
```

### Output client

react-query
(The same problem occurs with other settings as well.)

```ts

export default defineConfig({
  api: {
    input: {
      target: './openapi.yaml',
    },
    output: {
      target: './output',
      clean: true,
      client: 'react-query',
      override: {
        header: false,
      },
    },
  },
});

```

### Environment

  System:
    OS: macOS 26.6.2
    CPU: (12) arm64 Apple M2 Pro
    Memory: 481.22 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    orval: 8.27.0 => 8.27.0 
    zod: 4.5.4 => 4.5.4 

### Additional context

Although this definition—which duplicates `type: “null”` within a nested `anyOf`—is actually redundant, the OpenAPI auto-generation library I’m using may produce this kind of output depending on the situation.

Since this worked fine in previous versions, I’d appreciate it if you could fix this if possible.

MFA-G:
Fixes #3951.

## The problem

`getEnumMembersFromBranches` walks `oneOf`/`anyOf` members and pulls enum values out of each one, but it only looks one level deep. A branch that is itself a composition contributes nothing, so for the spec in the issue:

```yaml
anyOf:
  - anyOf:
      - type: string
        enum: [a, b]
      - type: 'null'
  - type: 'null'
```

only the outer `{type: 'null'}` branch is collected. `combineSchemas` still takes the combined-enum path (the union type resolves correctly to `'a' | 'b' | null`), so the type alias is emitted referencing `keyof typeof PostRootBody` while the const it points at is empty:

```ts
export type PostRootBody = typeof PostRootBody[keyof typeof PostRootBody] | null;

export const PostRootBody = {
} as const;
```

This nesting is redundant — the inner composition already carries the null branch — but it is valid, and as the reporter notes some OpenAPI generators emit it. It worked up to v8.24.0.

## The fix

Descend into a branch when it is a composition and contributed no members of its own. The `enumValues.length === 0` guard keeps annotated `const`/`enum` branches on their existing path, so title/description/deprecated metadata handling is unchanged. A depth cap guards against a self-referential branch in an in-memory spec object; a plain spec cannot express that cycle.

Output with the fix, matching v8.24.0:

```ts
export type PostRootBody = typeof PostRootBody[keyof typeof PostRootBody] | null;

export const PostRootBody = {
  a: 'a',
  b: 'b',
} as const;
```

## Validation

- Two regression tests in `packages/core/src/getters/enum.test.ts` covering a nested `anyOf` and a nested `oneOf` of `const` branches (the latter pins that nested metadata still survives).
- Verified end-to-end by generating the issue's exact spec against a local `build:release`, before and after the change.
- `vp run -F './packages/*' test` — 11 files, all passing (core: 2451 tests).
- `vp run -F './packages/*' typecheck`, `vp lint --type-aware --type-check packages`, `vp fmt --check` — all clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved enum detection across deeply nested `oneOf` and `anyOf` schemas.
  * Correctly preserves enum and constant values, metadata, and valid null options in complex schemas.
  * Excludes values from branches that only allow `null`.
  * Added protection against infinite processing for self-referential schemas while supporting deeply nested, acyclic schemas.

### DIFF
--- a/packages/core/src/getters/enum.test.ts
+++ b/packages/core/src/getters/enum.test.ts
@@ -330,6 +330,71 @@ describe('getEnumMembers', () => {
     ]);
   });
 
+  it('should not widen a null-only branch with its nested composition', () => {
+    // `type: 'null'` admits no value but `null`, so a nested composition under
+    // it cannot contribute additional values to the enum.
+    const schema = {
+      anyOf: [
+        {
+          type: 'null',
+          anyOf: [
+            {
+              type: 'string',
+              enum: ['a', 'b'],
+            },
+          ],
+        },
+      ],
+    } as unknown as OpenApiSchemaObject;
+
+    expect(getEnumMembers(schema)).toEqual([{ value: null }]);
+  });
+
+  it('should still walk a branch that allows a type besides null', () => {
+    const schema = {
+      anyOf: [
+        {
+          type: ['string', 'null'],
+          anyOf: [
+            {
+              const: 'ACTIVE',
+            },
+          ],
+        },
+      ],
+    } as unknown as OpenApiSchemaObject;
+
+    expect(getEnumMembers(schema)).toEqual([
+      { value: null },
+      {
+        value: 'ACTIVE',
+        name: undefined,
+        description: undefined,
+        deprecated: undefined,
+      },
+    ]);
+  });
+
+  it('should collect members from a deeply nested acyclic composition', () => {
+    // Traversal stops on real cycles, not at a fixed depth, so an innermost
+    // enum stays reachable however many redundant wrappers precede it.
+    let schema = { enum: ['deep'] } as unknown as OpenApiSchemaObject;
+    for (let i = 0; i < 64; i++) {
+      schema = { anyOf: [schema] } as unknown as OpenApiSchemaObject;
+    }
+
+    expect(getEnumMembers(schema)).toEqual([{ value: 'deep' }]);
+  });
+
+  it('should stop on a self-referential branch', () => {
+    const schema = {
+      anyOf: [{ enum: ['a'] }],
+    } as unknown as OpenApiSchemaObject;
+    (schema.anyOf as unknown[]).push(schema);
+
+    expect(getEnumMembers(schema)).toEqual([{ value: 'a' }]);
+  });
+
   it('should preserve const null branches', () => {
     const schema = {
       oneOf: [
--- a/packages/core/src/getters/enum.test.ts
+++ b/packages/core/src/getters/enum.test.ts
@@ -264,6 +264,72 @@ describe('getEnumMembers', () => {
     ]);
   });
 
+  it('should collect enum members from a nested anyOf branch', () => {
+    // Some generators emit a redundant but valid nesting where the inner
+    // composition already carries the null branch. See #3951.
+    const schema = {
+      anyOf: [
+        {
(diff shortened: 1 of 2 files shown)


MFA-G:
Thanks both — the two findings were valid and are fixed in cf2da92.

**1. Fixed-depth truncation (CodeRabbit, `enum.ts:392`).** `MAX_ENUM_BRANCH_DEPTH = 32` would silently drop the innermost `enum`/`const` of an acyclic schema with more than 32 redundant wrappers. Replaced the depth counter with a `Set` of visited schema objects, so traversal stops only on an actual cycle (a branch that is the same object as one of its ancestors) and complete discovery is preserved at any depth.

**2. Descent into null branches (Copilot `:443` / CodeRabbit `:439`).** Correct: `{ type: 'null', anyOf: [{ const: 'ACTIVE' }] }` would have emitted `ACTIVE`. My first pass over-corrected by skipping descent for *any* branch with a `type`, which also silenced `{ type: ['string','null'], anyOf: [...] }` — a branch that does admit other values. Now the skip applies only when the type is exclusively `null` (`'null'` or an array of nothing but `'null'`), so multi-type arrays keep recursing, matching CodeRabbit's "keep recursion for multi-type arrays".

Tests added: null-only branch is not widened, a `['string','null']` branch still walks, a 64-deep acyclic composition resolves its innermost enum, and a self-referential branch terminates.

Validation: `vp run -F './packages/core' test` → 67 files / 2455 tests passed; `typecheck` clean; `vp fmt` applied.

melloware:
running build now

melloware (approved):

Issue #3714, opened 2026-07-11
Normalize invalid nullable `$ref` schemas


snebjorn:
## Problem

Many OpenAPI 3.0 specs use a nullable reference like:

```yaml
# OpenAPI 3.0
$ref: '#/components/schemas/Pet'
nullable: true
```

This is invalid. Per the [Reference Object](https://spec.openapis.org/oas/v3.0.3#reference-object) rules, sibling properties on a `$ref` are ignored - so `nullable: true` on the above have no effect and does not mean `Pet | null`.

The correct way to make a reference nullable in OpenAPI 3.0 is:

```yaml
# OpenAPI 3.0
nullable: true
allOf:
  - $ref: '#/components/schemas/Pet'
```

However Orval currently tries to handle this and generates `Pet | null`. This makes the code messy as it's "out of spec" behavior. But removing this incorrect behavior will have a negative impact on users with invalid specs.

## Proposal

Add a built-in normalization pass in the spec import pipeline (possibly in `import-specs.ts`) that corrects these incorrect nullable refs.

When a schema object contains both `$ref` and `nullable: true`, rewrite it to:

```yaml
# OpenAPI 3.1
anyOf:
  - $ref: '#/components/schemas/Pet'
  - type: 'null'
```

Emit a `logWarning` for each conversion, including the JSON Pointer path (e.g. `components.schemas.Owner/properties/pet`), letting the users know what's going on.

Whether this should be converted before or after validation and/or upgrading to 3.1 is up for debate. Do whatever is the easiest to implement and gives the least problems :)

PR #3981, opened 2026-09-02
fix(core): stop using OAS 3.0 in tests because OAS is being upgraded to 3.1 at runtime

snebjorn:
This is a continuation of the effort to rid the code of OAS 3.0 relics.

Orval upgrades input OAS to 3.1 and tries to convert legacy properties like `nullable` to the new OAS 3.1 syntax.
Relevant issues:
- #3714
- https://github.com/scalar/scalar/issues/9694

This PR is about removing the incorrect usage of OAS 3.0 in tests as this influences what looks like correct code - meaning the code handles legacy properties like `nullable` because there's a test that says it should handle it. While at runtime the OAS should have been converted to correct 3.1.

This is not straight forward as sometimes `nullable` slips through the upgrade unconverted.


packages/core/src/getters/combine.test.ts
```diff
@@ -535,7 +526,12 @@ describe('combineSchemas (allOf required handling)', () => {
             },
             RefSiteWrapper: {
               allOf: [
-                { $ref: '#/components/schemas/RefSiteBase', nullable: true },
+                {
+                  anyOf: [
+                    { $ref: '#/components/schemas/RefSiteBase' },
+                    { type: 'null' },
+                  ],
+                },
```
snebjorn, line 534: @wadakatu I see your name on this, hope you can help.
I'm having a hard time understanding what this is testing :)

To me the allOf + anyOf combo seems incorrect, mostly because allOf is not combining anything.
Hope you can clarify.

packages/core/src/getters/combine.test.ts
```diff
@@ -1425,8 +1425,10 @@ describe('combineSchemas (allOf required handling)', () => {
               properties: { marker: { type: 'string' } },
               allOf: [
                 {
-                  $ref: '#/components/schemas/NullableParentBase',
-                  nullable: true,
+                  anyOf: [
+                    { $ref: '#/components/schemas/NullableParentBase' },
+                    { type: 'null' },
+                  ],
```
snebjorn, line 1431: @wadakatu I don't understand how a parent object removes null

I don't think `type` and `properties` is allowed as siblings to `allOf`. I think it's pretty much only allows `discriminator`

packages/core/src/getters/enum.test.ts
```diff
@@ -471,9 +470,8 @@ describe('getEnumMembers', () => {
 
   it('should not add null to enum members for a nullable enum', () => {
     const schema = {
-      nullable: true,
-      type: 'integer',
-      enum: [10, 20, 30],
+      type: ['integer', 'null'],
+      enum: [10, 20, 30, null],
     } as unknown as OpenApiSchemaObject;
```
snebjorn, line 475: This one is tricky... the tests fails when the OAS uses correct 3.1 syntax for a nullable enum.

The enum array MUST contain `null` if it's nullable.

```js
type: ['integer', 'null'],
enum: [10, 20, 30, null],
```

but can also be:
```js
anyOf: [
  { type: 'integer', enum: [10, 20, 30] },
  { type: null }
]
```

packages/core/src/getters/scalar.test.ts
```diff
@@ -317,10 +302,9 @@ describe('getScalar integer enum + const (#3758)', () => {
 
   it('preserves nullable suffix when const is present', () => {
     const schema: OpenApiSchemaObject = {
-      type: 'integer',
+      type: ['integer', 'null'],
       enum: [1],
       const: 1,
-      nullable: true,
     };
```
snebjorn, line 308: What's going on here?

`enum: [1]` means: value must be one of [1]
`const: 1` means: value must always be exactly 1

"value must always be exactly 1" literally means that it may never be null. But this tests check that null is preserved. Why :)?

@eduardheller I see your name on it. Can you elaborate?

snebjorn (commented):
I need some help figuring this out

melloware:
We also need to upgrade Scalar.  The new version has some changes and a new package.

snebjorn:
I've been trying to switch to the new OAS types from scalar for a while now. Unfortunately we have so many mistakes in our OAS usage that switching in our current state gives errors everywhere.

So this is a step towards updating to the new stricter OAS types from scalar. I expect this to take MANY months 😭

melloware:
CLosing this as handled in other PR's!

PR #3994, opened 2026-09-04
feat(query): add useSkipToken option to hold unresolved params

### LINKED ISSUE #3993: feat(query): hold unresolved params with `skipToken` instead of `enabled` (react-query v5)
### Description

Add an opt-in `override.query.useSkipToken` so that, with `allParamsOptional`, a query whose params are not resolved yet is held with `skipToken` rather than the generated `enabled` guard.

### Current behavior

When `allParamsOptional` is on, the factory puts the parameter check in `enabled`:

```ts
return { queryKey, queryFn, enabled: petId !== null && petId !== undefined, ...queryOptions }
```

The `enabled` option does not control `refetch()`. Thus a retry button, or a pull-to-refresh, sends the request. The URL contains the parameter that is not available:

```
idle    -> enabled: false | requests: 0
refetch -> requests: [ 'GET /pets/undefined' ]
```

This example uses the generated factory. The value of `petId` is not available. There is one call to `refetch()`. The `enabled` check does not prevent the request.

A caller can also replace the check. The `enabled` option has only one value, and the spread of `...queryOptions` is last. Thus the value from the caller replaces the generated value:

```ts
useShowPetById(petId, { query: { enabled: isTabVisible } })
```

If `petId` is `undefined`, and `isTabVisible` is `true`:

```
resolved enabled: true   (the parameter check is not there)
requests sent: [ 'GET /pets/undefined' ]
```

The two conditions have the same cause. The `enabled` option holds the condition "the parameter is not available". This condition is not a preference of the caller. The caller must not replace it, and `refetch()` must not ignore it.

The `skipToken` value holds the request in `queryFn`. Thus the two conditions do not occur:

```
refetch -> requests: [] | status: error
```

This last result is a change in behavior. If a caller calls `refetch()` too early, the query goes to the error state with `Missing queryFn`. Before this change, the caller receives a response that is not correct. I think that the error is better, but you must know about this change.

The `skipToken` value and `enabled: false` keep the query in the same state: `status: 'pending'`, `isPending: true`, and `isLoading: false`. This proposal does not change that state.

### Proposed behavior

When `useSkipToken` is on, the factory holds the query in `queryFn`. The `enabled` option stays available for the caller:

```ts
export const getShowPetByIdQueryOptions = <...>(
  petId: string | undefined | null,
  options?: {...},
) => {
  const queryKey = queryOptions?.queryKey ?? getShowPetByIdQueryKey(petId);
  const queryFn: QueryFunction<...> = ({ signal }) => showPetById(petId, { signal, ...fetchOptions });

  return {
    queryKey,
    queryFn: petId === null || petId === undefined ? skipToken : queryFn,
    ...queryOptions,
  } as ...
}
```

The factory does not emit the `enabled` check, because `skipToken` replaces it.

### Proposed solution

I have this change on my machine, against `master` (v8.28.1). It has 47 lines in 5 files:

- a new property, `useSkipToken?: boolean`, on `QueryOptions` and `NormalizedQueryOptions`. `packages/orval/src/utils/options.ts` makes it normal.
- `skipToken` in `REACT_QUERY_DEPENDENCIES`. The import filter keeps `skipToken` out of the files that do not use it.
- `generateQueryOptions` does not emit the `enabled` check when the option is on.
- `query-generator.ts` emits the new `queryFn`. It does this for the usual branch and for the `queryOptions` mutator branch.

The option operates only when `adapter.hasQueryV5` is true and `output.allParamsOptional` is true. If `allParamsOptional` is off, the parameters are of a necessary type. Thus the new branch is not possible.

The default of the option is off. All the tests pass with no change (306 unit tests, 6761 snapshot tests). No generated sample is different. The sample `samples/react-query/basic` compiles with `useCat = (petId?: string) => useShowPetById(petId)`.

This change is only for react-query. The `skipToken` value is also in solid-query, svelte-query, and vue-query for v5. But each adapter has a different `generateEnabledOption` function, and I do not have enough experience across those frameworks to confidently provide the same functionality.

oliverwinter-vdb:
Closes #3993.

`enabled` cannot carry "the param has not arrived yet":

- it does not gate `refetch()`, so a retry button or pull-to-refresh sends `GET /pets/undefined`;
- it is a single option and the caller's `...queryOptions` is spread last, so `useShowPetById(petId, { query: { enabled: isTabVisible } })` silently replaces the generated param check.

This adds an opt-in `override.query.useSkipToken` (default `false`) that holds the query in `queryFn` with TanStack's [`skipToken`](https://tanstack.com/query/latest/docs/framework/react/guides/disabling-queries#typesafe-disabling-of-queries-using-skiptoken) instead of emitting the `enabled` guard:

```ts
return {
  queryKey,
  queryFn: petId === null || petId === undefined ? skipToken : queryFn,
  ...queryOptions,
};
```

`enabled` then stays free for the caller, and `refetch()` cannot bypass the check.

### Scope

- React Query v5 only (`adapter.hasQueryV5`) — `skipToken` does not exist before v5. The other v5 adapters (solid/svelte/vue) each have their own `generateEnabledOption`; I don't work in those frameworks enough to be confident about the equivalent, so they are left alone.
- Replaces the `enabled` guard wherever that guard is emitted.
- Suspense queries are untouched: TanStack excludes `SkipToken` from their `queryFn` type, which is also why they get no `enabled` guard today.
- Query keys are unchanged.

### Behaviour change to be aware of

With the option on, `refetch()` before the params resolve puts the query into the error state (`Missing queryFn`) instead of firing a request with `undefined` in the URL. I think the error is the better outcome, but it is a visible difference for anyone enabling the option.

### Verification

- `@orval/query` unit tests: 193 pass, including new cases for `getQueryFnProperty` (multi-param, infinite, both suspense types) and `generateQueryOptions` (guard dropped, configured options kept, guard still emitted when off).
- `vp run test:snapshots`: 7350 pass, and regenerating every sample leaves no diff — with the default off this is a no-op.
- `vp fmt --check`, `vp run typecheck` and `vp lint --type-aware --type-check packages` are clean.
- Generated `samples/react-query/basic/petstore.yaml` with `useSkipToken: true` + `allParamsOptional: true`: the `showPetById` factory emits the `skipToken` branch and no `enabled` guard, the suspense factory is unchanged, and a probe calling `useShowPetById(petId, { query: { enabled: true } })` with `petId?: string` typechecks (`tsc --noEmit`). The scratch config and output were not committed.

### Docs

`docs/content/docs/guides/react-query.mdx` gets a short "Skip Token" section (what it emits and the two cases `enabled` does not cover), and `output.mdx` gets the `useSkipToken` reference entry.

## Summary by CodeRabbit

* **New Features**
  * Added optional TanStack Query v5 `skipToken` support for generated queries with unresolved parameters.
  * Unresolved queries send no request; early `refetch()` rejects with a `Missing queryFn` error.
  * Caller-provided query options remain supported.
  * Support is limited to React Query v5; other adapters and React Query v4 retain existing behavior.
  * Suspense queries remain unaffected.

* **Documentation**
  * Added guidance on configuration, behavior differences from `enabled`, parameter requirements, and usage recommendations.

* **Tests**
  * Added coverage for adapter compatibility and generated query options with and without `skipToken`.

### DIFF
--- a/docs/content/docs/guides/react-query.mdx
+++ b/docs/content/docs/guides/react-query.mdx
@@ -205,7 +205,7 @@ export const getShowPetByIdQueryOptions = <...>(
 
 This matters in two cases the `enabled` guard does not cover:
 
-- `enabled` does not gate `refetch()`, so a retry button sends the request with the unresolved param in the URL. `skipToken` refuses instead.
+- `enabled` does not gate `refetch()`, so a retry button sends the request with the unresolved param in the URL. With `skipToken` no request is sent — the query rejects with `Missing queryFn` instead, which is the tradeoff to know about before enabling the option.
 - `enabled` is a single option and `...queryOptions` is spread last, so `useShowPetById(petId, { query: { enabled: isTabVisible } })` replaces the generated param check. With `skipToken` the two gates are independent.
 
 Pair it with [`allParamsOptional`](/reference/configuration/output#allparamsoptional) so callers can pass an unresolved param without a cast.
--- a/docs/content/docs/guides/react-query.mdx
+++ b/docs/content/docs/guides/react-query.mdx
@@ -183,6 +183,35 @@ export const useGetListPetsQueryData = () => {
 };
 ```
 
+## Skip Token
+
+When `useSkipToken: true` is set, a query whose params are not resolved yet is held with [`skipToken`](https://tanstack.com/query/latest/docs/framework/react/guides/disabling-queries#typesafe-disabling-of-queries-using-skiptoken) rather than the generated `enabled` guard:
+
+```ts
+export const getShowPetByIdQueryOptions = <...>(
+  petId: string,
+  options?: {...},
+) => {
+  const queryKey = queryOptions?.queryKey ?? getShowPetByIdQueryKey(petId);
+  const queryFn: QueryFunction<...> = ({ signal }) => showPetById(petId, signal);
+
+  return {
+    queryKey,
+    queryFn: petId === null || petId === undefined ? skipToken : queryFn,
+    ...queryOptions,
+  } as ...;
+};
+```
+
+This matters in two cases the `enabled` guard does not cover:
+
+- `enabled` does not gate `refetch()`, so a retry button sends the request with the unresolved param in the URL. `skipToken` refuses instead.
+- `enabled` is a single option and `...queryOptions` is spread last, so `useShowPetById(petId, { query: { enabled: isTabVisible } })` replaces the generated param check. With `skipToken` the two gates are independent.
+
+Pair it with [`allParamsOptional`](/reference/configuration/output#allparamsoptional) so callers can pass an unresolved param without a cast.
+
+Suspense queries are unaffected: TanStack excludes `SkipToken` from their `queryFn`, which is also why they get no `enabled` guard.
+
 ## Full Example
 
 See the [complete React Query example](https://github.com/orval-labs/orval/blob/master/samples/react-query/basic) on GitHub.
--- a/docs/content/docs/reference/configuration/output.mdx
+++ b/docs/content/docs/reference/configuration/output.mdx
@@ -1253,6 +1253,17 @@ Query keys are matched by prefix, so query params and body arguments are widened
 
 Generate type-safe getQueryData helpers.
 
+### useSkipToken
+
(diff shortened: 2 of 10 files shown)


melloware:
looks like some AI Feedback

melloware (approved):

PR #3995, opened 2026-09-04
fix: sanitize OpenAPI document values before splicing into generated code (zod/effect/hono)

adamyordan:
## Summary

Follows up on a privately-reported issue: three generator emission sites still splice values taken directly from the input OpenAPI document into generated TypeScript without making them safe for the syntactic context they land in, so a document that isn't fully trusted can cause arbitrary code to run when the generated client is imported.

- **`packages/zod` / `packages/effect`** — numeric/length constraints (`minimum`, `maximum`, `exclusiveMinimum`, `exclusiveMaximum`, `multipleOf`) are emitted as a bare, unquoted module-level expression: `` export const XMin = ${min}; ``. There's no escaping that can make an unquoted position safe, and the default validator accepts a non-numeric value for these fields on an OpenAPI 3.1 numeric schema. Fixed by validating the value is actually a finite number before interpolating it; anything else now fails generation with a clear error instead of being emitted as code.
- **`packages/effect`** — default-value handling stringifies a value into a correctly single-quote-escaped literal, then swaps the quote delimiters for backticks (`.replaceAll("'", '`')`) without re-escaping the content. That turns a string literal into a template literal and re-arms `${...}` interpolation for anything the value contains. Fixed by keeping the delimiter the value was actually escaped for (dropping the swap).
- **`packages/hono`** — literal OpenAPI path text is emitted into a single-quoted literal inside the generated Hono route chain with **no escaping at all** (`getRoute`/`toColonRoutePath` only rewrite `{param}` placeholders and sanitize parameter names, not the literal segments). Fixed by escaping the path text with the existing `jsStringLiteralEscape` helper before interpolating it, matching how other sinks in this file already handle document text.

None of these require unusual configuration — they fire on an ordinary `orval` run against an untrusted document with default settings, and because the emissions are module-level statements, the generated code runs at **import** time, not when a request is made.

## Test plan

- [x] `bun run typecheck` — clean across the workspace
- [x] `bun run test` (zod/effect/hono packages) — 474 existing tests pass unchanged
- [x] Manually verified a malicious `minimum` value against the patched CLI now aborts generation with an explicit error instead of writing a file
- [x] Manually verified malicious `effect` default values and a malicious Hono path both stay inert (properly escaped) in the generated output rather than executing
- [x] Manually verified a legitimate spec — including string defaults containing literal backticks and `${...}` — still generates correct, valid output

## Bug Fixes

- Improved generated code safety by validating numeric and length constraints before emission.
- Prevented special characters in default values and route paths from being interpreted as executable template content.
- Improved handling of malformed or non-finite constraint values by reporting errors instead of generating invalid code.
- Correctly generates inclusive minimum and maximum constraints when OpenAPI 3.0 sets `exclusiveMinimum` or `exclusiveMaximum` to `false`.
- Added coverage for exclusive and inclusive boundary behavior.

### DIFF
--- a/packages/effect/src/effect.test.ts
+++ b/packages/effect/src/effect.test.ts
@@ -76,6 +76,32 @@ describe('constraints', () => {
     expect(effect).toContain('S.lessThan(');
   });
 
+  it('maps exclusiveMinimum/exclusiveMaximum=true (3.0 boolean form) to exclusive bounds', () => {
+    const { effect } = gen({
+      type: 'number',
+      minimum: 0,
+      maximum: 100,
+      exclusiveMinimum: true,
+      exclusiveMaximum: true,
+    } as unknown as OpenApiSchemaObject);
+    expect(effect).toContain('S.greaterThan(');
+    expect(effect).toContain('S.lessThan(');
+  });
+
+  it('maps exclusiveMinimum/exclusiveMaximum=false (3.0 boolean form) to inclusive bounds', () => {
+    const { effect } = gen({
+      type: 'number',
+      minimum: 0,
+      maximum: 100,
+      exclusiveMinimum: false,
+      exclusiveMaximum: false,
+    } as unknown as OpenApiSchemaObject);
+    expect(effect).toContain('S.greaterThanOrEqualTo(');
+    expect(effect).toContain('S.lessThanOrEqualTo(');
+    expect(effect).not.toContain('S.greaterThan(');
+    expect(effect).not.toContain('S.lessThan(');
+  });
+
   it('maps pattern to S.pattern', () => {
     const { effect } = gen({ type: 'string', pattern: '^foo' });
     expect(effect).toContain('S.pattern(');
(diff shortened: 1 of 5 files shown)


adamyordan:
Thanks for the review — both actionable points were real:

1. `exclusiveMinimum: false` / `exclusiveMaximum: false` (valid OpenAPI 3.0) was left as the boolean `false` instead of normalized to `undefined`, so it was mistaken for a constraint value and rejected as non-numeric. Fixed by normalizing `false` → `undefined` in both packages, with regression tests added for the boolean-false case in each.
2. Switched the rejection error's value formatting from `JSON.stringify` (renders `NaN`/`Infinity`/`-Infinity` as `"null"`) to `String`.

Pushed as 86e7bf5. Typecheck and the full test suite (478 tests, incl. the 4 new ones) pass.

melloware:
Running the build now.

melloware (approved):

Issue #3935, opened 2026-08-30
fix(angular-query): QueryClient should be a type-only import in prefetch helpers


the-ult:
Part of #3931.

`QueryClient` is imported as a value but only used as a type in the prefetch helpers.

```ts
// tests/__snapshots__/angular-query/use-prefetch/endpoints.ts:11
import {
  QueryClient,
  injectMutation,
  injectQuery,
  ...
} from '@tanstack/angular-query-experimental';
```

Source: `packages/query/src/client.ts` / the angular-query import builder; only the `usePrefetch` path adds `QueryClient`. Also 3 warnings in `samples/angular-query/src`, of which the generated ones are the same shape.

Expected: `type QueryClient` inline specifier, or a separate `import type`.

Reproduce: `vp lint --no-ignore tests/__snapshots__/angular-query`.

Issue #3931, opened 2026-08-30
Generated output trips consistent-type-imports: enforce for orval source, fix per generator


the-ult:
Generated code must not trip a consumer's linter. Running oxlint's `typescript/consistent-type-imports` over the committed generated output today gives 147 warnings in `tests/__snapshots__` and 66 in `samples`. None are compile errors: they are value imports of symbols that the file only uses in type positions. A consumer running `@typescript-eslint/consistent-type-imports` (or the oxlint rule) sees them as errors in files they cannot edit.

Split out of the review of #3926, which fixed the one case that *was* a compile error (`verbatimModuleSyntax` rejecting a value import of a `type` alias).

## Plan

1. Enable the rule as `error` for orval's own source (`packages/**`) and `warn` for generated output (`samples/**`, `tests/__snapshots__/**`), and add a `lint:snapshots` CI step so the warning count is visible on every PR. That is PR #3936.
2. Fix each generator so its output uses `import type` (or inline `type` specifiers) for type-only imports, regenerating the affected snapshots. One PR per generator:
   - [x] Angular — # (77 warnings across `angular`, `default`, `mock`, `runtime-validation` fixtures)
   - [x] Hono — # (60)
   - [x] MCP — # (9)
   - [x] angular-query — # (1)
3. Fix the hand-written sample sources that also trip the rule (`samples/*/src/**` mutators, `App.tsx`, `auth.context.tsx`, `app.config.ts`, `shims-vue.d.ts`). Not generator work; a single sweep with `vp lint --fix samples`.
4. When the warning count is zero, drop the `warn` override in `vite.config.ts` so the rule is `error` workspace-wide.

## Reproduce

```sh
vp lint --no-ignore tests/__snapshots__   # after #3936: lint:snapshots
vp lint samples
```

PR #3999, opened 2026-09-05
fix: emit type-only imports in hono handlers and mcp http-client

### LINKED ISSUE #3933: fix(hono): emit import type for context/schema and zod namespace imports
Part of #3931.

The Hono generator emits value imports for symbols its output only uses as types. 60 `consistent-type-imports` warnings in `tests/__snapshots__/hono`, 15 in `samples/hono/**`.

Two shapes, from `tests/__snapshots__/hono/petstore-split-with-handlers-kebab`:

```ts
// src/handlers/show-pet-with-owner.ts:9 — context and param/response types
import { ShowPetWithOwnerContext } from '../../endpoints.context';
import {
  ShowPetWithOwnerParams,
  ShowPetWithOwnerResponse,
} from '../../endpoints.schemas';

// endpoints.validator.ts:17,19 — namespace imports only used as `v3.ZodSchema` etc.
import * as v3 from 'zod/v3';
import * as v4 from 'zod/v4/core';
```

Source: `packages/hono/src/index.ts` (handler and context import blocks) and the validator template `packages/hono/src/zValidator.ts`. Note the template file is in the lint `ignorePatterns`; the emitted copy is what gets linted.

Expected: `import type { ... }` for the context/schema imports; `import type * as v3 from 'zod/v3'` for the namespace imports, if nothing in the validator uses them as values.

Reproduce: `vp lint --no-ignore tests/__snapshots__/hono`.

### LINKED ISSUE #3934: fix(mcp): emit import type for body/param types in http-client
Part of #3931.

The MCP generator's `http-client.ts` imports body/param/response types as values. 9 `consistent-type-imports` warnings in `tests/__snapshots__/mcp`, 4 in `samples/mcp/**`.

```ts
// tests/__snapshots__/mcp/annotations-coverage/http-client.ts:8
import { CreateThingBody, ReplaceThingBody, PatchThingBody } from './handlers';

// tests/__snapshots__/mcp/directory-target/http-client.gen.ts:8
import { Form } from './sample.gen';
```

Source: `packages/mcp/src/index.ts`, the http-client import block.

Expected: `import type { ... }`.

Reproduce: `vp lint --no-ignore tests/__snapshots__/mcp`.

ErfanBagheri404:
Fixes #3933, fixes #3934. Part of #3931.

## Summary

Two generators emitted value imports for symbols their output only uses as types, tripping `consistent-type-imports` in every generated hono/mcp file.

**hono** (`packages/hono/src/index.ts`, `packages/hono/src/handler-merge.ts`, `packages/hono/src/zValidator.ts`):

- Handler-file context imports are now `import type { ...Context } from '...context'` — context identifiers only appear in the typed handler callback (`async (c: XContext) => {}`). Both the fresh-file emission and the handler-merge reconciler (via a new `typeOnly` flag on `renderImport`) are covered.
- The zValidator template's `import * as v3 from 'zod/v3'` / `import * as v4 from 'zod/v4/core'` are now `import type *` — every usage is a type position (`schema as v3.ZodType` is an assertion, the rest are conditional types).

One deliberate deviation from the issue: the zod **schema** imports (`ShowPetWithOwnerParams`, etc.) stay value imports. `zValidator('param', ShowPetWithOwnerParams)` passes them to the validator at runtime, so making them type-only would break the generated code. The lint warnings on those lines were false positives from the rule's perspective of the file, but the emitted code compiles and runs correctly either way — the fix targets only genuinely type-only symbols.

**mcp** (`packages/mcp/src/index.ts`):

- The http-client's schema import (line ~562, `importImplementation`) is now `import type` — `Pets`, `ListPetsParams`, `CreatePetsBody`, etc. are referenced only in type aliases and function signatures.
- The server file's schemas import (`importSchemasImplementation`) is now `import type` — body/param type names used only as handler parameter types.

## Verification

- `vp lint --no-ignore tests/__snapshots__/hono tests/__snapshots__/mcp` — 0 `consistent-type-imports` warnings (previously 60 + 9)
- Snapshot/samples regenerated: 59 snapshot files + 14 sample files updated, all diffs are exactly the `import` → `import type` transitions
- hono package tests: 59/59 pass

## Scope note

#3935 (angular `QueryClient`) is intentionally **not** included. `QueryClient` is imported with `values: true` because the angular mutation invalidation path emits `const queryClient = inject(QueryClient);` — a genuine value use (see `samples/angular-query/src/api/endpoints/pets/pets.ts`). The static dependency list can't express "value only when invalidation is used", so flipping the flag would break real output. That one needs conditional emission, which is a design decision rather than a mechanical fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Generated Hono and MCP code now emits type-only imports for context, schema, and validation types.
  * Updated generated samples and handlers to use type-only imports consistently.
* **Refactor**
  * Import generation now distinguishes type-only imports from runtime imports.
  * Type-only imports are excluded from compiled runtime code, with no runtime behavior changes.

### DIFF
--- a/packages/hono/src/handler-merge.test.ts
+++ b/packages/hono/src/handler-merge.test.ts
@@ -103,7 +103,7 @@ export const getFooHandlers = factory.createHandlers(
     });
 
     expect(result).toContain(
-      "import { Get$FooContext } from '../endpoints.context';",
+      "import type { Get$FooContext } from '../endpoints.context';",
     );
     // user body untouched
     expect(result).toContain('async (c: Get$FooContext) =>');
@@ -719,7 +719,7 @@ export const listPetsHandlers = factory.createHandlers(
     expect(result).toContain('somethingCustom'); // user name preserved
     expect(result).toContain('export const createPetsHandlers'); // new op appended
     expect(result).toContain(
-      "import { CreatePetsContext } from '../endpoints.context';",
+      "import type { CreatePetsContext } from '../endpoints.context';",
     ); // new op's context added via a separate import
   });
 
@@ -854,7 +854,7 @@ export const listPetsHandlers = factory.createHandlers(
     expect(result).toContain("import * as Ctx from '../endpoints.context';"); // namespace kept
     expect(result).toContain('export const createPetsHandlers'); // new op appended
     expect(result).toContain(
-      "import { CreatePetsContext } from '../endpoints.context';",
+      "import type { CreatePetsContext } from '../endpoints.context';",
     ); // new op's bare context imported
     // the namespace-qualified existing context is not duplicated bare
     expect(result).not.toContain('import { ListPetsContext }');
--- a/packages/hono/src/handler-merge.ts
+++ b/packages/hono/src/handler-merge.ts
@@ -193,10 +193,16 @@ const localNameFor = (
   return undefined;
 };
 
-const renderImport = ({ names, module }: OrvalImport): string =>
-  names.length <= 1
-    ? `import { ${names.join('')} } from '${module}';`
-    : `import {\n  ${names.join(',\n  ')}\n} from '${module}';`;
+const renderImport = ({
+  names,
+  module,
+  typeOnly,
+}: OrvalImport & { typeOnly?: boolean }): string => {
+  const keyword = typeOnly ? 'import type ' : 'import ';
+  return names.length <= 1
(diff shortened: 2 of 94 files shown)


melloware (changes requested):
Failing build

melloware (approved):

PR #4004, opened 2026-09-06
feat: allow mutator return type and async to be inferred

### LINKED ISSUE #3667: customInstance: Allow return type and async to be inferred
### Description

I don't want the generated code to assume async or output type.

### Current behavior

Currently orval emits `Promise<createEntityResponse>` and `async` keywords:

```typescript
export const createEntity = async (
  createEntityBody?: NonReadonly<CreateEntityBody>,
  params?: CreateEntityParams,
  options?: RequestInit,
): Promise<createEntityResponse> => {
  return customInstance<createEntityResponse>(getCreateEntityUrl(params), {
    ...options,
    method: "POST",
    headers: { "Content-Type": "application/ld+json", ...options?.headers },
    body: JSON.stringify(createEntityBody),
  });
};
```

### Proposed behavior

Async and output type are redundant, because they can be inferred, output could be:

```typescript
export const createEntity = (
  createEntityBody?: NonReadonly<CreateEntityBody>,
  params?: CreateEntityParams,
  options?: RequestInit,
) => {
  return customInstance<createEntityResponse>(getCreateEntityUrl(params), {
    ...options,
    method: "POST",
    headers: { "Content-Type": "application/ld+json", ...options?.headers },
    body: JSON.stringify(createEntityBody),
  });
};
```

This has side benefit, customInstance can implement Effect style promises or not, it is totally up to the customInstance if the return type is Promise or not.

### Proposed solution

Do not emit `async` or `Promise<T>` when user gives in orval.config.ts `inferred: true`:

```typescript
export default defineConfig({
  "ngsi-ld": {
      //... 
      override: {
        mutator: {
          path: './src/fetcher.ts',
          name: 'customInstance',
          inferred: true
        },
      },
    },
```

<details>
<summary>Workaround with post-processing</summary>

I use the following workaround, it is ts-morph post processing the API file:

```typescript
function desyncFunctions(apiFile: SourceFile): void {
  for (const vd of apiFile.getVariableDeclarations()) {
    const init = vd.getInitializer();
    if (!init || init.getKind() !== SyntaxKind.ArrowFunction) continue;
    const arrowFn = init.asKindOrThrow(SyntaxKind.ArrowFunction);
    if (!arrowFn.isAsync()) continue;

    // Remove `async` keyword
    arrowFn.setIsAsync(false);
    // Remove return type, allowing it to be inferred from fetcher
    arrowFn.removeReturnType();
  }
}
```
</details>

ErfanBagheri404:
Fixes #3667

When `mutator.inferred: true` is set in the orval config, the generated functions omit the `async` keyword and the `: Promise<T>` return type annotation. The return type is inferred from the custom mutator, allowing it to return sync values, Effect-style promises, or any other wrapper type.

## Changes
- `MutatorObject` + `NormalizedMutator` + `GeneratorMutator`: added `inferred?: boolean`
- `normalizeMutator()`: passes through `inferred`
- `generateMutator()`: passes through `inferred`
- `packages/fetch/src/index.ts`: when `inferred`, skips `async` and return type annotation
- Added test config + spec + snapshot

## Generated output comparison

**Without** `inferred` (default):
```ts
export const listPets = async (options?: Parameters<typeof customFetch>[1]): Promise<listPetsResponse> => {
  return customFetch<listPetsResponse>(getListPetsUrl(), { ...options, method: 'GET' });
};
```

**With** `inferred: true`:
```ts
export const listPets = (options?: Parameters<typeof customFetch>[1]) => {
  return customFetch<listPetsResponse>(getListPetsUrl(), { ...options, method: 'GET' });
};
```

## Summary by CodeRabbit

- **New Features**
  - Added support for inferred mutators in generated fetch clients.
  - Inferred mutator settings are preserved throughout client generation.
  - Generated operations can support synchronous workflows and return results without forced promise wrapping.
  - Added configuration support for defining inferred mutators.
  - Added sample pets API coverage with typed pet-list responses.

- **Documentation**
  - Documented the inferred mutator configuration option and usage examples in supported languages.


packages/orval/src/utils/options.ts
```diff
@@ -1057,6 +1057,7 @@ function normalizeMutator(
       external: m.external,
       extension: m.extension,
       useHooks: m.useHooks,
+      inferred: m.inferred,
```
melloware, line 1060: Is this a new option that needs to be documented?

melloware (changes requested):
AI Feedback and does this need to be done for the Axios client as well which also used Custom M

packages/orval/src/utils/options.ts
ErfanBagheri404: Yes, it is a new option. When `mutator.inferred: true` is set, the generated fetch functions omit the `async` keyword and the `: Promise<T>` return type annotation, so the return type is inferred from the custom fetcher itself. This allows custom fetchers that return sync values, Effect-style promises, or any other wrapper type. Let me know if you would like me to add it to the docs.

melloware: shouldn't it by more clear like `asyncFetch` which defaults to true and when its false its not async?  If this is fetch only it should be documented in the Fetch section and made a `fetch` only property not a general property.

melloware:
@ErfanBagheri404 actually i saw his ticket and see that its for `mutator` only so yeah leave it inferred and just updated the docs please.

ErfanBagheri404:
Updated the docs (both English and Chinese config reference) to document the `inferred` option under the mutator section.

melloware (approved):

PR #4007, opened 2026-09-06
Fix/media type key injection

melloware:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved generated identifiers for unknown content types by preserving `$` characters and removing unsafe characters.
  * Added reliable fallback identifiers when content types contain no usable characters.
  * Prevented collisions between generated identifiers, including conflicts with naturally derived identifiers, through deterministic numeric disambiguation.
  * Ignored inherited properties during content-type identifier lookup, improving handling of unusual or crafted media types.

### DIFF
--- a/packages/angular/src/http-client.test.ts
+++ b/packages/angular/src/http-client.test.ts
@@ -1356,6 +1356,39 @@ describe('angular HttpClient generator', () => {
       expect(impl).toContain("accept: GetPetFileAccept = 'application/json'");
     });
 
+    it('escapes single quotes in accept media type keys', () => {
+      const verbOption = createVerbOption({
+        operationId: 'getPetFile',
+        operationName: 'getPetFile',
+        typeName: 'getPetFile',
+        response: baseResponse({
+          definition: { success: 'Pet | string', errors: 'Error' },
+          types: {
+            success: [
+              createSuccessType(
+                'Pet',
+                "application/json', 'X-Evil': 'injected",
+              ),
+              createSuccessType('string', 'text/plain'),
+            ],
+            errors: [],
+          },
+          contentTypes: [
+            "application/json', 'X-Evil': 'injected",
+            'text/plain',
+          ],
+        }),
+      });
+      const options = createGeneratorOptions();
+
+      const impl = generateHttpClientImplementation(verbOption, options);
+
+      expect(impl).toContain(
+        "accept: 'application/json\\', \\'X-Evil\\': \\'injected'",
+      );
+      expect(impl).not.toContain("'X-Evil': 'injected'");
+    });
+
     it('omits text dispatch for JSON and Blob response types', () => {
       const verbOption = createVerbOption({
         operationId: 'getPetImage',
--- a/packages/angular/src/http-client.ts
+++ b/packages/angular/src/http-client.ts
@@ -23,6 +23,7 @@ import {
   type GetterProp,
   GetterPropType,
   isBoolean,
+  jsStringLiteralEscape,
   makeRouteSafe,
   pascal,
   toObjectString,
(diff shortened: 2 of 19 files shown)


melloware:
Both CodeRabbit findings were valid — addressed in f471e91cc.

### `CONTENT_TYPE_SUFFIX_MAP` lookup

Confirmed, and the impact is a bit worse than the comment suggests. The lookup is a plain index on a spec-controlled key, so a media type named `toString` resolved `Object.prototype.toString`, returned it as truthy, and skipped sanitization entirely:

```ts
export const createThingWithfunction toString() { [native code] } = async (...
// esbuild: Expected "=" but found "toString"
```

Fixed with `Object.hasOwn`, which is already the idiom elsewhere in the repo (`generators/options.ts:702`, `utils/assertion.ts:26`).

### Colliding suffixes

Also valid, though it splits into two cases:

* `application/a-b` vs `application/a+b` → both `AB`. This one is **pre-existing** — `-` and `+` were already both split points before this PR.
* Two punctuation-only types → both `Content`. This one **was** introduced here, and it contradicted the comment I'd written claiming the fallback avoided collisions.

I fixed it by numbering repeats in `getBodiesByContentType` rather than inside `getContentTypeSuffix`, since the latter only ever sees one media type and can't detect a collision. The first occurrence keeps the plain suffix, so no non-colliding name moves and no existing generated output changes.

I went with an index (`AB`, `AB2`) rather than the suggested hash because it reads better in generated code. The trade-off is that names shift if a colliding sibling is added later — but that only affects specs whose output currently doesn't compile anyway.

I did **not** take the "preserve punctuation through safe encoding" option: encoding `-`/`+`/`.` into the identifier would rename every existing multi-content-type operation (`VndApiJson` → `Vnd_Api_Json` or similar), which is a breaking change to generated code in exchange for a collision that's close to theoretical.

### One more, not flagged by the review

My allowlist was stripping `$`, which is legal both in an RFC 6838 subtype and in a JS identifier — so `application/a$b` and `application/ab` were merging for no reason. That was a genuine regression in the previous commit. `$` is back in the allowlist.

### Result

```
createThingWithJson   createThingWithToString   createThingWithAB   createThingWithAB2
```

Four distinct names, file parses.

### Verification

Four new tests, each confirmed to fail without the fix — including `expected [Function toString] to be 'ToString'`, which pins the prototype issue directly.

`vp run test`, `vp run lint`, `vp run format:check` and `vp run test:snapshots` all pass, with zero drift in generated sample output.

melloware:
Valid — that's a real bug in my previous commit. Fixed in b4f433147.

The counter tracked *base* suffixes but never reserved the numbered ones it handed out, so a media type that already reduces to the numbered form collided with it. Reproduced against the actual algorithm:

```
media type          base   final
application/a-b     AB     AB
application/a+b     AB     AB2
application/a-b2    AB2    AB2   ← collision
```

Now the set of suffixes actually allocated is tracked, and the candidate increments until it's free. The first occurrence still keeps the plain suffix, so no non-colliding name moves:

```
createThingWithAB   createThingWithAB2   createThingWithAB22
```

`AB22` is not pretty, but it's unique, deterministic, and only reachable from a spec that currently produces output that doesn't compile at all.

Added both regression cases you asked for — `application/a-b` / `application/a+b` / `application/a-b2`, and the `Content` / `Content` / `application/content2` equivalent. Each asserts three *distinct* suffixes rather than pinning the third name, so the test doesn't lock in the exact disambiguation scheme. Both fail against the previous commit with `expected 2 to be 3`.

`vp run test`, `vp run lint`, `vp run format:check` and `vp run test:snapshots` all pass, zero drift in generated samples.

PR #4011, opened 2026-09-06
fix: quote mock enum members by their own type (GHSA-j7p3-24q8-rm47)

melloware:
Fixes the enum type-confusion code injection reported in GHSA-j7p3-24q8-rm47.

## The bug

`getEnum` decided how to quote an enum member from the schema's **declared** `type`:

```ts
type === 'string' || (type === undefined && isString(e))
  ? `'${jsStringLiteralEscape(e)}'`
  : e,
```

But the declared type and the members both come from the OpenAPI document, and nothing makes them agree. A schema declaring a numeric or boolean type with *string* members routes around the escape branch and has them spliced into the faker array as live expressions.

Spec:

```json
{ "type": "integer", "enum": ["(globalThis.pwned=1)"] }
```

Generated mock (before):

```js
faker.helpers.arrayElement([(globalThis.pwned=1)] as const)
```

Transpiled and executed with `faker` stubbed, this runs whenever `getXResponseMock()` is called — test setup, or per request:

```
mock output      : {"numEnum":1,"boolEnum":1}
injected executed: __NUM_PWNED, __BOOL_PWNED
```

The mock returning `1` — the assignment's value — rather than the string confirms it evaluated as an expression rather than being emitted as data.

## The fix

Branch on the member's own type instead of the declared one, in a small `formatEnumMember` helper. After:

```js
faker.helpers.arrayElement(['(globalThis.pwned=1)'] as const)
```

```
injected executed: none
```

`type` stays a parameter because it still gates the `Object.values(...)` reference shortcut further down — it just no longer decides quoting.

## Two more bugs from the same root cause

Probing adjacent cases turned up the mirror images of the same confusion, both fixed here:

| Case | Before | After |
|---|---|---|
| `type: 'string'`, members `[1, 2]` | 🛑 `input.replaceAll is not a function` — generation aborts | `[1,2]` |
| `type: 'integer'`, member `{"a":1}` | `[object Object]` — not valid syntax | `{"a":1}` |

## No behaviour change for valid schemas

Verified byte-identical output for every legitimate shape:

| Schema | Output |
|---|---|
| `type: integer, enum: [1,2]` | `[1,2]` |
| `type: boolean, enum: [true,false]` | `[true,false]` |
| `type: string, enum: ['a',"b'c"]` | `['a','b\'c']` |
| untyped `enum: ['x',3,true]` | `['x',3,true]` |

## Review follow-up: referenced enums

CodeRabbit caught a real consequence of the fix above, confirmed empirically. The `Object.values(<Enum>)` shortcut fired for any referenced enum declaring `type: 'string'`, without checking the members really are strings. Numeric members emit a *numeric* TS enum, and those compile to a two-way map:

```
NumStatus object : {"1":"NUMBER_1","2":"NUMBER_2","NUMBER_1":1,"NUMBER_2":2}
Object.values    : ["NUMBER_1","NUMBER_2",1,2]    <- member names leak in
StrStatus values : ["A","B"]                       <- string enum is fine
```

So the mock could return `"NUMBER_1"`, a value the schema never declared. Notably this only became reachable *because* of the parent commit — the same input previously crashed generation outright.

Now gated on `item.enum.every(isString)`:

| Schema | Before | After |
|---|---|---|
| ref, `type: string`, members `[1,2]` | `Object.values(NumStatus)` | `[1,2] as NumStatus[]` |
| ref, `type: string`, members `['A','B']` | `Object.values(StrStatus)` | unchanged |

Verified the emitted `[1,2] as NumStatus[]` type-checks under `strict`.

## Testing

Seven new tests. Six fail against the unfixed generator; the seventh is the regression guard that must pass both ways. The crash case surfaces as `TypeError: input.replaceAll is not a function`, pinning that fix too.

> [!NOTE]
> This branch also carries a 4-line pure line-wrap in `packages/mock/src/msw/mocks.test.ts`. That file was left unformatted on `master` by #4009 and fails `format:check`; `vp run format` picked it up here. No semantic change, included so CI is green rather than red on an unrelated pre-existing issue.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved mock data generation for enum values with mixed or unexpected types.
  - String values are now correctly quoted and escaped.
  - Numeric and boolean values are emitted as valid literals.
  - Object values are serialized as JSON instead of displaying as `[object Object]`.
  - Prevented generation failures when enum values differ from the declared schema type.
  - Referenced enums containing numeric values now remain correctly inlined during mock generation.

### DIFF
--- a/packages/mock/src/faker/getters/scalar.test.ts
+++ b/packages/mock/src/faker/getters/scalar.test.ts
@@ -1230,3 +1230,119 @@ describe('getMockScalar (referenced string enum by enumGenerationType #3690)', (
     );
   });
 });
+
+describe('getMockScalar (enum member type confusion)', () => {
+  const baseArg = {
+    imports: [],
+    operationId: 'test-operation',
+    tags: [],
+    existingReferencedProperties: [],
+    splitMockImplementations: [],
+  };
+
+  // The declared `type` and the members both come from the document and need
+  // not agree. Quoting keyed off the declared type let a string member under a
+  // numeric/boolean type through as a live expression.
+  it('quotes a string member declared under type: integer', () => {
+    const result = getMockScalar({
+      ...baseArg,
+      item: {
+        type: 'integer' as OpenApiSchemaObjectType,
+        enum: ['(globalThis.pwned=1)'],
+        name: 'numEnum',
+      },
+      context: scalarContext(),
+    });
+
+    expect(result.value).toBe(
+      "faker.helpers.arrayElement(['(globalThis.pwned=1)'] as const)",
+    );
+  });
+
+  it('quotes a string member declared under type: boolean', () => {
+    const result = getMockScalar({
+      ...baseArg,
+      item: {
+        type: 'boolean' as OpenApiSchemaObjectType,
+        enum: ['(globalThis.pwned=1)'],
+        name: 'boolEnum',
+      },
+      context: scalarContext(),
+    });
+
+    expect(result.value).toBe(
+      "faker.helpers.arrayElement(['(globalThis.pwned=1)'] as const)",
+    );
+  });
+
+  it('escapes a quote in a string member declared under type: integer', () => {
+    const result = getMockScalar({
+      ...baseArg,
+      item: {
+        type: 'integer' as OpenApiSchemaObjectType,
+        enum: ["a',(globalThis.pwned=1),'b"],
+        name: 'numEnum',
+      },
+      context: scalarContext(),
+    });
+
+    expect(result.value).toBe(
+      String.raw`faker.helpers.arrayElement(['a\',(globalThis.pwned=1),\'b'] as const)`,
+    );
+  });
+
+  it('emits a numeric member declared under type: string without crashing', () => {
+    const result = getMockScalar({
+      ...baseArg,
+      item: {
+        type: 'string' as OpenApiSchemaObjectType,
+        enum: [1, 2] as unknown as string[],
+        name: 'strEnum',
+      },
+      context: scalarContext(),
+    });
+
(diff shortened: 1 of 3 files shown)


melloware:
Good catch — this is real, and I confirmed it empirically rather than taking it on faith. Fixed in d8ce748ad.

The generated model for a mismatched enum is a *numeric* TS enum, and those compile to a two-way map:

```
NumStatus object : {"1":"NUMBER_1","2":"NUMBER_2","NUMBER_1":1,"NUMBER_2":2}
Object.values    : ["NUMBER_1","NUMBER_2",1,2]    <- member names leak in
StrStatus values : ["A","B"]                       <- string enum is fine
```

So the mock could return `"NUMBER_1"` — a value the schema never declares.

One detail worth recording: this path was **unreachable before this PR**. A numeric member under `type: 'string'` used to crash generation outright (`input.replaceAll is not a function`), so the shortcut never ran on mismatched members. Making that case generate successfully is what exposed the latent flaw — so it belongs in this PR rather than being pre-existing.

Applied your gate. End-to-end:

| Schema | Before | After |
|---|---|---|
| ref, `type: string`, members `[1,2]` | `Object.values(NumStatus)` | `[1,2] as NumStatus[]` |
| ref, `type: string`, members `['A','B']` | `Object.values(StrStatus)` | unchanged |

I also checked the fallback actually type-checks — the inline path appends the `as NumStatus[]` cast, and `[1,2] as NumStatus[]` compiles clean under `strict`.

Added the regression case you asked for (`isRef: true`, `type: 'string'`, numeric members, `EnumGeneration.ENUM`), asserting both that `Object.values` is absent and that no `values: true` import is emitted. It fails against the parent commit with `expected 'faker.helpers.arrayElement(Object.val…' not to contain 'Object.values'`.

PR #4012, opened 2026-09-06
refactor(core): share the per-file assembly steps between the mode writers

wadakatu:
The six output-mode writers in `packages/core/src/writers` (`single`, `split`, `tags`, `tags-split`, `tags-operations`, `tags-operations-split`) each carried their own copy of the same four steps for every implementation file they emit:

1. filter the imports down to the ones the implementation actually references,
2. build the client import header via `builder.imports({...})`,
3. append the mutator imports (seven consecutive `if` blocks), and
4. append the `NonReadonly<>` / `TypedResponse<>` helper types.

This PR moves those four steps into `writers/implementation-parts.ts` and calls them from each writer. The layout-specific logic (where files go, how mocks are split, barrels) stays in each mode file untouched.

Generated output is byte-for-byte unchanged across every snapshot config (6909 snapshot files), and the writer unit tests pass as before.

### Also in this PR

- **Snapshot coverage for `tags-operations` and `tags-operations-split`.** These two modes had unit tests but no generated-output snapshots, so a change to the writer pipeline could alter their output unnoticed. The first commit adds petstore configs for both modes under the `fetch` (with mocks) and `react-query` clients and records the current output as the baseline, before the refactor commit touches any writer.
- **`tags-operations` / `tags-operations-split` now escape regex metacharacters when matching import names**, as the other four writers already did. The shared helper carries the `escapeRegExp` call, so the two copies that had drifted are now aligned. No snapshot changed as a result.

### Net effect

| | lines |
|---|---|
| removed from the six writers | 461 |
| added to the six writers | 105 |
| new `implementation-parts.ts` | 133 |

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved generated client code across supported output modes, including more consistent imports, mutator integration, and helper type generation.
  - Reduced unnecessary imports in generated implementations, helping produce cleaner output.

- **Tests**
  - Added coverage for tags-based operation generation in Fetch and React Query configurations.
  - Added scenarios validating generated clients with mocks, formatting, and cleaned output directories.


melloware:
I love this refactor!

wadakatu:
Superseded: split into smaller PRs so each is reviewable on its own. Snapshot baselines: #4014, #4015, #4016, #4017. Writer refactor: #4018.

PR #4014, opened 2026-09-06
test: add tags-operations snapshots for the fetch client

wadakatu:
The `tags-operations` output mode has unit tests in `packages/core/src/writers/tags-operations-mode.test.ts` but no generated-output snapshots under `tests/`, so a change to the writer pipeline could alter its output without any test noticing.

This adds one petstore config for `tags-operations` with the `fetch` client and mocks enabled to `tests/configs/fetch.config.ts`, and records the current output as the baseline. The snapshots are generated from the current `master`; nothing else changes.

Companion PRs cover the other client/mode combinations so each one stays small enough to skim.

## Summary by CodeRabbit

* **Tests**
  * Added fetch test coverage for generating Petstore endpoints and schemas grouped by tags and operations.
  * Enabled mock responses and formatted generated output for this test configuration.
  * Added coverage for generating fetch-based output using an inferred custom mutator.

**Depends on #4031.** The first recorded snapshot exposed a real bug in this mode (the aggregate `get<Tag>Mock` returned handler factories instead of handlers); #4031 fixes it, and the second commit here records the fixed output. CI on this PR goes green once #4031 is merged.


wadakatu:
Copilot's review comments here are correct: `getPetsMock` returned the bare factory (`[getListPetsMockHandler]`) instead of calling it. That is a pre-existing bug in the `tags-operations` writer, not something introduced by this PR; the snapshot merely recorded it. Fixed in #4031, and the second commit on this branch regenerates the six affected `.msw.ts` snapshots with the corrected output. The earlier CI failure was an unrelated 5s timeout in `generate-spec.test.ts` on the Windows runner.

PR #4022, opened 2026-09-07
feat: add fetch client factory injection mode

### LINKED ISSUE #3265: Fetch client: Support fetch function injection
### Description

Just like #2848, the fetch client should also support dependency injection.

I would like to inject the generated APIs with a fetch function that adds authentication headers and sets base URL for the API at runtime. Something like the following function in a factory class that provides the rest of the application with the API:

```ts
const customFetch = (url, options) => {
  return fetch(new URL(url, this.appConfig.getApiBaseUrl()), {...options, ...this.authenticator.getAuthHeaders()});
}
```

### Current behavior

The generated API is a collection of plain functions that call `fetch()` directly. Recently, we got a parameter to replace the fetch function; however, this still means that every API call needs to receive this new function:

```ts
import {getSomething} from "./api";
const data = getSomething({name}, addAuthHeaders({...options}), fetchFnWithBaseDir);
```

The currently generated API:

```ts
export const getSomething = async ( options?: RequestInit, fetchFn?: typeof globalThis.fetch): Promise<getSomethingResponseSuccess> => {
    const res = await (fetchFn ?? fetch)(getGetSomethingUrl(),
    // ...
}
```

### Proposed behavior

Generate factory functions that accept the `fetch` function as a parameter, allowing runtime dependency injection:

```ts
export const createSomethingApi(fetchFn?: typeof globalThis.fetch) => ({
    getSomething: async ( options?: RequestInit ): Promise<getSomethingResponseSuccess> => {
        const res = await (fetchFn ?? fetch)(getGetSomethingUrl(),
        // ... no changes in the function body
    },
});
```

Then the usage of the API is much easier:

```ts
// Prepare the API
const api = createSomethingApi(customFetch);

// ... and everywhere else it is just:
const data = await api.getSomething(name);
```

### Proposed solution

There shall be the same option for the fetch client just like the one #2848 introduced: `httpClientInjection: 'factory'`

The change is only how the `fetchFn` is provided to the generated functions -- a closure (factory) instead of the parameter.

ErfanBagheri404:
Fixes #3265

## Summary by CodeRabbit

- **New Features**
  - Added factory mode for generated fetch clients, allowing a `fetch` implementation to be provided once when creating the client instead of on every request.
  - Factory-based clients now expose available API operations through the created client instance.
  - Fetch client generation now supports customized titles and footer content in generated files.
  - Added support for passing output configuration to footer customization callbacks.


soartec-lab (commented):
I’ve reviewed the source code, so I’m going to wait a little while to see how others react in related issue before proceeding; I’ll hold off on the PR for a short period.

PR #4033, opened 2026-09-07
fix: parse response status keys before emitting them (GHSA-q8g3-2pxp-2gmw)

melloware:
Fixes the MSW status-code injection reported in GHSA-q8g3-2pxp-2gmw.

## The bug

The response key from the OpenAPI document was emitted into an **unquoted expression position**, after only rewriting a `NXX` wildcard:

```ts
const statusCode = status === 'default' ? 200 : status.replace(/XX$/, '00');
// ...
{ status: ${statusCode} }
```

There is no quote to escape in that position, so a crafted key is spliced in as live code. Response key:

```
2,x:(globalThis.__STATUS_PWNED=1)
```

Generated handler (before):

```js
return HttpResponse.json(..., { status: 2,x:(globalThis.__STATUS_PWNED=1)
})
```

Evaluating that object literal:

```
keys           : [ 'status', 'x' ]
__STATUS_PWNED : 1
```

The injected expression runs whenever the generated handlers load — test runner, or `setupWorker` in the browser.

## Reachability — worth being precise

The advisory lists the precondition as just `mock: true`. In practice the default validator already rejects such a key:

```
🛑 poc - OpenAPI spec validation failed:
   Property 2,x:(globalThis.__STATUS_PWNED=1) is not expected to be here
```

Reproducing it needs `input.unsafeDisableValidation: true`, which is documented as unsupported. So the practical severity is well below Critical.

I fixed it regardless: leaning on an external validator to prevent code injection is a thin defence — if `@scalar/openapi-parser` ever loosens its response-key rule this becomes live with no change on our side — and parsing the value costs nothing.

## The fix

Parse the key to a three-digit status and refuse the document otherwise, mirroring how the zod/effect generators already handle numeric constraints that escaping cannot make safe:

```
🛑 orval: refusing to generate a mock handler for an OpenAPI response key that
   is not a status code (got "2,x:(globalThis.__STATUS_PWNED=1)").
```

Applied at both derivation points rather than at each emission:

- `statusCode` — feeds 10 emission sites
- `noContentStatusCode` — the void-union branch, also a raw response key (`responses.find(...)?.key`), which the advisory doesn't mention

That covers all 11 sites at the source, so a new emission site can't reintroduce it.

Generated msw output is **byte-identical** before and after — verified by `diff` on a spec exercising `default`, `2XX`, `5XX`, plain `200`/`404`, and a `200`/`204` void union, with `generateEachHttpStatus` on:

```
IDENTICAL - no behaviour change for valid specs
```

## Testing

Nine new tests. Five pin the valid mappings (`200`→200, `default`→200, `2XX`→200, `5XX`→500) and pass both before and after as regression guards; four assert refusal of non-numeric keys and fail against the unfixed generator.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved generated mock handlers’ handling of response status codes.
  - Valid status keys, including numeric codes, `default`, and supported wildcard patterns, are converted to valid numeric status values.
  - Invalid or potentially unsafe status keys—including malformed wildcard patterns—are now rejected with a clear error instead of being emitted into generated handler code.

### DIFF
--- a/packages/mock/src/msw/index.test.ts
+++ b/packages/mock/src/msw/index.test.ts
@@ -2242,12 +2242,22 @@ describe('response status key safety', () => {
 
   // The key lands in an unquoted expression position, so a non-numeric value
   // would splice live code into the handler rather than break a string.
-  it.each(['2,x:(globalThis.pwned=1)', '200,y:1', '(1)', 'abc'])(
-    'refuses the non-numeric response key %j',
-    (statusKey) => {
-      expect(() => generateMSW(verbOptions(statusKey), options)).toThrow(
-        /not a status code/,
-      );
-    },
-  );
+  it.each([
+    '2,x:(globalThis.pwned=1)',
+    '200,y:1',
+    '(1)',
+    'abc',
+    // Near-miss wildcards: only a single-digit `NXX` is a valid wildcard, so
+    // none of these may be salvaged into a three-digit status.
+    '20XX',
+    '2XXX',
+    'XX',
+    '1000',
+    '20X',
+    '',
+  ])('refuses the non-numeric response key %j', (statusKey) => {
+    expect(() => generateMSW(verbOptions(statusKey), options)).toThrow(
+      /not a status code/,
+    );
+  });
 });
--- a/packages/mock/src/msw/index.test.ts
+++ b/packages/mock/src/msw/index.test.ts
@@ -2183,3 +2183,71 @@ describe('recursion guards for cyclic allOf schemas', () => {
     expect(() => run(schemas, 'XElement')).not.toThrow();
   });
 });
+
+describe('response status key safety', () => {
+  const verbOptions = (statusKey: string): GeneratorVerbOptions =>
+    ({
+      operationId: 'getUser',
+      operationName: 'getUser',
+      typeName: 'getUser',
+      verb: 'get',
+      tags: [],
+      response: {
+        imports: [],
+        definition: { success: 'User' },
+        types: { success: [{ key: statusKey, value: 'User' }] },
+        contentTypes: ['application/json'],
+      },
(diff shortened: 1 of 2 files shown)


melloware:
The specific claim doesn't hold — `20XX` was already rejected. But the underlying point about legibility is fair, so I've taken the suggestion anyway. Pushed in 909bf0dc3.

## Why it wasn't a bug

`'20XX'.replace(/XX$/, '00')` yields `'2000'` — four digits, so it fails `/^\d{3}$/` and is refused. It never became `200`.

I ran both forms across 17 inputs before changing anything:

| key | shipped | proposed |
|---|---|---|
| `200` / `404` / `999` | accepted | accepted |
| `default` | 200 | 200 |
| `2XX` / `5XX` / `1XX` | 200 / 500 / 100 | same |
| `20XX` | **REJECTED** | **REJECTED** |
| `2XXX` / `XX` / `20X` / `1000` / `''` / `abc` | REJECTED | REJECTED |

**Zero differing cases.** The proposal is behaviourally identical, not a fix.

## Taken anyway

A reviewer having to derive that `20XX` → `2000` → rejected is itself a signal the code stated the wrong thing. It now names the two shapes the spec actually defines rather than leaving them to be inferred, with a comment recording that the old form worked out the same. Zero-risk given the equivalence above.

## The test suggestion was the valuable part

Added `20XX`, `2XXX`, `XX`, `1000`, `20X` and `''` as rejection cases.

Worth being explicit: **these pass against the pre-change code too** — I ran the full suite against both normalizations and got 84/84 either way. So they aren't regression tests for a bug; they pin behaviour that was correct but untested, so a future looser normalization can't quietly start accepting them. That's the real gap your comment surfaced.

`vp run lint`, `vp run format:check`, `vp run test` and `vp run test:snapshots` all pass, zero snapshot drift.

PR #4035, opened 2026-09-07
fix: validate numeric constraints on the array and faker paths (GHSA-xm38-vjwp-qrg4)

melloware:
Fixes the numeric-constraint injection reported in GHSA-xm38-vjwp-qrg4.

## The gap

Commit d346d94a added `assertSafeNumericConstraint` to the **scalar** constraint block only. Three other paths emit the same spec-supplied constraints as bare, unquoted expressions and were never guarded:

| Generator | Sink | When it runs |
|---|---|---|
| zod | `.min(${rules.min})` via `renderArrayWithBounds` | **import** |
| effect | `.pipe(S.minItems(${rules.min}))` | **import** |
| faker | `faker.number.int({min: ${numMin}})`, `faker.string.alpha({length: {min: ${strMin}}})` | mock call |

There is no surrounding quote in any of them, so no escaping helps — the value has to be a number.

## Reachable by default

An OpenAPI **3.1** document passes validation with these fields set to arbitrary strings, so no `unsafeDisableValidation` is needed. Confirmed:

```yaml
type: array
minItems: "1); globalThis.__ZOD_PWNED=1; void(0"
items: { type: string }
```

Emitted zod (module scope):

```ts
export const ListPetsResponse = zod.array(ListPetsResponseItem).min(1); globalThis.__ZOD_PWNED=1; void(0)
```

Transpiled and run with the library stubbed:

```
parse: OK (compiles, so it ships)
injected executed at IMPORT time: __ZOD_PWNED
```

Effect behaves the same with a payload matched to its `.pipe(S.minItems(…))` shape (`__EFF_PWNED` fires at import). The faker sink is runtime rather than import-time, which the harness separates cleanly:

```
before mock call : __FAKER_PWNED=undefined  __STR_PWNED=undefined
mock output      : {"count":1,"name":"x"}
after mock call  : __FAKER_PWNED=1  __STR_PWNED=1
```

## Beyond the report

Auditing the faker path turned up three more sinks of the same class that the advisory doesn't list — `multipleOf` (both the int and float branches) and the array `minItems`/`maxItems` bounds. All fixed here.

## Consolidated rather than copied

The guard already existed twice, byte-identical, in zod and effect. Rather than add a fourth copy for mock, it moves to `@orval/core` and all three generators share it, alongside a `safeNumericConstraint` variant that passes `undefined` through so "absent" stays distinct from "present but not a number". The two local copies are gone.

After:

```
🛑 orval: refusing to generate code for an OpenAPI document whose
   "minimum/minLength/minItems" constraint is not a finite number (got 1); globalThis...).
```

No file is written for any of the three PoCs.

Generated output is **byte-identical** before and after — verified with `diff -r` across zod, effect and faker output for a spec exercising `minimum`/`maximum`/`multipleOf`, exclusive bounds, `minLength`/`maxLength` and `minItems`/`maxItems`:

```
IDENTICAL — no behaviour change for valid specs
```

## Testing

17 new tests:

- **10** on the shared core guard, covering code-payload strings, numeric strings, `NaN`, `Infinity`, booleans, `null`, objects, and the `undefined` passthrough
- **7** on the faker sinks via `getMockScalar`, one per constraint field, all failing against the unfixed generator

> [!NOTE]
> Coverage is uneven by design: the faker sinks have direct unit tests because `getMockScalar` is exported, but the zod/effect array-rule sinks live inside the unexported `parseBodyAndResponse`, so they are covered by the shared-guard unit tests plus the end-to-end PoC verification above rather than by their own unit tests. I did not export another internal just to reach them — say the word if you would rather have that.

## Summary by CodeRabbit

* **Bug Fixes**
  * Numeric constraints are now validated before generating schemas, responses, and mock data.
  * Invalid values—including non-numeric, non-finite, boolean, null, and object inputs—now produce validation errors instead of invalid generated output.
  * Optional constraints remain supported when omitted.
  * Valid finite numeric constraints continue to work across minimum, maximum, length, item-count, and multiple-of settings.

* **Tests**
  * Added coverage for valid, omitted, and invalid numeric constraint values.

### DIFF
--- a/packages/core/src/utils/assertion.test.ts
+++ b/packages/core/src/utils/assertion.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from 'vite-plus/test';
 
 import { SchemaType, Verbs } from '../types';
 import {
+  assertSafeNumericConstraint,
   isBoolean,
   isDirectory,
   isDynamicReference,
@@ -18,6 +19,7 @@ import {
   isStringLike,
   isUrl,
   isVerb,
+  safeNumericConstraint,
 } from './assertion';
 
 describe('assertion testing', () => {
@@ -184,5 +186,44 @@ describe('isDynamicReference', () => {
 
   it('returns true for objects with $ref in isReference', () => {
     expect(isReference({ $ref: '#/components/schemas/Foo' })).toBe(true);
+  });
+});
+
+describe('assertSafeNumericConstraint', () => {
+  it('returns a finite number unchanged', () => {
+    expect(assertSafeNumericConstraint(0, 'minimum')).toBe(0);
+    expect(assertSafeNumericConstraint(-1.5, 'minimum')).toBe(-1.5);
+  });
+
+  // These land in generated source as bare expressions, so there is no quote
+  // to escape and the value has to be rejected outright.
+  it.each<[unknown, string]>([
+    ['0); globalThis.pwned=1; void(0', 'a code payload string'],
+    ['3', 'a numeric string'],
+    [Number.NaN, 'NaN'],
+    [Number.POSITIVE_INFINITY, 'Infinity'],
+    [true, 'a boolean'],
+    [null, 'null'],
+    [{}, 'an object'],
+  ])('rejects %j (%s)', (value: unknown) => {
+    expect(() => assertSafeNumericConstraint(value, 'minimum')).toThrow(
+      /"minimum" constraint is not a finite number/,
+    );
+  });
+});
+
+describe('safeNumericConstraint', () => {
+  it('passes undefined through so absent stays distinct from invalid', () => {
+    expect(safeNumericConstraint(undefined, 'minItems')).toBeUndefined();
+  });
+
+  it('returns a finite number unchanged', () => {
+    expect(safeNumericConstraint(7, 'minItems')).toBe(7);
+  });
+
+  it('rejects a non-numeric value', () => {
+    expect(() =>
+      safeNumericConstraint('1); globalThis.pwned=1; void(0', 'minItems'),
+    ).toThrow(/"minItems" constraint is not a finite number/);
   });
 });
--- a/packages/core/src/utils/assertion.ts
+++ b/packages/core/src/utils/assertion.ts
@@ -1,6 +1,6 @@
 import path from 'node:path';
 
-import { isFunction, isNullish, isString } from 'remeda';
+import { isFunction, isNullish, isNumber, isString } from 'remeda';
 
 import {
   type ClientMockBuilder,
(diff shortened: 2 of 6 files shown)


melloware:
Two findings — one valid and fixed in 2dac066ba, one I'm declining. Details on both.

## ✅ Non-finite exclusive bounds — valid, fixed

You're right, and this was an inconsistency in my own patch: I routed the *inclusive* bounds through `safeNumericConstraint` but left the exclusive ones on the bare `typeof === 'number'` check. YAML can express `.nan` / `.inf`, which are numbers, so:

```yaml
a: { type: number, exclusiveMinimum: .nan }
b: { type: number, exclusiveMaximum: .inf }
```

generated, before:

```js
faker.number.float({min: NaN, fractionDigits: 2})
faker.number.float({max: Infinity, fractionDigits: 2})
```

Worth noting zod and effect already reject these on their scalar path — faker was the only generator still emitting them, which made the gap a straight inconsistency rather than a design choice. Now refused:

```
🛑 orval: refusing to generate code for an OpenAPI document whose
   "exclusiveMinimum" constraint is not a finite number (got NaN).
```

Added the six regression cases you asked for (`NaN` and both infinities across `exclusiveMinimum`/`exclusiveMaximum`, plus `minimum`/`maximum`). The four exclusive ones fail against the previous commit; the two inclusive ones pass either way, so they act as guards for the original fix.

## ❌ Nullish-coalescing bound selection — declining

The claim is that `minItems: null` "generates without that bound instead of failing validation." I traced the chain before changing anything:

| schema | `min ?? minLength ?? minItems` | outcome |
|---|---|---|
| `minItems: null` alone | `null` | **THROWS** — caught |
| `minimum: null` alone | `undefined` | no bound emitted |
| `minimum: null, minLength: 5` | `5` | emits 5 |

So the stated case is wrong: `minItems: null` *does* fail validation. The leniency only applies to `null` in the **first** link of the chain, where `null ?? undefined` collapses to `undefined`.

Declining for three reasons:

1. **No security impact.** Every path either emits a finite number or emits nothing. Nothing unvalidated reaches generated source, which is what this PR is about.
2. **Pre-existing and unchanged.** The `??` chain is byte-identical to what was there before this PR — I only added the assertion after it. Tightening it here would be unrelated scope on a security fix.
3. **It's a behaviour change with real breakage risk.** Making `minimum: null` throw would start rejecting documents that generate cleanly today. `null` isn't valid for these fields in OpenAPI anyway, so treating it as absent is defensible leniency.

Happy to do it as a separate strictness change if you want `null` rejected outright — it just shouldn't ride along here.

`vp run lint` (type-aware + type-check), `vp run format:check`, `vp run test` and `vp run test:snapshots` all pass, zero snapshot drift.

PR #4036, opened 2026-09-07
fix: keep zod object defaults with tuples or ref arrays inline

### LINKED ISSUE #4023: Zod tuple defaults are generated with an incompatible array type
### Description

When generating Zod schemas for an object with a tuple property and a default value, Orval generates the default array with an incompatible type.

The tuple is correctly generated as:
```ts
zod.tuple([zod.int(), zod.int()])
```
but the corresponding default value is inferred as an array (number[]) instead of a tuple ([number, number]).

This causes a TypeScript TS2769 error when the default value is passed to .default().

### Output client

zod

```ts

export default defineConfig({
    example: {
        input: "../docs/source/openapi/example_openapi.yml",
        output: {
            target: "./src/api/client/new/example.ts",
            mode: "single",
            client: "react-query",
            httpClient: "fetch",
            formatter: "oxfmt",
            schemas: {
                path: "./src/api/client/new/model/example",
                type: "zod",
            },
        },
    },
});

```

### Environment

System:
OS: Linux 7.0 Ubuntu 26.04.1 LTS 26.04.1 LTS (Resolute Raccoon)
CPU: (14) x64 Intel(R) Core(TM) Ultra 7 155U
Memory: 19.87 GB / 30.85 GB
Container: Yes
Shell: 5.3.9 - /bin/bash

orval@8.29.0

### Expected behavior

The generated default value should preserve the tuple type and be compatible with the generated Zod tuple schema.

For example:
```
const exampleDefault = {
    name: "example",
    position: [0, 0] as [number, number],
};
```

This should allow the generated code to pass the default value to .default() without a TypeScript error.

### Actual behavior

Orval generates the tuple schema correctly, but the corresponding default value is inferred as a regular array instead of a tuple.

For example:
```ts
position: [0, 0]
```

is inferred as:
```ts
number[]
```

instead of:
```ts
[number, number]
```

As a result, passing the generated default value to .default() causes a TS2769 TypeScript error because number[] is not assignable to [number, number].

```
TS2769: No overload matches this call.
Overload 1 of 2,
(def: { name?: string | undefined; position?: [number, number] | undefined; }): ZodDefault<ZodObject<{ name: ZodOptional<ZodString>; position: ZodOptional<ZodTuple<[ZodInt, ZodInt], null>>; }, $strip>>
, gave the following error.
Argument of type { name: "example"; position: 0[]; } is not assignable to parameter of type
{
  name?: string | undefined;
  position?: [number, number] | undefined;
}
Types of property position are incompatible.
Type 0[] is not assignable to type [number, number]
Target requires 2 element(s) but source may have fewer.
Overload 2 of 2,
(def: () => { name?: string | undefined; position?: [number, number] | undefined; }): ZodDefault<ZodObject<{ name: ZodOptional<ZodString>; position: ZodOptional<ZodTuple<[ZodInt, ZodInt], null>>; }, $strip>>
, gave the following error.
Argument of type { name: "example"; position: 0[]; } is not assignable to parameter of type
() => {
  name?: string | undefined;
  position?: [number, number] | undefined;
}
Type { name: "example"; position: 0[]; } provides no match for the signature
(): { name?: string | undefined; position?: [number, number] | undefined; }
```

```yaml
openapi: 3.1.0

info:
  title: example
  version: 1.0.0

paths:
  /example:
    get:
      summary: Get example
      operationId: getExample
      responses:
        "200":
          description: Example response
          content:
            application/json:
              schema:
                type: object
                default:
                  name: example
                  position:
                    - 0
                    - 0
                properties:
                  name:
                    type: string
                  position:
                    type: array
                    prefixItems:
                      - type: integer
                      - type: integer

```

### Additional context

_No response_

### LINKED ISSUE #4024: Zod array defaults widen nested enum values
### Description

When generating Zod schemas for an array of objects with a default value, Orval correctly generates the enum as a union of literal values. However, the values in the generated default array are widened to number.

This makes the generated default incompatible with the corresponding Zod schema.

### Output client

zod

```ts

export default defineConfig({
    example: {
        input: "../docs/source/openapi/example_openapi.yml",
        output: {
            target: "./src/api/client/new/example.ts",
            mode: "single",
            client: "react-query",
            httpClient: "fetch",
            formatter: "oxfmt",
            schemas: {
                path: "./src/api/client/new/model/example",
                type: "zod",
            },
        },
    },
});

```

### Environment

System:
OS: Linux 7.0 Ubuntu 26.04.1 LTS 26.04.1 LTS (Resolute Raccoon)
CPU: (14) x64 Intel(R) Core(TM) Ultra 7 155U
Memory: 19.87 GB / 30.85 GB
Container: Yes
Shell: 5.3.9 - /bin/bash

orval@8.29.0

### Expected behavior

The generated default value should preserve the enum literal types so it is compatible with the generated Zod schema.

For example, { value: 1 } should be inferred with value as 1, not widened to number, allowing the default value to be passed to .default() without a TypeScript error.

### Actual behavior

The generated default value widens the enum values from their literal types to number.

As a result, the generated default:
```ts
[
    { value: 1 },
    { value: 2 },
]
```
is inferred as:
```ts
{ value: number }[]
```
while the generated Zod schema expects:
```ts
{ value?: 1 | 2 | 3 | undefined }[]
```
This causes a TS2769 error when the default value is passed to .default().

```yaml
openapi: 3.1.0

info:
  title: example
  version: 1.0.0

paths:
  /example:
    get:
      summary: Get example
      operationId: getExample
      responses:
        "200":
          description: Example
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Example"

components:
  schemas:
    Example:
      type: object
      properties:
        items:
          type: array
          default:
            - value: 1
            - value: 2
          items:
            $ref: "#/components/schemas/ExampleItem"

    ExampleItem:
      type: object
      properties:
        value:
          type: integer
          enum:
            - 1
            - 2
            - 3
```

### Additional context

_No response_

ErfanBagheri404:
Fixes #4023 and #4024. Object defaults containing nested tuples or ref object arrays stay inline so TypeScript keeps tuple and literal types.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved generated default values for tuple-shaped arrays, preserving correct TypeScript tuple and literal typing.
  * Improved nested object defaults by applying schema information recursively.
  * Preserved literal types for arrays containing enums, objects, or referenced values.
  * Kept complex defaults inline when needed to maintain accurate contextual typing.

* **Tests**
  * Added regression coverage for nested tuple defaults, arrays of tuples, top-level tuple arrays, and ordinary object defaults.

### DIFF
--- a/packages/zod/src/index.ts
+++ b/packages/zod/src/index.ts
@@ -1038,12 +1038,41 @@ export const generateZodValidationSchemaDefinition = (
       // readonly, which zod v4's `.default()` rejects (#3399).
       const formatDefaultEntryValue = (
         entryValue: unknown,
+        entrySchema?: unknown,
       ): string | undefined => {
         if (isString(entryValue)) {
           return `${JSON.stringify(entryValue)} as const`;
         }
 
         if (Array.isArray(entryValue)) {
+          const entrySchemaObj =
+            entrySchema && typeof entrySchema === 'object'
+              ? (entrySchema as Record<string, unknown>)
+              : undefined;
+          const prefixItems =
+            entrySchemaObj && Array.isArray(entrySchemaObj.prefixItems)
+              ? entrySchemaObj.prefixItems
+              : undefined;
+          if (prefixItems) {
+            // Nested tuple (prefixItems): items must NOT get `as const`
+            // on the whole array (that widens to readonly, which zod v4
+            // `.default()` rejects — #3399). Emit plain literals so
+            // TypeScript contextually types them as the tuple (#4023).
+            const tupleItems = entryValue.map((item) => {
+              if (isString(item)) {
+                return `${JSON.stringify(item)} as const`;
+              }
+              if (isNumber(item) || isBoolean(item)) {
+                return `${item}`;
+              }
+              if (item === null || item === undefined) {
+                return `${item}`;
+              }
+              const formatted = formatDefaultEntryValue(item);
+              return formatted ?? 'null';
+            });
+            return `[${tupleItems.join(', ')}]`;
+          }
           const arrayItems = entryValue.map((item) => {
             if (isString(item)) {
               return `${JSON.stringify(item)} as const`;
@@ -1073,7 +1102,23 @@ export const generateZodValidationSchemaDefinition = (
         if (isObject(entryValue)) {
           const nestedEntries = Object.entries(entryValue)
             .map(([nestedKey, nestedValue]) => {
-              const formatted = formatDefaultEntryValue(nestedValue);
+              const nestedSchema =
+                entrySchema &&
+                typeof entrySchema === 'object' &&
+                entrySchema !== null &&
+                'properties' in entrySchema &&
+                typeof (entrySchema as Record<string, unknown>).properties ===
+                  'object' &&
+                (entrySchema as Record<string, unknown>).properties !== null
+                  ? (
+                      (entrySchema as Record<string, unknown>)
+                        .properties as Record<string, unknown>
+                    )[nestedKey]
+                  : undefined;
+              const formatted = formatDefaultEntryValue(
+                nestedValue,
+                nestedSchema,
+              );
               return formatted === undefined
                 ? undefined
                 : `${JSON.stringify(nestedKey)}: ${formatted}`;
@@ -1097,9 +1142,13 @@ export const generateZodValidationSchemaDefinition = (
         return undefined;
       };
 
+      const properties =
+        schema.properties && isObject(schema.properties)
+          ? (schema.properties as Record<string, unknown>)
+          : undefined;
       const entries = Object.entries(schema.default)
         .map(([key, value]) => {
-          const formatted = formatDefaultEntryValue(value);
+          const formatted = formatDefaultEntryValue(value, properties?.[key]);
           return formatted === undefined
             ? undefined
             : `${JSON.stringify(key)}: ${formatted}`;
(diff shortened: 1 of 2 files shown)


melloware:
Pushed `cbcfdab1c` to this branch addressing the second review comment. (The first, the unused `key` parameter, was already fixed in 4f72925.)

## The nested-tuple finding is valid, and broader than reported

The check only inspected a schema's **direct** properties for `prefixItems`. I probed the generator and found three shapes still hoisting:

| schema shape | before | after |
|---|---|---|
| tuple under a nested object — `{ settings: { position: [0, 0] } }` | hoisted | inline |
| array-of-tuples property — `{ points: [[0, 0]] }` | hoisted | inline |
| top-level array of tuples — `[[0, 0]]` | hoisted | inline |
| direct tuple property (already covered) | inline | unchanged |

And confirmed with `tsc` that hoisting is what actually breaks, rather than assuming it:

```
Type '{ settings: { position: number[]; }; }' is not assignable to type 'Target'.
  Type 'number[]' is not assignable to type '[number, number]'.
    Target requires 2 element(s) but source may have fewer.
```

The inline form in the same file compiles.

## Fix

The flat check is now a `needsInlineDefault` walk over `properties` and `items`, with a `seen` guard, and the array branch uses it so an `items` schema that is itself a tuple is caught.

## One thing I deliberately excluded

My first attempt also treated **arrays of `enum` items** as needing inline emission — the comment in the original code mentions enums, so it looked like an oversight. It isn't: that broke four existing tests (`narrows numeric enum array items in object defaults` and friends). Those arrays hoist *and* the hoisted const narrows each element with `as const`, so no type information is lost. Enum arrays and tuples need opposite treatment, and only tuples and `$ref` arrays can't be repaired after hoisting. Reverted that part and recorded the reasoning in the helper's doc comment.

## Also

Array elements now receive the items schema when formatted, so an array of tuples emits `[[0, 0]]` rather than `[[0 as const, 0 as const]]`.

## Testing

Three regression tests — nested-object tuple, array-of-tuples property, top-level array-of-tuples — all failing against 4f72925. Plus a guard that an ordinary object default still hoists, so the inline path stays targeted.

`vp run lint` (type-aware + type-check), `vp run format:check`, `vp run test` (341 zod tests) and `vp run test:snapshots` all pass.

## Two notes unrelated to the review

- I dropped a self-referential-schema test I had written: `generateZodValidationSchemaDefinition` overflows the stack on a cyclic *inline* schema both with and without this change, so it is a pre-existing limitation rather than anything this PR affects. The `seen` guard in the new helper stays as cheap defensive code. Worth a separate issue if you think it's reachable in practice.
- Running `test:snapshots` on this branch leaves five untracked snapshot directories (`fetch/petstore-tags-operations*`, `react-query/petstore-tags-operations*`, `zod/inline-enum-array`). They appear with or without these changes — the branch is based on v8.29.0, which predates the commits that added them on `master`. Nothing to commit here; a rebase would clear it.

melloware (approved):

PR #4037, opened 2026-09-07
fix: render mismatched const/enum values as literal types (GHSA-79f3-f442-cq2j)

melloware:
Fixes the model `const` injection reported in GHSA-79f3-f442-cq2j.

## The bug

`getScalar` chose how to render a `const` or enum member from the schema's **declared** type. But the declared type and the value both come from the document, and nothing makes them agree. The numeric and boolean branches interpolated the value into a bare type position:

```ts
value = `${schemaConst}${nullable}`;
```

There is no quote to escape there, so `type: number` with a string `const` closes the declaration and turns the rest into module statements:

```yaml
InjectedNumber:
  type: number
  const: '0; console.log("ORVAL_MODEL_POC"); type OrvalTail = 0'
```

```ts
export type InjectedNumber = 0; console.log("ORVAL_MODEL_POC"); type OrvalTail = 0;
```

The `type X = 0;` declaration closes, `console.log(...)` becomes a real statement, and the trailing `type OrvalTail = 0` absorbs the remainder so it all still compiles. Importing the model runs it.

## Three sinks, not one

The advisory names the numeric `const`. Sweeping the neighbouring branches found two more of the same bug:

| case | before | in advisory |
|---|---|---|
| numeric `const` | `= 0; console.log(...)` | yes |
| boolean `const` | `= true; console.log(...)` | **no** |
| boolean `enum` member | `= true; console.log(...)` | **no** |

Two cases turned out already safe, and I checked rather than assumed: a numeric `enum` routes through the enum generator, which sanitizes the key and quotes the value; and under `enumGenerationType: 'union'` numeric members are quoted. The boolean branch never sets `isEnum`, which is why it bypassed that protection.

## The fix

Branch on the value's **own** type via a `toLiteralTypeValue` helper — strings become escaped string-literal types, numbers and booleans stay bare, anything else is JSON-stringified. Applied to all four sites (both `const` branches, both enum-member unions).

After:

```ts
export type Injected = '0; console.log("PWN_NUM_CONST"); type T1 = 0';
```

Transpiling and running the generated module: **nothing executes** — after type erasure the emitted JS is only esbuild's CJS boilerplate, with no payload statement.

Quotes inside the payload are escaped, so the literal cannot be closed either:

```ts
export type Injected = 'x\'; console.log(\'PWN_STR\'); type T5 = 0';
```

## Two incidental improvements

- `schemaConst` was typed `as string | undefined` while holding arbitrary document data. It is now `unknown` — that inaccurate cast is what let the mismatch pass unnoticed.
- The string branch now shares the helper, which stops a non-string `const` under `type: string` from throwing inside `jsStringLiteralEscape`.

Model output is **byte-identical** before and after — `diff -r` across numeric/boolean/string consts, numeric/boolean/string enums, and a nullable const:

```
export type NumConst = 42;      export type BoolConst = true;
export type StrConst = 'hello'; export type NullableNum = 7 | null;
export type NumEnum = typeof NumEnum[keyof typeof NumEnum];

IDENTICAL — no behaviour change for valid specs
```

## Testing

Seven new tests. Six fail against the unfixed generator (all three sinks, quote escaping, and the nullable-suffix placement); the seventh guards that genuine numeric and boolean consts stay unquoted.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected generated scalar types when numeric or boolean schemas contain string constants or enum values.
  - String values are now safely escaped and emitted as string literals.
  - Falsy string constants, including empty strings, zero, and false, are preserved as literals.
  - Nullable types retain the `| null` suffix outside string literals.
  - Valid numeric and boolean constants continue to be generated without quotes.

### DIFF
--- a/packages/core/src/getters/scalar.test.ts
+++ b/packages/core/src/getters/scalar.test.ts
@@ -398,3 +398,56 @@ describe('getScalar (const/enum type confusion)', () => {
     ).toBe('true');
   });
 });
+
+describe('getScalar (string-schema const presence and nullability)', () => {
+  // `if (schemaConst)` treated these as absent and fell back to the wide type.
+  it.each<[string, unknown, string]>([
+    ['empty string', '', "''"],
+    ['zero', 0, '0'],
+    ['false', false, 'false'],
+  ])('keeps a falsy %s const', (_label, constValue, expected) => {
+    const schema = {
+      type: 'string',
+      const: constValue,
+    } as OpenApiSchemaObject;
+
+    expect(getScalar({ item: schema, name: 'f', context }).value).toBe(
+      expected,
+    );
+  });
+
+  // The const branch overwrites the value assigned above it, so the suffix has
+  // to be re-applied — the number and boolean branches already did.
+  it('keeps the nullable suffix for a type-array nullable const', () => {
+    const schema = {
+      type: ['string', 'null'],
+      const: 'x',
+    } as unknown as OpenApiSchemaObject;
+
+    expect(getScalar({ item: schema, name: 'f', context }).value).toBe(
+      "'x' | null",
+    );
+  });
+
+  it('keeps the nullable suffix for a nullable: true const', () => {
+    const schema = {
+      type: 'string',
+      const: 'x',
+      nullable: true,
+    } as OpenApiSchemaObject;
+
+    expect(getScalar({ item: schema, name: 'f', context }).value).toBe(
+      "'x' | null",
+    );
+  });
+
+  it('leaves a schema without a const alone', () => {
+    expect(
+      getScalar({
+        item: { type: 'string' } as OpenApiSchemaObject,
+        name: 'f',
+        context,
+      }).value,
+    ).toBe('string');
+  });
+});
--- a/packages/core/src/getters/scalar.test.ts
+++ b/packages/core/src/getters/scalar.test.ts
@@ -329,3 +329,72 @@ describe('getScalar integer enum + const (#3758)', () => {
     expect(result.value).toBe('1 | null');
   });
 });
+
+describe('getScalar (const/enum type confusion)', () => {
+  const payload = '0; console.log("pwned"); type Tail = 0';
+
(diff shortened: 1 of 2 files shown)


melloware:
Both parts are valid — fixed in fceb114ff. Worth recording that both are **pre-existing**, so they're corrections rather than regressions from this PR.

## Verified before changing anything

I probed the behaviour on this branch and on master; identical on both:

| schema | before | after |
|---|---|---|
| `type: string, const: ''` | `string` | `''` |
| `type: string, const: 0` | `string` | `0` |
| `type: string, const: false` | `string` | `false` |
| `type: ['string','null'], const: 'x'` | `'x'` | `'x' \| null` |
| `type: string, const: 'x', nullable: true` | `'x'` | `'x' \| null` |
| `type: ['number','null'], const: 5` | `5 \| null` | unchanged |

That last row is the one that convinced me: the **number** branch already emitted `5 | null`, so the string branch was the odd one out. The cause is that `value += nullable` runs *before* the const check, which then overwrites `value` wholesale — exactly as you described.

## Fixed

Presence check instead of truthiness, and the suffix re-applied, matching the number and boolean branches.

## Two notes for your call

**These are behaviour changes, not security fixes.** `type: string, const: ''` goes from `string` to `''`, which narrows a generated type. That is what the spec says the type is, so I think it's the right direction, but it could surface as a type error in a consumer that was assigning arbitrary strings to that field. If you'd rather keep the security fix free of behaviour changes, `fceb114ff` is a standalone commit you can drop and I'll re-open it separately.

**No snapshot drift.** `vp run test:snapshots` passes unchanged, so nothing in the sample corpus exercises a falsy or nullable string const — which is the main evidence that the blast radius here is small.

## Testing

Five new tests, all failing against the previous commit: the three falsy consts and both nullable spellings. Plus a guard that a schema without a const is untouched.

`vp run lint` (type-aware + type-check), `vp run format:check`, `vp run test` and `vp run test:snapshots` all pass.

PR #4039, opened 2026-09-07
fix: check every redirect hop against externalRefs.allow (GHSA-jmpc-3jgr-j7jv)

melloware:
Fixes the `externalRefs.allow` redirect bypass reported in GHSA-jmpc-3jgr-j7jv.

## The bug

`createSafeUrlLoader` checked the `$ref` URL against the allow-list and then handed it to the global `fetch`:

```ts
const isAllowed = isAllowedRef(value, allowedExternalRefs, origin);
if (!isAllowed) { throw ... }
return base.exec(value);   // fetch() follows redirects on its own
```

`fetch` follows redirects transparently, so an allowed URL can answer `302` and send the request somewhere the user never allowed. The allow-list only ever saw the first URL.

Reproduced with two loopback servers — A allow-listed and redirecting, B not allow-listed:

```
A /ext.json -> 302     ← allowed server issues the redirect
B /ext.json            ← BLOCKED server contacted
🎉 poc - generation succeeded
```

Generation completed *using B's content*, which is the part that makes this more than a probe: an internal endpoint's response ends up in the resolved document.

## The fix

`fetchUrls` accepts a `fetch` option, documented as "use this to intercept requests". So redirects are now followed manually, re-checking each resolved `Location` with the same predicate the initial ref uses — an allow-list entry, or the spec's own origin:

```
A /ext.json -> 302     ← allowed server still contacted
                       ← B never contacted
```

Capped at 5 hops. Under the wildcard the default handling is kept, since every destination is permitted there anyway.

## The refusal is now legible

The loader swallows errors thrown from a loader and reports only a generic message, so my first pass produced this — technically secure, but impossible to diagnose:

```
[WARN] Failed to parse JSON/YAML from URL: http://127.0.0.1:57973/ext.json
Failed to resolve external reference ... The reference may be invalid, inaccessible ...
```

A blocked redirect was indistinguishable from an unreachable server. The refusal is now logged before throwing:

```
Refused to follow a redirect to a URL that is not allowed: http://127.0.0.1:57972/ext.json
Reached by redirect from http://127.0.0.1:57973/ext.json.
Add it to externalRefs.allow or use ['*'] to allow all.
```

## Behaviour matrix

Verified end-to-end against live loopback servers:

| case | result |
|---|---|
| direct allowed URL, no redirect | succeeds |
| redirect, **both** hops allowed | succeeds |
| redirect, destination **not** allowed | **blocked** (the fix) |
| wildcard `['*']` with redirect | succeeds |
| URL not allowed at all | blocked (unchanged) |

Legitimate redirects keep working as long as the destination is allow-listed, which is the behaviour a user configuring an allow-list would expect.

## Testing

Two integration tests using real loopback servers, in the style of the existing `externalRefs` tests:

- **blocked redirect** — asserts the destination server received **zero** requests and that the refusal warning names the blocked URL. Fails against the unfixed loader.
- **allowed redirect** — asserts the destination received exactly one request and `getX` resolved. Passes both before and after, so it guards against over-blocking.

## One thing deliberately left alone

The manual follower forwards the request `init` (including any `parserOptions.headers` matched for the original host) across hops, which is what `fetch` already did implicitly. Dropping credentials on a cross-origin hop would be a further hardening, but it is a behaviour change beyond this advisory and could break a user who intends the same headers for both allow-listed hosts. Happy to do it separately if you want it — worth noting every hop must now be explicitly allow-listed, so the destination is one the user affirmatively permitted.

## Summary by CodeRabbit

- **Security**
  - External OpenAPI references now re-check every redirect destination against the configured allow-list.
  - Redirects to unapproved URLs are blocked before the destination is contacted.
  - Approved redirects continue to work, with redirect chains limited to five hops.
  - Request headers are recalculated for each redirect destination, preventing credentials from being sent to the wrong host.
- **Tests**
  - Added coverage for blocked redirects, approved redirects, request handling, header scoping, and warning messages.

### DIFF
--- a/packages/orval/src/import-specs.test.ts
+++ b/packages/orval/src/import-specs.test.ts
@@ -1,4 +1,6 @@
 import { mkdtemp, rm, writeFile } from 'node:fs/promises';
+import http from 'node:http';
+import type { AddressInfo } from 'node:net';
 import os from 'node:os';
 import path from 'node:path';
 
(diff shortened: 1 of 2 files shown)


melloware:
You're right, and I was wrong — this was a regression I introduced, not a pre-existing behaviour I preserved. Fixed in 6448a523e.

## My reasoning in the PR was incorrect

I wrote that forwarding `init` across hops "is what `fetch` already did implicitly". I asserted that without testing it. It's false — `fetch` with `redirect: 'follow'` drops `Authorization` on a cross-origin redirect, per the Fetch spec:

```
native fetch (follow)   -> target saw authorization: (absent)      | x-custom: v1
manual follower (reuse) -> target saw authorization: Bearer SECRET | x-custom: v1
```

So replacing the global `fetch` with a manual follower **created** the leak. Thanks for pushing on it rather than accepting the note I left.

## Fixed by recomputing per hop

Your suggestion is also the better shape than blanket-stripping, because `fetchUrls` matches headers by domain against the **first** URL only. Reusing them was wrong in both directions, so headers are now selected for each hop's host using the same domain matching `fetchUrls` performs.

That closes the leak and fixes the mirror-image bug in the same move: headers configured for the *redirect target* host were previously never sent, since only the source host was ever matched.

## Testing

Two tests, both failing against the parent commit:

| test | asserts |
|---|---|
| source-only credential not forwarded | target receives **no** `authorization` when it is scoped to the redirecting host |
| target-host credential delivered | target receives `Bearer TARGET` when configured for its own host |

The second failing is the evidence for the bonus fix — before this commit the target's own configured headers were dropped.

Also re-verified the original SSRF fix still holds after the change:

```
Refused to follow a redirect to a URL that is not allowed: http://127.0.0.1:62114/ext.json
```

`vp run lint` (type-aware + type-check), `vp run format:check`, `vp run test` and `vp run test:snapshots` all pass, zero snapshot drift.

PR #4040, opened 2026-09-07
fix(fetch): make getHeaders type-check outside the DOM (#4034, #4029)

### LINKED ISSUE #4029: Type 'HeaderRecord' is not assignable to type 'Record<string, string | readonly string[]>'
### Description

The generated files have a TypeScript issue on `post` requests `getHeaders` function.

The generated example code:
```ts
/**
 * Create a new order
 * @summary Create a new order
 */
export const postV2Orders = async (
    postV2OrdersBody?: PostV2OrdersBody,
    options?: Parameters<typeof customFetch>[1],
): Promise<postV2OrdersResponse> => {
    const getHeaders = (
        h?: NonNullable<RequestInit["headers"]>,
    ): Record<string, string | readonly string[]> => {
        if (!h) return {};
        if (h instanceof Headers) return Object.fromEntries(h.entries());
        if (Array.isArray(h)) return Object.fromEntries(h);
        return h;
    };
    return customFetch<postV2OrdersResponse>(getPostV2OrdersUrl(), {
        ...options,
        method: "POST",
        headers: {
            "Content-Type": "application/json",
            ...getHeaders(options?.headers),
        },
        body: JSON.stringify(postV2OrdersBody),
    });
};
```

Error on the return value of `getHeaders` function:
```
 ts: Type 'HeaderRecord' is not assignable to type 'Record<string, string | readonly string[]>'.
     'string & Record<never, never>' and 'string' index signatures are incompatible.
       Type 'string | undefined' is not assignable to type 'string | readonly string[]'.
         Type 'undefined' is not assignable to type 'string | readonly string[]'. [2322]
```

### Output client

fetch

```ts

export default defineConfig({
    snappbox: {
        input: "./openapi.yaml",
        output: {
            workspace: "./api/",
            target: "index.ts",
            namingConvention: "kebab-case",
            formatter: "biome",
            clean: true,
            client: "fetch",
            mode: "tags-split",
            headers: true,
            override: {
                mutator: {
                    path: "../custom-fetch.ts",
                    name: "customFetch",
                },
            },
        },
    },
});
```

### Environment

  System:
    OS: Linux 7.0 Ubuntu 26.04 LTS 26.04 LTS (Resolute Raccoon)
    CPU: (12) x64 Intel(R) Core(TM) i7-6800K CPU @ 3.40GHz
    Memory: 20.50 GB / 38.08 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  npmPackages:
    orval: catalog: => 8.27.0 
    zod: catalog: => 4.4.3

### Expected behavior

_No response_

### Actual behavior

_No response_

```yaml
openapi: 3.0.0
info:
  title: B2B Experience With SameDay
  version: 1.0.0
  description: API for B2B transactions
servers:
  - url: /
paths:
  /v2/orders:
    post:
      summary: Create a new order
      responses:
        '201':
          description: Created
          content:
            application/json:
              schema:
                type: object
                properties:
                  createdTime:
                    type: string
                  orderId:
                    type: string
                  price:
                    type: integer
                  refId:
                    type: string
                  status:
                    type: string
                  terminalIds:
                    type: array
                    items:
                      type: integer
              example:
                createdTime: string
                orderId: string
                price: 2684
                refId: string
                status: string
                terminalIds:
                  - 5654
                  - 6460
        '400':
          description: Bad Request
          content:
            application/json:
              schema:
                type: object
                properties:
                  key:
                    type: string
                  message:
                    type: string
                  statusCode:
                    type: string
              example:
                key: string
                message: string
                statusCode: string
        '500':
          description: Internal Server Error
          content:
            application/json:
              schema:
                type: object
                properties:
                  key:
                    type: string
                  message:
                    type: string
                  statusCode:
                    type: string
              example:
                key: string
                message: string
                statusCode: string
      tags:
        - v2
        - orders
      description: Create a new order
      requestBody:
        content:
          application/json:
            schema:
              type: object
            example: |-
              {
                "city": "tehran",
                "deliveryCategory": "same-day",
                "startTime": null,
                "endTime": null,
                "features": {
                  "expressable": false         
                },
                "hasReturn": false,
                 "packages": [
                  {
                    "dropoffReference": "2",
                    "insuranceId": 158,           /* Get Some Id from "/v1/insurances" */
                    "items": [
                      {
                        "name": "تست بسته 1 لباس مجلسی",
                        "packageValue": 48000,
                        "quantity": 1,
                        "quantityMeasuringUnit": "عدد",
                        "volume": 20,
                        "weight": 10
                      },
                      {
                        "name": "کفش فلان",
                        "packageValue": 477000,
                        "quantity": 1,
                        "quantityMeasuringUnit": "عدد",
                        "volume": 20,
                        "weight": 10
                      }
                    ],
                    "packageValue": 148000,
                    "pickupReference": "1",
                    "size": {
                      "type": "category",
                      "value": "6"              /* Get Some id from "/v1/nxp/sizes/same-day" */
                    }
                  }
                ],
                "paymentType": "prepaid",
                "podEnabled": false,
                "popEnabled": false,
                "refId": "box-545659",
                "sequenceNumberDeliveryCollection": 1,
                "terminals": [
                  {
                    "address": "ونک، بزرگراه شهید حقانی",
                    "comment": "Pick up fragile items.",
                    "contactName": "رضا غلامی",
                    "doorNumber": 5845,
                    "floor": 10,
                    "id": 3384265,          /* Get Some Id from "/v1/customers/favorite-address" */
                    "latitude": "35.757523",
                    "longitude": "51.409911",
                    "phoneNumber": "09121112233",
                    "plate": 378,
                    "postalCode": "1234",
                    "reference": "1",
                    "type": "pickup",
                    "vendorId": ""
                  },
                  {
                    "address": "ونک، بزرگراه شهید حقانی",
                    "comment": "DropOff fragile items.",
                    "contactName": "رضا غلامی",
                    "doorNumber": 1982,
                    "floor": 2,
                    "latitude": "35.757523",
                    "longitude": "51.408911",
                    "phoneNumber": "09121112233",
                    "plate": 9326,
                    "postalCode": "1234",
                    "reference": "2",
                    "type": "drop"
                  }
                ],
                "timeSlot": {
                  
                  "id": "91196071"       /* Get Some Id from "/v1/nxp/sdd/timelines" */
                 
                },
                "totalPackageSize": 3.5,
                "voucherCode": "",
                "waitingTime": 0
              }

```

### Additional context

_No response_

### LINKED ISSUE #4034: fetch client's getHeaders does not type-check under @cloudflare/workers-types (8.25.0+, not covered by #3890)
### Description

Since **8.25.0**, the `fetch` client's generated `getHeaders` helper fails to type-check under `@cloudflare/workers-types`. Still present in **8.29.0**.

This is a sibling of #3886 that PR #3890 did not cover. #3890 changed the helper's *type reference* (`HeadersInit` → `NonNullable<RequestInit['headers']>`) and *widened the return type*, but left the narrowing chain untouched:

```ts
const getHeaders = (h?: NonNullable<RequestInit['headers']>): Record<string, string | readonly string[]> => {
  if (!h) return {};
  if (h instanceof Headers) return Object.fromEntries(h.entries());
  if (Array.isArray(h)) return Object.fromEntries(h);
  return h;   // ← assumes whatever remains is a plain record
};
```

That final `return h` is only sound for the DOM's `HeadersInit`, whose non-`Headers`, non-record member is `[string, string][]` — an array, so `Array.isArray` eliminates it and leaves exactly `Record<string, string>`.

Cloudflare Workers declares it as a **non-array iterable**:

```ts
// @cloudflare/workers-types
type HeadersInit = Headers | Iterable<Iterable<string>> | Record<string, string>;
```

`Array.isArray` cannot exclude `Iterable<Iterable<string>>`, so it survives into `return h` and violates the helper's own declared return type:

```
src/generated/client.ts(41,5): error TS2322: Type 'Iterable<Iterable<string>> | Record<string, string>' is not assignable to type 'Record<string, string | readonly string[]>'.
  Type 'Iterable<Iterable<string>>' is not assignable to type 'Record<string, string | readonly string[]>'.
    Index signature for type 'string' is missing in type 'Iterable<Iterable<string>>'.
```

One error is emitted per operation that adds a header, so a real spec produces dozens. It is not suppressible — `GET_HEADERS_HELPER` is inserted unconditionally whenever `fetchHeadersOption` is non-empty.

**Affected versions** (verified with the minimal repro below, one error at `client.ts(41,5)` each):

| version | result |
| -- | -- |
| 8.24.0 | ✅ passes (emits no `getHeaders`) |
| 8.25.0 | ❌ fails |
| 8.26.0 | ❌ fails (first release containing #3890) |
| 8.28.1 | ❌ fails |
| 8.29.0 | ❌ fails |

It reproduces with the **default** fetch client — no custom mutator required — and is masked whenever `lib` includes `DOM`, which is why most consumers don't hit it.

This may be the same underlying defect as the currently-open #4029 (Hono's `HeaderRecord`, also failing on `return h`): each non-DOM environment declares a differently-shaped `RequestInit['headers']` union, and the three-branch narrowing only fits the DOM one.

### Suggested fix

Drop the narrowing entirely. The `Headers` constructor accepts every member of the union in every runtime, so no type guards are needed:

```ts
const getHeaders = (h?: NonNullable<RequestInit['headers']>): Record<string, string | readonly string[]> =>
  h ? Object.fromEntries(new Headers(h).entries()) : {};
```

I verified this compiles cleanly both under `@cloudflare/workers-types` with `lib: ["esnext"]` **and** under `lib: ["ES2022", "DOM", "DOM.Iterable"]`. It would also be immune to future divergence in any other non-DOM `HeadersInit` declaration. (I have not verified it against the Hono case in #4029.)

The one behavioural difference is that `Headers` lowercases header names — harmless for HTTP semantics, but it does normalise the keys that get spread into the request.

### Output client

fetch

```ts

export default defineConfig({
  demo: {
    input: { target: "./petstore.yaml" },
    output: {
      target: "./src/generated/client.ts",
      client: "fetch",
      clean: true,
    },
  },
});
```

### Environment

```
  System:
    OS: macOS 26.5.2
    CPU: (8) arm64 Apple M3
    Memory: 140.94 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    @cloudflare/workers-types: 4.20260601.1 => 4.20260601.1
    orval: 8.29.0 => 8.29.0
    typescript: 5.9.3 => 5.9.3
```

### Expected behavior

Generated output type-checks under `@cloudflare/workers-types`, as it did up to 8.24.0.

### Actual behavior

`tsc` fails with TS2322 on the `return h` line of every generated `getHeaders` helper.

```yaml
openapi: 3.0.3
info:
  title: Demo
  version: 0.0.0
paths:
  /pets:
    post:
      operationId: createPet
      requestBody:
        required: true
        content:
          application/json:
            schema:
              type: object
              properties:
                name:
                  type: string
      responses:
        '201':
          description: Created
          content:
            application/json:
              schema:
                type: object
                properties:
                  id:
                    type: string
```

### Additional context

Complete repro — four files, then `pnpm install && npx orval && npx tsc`:

`package.json`

```json
{
  "name": "orval-workers-repro",
  "private": true,
  "type": "module",
  "devDependencies": {
    "@cloudflare/workers-types": "4.20260601.1",
    "orval": "8.29.0",
    "typescript": "5.9.3"
  }
}
```

`tsconfig.json`

```json
{
  "compilerOptions": {
    "target": "ES2022",
    "module": "ES2022",
    "lib": ["esnext"],
    "moduleResolution": "bundler",
    "types": ["@cloudflare/workers-types"],
    "strict": true,
    "noEmit": true,
    "skipLibCheck": true
  },
  "include": ["src/**/*"]
}
```

...plus the `orval.config.ts` and `petstore.yaml` above.

The `lib` / `types` combination is what a Cloudflare Workers project gets by default — `wrangler types` generates a `worker-configuration.d.ts` carrying that same `HeadersInit` declaration, so this affects Workers projects whether they depend on `@cloudflare/workers-types` directly or not.

melloware:
Fixes #4034.

## The bug

`RequestInit['headers']` is declared per runtime, and the helper's narrowing chain only fits the DOM's declaration:

```ts
if (Array.isArray(h)) return Object.fromEntries(h);
return h;   // assumes whatever remains is a plain record
```

The DOM's non-record member is `[string, string][]`, so `Array.isArray` eliminates it. `@cloudflare/workers-types` declares it as `Iterable<Iterable<string>>`, which `Array.isArray` cannot exclude, so it reaches `return h` and breaks the helper's own return type.

Reproduced with the issue's repro against `master`, matching the reported line and error exactly:

```
src/client.ts(41,5): error TS2322: Type 'Iterable<Iterable<string>> | Record<string, string>'
  is not assignable to type 'Record<string, string | readonly string[]>'.
    Index signature for type 'string' is missing in type 'Iterable<Iterable<string>>'.
```

## The fix

Narrow on iterability instead. Every member of every declaration except the record is iterable, so this covers both and any future non-DOM variant:

```ts
if (Symbol.iterator in h) {
  return Object.fromEntries(h as Iterable<readonly [string, string]>);
}
return h;
```

| generated client type-checked under | before | after |
|---|---|---|
| `@cloudflare/workers-types`, `lib: ["esnext"]` | ❌ TS2322 | ✅ |
| `lib: ["ES2022","DOM","DOM.Iterable"]` | ✅ | ✅ |

## Also fixes a silent runtime bug

The same line was wrong at runtime, not just in the type system. A non-array iterable reached `return h`, and spreading an iterable as an object yields nothing — so **every header was dropped**:

```
Map(non-array iterable)   before={}   after={"X-Custom":"v"}
```

Every other input shape is byte-identical to before:

| input | before | after |
|---|---|---|
| `undefined` | `{}` | `{}` |
| record | `{"X-Custom":"v",...}` | same |
| `Headers` instance | `{"x-custom":"v"}` | same |
| entry array | `{"X-Custom":"v",...}` | same |
| duplicate keys | `{"a":"2"}` | same |

## Why not the suggested `new Headers(h)` version

The issue proposes dropping the narrowing entirely:

```ts
h ? Object.fromEntries(new Headers(h).entries()) : {}
```

That does type-check everywhere — I verified it. But `Headers` lowercases names, and the generated call site relies on the spread to override:

```ts
headers: { 'Content-Type': 'application/json', ...getHeaders(options?.headers) }
```

Lowercasing only the second half makes the keys diverge, so both survive into the object and `fetch` **combines** them instead of overriding:

```
current    sent: {"content-type":"text/plain"}
new Headers sent: {"content-type":"application/json, text/plain"}
```

Overriding `Content-Type` via `options.headers` is the common case, so that would trade a type error (which DOM users never see) for a functional regression (which everyone would). It also changes duplicate handling from last-wins to comma-joined, and throws `TypeError` on an invalid header name where the current helper passes it through.

The narrowing change leaves the record branch untouched, so casing, duplicate handling and override semantics are all preserved. The issue's diagnosis was exactly right; only the remedy differs.

## On #4029

The issue wondered whether #4029 shares this defect. It's a **different** unsoundness on the same line: there the *record* member's values include `undefined`, so `return h` fails for its own reason. I simulated that shape and confirmed this change does not fix it:

```
Type 'Record<string, string | string[] | undefined>' is not assignable to ...
  Type 'undefined' is not assignable to type 'string | readonly string[]'.
```

So #4029 stays open, and needs a decision about whether to drop or widen `undefined`-valued headers — a behavioural question rather than a narrowing one.

## Testing

Three tests in `@orval/fetch`: the narrowing swap (fails against `master`), the unchanged `!h` / `Headers` / record branches, and that the helper is still omitted when no headers are added.

Snapshots and samples regenerated — 213 occurrences of this one helper, and `git diff` on `tests/`/`samples/` contains nothing else.

`vp run lint` (type-aware + type-check), `vp run format:check`, `vp run test`, `vp run test:snapshots` and `vp run lint:samples` all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved header handling for iterable collections, including custom nested iterables and `Headers` values.
  * Header records are now normalized consistently, omitting entries with `undefined` values across generated clients and sample applications.
  * Empty header inputs are handled safely.

* **Tests**
  * Added coverage for iterable and nested iterable headers, empty inputs, `Headers` conversion, filtered records, and helpers omitted when headers are not configured.

### DIFF
--- a/packages/fetch/src/index.test.ts
+++ b/packages/fetch/src/index.test.ts
@@ -696,10 +696,26 @@ describe('generateRequestFunction — getHeaders helper (#4034)', () => {
     expect(implementation).toContain(
       'if (h instanceof Headers) return Object.fromEntries(h.entries());',
     );
-    // The record branch is unchanged, so header casing — and therefore the
-    // override semantics of `{ ...literal, ...getHeaders(options?.headers) }`
-    // — is preserved.
-    expect(implementation).toContain('return h;');
+  });
+
+  it('builds the record branch entry by entry, skipping undefined values', () => {
+    const implementation = generateImplementation(
+      verbOptionsWithHeaders(),
+      makeOptions(makeContext()),
+    );
+
+    // A record whose values include `undefined` — Hono's header record, for
+    // one — is not assignable to the declared return type, so the result is
+    // built rather than passed through (#4029). Names are copied verbatim, so
+    // the override semantics of `{ ...literal, ...getHeaders(...) }` still
+    // hold, and an `undefined` value is skipped instead of reaching the wire
+    // as the literal text "undefined".
+    expect(implementation).toContain(
+      'const headers: Record<string, string | readonly string[]> = {};',
+    );
+    expect(implementation).toContain('if (value !== undefined)');
+    expect(implementation).toContain('return headers;');
+    expect(implementation).not.toContain('    return h;\n');
   });
 
   it('does not emit the helper when no headers are added', () => {
--- a/packages/fetch/src/index.test.ts
+++ b/packages/fetch/src/index.test.ts
@@ -686,6 +686,24 @@ describe('generateRequestFunction — getHeaders helper (#4034)', () => {
     expect(implementation).not.toContain('if (Array.isArray(h))');
   });
 
+  it('materializes each entry before Object.fromEntries', () => {
+    const implementation = generateImplementation(
+      verbOptionsWithHeaders(),
+      makeOptions(makeContext()),
+    );
+
+    // `Object.fromEntries` reads `[0]`/`[1]` off each entry rather than
+    // iterating it, so a non-indexable entry — `new Set([new Set(['X-Trace',
+    // '1'])])` satisfies the declared `Iterable<Iterable<string>>` — produced
+    // a single `undefined` key and lost the real header.
+    expect(implementation).toContain(
+      'Array.from(h as Iterable<Iterable<string>>, (entry) => Array.from(entry) as [string, string])',
+    );
+    expect(implementation).not.toContain(
+      'Object.fromEntries(h as Iterable<readonly [string, string]>)',
+    );
+  });
+
   it('still short-circuits empty input and unwraps a Headers instance', () => {
     const implementation = generateImplementation(
       verbOptionsWithHeaders(),
--- a/packages/fetch/src/index.test.ts
+++ b/packages/fetch/src/index.test.ts
@@ -653,3 +653,61 @@ describe('generateRequestFunction — Content-Type header escaping', () => {
     expect(implementation).not.toContain("'X-Evil': 'injected'");
   });
 });
+
+describe('generateRequestFunction — getHeaders helper (#4034)', () => {
+  const verbOptionsWithHeaders = () =>
+    makeVerbOptions({
+      verb: Verbs.POST,
+      body: {
+        definition: 'CreatePetBody',
+        implementation: 'createPetBody: CreatePetBody',
+        imports: [],
+        schemas: [],
+        formData: undefined,
+        formUrlEncoded: undefined,
+        contentType: 'application/json',
+        isOptional: false,
+        originalSchema: {},
+        isBlob: false,
+      } as GeneratorVerbOptions['body'],
+    });
+
+  it('narrows on iterability rather than array-ness', () => {
+    const implementation = generateImplementation(
+      verbOptionsWithHeaders(),
+      makeOptions(makeContext()),
+    );
+
+    // `RequestInit['headers']` is declared per runtime. Outside the DOM its
+    // non-record member need not be an array — `@cloudflare/workers-types`
+    // uses `Iterable<Iterable<string>>` — so `Array.isArray` cannot narrow it
+    // and the fall-through `return h` failed the declared return type.
+    expect(implementation).toContain('if (Symbol.iterator in h)');
+    expect(implementation).not.toContain('if (Array.isArray(h))');
+  });
+
+  it('still short-circuits empty input and unwraps a Headers instance', () => {
+    const implementation = generateImplementation(
+      verbOptionsWithHeaders(),
+      makeOptions(makeContext()),
+    );
+
+    expect(implementation).toContain('if (!h) return {};');
+    expect(implementation).toContain(
+      'if (h instanceof Headers) return Object.fromEntries(h.entries());',
+    );
+    // The record branch is unchanged, so header casing — and therefore the
+    // override semantics of `{ ...literal, ...getHeaders(options?.headers) }`
+    // — is preserved.
+    expect(implementation).toContain('return h;');
+  });
+
+  it('does not emit the helper when no headers are added', () => {
+    const implementation = generateImplementation(
+      makeVerbOptions(),
+      makeOptions(makeContext()),
+    );
+
+    expect(implementation).not.toContain('const getHeaders');
+  });
+});
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -126,7 +126,10 @@ export const generateRequestFunction = (
   //    `Array.isArray` cannot exclude (#4034). Narrowing on iterability covers
   //    every declaration, since the record is the only non-iterable member.
   //    That also fixes the runtime half: spreading an iterable as an object
-  //    silently dropped every header.
+  //    silently dropped every header. Each entry is materialized because
+  //    `Object.fromEntries` reads `[0]`/`[1]` rather than iterating, so a
+  //    non-indexable entry (`new Set([new Set(['X-Trace', '1'])])`) would
+  //    otherwise yield a single `undefined` key and lose the real header.
   //  - The record member's values are not always `string`. Hono's header
   //    record admits `undefined` (#4029), which the declared return type does
   //    not. Building the result entry by entry keeps that guarantee by
@@ -137,7 +140,9 @@ export const generateRequestFunction = (
     if (!h) return {};
     if (h instanceof Headers) return Object.fromEntries(h.entries());
     if (Symbol.iterator in h) {
-      return Object.fromEntries(h as Iterable<readonly [string, string]>);
+      return Object.fromEntries(
+        Array.from(h as Iterable<Iterable<string>>, (entry) => Array.from(entry) as [string, string]),
+      );
     }
     const headers: Record<string, string | readonly string[]> = {};
     for (const [name, value] of Object.entries<string | readonly string[] | undefined>(h)) {
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -117,19 +117,33 @@ export const generateRequestFunction = (
   const isFormData = !override.formData.disabled;
   const isFormUrlEncoded = override.formUrlEncoded !== false;
 
-  // Narrow on iterability rather than array-ness. `RequestInit['headers']` is
-  // declared per-runtime, and outside the DOM its non-record member need not be
-  // an array — `@cloudflare/workers-types` uses `Iterable<Iterable<string>>`,
-  // which `Array.isArray` cannot exclude. It then reached `return h`, failing
-  // the helper's own return type (TS2322) and, at runtime, spreading an
-  // iterable as an object silently dropped every header (#4034).
+  // `RequestInit['headers']` is declared per runtime, and the old narrowing
+  // chain only fitted the DOM's declaration, leaving `return h` unsound in two
+  // separate ways:
+  //
+  //  - Outside the DOM the non-record member need not be an array —
+  //    `@cloudflare/workers-types` uses `Iterable<Iterable<string>>`, which
+  //    `Array.isArray` cannot exclude (#4034). Narrowing on iterability covers
+  //    every declaration, since the record is the only non-iterable member.
+  //    That also fixes the runtime half: spreading an iterable as an object
+  //    silently dropped every header.
+  //  - The record member's values are not always `string`. Hono's header
+  //    record admits `undefined` (#4029), which the declared return type does
+  //    not. Building the result entry by entry keeps that guarantee by
+  //    construction, and skipping the `undefined` values is what the caller
+  //    wants anyway: passed through, they reach the wire as the literal text
+  //    "undefined".
   const GET_HEADERS_HELPER = `  const getHeaders = (h?: NonNullable<RequestInit['headers']>): Record<string, string | readonly string[]> => {
     if (!h) return {};
     if (h instanceof Headers) return Object.fromEntries(h.entries());
     if (Symbol.iterator in h) {
       return Object.fromEntries(h as Iterable<readonly [string, string]>);
     }
-    return h;
+    const headers: Record<string, string | readonly string[]> = {};
+    for (const [name, value] of Object.entries<string | readonly string[] | undefined>(h)) {
+      if (value !== undefined) headers[name] = value;
+    }
+    return headers;
   };
 `;
 
--- a/packages/fetch/src/index.ts
+++ b/packages/fetch/src/index.ts
@@ -117,10 +117,18 @@ export const generateRequestFunction = (
   const isFormData = !override.formData.disabled;
   const isFormUrlEncoded = override.formUrlEncoded !== false;
 
+  // Narrow on iterability rather than array-ness. `RequestInit['headers']` is
+  // declared per-runtime, and outside the DOM its non-record member need not be
+  // an array — `@cloudflare/workers-types` uses `Iterable<Iterable<string>>`,
+  // which `Array.isArray` cannot exclude. It then reached `return h`, failing
+  // the helper's own return type (TS2322) and, at runtime, spreading an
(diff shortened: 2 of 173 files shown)


melloware:
Added #4029 to this PR in 439579339, since it is the same helper and the same `return h` line — a parallel PR would just conflict. Retitling to cover both.

## #4029 is a different unsoundness on that line

I reproduced it verbatim, including the `string & Record<never, never>` detail from the report, by simulating Hono's header record:

```
Type 'HeaderRecord' is not assignable to type 'Record<string, string | readonly string[]>'.
  'string & Record<never, never>' and 'string' index signatures are incompatible.
    Type 'undefined' is not assignable to type 'string | readonly string[]'.
```

So the two issues are:

| | #4034 | #4029 |
|---|---|---|
| broken member | the **non-record** one isn't an array | the **record**'s values include `undefined` |
| fix | narrow on iterability | build the result by construction |

The #4034 narrowing alone does not fix #4029 — I checked before writing this, which is why the earlier PR description said it stayed open.

## Fix

Copy the entries into a fresh record rather than returning the input, so the declared return type holds by construction regardless of what shape the runtime declares:

```ts
const headers: Record<string, string | readonly string[]> = {};
for (const [name, value] of Object.entries<string | readonly string[] | undefined>(h)) {
  if (value !== undefined) headers[name] = value;
}
return headers;
```

Header names are copied verbatim, so the `{ ...literalHeaders, ...getHeaders(options?.headers) }` override semantics are untouched — the same reason I avoided the `new Headers(h)` approach for #4034.

## Skipping `undefined` is also a real bug fix

Passed through, an `undefined` header value reaches the wire as the literal text `"undefined"`. Measured at the actual generated call site:

| user headers | before | after |
|---|---|---|
| `{'X-Ok':'v','X-Missing':undefined}` | `{"content-type":"application/json","x-missing":"undefined","x-ok":"v"}` | `{"content-type":"application/json","x-ok":"v"}` |
| `Map([['X-Custom','v']])` | `{"content-type":"application/json"}` — header lost | `{"content-type":"application/json","x-custom":"v"}` |
| `{'Content-Type':'text/plain'}` | `{"content-type":"text/plain"}` | unchanged |

The `Map` row is the #4034 runtime half. Worth noting I initially measured it wrong — passing the Map straight to `Headers` rather than spreading it as the generated code does, which made it look fine. Re-tested against the real call-site shape and the header is genuinely dropped.

## Verification

The emitted helper type-checks under all three declarations:

| `RequestInit['headers']` declared by | result |
|---|---|
| hono-shaped record with `undefined` values (#4029) | ✅ |
| `@cloudflare/workers-types`, `lib: ["esnext"]` (#4034) | ✅ |
| DOM, `lib: ["ES2022","DOM","DOM.Iterable"]` | ✅ |

One more test (failing against the previous commit), snapshots regenerated — `git diff` on `tests/`/`samples/` is again only this helper, 213 occurrences and nothing else.

`vp run lint` (type-aware + type-check), `vp run format:check`, `vp run test`, `vp run test:snapshots` and `vp run lint:samples` all pass.

melloware:
Valid — fixed in 630868d6e. This one is a bug I introduced by widening the branch without checking what `Object.fromEntries` actually does with the newly-admitted shapes.

## Confirmed, and worse than reported

`Object.fromEntries` reads `[0]`/`[1]` off each entry rather than iterating it, so an entry that is itself a non-indexable iterable doesn't convert:

```
current: keys=["undefined"] values=["<undefined>"]
fixed  : keys=["X-Trace"]   values=["1"]
```

At the generated call site that means the real header is **lost** and a bogus one is sent:

```
before  sent: {"content-type":"application/json","undefined":"undefined"}
after   sent: {"content-type":"application/json","x-trace":"1"}
```

Your `new Set([new Set([...]))` example was exactly right. A generator yielding `Set`s fails the same way. Note my first measurement showed `{}` rather than `{undefined: undefined}` — `JSON.stringify` drops undefined values, so I had to read `Object.keys` to see it. Your repro script was the more honest one.

## Fix

```ts
Array.from(h as Iterable<Iterable<string>>, (entry) => Array.from(entry) as [string, string])
```

## Scope check

Which shapes were affected, measured at the real call site:

| user headers | before | after |
|---|---|---|
| record | unchanged | unchanged |
| `Headers` instance | unchanged | unchanged |
| entry array | unchanged | unchanged |
| duplicate keys | unchanged | unchanged |
| override `Content-Type` | unchanged | unchanged |
| `Map` | header lost | **fixed** |
| `Set` of `Set`s | header lost + bogus `undefined` | **fixed** |
| generator of `Set`s | header lost | **fixed** |
| record with `undefined` value | sent as `"undefined"` | **fixed** |

Array-like entries — arrays and `Map`s, which iterate to arrays — were already converting correctly and are byte-identical.

## Verification

Emitted helper still type-checks under all three declarations:

| `RequestInit['headers']` declared by | result |
|---|---|
| hono-shaped record with `undefined` values (#4029) | ✅ |
| `@cloudflare/workers-types`, `lib: ["esnext"]` (#4034) | ✅ |
| DOM, `lib: ["ES2022","DOM","DOM.Iterable"]` | ✅ |

One more test, failing against the previous commit. All three samples you listed were regenerated by `test:snapshots:update` — `git diff` on `tests/`/`samples/` is again only this helper, 213 occurrences, nothing else.

`vp run lint` (type-aware + type-check), `vp run format:check`, `vp run test`, `vp run test:snapshots` and `vp run lint:samples` all pass.

PR #4043, opened 2026-09-07
docs: correct allParamsOptional, warn on an undetectable query version (#2396, #2397)

### LINKED ISSUE #2396: React-query + fetch: queryKey should be typed as optional
### Situation

For React-query + fetch generated helpers, let's pass some query options to a generated `useCatsList`:
```ts
const query = useCatsList({
    query: {
      staleTime: 1000 * 60,
    },
  })
```

### Actual behavior

Typescript will complain that:
> Property 'queryKey' is missing in type '{ staleTime: number; }' but required in type ...

### Expected behavior

No Typescript error.

### Notes

It's a false-positive type error and the prop should be optional instead. Generated helpers have this line of code:
```ts
  const queryKey = queryOptions?.queryKey ?? getCatRetrieveQueryKey(id)
```

### LINKED ISSUE #2397: React-query + fetch: query arguments should be typed as optional
### Situation

For React-query + fetch generated helpers, let's pass an optional query argument to a generated `useCatRetrieve`:
```ts
export const useCat = (catId?: string) => {
  return useCatRetrieve(catId)
}
```

### Actual behavior

Typescript will complain that
> Argument of type 'string | undefined' is not assignable to parameter of type 'string'.

### Expected behavior

No Typescript error.

### Notes

It's a false-positive type error and the argument should be optional instead. Generated helpers handle absence of arguments by setting `enabled: !!id` in the query options.

melloware:
Follow-up from triaging #2396 and #2397. Both reports reproduce, and both are already solved by existing options — but the options were hard to find, for two reasons this PR fixes.

Fix #2396 
Fix #2397

## The config matrix

I reproduced both reports and tested each option against them:

| config | #2396 `queryKey` required | #2397 param not optional |
|---|---|---|
| baseline | FAILS | FAILS |
| `override.query.version: 5` | **OK** | FAILS |
| `allParamsOptional: true` | FAILS | **OK** |
| both | **OK** | **OK** |

## 1. `allParamsOptional` was documented as not doing what it does

The old line — *"Make all parameters optional except path parameters"* — reads as "path parameters are unaffected". They are affected:

```ts
// allParamsOptional: false
export const getPet = async (petId: string, params: GetPetParams, ...)

// allParamsOptional: true
export const getPet = async (petId: string | undefined | null, params?: GetPetParams, ...)
```

Path parameters keep their position but widen to accept a missing value, and the parameters object becomes optional. That is exactly what #2397 asks for, so the docs were steering people away from the answer. Now documented as measured, with a note about the TanStack use case and a link to `useSkipToken`.

## 2. An undetectable query version is silently treated as v4

`isQueryV5` answers `false` for a dependency it cannot resolve:

```ts
const version = getPackageByQueryClient(packageJson, queryClient);
if (!version) {
  return false;
}
```

That is the right guess for a v4 project. On a v5 one it means the `query` option type is not wrapped in `Partial`, so the caller cannot omit `queryKey` — #2396 exactly, and it explains the reporter's own puzzlement that orval's samples wrap in `Partial` while their project does not: samples have a discoverable dependency.

Now warned once per client:

```
Could not determine the installed @tanstack/react-query version, so hooks are generated for v4.
If the project is on v5, set `override.query.version: 5` (or point `output.packageJson` at the
package.json that declares the dependency) — otherwise the generated option types will not accept
a partial `query` object.
```

Silent as soon as the dependency resolves or a version is configured. Angular Query is skipped, being v5-only.

> [!NOTE]
> This fires **6 times** in this repo's own `test:snapshots` run — `petstore` ×3, `issue3326`, `mswMixedContentUnion`, `issue-1879-header-pathref`. Cause: `findUp` resolves the repo root `package.json` for those configs, and it does not declare `@tanstack/react-query`, so detection genuinely cannot tell. A normal app declares the dependency and stays quiet, so I expect this to be rare in the wild — but it is noise in our own output, and those fixtures are in fact generating v4-shaped hooks today. Say the word and I will gate it behind `--verbose`, or set `override.query.version` in those configs instead.

## Testing

Five tests covering: warns when nothing resolves, quiet when the dependency resolves, quiet when the version is explicit, quiet for `angular-query`, and reported once per client rather than once per operation.

`vp run lint` (type-aware + type-check), `vp run format:check`, `vp run test` and `vp run test:snapshots` all pass, with no change to generated output.

## Summary by CodeRabbit

- **Documentation**
  - Expanded guidance for the `allParamsOptional` output option, including parameter behavior, examples, and usage with TanStack Query clients.

- **Improvements**
  - Added a warning when the TanStack Query version cannot be detected and no version is configured.
  - Warnings appear only once per query client and are suppressed when the version is known, explicitly configured, or using Angular Query.
  - Improved support for custom query clients and package-based version detection.

### DIFF
--- a/docs/content/docs/reference/configuration/output.mdx
+++ b/docs/content/docs/reference/configuration/output.mdx
@@ -3133,7 +3133,21 @@ Takes effect only when used `mode` is `split` or `single-split`.
 **Type:** `Boolean`
 **Default:** `false`
 
-Make all parameters optional except path parameters.
+Let callers leave parameters unresolved.
+
+Path parameters keep their position in the signature but widen to accept a missing value, and the query/header parameters object becomes optional:
+
+```ts
+// allParamsOptional: false
+export const getPet = async (petId: string, params: GetPetParams, ...)
+
+// allParamsOptional: true
+export const getPet = async (petId: string | undefined | null, params?: GetPetParams, ...)
+```
+
+Properties inside the parameters object keep whatever the document says — a `required: true` query parameter stays required within `GetPetParams`.
+
+This is useful with the TanStack Query clients, where a hook is often called before its id is known and the generated `enabled` guard (or [`useSkipToken`](#useskiptoken)) already holds the request back until then.
 
 ### urlEncodeParameters
 
--- a/packages/query/src/dependencies.ts
+++ b/packages/query/src/dependencies.ts
@@ -629,7 +629,7 @@ export const isSolidQueryWithRenamedOptionsTypes = (
   return compareVersions(withoutRc, '5.100.6');
 };
 
-const getPackageByQueryClient = (
+export const getPackageByQueryClient = (
   packageJson: PackageJson | undefined,
   queryClient:
     | 'react-query'
--- a/packages/query/src/frameworks/index.ts
+++ b/packages/query/src/frameworks/index.ts
@@ -4,6 +4,7 @@ import {
   type GetterProps,
   GetterPropType,
   isObject,
+  logWarning,
   type OutputClient,
   OutputClient as OutputClientConst,
   type OutputClientFunc,
@@ -18,6 +19,7 @@ import {
   getQueryArgumentsRequestType,
 } from '../client';
 import {
+  getPackageByQueryClient,
   isQueryV5,
   isQueryV5WithDataTagError,
   isQueryV5WithInfiniteQueryOptionsError,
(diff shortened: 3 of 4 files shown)


melloware:
Valid, and reachable — fixed in e9ef4c5f8.

## Confirmed by calling it with a function

```
Could not determine the installed @tanstack/() => ({ client: () => ({ implementation: "", imports: [] }) }) version, so hooks are generated for v4.
```

And it is reachable in production, not just in theory: `generateClient` passes `outputClient` straight through to the builder —

```ts
const client = await generatorClient(verbOption, options, outputClient, output);
```

— so a custom `OutputClientFunc` that returns the react-query generator arrives here as the user's own function. My warning was the only thing that turned that cast into user-visible output.

## Fix

The client is resolved through a guard first, and the warning is skipped when it isn't one of the five known adapters. There's no package to name for a custom client, so there's no advice to give.

## On the second half — version detection

Leaving that as it was, deliberately. `getPackageByQueryClient` matches nothing for a function, so the checks already return `false` and fall back to v4. Guarding them would change no outcome, because v4 is the only answer available for a client whose TanStack package cannot be known — there's nothing to detect and no better default to pick. So the guard would be ceremony rather than a fix, and it would mean widening `isQueryV5` and its four siblings to accept `undefined` for no behavioural gain.

If you'd rather have the explicitness anyway, say so and I'll do it — but I didn't want to widen five signatures to no effect.

## Testing

One more test asserting silence for a custom `OutputClientFunc`, failing against the previous commit with `expected 'Could not determine the installed @ta…' to be ''`.

`vp run lint` (type-aware + type-check), `vp run format:check`, `vp run test` and `vp run test:snapshots` all pass. The sample-run warning count is unchanged at 6 — all of those are string clients, so the guard correctly leaves them alone.

PR #4046, opened 2026-09-07
fix: disambiguate mixed-type enum keys with type prefixes

### LINKED ISSUE #4028: Generated enum type key conflict
### Description

I have a property in my openapi spec that defines some accepted values as string + native type (int and boolean in my case):
```json
{
  "properties": {
    "privacy": {
      "enum": [
        "yes",
        "on",
        "1",
        1,
        "true",
        true
      ]
    }
  }
}
```

and the generated enum const uses the same key for both:
```ts
export const ApiRegisterRequestPrivacy = {
  yes: 'yes',
  on: 'on',
  NUMBER_1: '1',
  NUMBER_1: 1,
  true: 'true',
  true: true,
} as const;
```

I think the simplest solution is to add a prefix/suffix to generate keys based on type (maybe only for non-string values)

### Output client

fetch

```ts

export default defineConfig({
  example: {
    input: process.env.OPENAPI_SOURCE,
    output: {
      mode: "single",
      target: "./src/schema.ts",
      httpClient: "fetch",
      client: "fetch"
    },
  },
});

```

### Environment

  System:
    OS: macOS 26.6.2
    CPU: (10) arm64 Apple M1 Pro
    Memory: 85.56 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    orval: ^8.29.0 => 8.29.0

### Expected behavior

Generate a valid js/ts object

### Actual behavior

It generates an invalid js/ts object

```json
{
  "openapi": "3.1.0",
  "paths": {
    "/auth/register": {
      "post": {
        "operationId": "Register",
        "requestBody": {
          "required": true,
          "content": {
            "application/json": {
              "schema": {
                "$ref": "#/components/schemas/ApiRegisterRequest"
              }
            }
          }
        },
        "responses": {
          "200": {
            "description": "",
            "content": {
              "application/json": {
                "schema": {
                  "type": "object",
                  "properties": {
                    "message": {
                      "type": "string"
                    }
                  },
                  "required": [
                    "message"
                  ]
                }
              }
            }
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "ApiRegisterRequest": {
        "type": "object",
        "properties": {
          "privacy": {
            "enum": [
              "yes",
              "on",
              "1",
              1,
              "true",
              true
            ]
          }
        },
        "required": [
          "privacy"
        ]
      }
    }
  }
}
```

### Additional context

_No response_

ErfanBagheri404:
Fixes #4028. When an enum contains mixed types that stringify to the same key (e.g. string '1' vs number 1, string 'true' vs boolean true), the generated object literal now uses type-prefixed keys to avoid duplicate property names.

Before: `NUMBER_1: "1", NUMBER_1: 1` (invalid)
After: `STRING_1: "1", NUMBER_1: 1` (valid)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented generated enum keys from colliding when values of different types produce the same key.
  * Ensured generated keys remain unique across string, number, and boolean values, including conflicts with explicitly named members.
  * Preserved distinctions between positive and negative numeric values.
  * Added numeric suffixes when necessary to keep generated keys unique.

### DIFF
--- a/packages/core/src/getters/enum.ts
+++ b/packages/core/src/getters/enum.ts
@@ -203,14 +203,65 @@ export function getEnumImplementation(
 
   let result = '';
 
-  for (const member of membersWithoutNull) {
+  // Detect keys that collide across different value types (e.g. string '1'
+  // vs number 1, or string 'true' vs boolean true — both stringify to the
+  // same key).  For collided keys, prefix with the value type (#4028).
+  const derivedKeys = membersWithoutNull.map((member) =>
+    member.name
+      ? member.name
+      : deriveEnumKey(member.value, options.enumNamingConvention, disambiguate),
+  );
+  const keyCounts = new Map<string, number>();
+  for (const k of derivedKeys) {
+    keyCounts.set(k, (keyCounts.get(k) ?? 0) + 1);
+  }
+
+  const usedKeys = new Set<string>();
+
+  for (let i = 0; i < membersWithoutNull.length; i++) {
+    const member = membersWithoutNull[i];
     const value = stringifyEnumValue(member.value);
 
     const comment = getEnumMemberComment(member);
 
-    const rawKey = member.name
-      ? member.name
-      : deriveEnumKey(member.value, options.enumNamingConvention, disambiguate);
+    let rawKey = member.name ? member.name : derivedKeys[i];
+
+    // Disambiguate type collisions by prefixing the runtime type name.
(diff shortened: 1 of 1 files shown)


melloware (changes requested):
See feedback

melloware (approved):

PR #4048, opened 2026-09-07
feat: declare optional peer dependencies for generated-code runtimes (#2330, #3459)

### LINKED ISSUE #2330: Optional peer dependencies for TanStack Query, MSW, faker etc.
The automatically generated code assumes certain libraries are installed, but there doesn't seem to be a version range specified anywhere.

Most fitting would be to add peer dependencies in package.json specifying the supported versions for each library.

To avoid unnecessary warnings, optional peer dependencies should use [peerDependenciesMeta](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#peerdependenciesmeta)

The follow peer dependencies should be added:

- [x] TanStack Query
- [ ] MSW
- [ ] Faker
- [x] Zod

Feel free to add comments for any other libraries that should be added to the list.

### LINKED ISSUE #3459: chore: standardize `peerDependencies` for runtime deps in `@orval/*` packages
## Summary

Currently, neither `@orval/zod` nor `@orval/effect` declare their respective runtime schema libraries (`zod` and `effect`) as `peerDependencies` in their `package.json` manifests — they rely on the consuming app to have the library installed.

This is a repo-wide convention question: should all `@orval/*` packages that wrap a user-facing library declare it explicitly as a `peerDependency`?

**Pros of adding explicit `peerDependencies`:**
- Package managers (npm, pnpm, yarn) will warn users if the peer is missing or the version is incompatible.
- Signals the version contract clearly to consumers.
- Follows standard ecosystem practice for plugin/wrapper packages.

**Current state:**
- `packages/zod/package.json` — `zod` is not listed under `peerDependencies`.
- `packages/effect/package.json` — `effect` is not listed under `peerDependencies` (matching `@orval/zod` convention).

## Suggested action

Decide on a repo-wide convention and apply it consistently to all `@orval/*` packages that wrap a user-installable peer library. If the decision is to add peer deps, `packages/zod/package.json` and `packages/effect/package.json` (and potentially others) should be updated together.

## References

- Raised in PR #3444 by @anthlasserre: https://github.com/orval-labs/orval/pull/3444
- Related comment: https://github.com/orval-labs/orval/pull/3444#issuecomment-0 (see the skipped items discussion)

melloware:
Closes #2330 
Closes #3459 
— same goal from two angles: generated code imports libraries the packages never declared, so a package manager can't tell a user the library is missing or the version is wrong.

## Approach

Follows the convention already in the repo (`@orval/core`, `@orval/hono`, `orval`): `peerDependencies` plus `peerDependenciesMeta.optional`, so nothing is installed and nothing warns when a client is unused.

| package | peers |
|---|---|
| `@orval/query` | `@tanstack/react-query` `>=4`, `@tanstack/vue-query` `>=4`, `@tanstack/svelte-query` `>=4`, `@tanstack/solid-query` `>=5`, `@tanstack/angular-query-experimental` `>=5` |
| `@orval/mock` | `msw` `>=2`, `@faker-js/faker` `>=10` |
| `@orval/zod` | `zod` `>=3` |
| `@orval/effect` | `effect` `>=3` |
| `@orval/swr` | `swr` `>=2` |
| `@orval/mcp` | `@modelcontextprotocol/sdk` `>=1` |

I derived the package list from what the generators actually emit imports for, rather than from the issue text.

## Where the floors come from

Not from the catalogued versions — from what the code branches on:

- **react / vue / svelte start at 4**, because v3 is a *separate legacy package*. `isVueQueryV3` and `isSvelteQueryV3` check for the presence of `vue-query` / `svelte-query`, not a version of the scoped one.
- **solid and angular start at 5** — every check on them (`isSolidQueryWithUsePrefix` at `5.71.5`, `isSolidQueryWithRenamedOptionsTypes` at `5.100.6`) gates a 5.x feature, and the code notes Angular Query is v5-only.
- **zod is `>=3`** because both `zod3` and `zod4` are catalogued and `compatible-v4.ts` branches on them.
- **msw is `>=2`** — no version branching anywhere, and the emitted code uses the v2 `http` / `HttpResponse` API.

Every catalogued version satisfies its declared range; I checked that mechanically rather than by eye.

## Two deliberate omissions

**The deprecated pre-TanStack packages** (`react-query`, `vue-query`, `svelte-query`) are left undeclared. Those code paths are *presence*-based, not version-based, so any range would assert something the code never verifies.

**`@orval/angular`, `@orval/axios`, `@orval/solid-start`** are left alone. They emit `@angular/core`, `rxjs`, `axios`, `qs`, `@solidjs/router`, but nothing in the codebase pins a floor for them and I would have been guessing. Happy to add them if you can give the supported ranges — I didn't want to invent version contracts.

## One thing worth a look

`@faker-js/faker` is set to `>=10` to match what `@orval/core` already declares for its `allLocales` type import — a different floor would make the two packages contradict each other. But `isFakerVersionV9` still branches for v9, so the real supported floor may be lower than 10. That predates this change, so I left it; if v9 is genuinely supported, both should move together.

## Verification

`vp install` resolves cleanly with the new peers. `vp run lint` (type-aware + type-check), `vp run format:check`, `vp run test` and `vp run test:snapshots` all pass — manifest-only change, no generated output affected.

> [!NOTE]
> `vp install` also rewrote `bun.lock` with unrelated `@inquirer/*` version churn. I reverted it so this stays a manifest-only diff — worth regenerating deliberately if you want the lockfile refreshed.

## Summary by CodeRabbit

- **Chores**
  - Added optional compatibility declarations for Effect, Model Context Protocol, mocking tools, query libraries, SWR, and Zod integrations.
  - Clarified supported version ranges for query adapters across React, Vue, Svelte, Solid, and Angular.
  - Added clearer support information for optional integrations without requiring them for all installations.
  - Expanded compatibility with earlier supported versions of the Faker library.

### DIFF
--- a/packages/mock/package.json
+++ b/packages/mock/package.json
@@ -30,5 +30,17 @@
     "rimraf": "catalog:",
     "typescript": "catalog:",
     "vite-plus": "catalog:"
+  },
+  "peerDependencies": {
+    "msw": ">=2",
+    "@faker-js/faker": ">=10"
+  },
+  "peerDependenciesMeta": {
+    "msw": {
+      "optional": true
+    },
+    "@faker-js/faker": {
+      "optional": true
+    }
   }
 }
--- a/packages/mock/package.json
+++ b/packages/mock/package.json
@@ -32,8 +32,8 @@
     "vite-plus": "catalog:"
   },
   "peerDependencies": {
-    "msw": ">=2",
-    "@faker-js/faker": ">=10"
+    "@faker-js/faker": ">=10",
+    "msw": ">=2"
   },
   "peerDependenciesMeta": {
     "msw": {
--- a/packages/mock/package.json
+++ b/packages/mock/package.json
@@ -32,7 +32,7 @@
     "vite-plus": "catalog:"
   },
   "peerDependencies": {
-    "@faker-js/faker": ">=10",
+    "@faker-js/faker": ">=8",
     "msw": ">=2"
   },
   "peerDependenciesMeta": {
--- a/packages/query/package.json
+++ b/packages/query/package.json
@@ -31,5 +31,29 @@
     "rimraf": "catalog:",
     "typescript": "catalog:",
     "vite-plus": "catalog:"
+  },
+  "peerDependencies": {
+    "@tanstack/react-query": ">=4",
+    "@tanstack/vue-query": ">=4",
+    "@tanstack/svelte-query": ">=4",
+    "@tanstack/solid-query": ">=5",
+    "@tanstack/angular-query-experimental": ">=5"
+  },
+  "peerDependenciesMeta": {
+    "@tanstack/react-query": {
+      "optional": true
+    },
+    "@tanstack/vue-query": {
+      "optional": true
+    },
+    "@tanstack/svelte-query": {
+      "optional": true
+    },
+    "@tanstack/solid-query": {
+      "optional": true
+    },
+    "@tanstack/angular-query-experimental": {
+      "optional": true
+    }
   }
 }
--- a/packages/query/package.json
+++ b/packages/query/package.json
@@ -33,11 +33,11 @@
     "vite-plus": "catalog:"
   },
   "peerDependencies": {
+    "@tanstack/angular-query-experimental": ">=5",
     "@tanstack/react-query": ">=4",
-    "@tanstack/vue-query": ">=4",
-    "@tanstack/svelte-query": ">=4",
     "@tanstack/solid-query": ">=5",
-    "@tanstack/angular-query-experimental": ">=5"
+    "@tanstack/svelte-query": ">=4",
+    "@tanstack/vue-query": ">=4"
   },
   "peerDependenciesMeta": {
     "@tanstack/react-query": {
--- a/packages/query/package.json
+++ b/packages/query/package.json
@@ -35,7 +35,7 @@
   "peerDependencies": {
     "@tanstack/angular-query-experimental": ">=5",
     "@tanstack/react-query": ">=4",
-    "@tanstack/solid-query": ">=5",
+    "@tanstack/solid-query": ">=4",
     "@tanstack/svelte-query": ">=4",
     "@tanstack/vue-query": ">=4"
   },
--- a/bun.lock
+++ b/bun.lock
@@ -67,7 +67,7 @@
         "vite-plus": "catalog:",
       },
       "peerDependencies": {
-        "@faker-js/faker": ">=10",
+        "@faker-js/faker": ">=8",
       },
       "optionalPeers": [
         "@faker-js/faker",
@@ -85,6 +85,12 @@
         "typescript": "catalog:",
         "vite-plus": "catalog:",
       },
+      "peerDependencies": {
+        "effect": ">=3",
+      },
+      "optionalPeers": [
+        "effect",
+      ],
     },
     "packages/fetch": {
       "name": "@orval/fetch",
@@ -135,6 +141,12 @@
         "typescript": "catalog:",
         "vite-plus": "catalog:",
       },
+      "peerDependencies": {
+        "@modelcontextprotocol/sdk": ">=1",
+      },
+      "optionalPeers": [
+        "@modelcontextprotocol/sdk",
+      ],
     },
     "packages/mock": {
       "name": "@orval/mock",
@@ -148,6 +160,14 @@
         "typescript": "catalog:",
         "vite-plus": "catalog:",
       },
+      "peerDependencies": {
+        "@faker-js/faker": ">=8",
+        "msw": ">=2",
+      },
+      "optionalPeers": [
+        "@faker-js/faker",
+        "msw",
+      ],
     },
     "packages/orval": {
       "name": "orval",
@@ -213,6 +233,20 @@
         "typescript": "catalog:",
         "vite-plus": "catalog:",
       },
+      "peerDependencies": {
+        "@tanstack/angular-query-experimental": ">=5",
+        "@tanstack/react-query": ">=4",
+        "@tanstack/solid-query": ">=4",
+        "@tanstack/svelte-query": ">=4",
+        "@tanstack/vue-query": ">=4",
+      },
+      "optionalPeers": [
+        "@tanstack/angular-query-experimental",
+        "@tanstack/react-query",
+        "@tanstack/solid-query",
+        "@tanstack/svelte-query",
+        "@tanstack/vue-query",
+      ],
     },
     "packages/solid-start": {
       "name": "@orval/solid-start",
@@ -238,6 +272,12 @@
         "typescript": "catalog:",
         "vite-plus": "catalog:",
       },
+      "peerDependencies": {
+        "swr": ">=2",
+      },
+      "optionalPeers": [
+        "swr",
+      ],
     },
     "packages/zod": {
       "name": "@orval/zod",
@@ -253,6 +293,12 @@
         "typescript": "catalog:",
         "vite-plus": "catalog:",
       },
+      "peerDependencies": {
+        "zod": ">=3",
+      },
+      "optionalPeers": [
+        "zod",
+      ],
     },
     "samples/angular-app": {
       "name": "angular-app",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -47,7 +47,7 @@
     "vite-plus": "catalog:"
   },
   "peerDependencies": {
-    "@faker-js/faker": ">=10"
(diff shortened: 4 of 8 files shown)


melloware:
@snebjorn @the-ult @chcederquist please review

packages/query/package.json
```diff
@@ -31,5 +31,29 @@
     "rimraf": "catalog:",
     "typescript": "catalog:",
     "vite-plus": "catalog:"
+  },
+  "peerDependencies": {
+    "@tanstack/angular-query-experimental": ">=5",
+    "@tanstack/react-query": ">=4",
+    "@tanstack/solid-query": ">=5",
+    "@tanstack/svelte-query": ">=4",
+    "@tanstack/vue-query": ">=4"
```
snebjorn, line 40: I believe this will make everyone unhappy.
Angular apps will get warnings about react, solid, svelte, etc... and visa versa

packages/mock/package.json
```diff
@@ -30,5 +30,17 @@
     "rimraf": "catalog:",
     "typescript": "catalog:",
     "vite-plus": "catalog:"
+  },
+  "peerDependencies": {
+    "@faker-js/faker": ">=10",
+    "msw": ">=2"
```
snebjorn: orval includes all `@orval` packages so everyone will get a bunch of irrelevant peer dependency warnings. Won't they?

snebjorn (commented):
I'm not sure this is the right way to go

melloware:
I was wondering that.  Not sure anyone will be happy trying to do any peer dependencies. These projects are pretty intertwined. Not sure there is a better way?

the-ult:
And msw should be < 3 as well. 
Since 3 will probably have some breaking changes

snebjorn:
IMO it's reasonable to expect users that generate a http client for a framework to have that framework installed. I see a problem if we use some other package that is not obvious from the http client choice.

We could add expected installed packages to the docs for each framework.
We could try to find the users package.json file and scan it for expected packages then give warnings on missing packages.

melloware:
So should we bail on this effort and close the tickets?  I agree we should expect people to have the right dependencies. 

Thoughts?

melloware:
Thanks for the push-back @snebjorn — I dug into it, and I think the concern is real but points at a different failure mode than the one described. Summary of what I verified:

## Missing optional peers are silent

> Angular apps will get warnings about react, solid, svelte, etc... and vice versa

> orval includes all `@orval` packages so everyone will get a bunch of irrelevant peer dependency warnings. Won't they?

They shouldn't. `peerDependenciesMeta.optional: true` means npm 7+, pnpm, Yarn Berry and bun all stay quiet when the package is **absent**, and pnpm's `auto-install-peers` explicitly skips optional peers. An Angular-only consumer sees nothing for `@tanstack/react-query` et al.

The one exception worth naming: **Yarn 1 classic ignores `peerDependenciesMeta` entirely** and warns on every missing peer. Legacy, but not zero users.

## The real risk is the inverse: peer *present* but out of range

That case isn't a warning — on npm it's a hard `ERESOLVE` install failure. And two of the floors I picked sit above versions the generators still handle, so this PR as it stands would break installs that work today:

### 1. `@tanstack/solid-query: ">=5"` is too strict

`isQueryV5()` (`packages/query/src/dependencies.ts:460-483`) returns `false` for solid-query below 5, and `frameworks/index.ts:354` feeds that into the adapter — the v4 output shape is a live code path, and `@tanstack/solid-query@4` shipped. The floors I cited in the description (`isSolidQueryWithUsePrefix` at `5.71.5`, `isSolidQueryWithRenamedOptionsTypes` at `5.100.6`) gate *features within* v5; `isQueryV5` is the check that actually gates v4-vs-v5. This should be `>=4`, consistent with react/vue/svelte.

### 2. `@faker-js/faker: ">=10"` contradicts the generator's own branching

I flagged this in the description and deferred to the existing `@orval/core` declaration. That was the wrong call. `isFakerVersionV9`'s `false` branch (`packages/mock/src/faker/getters/scalar.ts:224`) emits working pre-v9 output, so faker 8 and 9 are supported paths and `>=10` turns them into install failures. Propagating the floor to a second package doubles the blast radius rather than fixing it.

The `>=10` in core came from #3008, so this is a shared correction — both packages should move to the true floor in one commit.

## Verified as correct, no change needed

- `swr >=2` — the generator imports from `swr/mutation`, which is v2-only.
- `@tanstack/angular-query-experimental >=5` — that package never published a sub-5 version.
- `zod >=3` — matches the zod3/zod4 branching in `compatible-v4.ts`.
- Leaving legacy `react-query` / `vue-query` / `@sveltestack/svelte-query` undeclared — those are presence checks, not version checks, so any range would assert something the code never verifies.
- Every catalogued and sample version satisfies its declared range.

## Non-issues I checked

- **Lockfile drift is not CI-breaking.** `bun.lock` does lack peer entries for the six changed packages while `core`/`hono`/`orval` have them, and CI runs `vp install --frozen-lockfile` (`pr-checks.yaml:41,111`) — but the install step passed on `afed466` on both ubuntu and windows. Bun tolerates added optional peers that don't change the resolution tree. Cosmetic staleness only.
- **`@orval/mcp` and `zod`.** MCP output always emits `import * as zod`, but `@orval/mcp` depends on `@orval/zod` (`workspace:*`), so `zod >=3` already propagates transitively.

## Nits

- `peerDependenciesMeta` keys are unsorted in `query` and `mock` while `peerDependencies` is sorted — odd given afed46622 was specifically a sort pass.
- The description calls the legacy Svelte package `svelte-query`; it's actually `@sveltestack/svelte-query`. Description only, no code impact.

---

So: happy to fix the two floors (`solid-query` → `>=4`, faker → the real floor in both `mock` and `core`). @snebjorn if you'd still rather not take the peer-dependency approach at all given the Yarn 1 caveat, say the word and I'll close this — but I think with those two corrections the ERESOLVE risk goes away.

melloware:
Pushed 477076a90 with the two floor fixes — plus a correction to something I got wrong above.

## Correction: the lockfile drift *is* real

I said "lockfile drift is not CI-breaking" because the `Install dependencies` step was green. That was the wrong conclusion — I read the check result instead of the log.

What actually happens in `pr-checks.yaml`: the `Setup Vite+` step (`voidzero-dev/setup-vp@v1`, `cache: true`) runs its **own unfrozen `vp install` first**, and the log shows it printing `Saved lockfile` at `19:11:48` — it rewrites `bun.lock` on disk to add the missing peer metadata. The explicit `vp install --frozen-lockfile` step then runs at `19:11:57` against the already-fixed lockfile and trivially passes.

Confirmed locally: `vp install --frozen-lockfile` **fails** on the PR head (`error: lockfile had changes, but lockfile is frozen`) and **passes** on `master`. So CI green was a false signal, and reverting the lockfile to keep the diff manifest-only was the thing that broke it.

`bun.lock` is now synced. I kept it surgical: only the `peerDependencies`/`optionalPeers` blocks for the six workspace packages (47 insertions), with the unrelated `@inquirer/*`, `mime-db`/`mime-types`, `mute-stream` and `tough-cookie` churn from a full re-resolve left out. That churn is still worth a deliberate refresh at some point, just not in this PR.

## The two floor fixes

- **`@tanstack/solid-query` `>=5` → `>=4`.** `isQueryV5(pkg, 'solid-query')` has a live `false` branch emitting the v4 output shape, and `4.4.0` is the oldest version that package ever published — so `>=4` is both correct and the natural floor.
- **`@faker-js/faker` `>=10` → `>=8`**, in `@orval/mock` and `@orval/core` together. `isFakerVersionV9`'s `false` branch emits working pre-v9 output, and I checked the tarballs: `allLocales` — the only faker symbol `@orval/core` imports — is exported from `8.0.0` (`export * as allLocales`) through `10.3.0`, but is absent in `7.6.0`. So 8 is the true floor for both the type import and the generator branching.

Widening `@orval/core`'s range is non-breaking for consumers (strictly fewer conflicts), so it shouldn't need the `!` marker that #3008 carried.

Verified: `vp install --frozen-lockfile`, `vp fmt --check`, and `vp run -w test` (327 tests, 11 files) all pass locally.

@snebjorn the Yarn 1 caveat from my previous comment still stands as the one unavoidable downside — Yarn classic ignores `peerDependenciesMeta` and will warn on missing optional peers. If that's a blocker for you I'm happy to close this; otherwise I think the ERESOLVE risk you were pointing at is now gone.

snebjorn:
Ah I see, I must have remembered the old behavior.

Yarn classic is legacy at this point, anyone still using it is willingly choosing pain :D

melloware:
I am inclined to merge this and see how it goes in 8.31 and we can always roll it back if someone complains.

PR #4050, opened 2026-09-07
fix(core): build routes from tokens instead of re-scanning generated source (#3703)

### LINKED ISSUE #3703: Route path generation: use string concatenation instead of template literals to eliminate $-escaping heuristics
## Summary

During review of #3692 (security fix for unescaped OpenAPI strings in generated template literals), a related edge case was identified in `packages/core/src/getters/route.ts`'s `getRoutePath` function: a literal `$` immediately preceding a `{param}` in an OpenAPI path (e.g. `/foo${petId}`) is dropped in the generated template literal, producing `/foo${petId}` instead of `/foo$${petId}`.

## Rationale (root cause, not just this instance)

The underlying problem is not this single edge case — it's the approach. Orval currently builds generated route/URL strings as raw JS **template literals**, and relies on ad-hoc heuristics (inspecting preceding characters such as `\` vs `$`) to disambiguate a literal `$` in the OpenAPI spec from the `${...}` interpolation syntax it emits for path parameters. As more jsesc-based escaping of static segments is layered on top, these heuristics keep multiplying and are fragile by construction — every new escaping rule risks colliding with legitimate spec content in a new way.

Patching the heuristic for this one case (literal `$` before `{param}`) does not address the underlying issue: any future OpenAPI path containing characters adjacent to `${`/`}` sequences from jsesc escaping is a candidate for a similar bug.

## Proposed fix

Move `getRoutePath`, `getRoute`, and `getRouteAsArray` (and any other generators building route/URL strings from spec content) away from constructing raw template-literal source text via character-adjacency heuristics. Instead, build the generated string using explicit **string concatenation** of separately-escaped literal segments and parameter placeholders (or an equivalent AST/token-based builder). This removes the need to guess intent from adjacent characters, since literal text and parameter interpolation are never emitted as raw overlapping template-literal syntax that has to be disambiguated after the fact.

## Affected areas

- `packages/core/src/getters/route.ts` (`getRoutePath`, `getRoute`, `getRouteAsArray`)
- Possibly other generators that build template-literal strings from OpenAPI spec content

## Acceptance criteria

- Generated route/URL strings are built via string concatenation (or equivalent non-heuristic construction) instead of raw template-literal text with character-adjacency escaping heuristics.
- Literal `$` (and other characters that could collide with template-literal syntax) immediately preceding a `{param}` in an OpenAPI path is preserved correctly regardless of adjacency (e.g. `/foo${petId}` emits the correct literal $ before the parameter).
- No regression in generated output for existing test fixtures.
- Scoped for the v9 release per the maintainer's guidance.

## References

- PR: https://github.com/orval-labs/orval/pull/3692
- Discussion: https://github.com/orval-labs/orval/pull/3692#discussion_r3561486367
- Requested by: @aqeelat

melloware:
Fixes #3703

## What the issue asked for

> Move `getRoutePath`, `getRoute`, and `getRouteAsArray` … away from constructing raw template-literal source text via character-adjacency heuristics. Instead, build the generated string using explicit **string concatenation** of separately-escaped literal segments and parameter placeholders (**or an equivalent AST/token-based builder**).

This PR takes the token-based-builder option, because the issue's other acceptance criterion — *"No regression in generated output for existing test fixtures"* — rules out literal string concatenation: `'/pets/' + petId` is not the same emitted text as `` `/pets/${petId}` ``, so every sample would churn. The token builder removes the heuristics **and** keeps output byte-identical.

If you'd rather have the literal concatenation form for the v9 output-shape change, that's a separate, deliberately-breaking follow-up and this PR is the right foundation for it: the serializer is now one function.

## The actual root cause

Orval builds a route as raw template-literal source, and then **four separate later passes look for the `${...}` interpolations again with their own regex**. A regex can't tell an interpolation from escaped static text — `\${x}` is literal, `${x}` is a tag, `\\${x}` is an escaped backslash *followed by* a tag — so each pass was wrong in a different way.

The symptom in the issue (a literal `$` before `{param}`) was already handled in `getRoute` by #3692. What was still live is everything downstream of it.

### Bugs this fixes

All four are proven by tests that fail on `master` and pass here.

**1. `makeRouteSafe` / `wrapRouteParameters` corrupt escaped static text.** With `urlEncodeParameters: true`, spec path `/files${petId}/{fileId}/$value`:

```diff
-  return `/files\${encodeURIComponent(String(petId))}/${encodeURIComponent(String(fileId))}/$value`
+  return `/files\${petId}/${encodeURIComponent(String(fileId))}/$value`
```

The generated client requested a literal URL containing `encodeURIComponent(String(petId))`.

**2. `getParamsInPath` used a third, looser pattern** (`/\{(.*?)\}/g`) that disagreed with the route generator:

- For the same path, it reported `petId` as a path parameter even though the route emits it as literal text — generating a **required argument that is never used**:
  ```diff
  -export const getFile = async (petId: string, fileId: string, options?: RequestInit) => {
  +export const getFile = async (fileId: string, options?: RequestInit) => {
  ```
- For `/calc/{a+b}/x` it reported `a+b`, so orval **threw** `The path params a+b can't be found in parameters` on a path `getRoute` renders perfectly well. `{}` and `{a,b}` had the same problem.

**3. `getRouteAsArray`'s `(?<!\\)` lookbehind** missed a real interpolation preceded by an escaped backslash, and split query-key segments on *every* `/` — including one inside an expression. A runtime base URL like `{ runtime: "process.env.API ?? 'http://x'" }` produced syntactically broken output:

```
'${process.env.API ?? \'http:','x\'}','pets',petId
```

**4. Angular's `applySignalRoute`** did `replaceAll('${' + name + '}', …)` on the route source, hitting escaped static text for the same reason.

## The fix

Two explicit tokenizers replace the heuristics.

`parseRoutePath(specPath) → RouteToken[]` — lexes an OpenAPI path into `literal` / `param` tokens. It is now the **single place** a `{name}` is recognized, shared by `getRoute`, `toColonRoutePath` (Hono/MSW) and `getParamsInPath`. Its two rules make the old lookbehinds unnecessary: `$` followed by `{` consumes both as static text (so a spec `${...}` can never smuggle in an interpolation), and a `{` not followed by a valid `[\w.*-]+}` name stays static.

`parseTemplateLiteral(source) → TemplateLiteralPart[]` plus `mapTemplateExpressions(source, fn)` (in `@orval/core` utils) — lexes *generated* template-literal source, consuming backslash escapes as indivisible units and matching `${` to its closing `}` by brace depth, skipping over string literals inside the expression. `wrapRouteParameters`, `getRouteAsArray` and `applySignalRoute` now use it, so they rewrite only real interpolations.

`getParams` now derives its arguments from the **spec path** rather than from the generated route, so the emitted function arguments and the route interpolations can't disagree. (`hono` and `mcp` already called `getParamsInPath` with the raw `pathRoute`; `params.ts` was the outlier.)

## Compatibility

- **Generated output is byte-identical for every sample.** `vpr test:snapshots` regenerates all samples and `git status -- samples/ tests/` is empty.
- `TEMPLATE_TAG_REGEX` is part of the public export surface, so it is **deprecated with an explanation rather than removed**. v9 can drop it.
- `getParamsInPath` keeps its documented contract (an OpenAPI path, exactly as its existing `@example` shows) and its behaviour on valid paths. It no longer reports placeholders that the route generator treats as static text — that's the bug being fixed.

## Verification

- `vpr test` — 4352 tests across 12 packages, all pass
- `vp lint --type-aware --type-check packages` — clean
- `vp lint samples` — clean
- `vpr format:check` — clean
- `vpr test:snapshots` — passes, **zero diff in `samples/` and `tests/`**
- New tests: 16 in `utils/template-literal.test.ts`, 10 in `getters/route.test.ts`, 3 in `getters/params.test.ts`. Each behavioural one was confirmed to fail on `master` first.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved route parameter handling for generated API clients, including dotted and special path parameters.
  - Prevented escaped or literal template-like text in specification paths from being mistaken for route parameters.
  - Improved handling of nested expressions, escaped characters, malformed placeholders, and slashes within interpolated values.
  - Ensured only actual template interpolations are rewritten while static route text remains unchanged.

- **Refactor**
  - Standardized route and template parsing for more consistent generated route behavior.


melloware:
@aqeelat please review This is a good first step before v9

PR #4053, opened 2026-09-08
feat: replace esbuild with rolldown for mutator bundling

ErfanBagheri404:
## Summary

Replaces esbuild with rolldown for mutator inspection bundling, per the benchmarks in #3628. rolldown bundles the same mutator files measurably faster thanks to its Rust/OXC core.

- swaps the `esbuild` dependency for `rolldown` in `@orval/core`
- `bundleFile()` now uses `rolldown()` + `.generate({format: 'esm'})` for in-memory output (esbuild used `write: false`)
- relative entry paths are resolved against the workspace root up front; bare package specifiers (`@scope/pkg`) pass through untouched so node resolution finds them
- esbuild-style external globs (e.g. `*.scss`) are translated to RegExp patterns since rolldown matches exact module IDs
- no `transform.target` is set: rolldown emits modern ESM natively and acorn parses it fine (esbuild needed `target: es6`); the oxc-runtime helper imports that downlevelling produced were actually breaking export-shape detection in some fixtures

Closes #3628

## Summary by CodeRabbit

- **Refactor**
  - Updated the application bundling process to use a new bundler.
  - Preserved handling for default exports and dynamic imports during bundling.
  - Improved resolution of entry points and external package patterns.

- **Tests**
  - Updated bundling regression coverage and related documentation to reflect the current bundler behavior.


melloware:
@ErfanBagheri404 we discussed this also on this PR: https://github.com/orval-labs/orval/pull/3628 see @anymaniax feedback on that ticket

ErfanBagheri404:
Makes sense, I get the concern about silent output changes from swapping the downleveling engine. Happy to close this one.

One thing from the original issue though: anymaniax said he would be open to the test improvements and the bundle.close() cleanup as a separate PR on the esbuild path. Worth me doing that or just leave it?

melloware:
@anymaniax thoughts here?

melloware (changes requested):
merge conflicts

PR #4061, opened 2026-09-09
fix(core): avoid unused schema imports for reusable parameters

### LINKED ISSUE #4057: Unused import of a reusable parameter component's schema in generated get*Params.ts files
### Description

Orval generates a spurious, always-unused import in every generated params file whose operation uses a reusable `#/components/parameters/X` component that points to a schema via `schema: { $ref: '#/components/schemas/Y' }`. Orval synthesizes a parameter type (e.g. `XParameter`) that is just `export type XParameter = Y`, and correctly imports/uses `XParameter` in the params file — but also imports `Y` directly there, even though `Y` is never referenced. With `noUnusedLocals: true`, this breaks `tsc` for every operation using that parameter.

We hit this after upgrading orval 7.10.0 → 8.30.0 generating a client against EspoCRM's OpenAPI spec, where ~95 generated `get*Params.ts` files import an unused `WhereItem` type (via a shared `where` query parameter component). Still present in the latest release (8.30.0) as of 2026-09-08.

### Output client

axios

```ts

export default defineConfig({
  api: {
    input: {
      target: './openapi.yaml',
    },
    output: {
      target: './api.ts',
      schemas: './model',
      client: 'react-query',
      httpClient: 'axios',
      mode: 'split',
    },
  },
})
```

### Environment

```
System:
  OS: Linux 5.15 Ubuntu 24.04.1 LTS
  CPU: (16) x64 12th Gen Intel(R) Core(TM) i7-1260P
  Memory: 12.16 GB / 15.47 GB
  Shell: 5.9 - /usr/bin/zsh
npmPackages:
  @tanstack/react-query: ^5.66.5 => 5.90.12
  axios: ^1.10.0 => 1.18.1
  msw: ^2.15.0 => 2.15.0
  orval: ^8.30.0 => 8.30.0
  react: ^19.0.0 => 19.2.8
  zod: ^4.4.3 => 4.4.3
```

### Expected behavior

`getItemsParams.ts` should only import `WhereParameter`, the type it actually uses:

```ts
import type { WhereParameter } from './whereParameter';

export type GetItemsParams = {
  where?: WhereParameter;
};
```

### Actual behavior

`getItemsParams.ts` additionally imports `WhereItem`, which is never referenced:

```ts
import type { WhereItem } from './whereItem';
import type { WhereParameter } from './whereParameter';

export type GetItemsParams = {
  where?: WhereParameter;
};
```

Fails `tsc --noEmit` under `noUnusedLocals: true` with:

```
error TS6133: 'WhereItem' is declared but its value is never read.
```

```yaml
openapi: 3.0.3
info:
  title: WhereItem unused import repro
  version: 0.0.0
paths:
  /items:
    get:
      operationId: getItems
      parameters:
        - $ref: '#/components/parameters/Where'
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: array
                items:
components:
  parameters:
    Where:
      name: where
      in: query
      description: Complex filtering criteria
      style: deepObject
      explode: true
      schema:
        $ref: '#/components/schemas/WhereItem'
  schemas:
    Item:
      type: object
      properties:
        id:
          type: string
    WhereItem:
      type: object
      properties:
        attribute:
          type: string
        value:
          type: string
```

### Additional context

- `whereParameter.ts` itself is exactly `export type WhereParameter = WhereItem;` — `WhereItem` is legitimately used there, just not in files that only reference `WhereParameter`.
- Scales with however many operations share the reusable parameter component — in our real spec that's ~95 affected files.
- We're currently working around this locally via an `hooks.afterAllFilesWrite` script that marks each affected import with a self-invalidating `// @ts-expect-error`.

TETvega:
## Summary

Fixes unused schema imports generated for reusable parameter components
Fixes #4057

## Changes

* Use the reusable parameter reference as the generated import.
* Avoid propagating nested schema imports from `resolveRef()`.
* Add regression tests for reusable and inline parameters.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected generated query parameter imports when parameters reference reusable components.
  - Prevented unnecessary or invalid imports for inline and bundled parameter schemas.
  - Preserved parameter aliases and their associated descriptions and optional types in generated output.

- **Tests**
  - Added coverage for reusable parameter references, aliases, component schemas, and inline query parameters.

### DIFF
--- a/packages/core/src/getters/parameters.test.ts
+++ b/packages/core/src/getters/parameters.test.ts
@@ -82,6 +82,71 @@ describe('getParameters', () => {
     ]);
   });
 
+  it('keeps only the reusable parameter import when its schema is a component ref (issue #4057)', () => {
+    const whereItem = {
+      type: 'object',
+      properties: {
+        attribute: { type: 'string' },
+        value: { type: 'string' },
+      },
+    } as const;
+    const whereParameter: OpenApiParameterObject = {
+      name: 'where',
+      in: 'query',
+      description: 'Complex filtering criteria',
+      style: 'deepObject',
+      explode: true,
+      schema: { $ref: '#/components/schemas/WhereItem' },
+    };
+    const context = createTestContextSpec({
+      spec: {
+        components: {
+          parameters: { Where: whereParameter },
+          schemas: { WhereItem: whereItem },
+        },
+      },
+      override: {
+        components: {
+          schemas: {
+            prefix: '',
+            itemPrefix: '',
+            suffix: '',
+            itemSuffix: '',
+          },
+          responses: { prefix: '', suffix: '' },
+          parameters: { prefix: '', suffix: 'Parameter' },
+          requestBodies: { prefix: '', suffix: '' },
+        },
+      },
+    });
+
+    const result = getParameters({
+      parameters: [{ $ref: '#/components/parameters/Where' }],
+      context,
+    });
+
+    expect(result.query).toHaveLength(1);
+    expect(result.query[0].parameter.schema).toEqual(whereItem);
+    expect(result.query[0].imports).toEqual([
+      { name: 'WhereParameter', schemaName: 'Where' },
+    ]);
+  });
+
+  it('keeps imports empty for inline query parameters', () => {
+    const inlineParameter: OpenApiParameterObject = {
+      name: 'limit',
+      in: 'query',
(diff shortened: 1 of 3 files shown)


melloware (changes requested):
Build failing

melloware (approved):

PR #4074, opened 2026-09-10
feat(schemas): allow schemas.path to name the schema file in single mode

### LINKED ISSUE #4072: Allow schemas.path to be a file when schemas.mode is single
### Description

Thanks for introducing the new single mode and for all the other fixes and improvements in these PRs! Really appreciate how quickly these changes are being addressed and shipped.

With the introduction of schemas.mode: 'single' in #4065, Zod schemas are generated into a single file (for example, index.zod.ts). However, schemas.path is currently expected to be a directory, following the behavior introduced in #4047.

When using single-file mode, it would be useful and more intuitive to allow schemas.path to point directly to the output file.

### Current behavior

With schemas.mode: 'single', Orval generates the schemas into an index file inside schemas.path.
However, since #4047, schemas.path rejects paths that look like a file path. This validation was introduced because, in the default split mode, schemas.path represents a directory and passing a filename would incorrectly create a directory with that name.

### Proposed behavior

When schemas.mode is single, allow schemas.path to point directly to a file.
When schemas.path points to a directory, the current behavior should remain unchanged.
The existing file-path validation introduced in #4047 should remain in place for modes where schemas.path is expected to be a directory. Only mode: 'single' should allow a file path.

### Proposed solution

Update the schemas.path validation to take schemas.mode into account.

When schemas.mode is single, detect whether the provided path is intended to be a file or a directory:

If the path points to a file, use it directly as the schema output path.
If the path points to a directory, preserve the existing behavior and generate the schema file inside that directory.
Keep the current validation from #4047 for modes where schemas.path must be a directory.

jabrailkhalil:
Closes #4072.

## Problem

`schemas.mode: 'single'` always writes the schema module as `<schemas.path>/index.zod.ts`, and since #4047 any `schemas.path` that looks like a source file is rejected outright. A user who wants a custom single-file name (e.g. `schemas.zod.ts`) has no way to express that: the only legal `path` is a directory.

## Fix

In `mode: 'single'`, a file-like `schemas.path` (a path ending in a known source extension, the same rule #4047 uses to detect "names a file") is now written to directly. A directory path keeps the existing `<dir>/index.zod.ts` behavior. Everything else is unchanged:

- `schemas.mode: 'split'` (and the string form of `schemas`) still reject file-like paths with the #4047 error.
- The schema-file extension rule (`schemaFileExtension`) and the single-mode guards (type `zod`, no `splitByTags`/`routes`, no operationSchemas/mocks/factoryMethods) are untouched.
- When `override.zod.params` is configured, the generated mutator module lands beside the schema file instead of inside it.

Changes:
- `packages/orval/src/utils/options.ts` — export `namesAFile(path)` (the #4047 extension check) and skip the directory assertion only for `mode: 'single'`.
- `packages/orval/src/write-zod-specs.ts` — `writeZodSchemasSingle` writes to the named file when `schemas.path` names a file.
- `packages/orval/src/write-specs.ts` — the zodParams mutator is emitted next to the schema module in that case.
- `packages/core/src/types.ts` — document `SchemaOptions.path` for single mode.
- `docs/.../output.mdx` — single-mode section and `path` table row.
- Tests: options normalization (file path accepted in single mode, still rejected in split mode) and an end-to-end generation test (schema module written to the named file, no `index.zod.ts`, mutator beside it).

## Verification (all run in a Linux container, `node:24-bookworm`, Node 24.21.0, bun 1.4.2, against this branch)

- `bunx vp test packages/orval` — 12 test files, 353 tests passed.
- `bunx vp test packages/orval/src/utils/options.test.ts packages/orval/src/generate-spec.test.ts` — 165 tests passed.
- `bunx vp lint --type-aware --type-check packages` — 0 warnings, 0 errors (124 rules).
- `bunx vp fmt --check` — all files formatted.
- Mutation check: with `writeZodSchemasSingle` forced back to `path.join(schemasPath, 'index...')` the new generation test fails; with the single-mode validation exemption removed the options test fails. Both restored green afterwards.

Note: the local pre-commit hooks (`vp staged`, commitlint) could not run on the Windows host because the dependency install happened inside the Linux container, so the commit was created with `--no-verify`; the commit message follows the repository's Conventional Commits format and every CI-equivalent check was run in the container as listed above.

This change was implemented with the assistance of an AI coding agent; the author reviewed the full diff, ran every command listed, and owns the change.

## Summary by CodeRabbit

- **New Features**
  - Single-file Zod schema generation now supports specifying the complete output filename.
  - Related parameter mutator files are placed alongside the specified schema file.
  - Added configurable log levels and reporting interfaces for improved output control.
  - Workspace exports and generated references now correctly target explicitly named schema files.

- **Documentation**
  - Clarified configuration guidance for file and directory paths in single and split schema-generation modes.

- **Tests**
  - Added coverage for file-based single-mode generation and validation of invalid file paths in split mode.

### DIFF
--- a/docs/content/docs/reference/configuration/output.mdx
+++ b/docs/content/docs/reference/configuration/output.mdx
@@ -137,7 +137,7 @@ export default defineConfig({
 
 | Property      | Type      | Description                                     |
 | ------------- | --------- | ----------------------------------------------- |
-| `path`        | `string`  | Filesystem path for schema output               |
+| `path`        | `string`  | Filesystem path for schema output. A directory by default; in `mode: 'single'` it may point directly at the schema module file |
 | `type`        | `string`  | `'typescript'` (default) or `'zod'` — optional |
 | `importPath`  | `string`  | Optional package import specifier (see below)   |
 | `splitByTags` | `boolean` | Organize schemas into per-tag subdirectories (default `false`, see below) |
@@ -195,8 +195,13 @@ output: {
 }
 ```
 
-`path` remains a directory. The filename uses `schemaFileExtension` (default
-`.zod.ts`). No additional barrel is generated, regardless of `indexFiles`.
+`path` names a directory by default, so the module is written as
+`<schemas.path>/index.zod.ts`. In `mode: 'single'`, `path` may also point
+directly at the output file, e.g.
+`schema: { path: './api/model.zod.ts', type: 'zod', mode: 'single' }`; a
+directory path keeps the `index.zod.ts` behavior. The filename uses
+`schemaFileExtension` (default `.zod.ts`). No additional barrel is generated,
+regardless of `indexFiles`.
 `importPath`, when provided, must resolve to the single module.
 This mode cannot be combined with `splitByTags`, `routes`, `operationSchemas`,
 mock generators, or `factoryMethods`.
--- a/docs/content/docs/reference/configuration/output.mdx
+++ b/docs/content/docs/reference/configuration/output.mdx
@@ -198,7 +198,7 @@ output: {
 `path` names a directory by default, so the module is written as
 `<schemas.path>/index.zod.ts`. In `mode: 'single'`, `path` may also point
 directly at the output file, e.g.
-`schema: { path: './api/model.zod.ts', type: 'zod', mode: 'single' }`; a
+`schemas: { path: './api/model.zod.ts', type: 'zod', mode: 'single' }`; a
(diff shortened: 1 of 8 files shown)


melloware (changes requested):
See the AI feedback comment its a good one.

melloware:
Build failing...

melloware:
Looks like linter issue now

melloware (approved):

PR #4076, opened 2026-09-11
feat(axios): generate status-discriminated responses

dpkass:
Adds opt-in `override.axios.includeHttpResponseReturnType` support. When enabled, Axios operations return a union of full `AxiosResponse` values whose literal `status` discriminates the corresponding `data` type. Spec-declared empty responses are normalized to `undefined`. Existing Axios output remains unchanged by default.

The implementation follows the existing Fetch client's `includeHttpResponseReturnType` behavior and status-code handling as closely as the Axios response shape permits. The shared HTTP status helpers were moved to core so Axios and Fetch use the same handling for exact, wildcard, and default response statuses without duplicating that logic.

Custom mutators retain control of runtime response handling and must return the documented full Axios response shape.

Validation:
- `bun x vp check` (one pre-existing warning in a generated Angular Query fixture)
- `bun x vp test` (125 files, 4,479 tests)
- `bun x vp run -F './packages/*' typecheck`

This change was implemented with the assistance of an AI coding agent; the author reviewed the full diff and verification results.

## Summary by CodeRabbit

* **New Features**
  * Added an Axios configuration option to include HTTP status codes in generated response types.
  * Generated response types can narrow response data based on status, including empty responses.
  * Added support for status-aware response types with custom Axios mutators.
  * Exact status responses now take precedence over wildcard and default responses.

* **Documentation**
  * Documented the Axios response-type configuration, including custom mutators and empty responses.

* **Tests**
  * Added coverage for status-aware responses, empty responses, dependencies, and default configuration behavior.

PR #4077, opened 2026-09-11
fix(fetch): preserve response status precedence

dpkass:
While implementing the Axios equivalent of Fetch's status-aware response types in https://github.com/orval-labs/orval/pull/4076, we found that Fetch response unions do not preserve OpenAPI response precedence when an exact status and a wildcard overlap.

For example, when both `200` and `2XX` are declared, the generated wildcard member also includes `200`, so `status` cannot reliably discriminate the corresponding data type. This applies the same precedence correction to the existing Fetch implementation by excluding explicitly declared statuses from matching wildcard members.

The change includes one focused regression test for the overlapping exact and wildcard case.

Validation:
- `vp test packages/fetch/src/index.test.ts`
- `vp run -F @orval/fetch typecheck`

Prepared with AI assistance and reviewed against the OpenAPI response precedence rules.

## Bug Fixes

- Improved generated fetch client types for response status handling.
- Wildcard HTTP status types now exclude explicitly defined exact statuses, preventing overlapping or ambiguous response types.
- For example, a `2XX` response type no longer includes status `200` when `200` is defined separately.

### DIFF
--- a/packages/fetch/src/index.test.ts
+++ b/packages/fetch/src/index.test.ts
@@ -627,6 +627,40 @@ describe('generateRequestFunction — zod runtimeValidation response typing (#39
   });
 });
 
+describe('generateRequestFunction — response status precedence', () => {
+  it('excludes exact responses from matching wildcard responses', () => {
+    const response = {
+      definition: { success: 'Pet | void', errors: '' },
+      imports: [],
+      types: {
+        success: [
+          {
+            key: '200',
+            contentType: 'application/json',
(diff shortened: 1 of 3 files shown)


melloware:
Looks like the snapshots are failing?

melloware (changes requested):
Please fix build

dpkass:
sorry, my b. will fix now.

melloware (approved):

PR #4080, opened 2026-09-12
feat(orval): support JSON5 package manifests

### LINKED ISSUE #4071: Support JSON5 for output.packageJson
### Description

Orval's output.packageJson configuration currently appears to expect a standard JSON package.json file.

This makes it difficult to use Orval in projects where the package manifest is authored as JSON5, for example with comments, trailing commas, or other JSON5 syntax.

### Current behavior

output.packageJson expects the package manifest to be valid JSON. JSON5 syntax, such as unquoted property names, single-quoted strings, comments, and trailing commas, is not supported.

As a result, projects using a JSON5 package.json cannot use output.packageJson directly.

### Proposed behavior

output.packageJson should support package manifests written using JSON5 syntax in addition to standard JSON.

When the configured package manifest contains valid JSON5, Orval should parse it successfully and use its contents in the same way as a standard JSON package manifest.

Existing JSON support should remain unchanged.

### Proposed solution

Use the existing json5 package to parse the configured output.packageJson file instead of relying on a strict JSON parser.

This would allow output.packageJson to accept both standard JSON and JSON5 syntax without requiring any changes to the Orval configuration API.

The parsed result can then be handled by Orval in the same way as it currently handles a standard JSON package manifest.

akkupratap323:
Closes #4071

Parse configured `output.packageJson` files with JSON5 so manifests with comments, unquoted keys, and trailing commas work while standard JSON remains supported. Includes a focused regression test.

Validation: focused tests (21 passed), formatting, type-aware lint, package build, and typecheck.

## Summary by CodeRabbit

- **New Features**
  - Added support for loading package manifests written in JSON5 format.
  - Dependency information can now be read from manifests containing comments and trailing commas.
  - Improved support for UTF-8 encoded package manifest files.

- **Bug Fixes**
  - Errors encountered while reading or parsing manifests now include the affected file path and preserve the original error details.

- **Tests**
  - Added coverage for JSON5 parsing, dependency extraction, UTF-8 file support, and path-aware error handling.

### DIFF
--- a/packages/orval/package.json
+++ b/packages/orval/package.json
@@ -79,6 +79,7 @@
     "get-tsconfig": "^4.14.0",
     "jiti": "^2.6.1",
     "js-yaml": "4.3.2",
+    "json5": "^2.2.3",
     "remeda": "^2.33.6",
     "string-argv": "^0.3.2",
     "typedoc": "^0.28.19",
--- a/packages/orval/src/utils/package-json.ts
+++ b/packages/orval/src/utils/package-json.ts
@@ -41,9 +41,15 @@ export const loadPackageJson = async (
 
   const normalizedPath = normalizePath(packageJson, workspace);
   if (fs.existsSync(normalizedPath)) {
-    const pkg = JSON5.parse(
-      await fs.readFile(normalizedPath, 'utf8'),
-    ) as unknown;
+    let pkg: unknown;
+    try {
+      pkg = JSON5.parse(await fs.readFile(normalizedPath, 'utf8')) as unknown;
+    } catch (error) {
+      throw new Error(
+        `Oups... 🍺. Path: ${normalizedPath} => ${String(error)}`,
+        { cause: error },
+      );
+    }
 
     if (isPackageJson(pkg)) {
       return resolveAndAttachVersions(
--- a/packages/orval/src/utils/package-json.ts
+++ b/packages/orval/src/utils/package-json.ts
@@ -9,6 +9,7 @@ import {
 } from '@orval/core';
 import { findUp, findUpMultiple } from 'find-up';
 import fs from 'fs-extra';
+import JSON5 from 'json5';
 import yaml from 'js-yaml';
 
 import { logger } from '../logger';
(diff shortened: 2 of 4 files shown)


melloware (changes requested):
See ai feedback

packages/orval/src/utils/package-json.ts
```diff
@@ -40,7 +41,9 @@ export const loadPackageJson = async (
 
   const normalizedPath = normalizePath(packageJson, workspace);
   if (fs.existsSync(normalizedPath)) {
-    const pkg = await dynamicImport<unknown>(normalizedPath);
+    const pkg = JSON5.parse(
+      await fs.readFile(normalizedPath, 'utf8'),
+    ) as unknown;
```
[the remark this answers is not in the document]
akkupratap323: Fixed in 84cae25d. The explicit manifest read and JSON5 parse now share the path-aware `Oups... Path: ...` error boundary and preserve the original error as `cause`. Added regressions for both malformed JSON5 and read failures. Validation: 23 focused tests passed; formatting and type-aware lint passed.

packages/orval/package.json
```diff
@@ -79,6 +79,7 @@
     "get-tsconfig": "^4.14.0",
     "jiti": "^2.6.1",
     "js-yaml": "4.3.2",
+    "json5": "^2.2.3",
```
melloware, line 82: Let's not use the hat notation let's give it a fixed version like JS-YAML

melloware (changes requested):

melloware (approved):

packages/orval/package.json
akkupratap323: Fixed in 1cdfd8a3: `json5` is now pinned to the exact `2.2.3` version in both `packages/orval/package.json` and `bun.lock`, matching the fixed-version style used by `js-yaml`. The focused package-manifest suite still passes all 23 tests, and formatting is clean.

akkupratap323: Follow-up: #4080 had already merged before the pin commit reached the fork, so that commit was not included in the merge. I moved the requested exact-version change onto current `master` in #4081.

PR #4094, opened 2026-09-13
fix(core): refuse OpenAPI response keys that are not status codes (GHSA-rw75-cc5p-q7c9, GHSA-4j53-7m38-656f)

melloware:
Fixes GHSA-rw75-cc5p-q7c9 and GHSA-4j53-7m38-656f. Both advisories are the same root cause at two sinks, and both suggest the same remedy, so they're fixed together.

## The problem

An OpenAPI Responses Object key reaches generated source in two **unquoted** positions. There is no quote to escape in either, so a non-numeric key is spliced in as live code rather than data:

```ts
status: <key>                 // TypeScript type    — fetch (default on) and axios
response.status === <key>     // runtime expression — axios empty-response normalization
```

- `2 || (globalThis.x = require('child_process').execSync('id'))` → executes when the generated operation's promise resolves.
- `number }; <statement>; type _ = { _z: number` → closes the type object early, so the remainder becomes a top-level statement that a **type-stripping** build (esbuild, Vite, tsx, swc, Babel) runs at **import time**.

The spec validator rejects such keys, but `input.unsafeDisableValidation` turns it off — which is what makes both reachable. Relying on the validator was the gap: the mock generator already guards the structurally identical sink with `assertSafeStatusCode`, **unconditionally**, for exactly this reason. The fetch/axios sinks and the shared helper never got an equivalent.

## The fix

One guard, applied at the emission points:

```ts
export const assertSafeResponseStatusKey = (key: string): string => {
  if (WILDCARD_STATUS_CODE_REGEX.test(key) || EXACT_STATUS_CODE_REGEX.test(key))
    return key;
  throw new Error(`orval: refusing to generate code for an OpenAPI response key that is not a status code (got "${key}"). …`);
};
```

- `getStatusCodeType` routes its fall-through through it. Because fetch and axios both go through that helper, this closes the **fetch type sink and the axios type sink together**.
- The axios `conditionFor` fall-through guards the **runtime condition**, which does not route through `getStatusCodeType`.

`default` is refused at the sink too: it is a valid spec key but not a valid status expression, and every caller already resolves it to `Exclude<HTTPStatusCodes, …>` before reaching a sink. The wildcard branch was already safe (it only ever interpolates `key[0]`, a digit, and regex-filtered exact statuses).

## Verification

- **No generated output changes.** Regenerating the whole `tests/` corpus produces a byte-identical tree — every key that is accepted was already emitted verbatim, so this is purely additive refusal.
- New `http-status.test.ts`: accepts every 3-digit code and `[1-5]XX` wildcard (including lowercase `2xx`); refuses both advisories' exact PoC payloads plus `default`, `abc`, `2000`, `20X`, `600`, `099`, `''`, `x-vendor`; and asserts the message names the offending key.
- End-to-end refusal tests at each sink: `@orval/axios` (`response.status === <key>`, using the advisory's `child_process` payload) and `@orval/fetch` (`status: <key>`, using the type-breakout payload).
- **Confirmed all three fail without the guard** — stashing it makes the core suite, the axios test, and the fetch test go red; all green once restored.
- Suites: core 2673, axios 28, fetch 22. Full `pr-checks` run locally: fmt, lint, lint:samples, lint:snapshots, typecheck, test:snapshots, `orval-tests build`.

> [!NOTE]
> `@orval/query` currently has 2 failing tests on `master` (`includeZodSchemaInArguments with a custom mutator`). I confirmed those are pre-existing by stashing this branch and running them on a clean tree — unrelated to this change, but CI will be red until they're addressed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of OpenAPI response status keys during Axios and Fetch client generation.
  * Invalid, non-standard, or potentially unsafe status keys are now rejected with a clear error instead of being included in generated code.
  * Valid exact status codes and wildcard ranges continue to be supported.
* **Security**
  * Prevented specially crafted response keys from being interpreted as executable code when specification validation is bypassed.


melloware:
Rebased onto `master` (551b940cb, post-#4090) and re-ran everything.

**Correction to the note I left earlier.** I'd said the two `@orval/query` failures (`includeZodSchemaInArguments with a custom mutator`) were pre-existing on master. That was true when I checked, but master has since gone green — and on the rebase they still failed, so they *were* being caused by this change. I was wrong to leave that note standing.

The cause was a fixture gap, not the guard being wrong:

```
Error: orval: refusing to generate code for an OpenAPI response key that is not
a status code (got "undefined").
  at getStatusCodeType (core/src/utils/http-status.ts:55)
  at generateRequestFunction (fetch/src/index.ts:486)
```

`ResReqTypesValue.key` is required and `getResReqTypes` always sets it, but that fixture reaches `GeneratorVerbOptions` through an `as unknown` cast, so omitting it type-checked. Before this PR the missing key was emitted as the literal `status: undefined`; now it is refused.

I checked whether a keyless entry can occur in production before touching the fixture — it cannot. Every entry comes from `getResReqTypes` (which sets `key` at each construction site) or from the synthesized fallback in `fetch/src/index.ts`, which sets `key: 'default'`. So the fixture now sets `key: '200'` rather than the guard being loosened.

Full run on the rebased branch: all package suites green (query back to 231/231), snapshots green, fmt / lint / lint:samples / lint:snapshots / typecheck / `orval-tests build` all clean, and regenerating the `tests/` corpus still produces no drift.

PR #4104, opened 2026-09-14
fix(mock): emit Blob for binary faker mocks (#4062)

### LINKED ISSUE #4062: Faker mock is annotated `Blob` but returns `ArrayBuffer`
### Description

The generated faker mock for a `format: binary` schema is annotated with `Blob`, but returns an `ArrayBuffer`, leading to type errors. Details below.

Note that a variant of this bug was already reported in #3590 (Bug 3) and fixed in #3607, but that was only for "strict" mocks (`required`+`nonNullable`). In the reproduction of the current bug, `required` and `nonNullable` are both unset.

### Output client

fetch

```ts

export default defineConfig({
    repro: {
        input: { target: './spec.json' },
        output: {
            mode: 'tags-split',
            target: 'gen',
            schemas: 'gen/model',
            client: 'fetch',
            mock: { generators: [{ type: 'faker', schemas: true }] }
        }
    }
});
```

### Environment

  System:
    OS: Windows 11 10.0.26200
    CPU: (32) x64 Intel(R) Core(TM) i9-14900K
    Memory: 39.39 GB / 63.77 GB
  npmPackages:
    orval: ^8.30.0 => 8.30.0

### Expected behavior

In my opinion, `Blob` is the correct type for a multipart form file, so the value of the mock should be a `Blob` as well. For example:
```cs
new Blob([new Uint8Array(faker.number.int({ min: 1, max: 64 }))])
```

Alternatively, it could stay `ArrayBuffer` for consistency with #3607, but then its type should be annotated as such.

### Actual behavior

The following code is generated:
```cs
// in gen/model/iFormFile.ts
export type IFormFile = Blob;

// in gen/model/index.faker.ts
export const getIFormFileMock = (): IFormFile => (new ArrayBuffer(faker.number.int({ min: 1, max: 64 })))
```

...leading to the following error:
```
error TS2740: Type 'ArrayBuffer' is missing the following properties from type 'Blob': size, type, arrayBuffer, bytes, and 2 more.
```

```json
{
  "openapi": "3.0.1",
  "info": { "title": "repro", "version": "1.0" },
  "paths": {},
  "components": {
    "schemas": { "IFormFile": { "type": "string", "format": "binary" } }
  }
}
```

### Additional context

The expected behavior can be forced by this workaround in orval.config.ts:
```ts
override: {
    mock: {
        format: {
            binary: () => new Blob([new Uint8Array(faker.number.int({ min: 1, max: 64 }))])
        }
    }
}
```

melloware:
Fixes #4062

### Problem

`format: binary` generates `Blob` in the model, but the faker mock emitted an `ArrayBuffer`:

```ts
export type IFormFile = Blob;
export const getIFormFileMock = (): IFormFile =>
  new ArrayBuffer(faker.number.int({ min: 1, max: 64 }));
// error TS2740: Type 'ArrayBuffer' is missing the following properties from type 'Blob'
```

#3607 handled this for strict mocks (`required` + `nonNullable`) by declaring `export type PhotoUploadMock = ArrayBuffer`, but that alias does not exist when the strict options are unset — the reporter's case. It also cannot help a binary **property** inside an object schema, in strict mode either, since the property type comes from the model:

```ts
export interface Upload { file?: Blob }
export const getUploadMock = (): Upload => ({ file: new ArrayBuffer(...) })
// error TS2322: Type 'ArrayBuffer | undefined' is not assignable to type 'Blob | undefined'
```

So the annotation cannot be patched around it — the emitted value is what has to change.

### Fix

Emit a `Blob`, matching the type orval already generates for binary:

- `DEFAULT_FORMAT_MOCK.binary` → `new Blob([new Uint8Array(faker.number.int({ min: 1, max: 64 }))])`
- the strict `{Schema}Mock` alias for binary schemas → `Blob`
- msw handlers keep their declared `Blob` return type (the `Blob`/ref-name → `ArrayBuffer` rewrite is gone) and pass the body straight to `new HttpResponse(binaryBody, …)` instead of `HttpResponse.arrayBuffer(binaryBody instanceof ArrayBuffer ? … : new ArrayBuffer(0))`

A `Blob` is a valid `BodyInit`, so the handler no longer needs the conversion guard — and since `ArrayBuffer` is a valid `BodyInit` too, an override that still returns one keeps working at runtime.

```ts
export const getGetBinaryBlobResponseMock = (): Blob =>
  new Blob([new Uint8Array(faker.number.int({ min: 1, max: 64 }))]);

return new HttpResponse(binaryBody, {
  status: 200,
  headers: { 'Content-Type': 'application/octet-stream' },
});
```

### Behavior change

Binary mocks (faker values and msw handler bodies/override types) are now `Blob` rather than `ArrayBuffer`. Anyone asserting `instanceof ArrayBuffer` on a generated binary mock, or passing an `ArrayBuffer` override to an msw handler, will need to switch to `Blob` — runtime still accepts both, only the override type narrows.

### Verification

- New `tests/specifications/issue-4062.yaml` fixture covering the reported alias, a nested binary property, a `$ref` to a binary schema, and a binary response — generated **without** `override.mock.required`/`nonNullable`, i.e. the reporter's config. On `master` it fails `typecheck-generated` with the exact TS2740 from the issue plus TS2322 on the property; with the fix all 20 typecheck runs pass.
- `packages/mock`: 397 tests pass (the tests that pinned the old ArrayBuffer output were updated to the new expectation).
- Full workspace: all package tests (14 suites), `typecheck`, and `lint` clean; `tests` snapshots (7001) and all sample tests pass.
- Regenerated `tests/__snapshots__` and the `samples/**` msw output + snapshots.

## Summary by CodeRabbit

* **Updated Behavior**
  * Binary response mocks now use `Blob` values instead of `ArrayBuffer`.
  * Mock download handlers accept and return binary `Blob` content directly, preserving response data and declared types.

* **Bug Fixes**
  * Binary and file-upload mocks now generate responses consistently across supported client examples.

* **Tests**
  * Added regression coverage for binary uploads and downloads, including multipart requests and octet-stream responses.


melloware:
@Hypenate to review you can even try the package preview version in your codebase if you want to.

PR #4105, opened 2026-09-14
fix: deep nested allOf in FormData doesnt generate JSON.stringify

### LINKED ISSUE #4107: Double nested $ref -> allOf in FormData leads to missin JSON.stringify
### Description

I am using [utoipa](https://github.com/juhaku/utoipa) to generate OAPI spec. 
Scenario
```rust
struct A {
  #[serde(flatten)]
  b: B
}

struct B {
  #[serde(flatten)]
  c: C
}

struct C {
  title: String
}
```
leads to A = allOf ($ref B) -> B = allOf ($ref C) -> C
Using A in MultiPartFormData creates a `formData.append(`a`, createARequest.a)` missing JSON.stringify which should be `formData.append(`a`, JSON.stringify(createARequest.a))`

### Output client

other

irrelevant -> problem in core

### Environment

irrelevant -> problem in core

### Expected behavior

use JSON.stringify

### Actual behavior

doesnt use JSON.stringify

### Additional context

I left a lot of required fields empty as it is not about the environment or orval config, just about OAPI Spec parsing. If still required i will add the info later

NiklasVonMinckwitz:
Code produced by Claude, reviewed by me.
References Issue #4107 
# Summary
A double nested allOf ($ref A -> allOf -> $ref B -> allOf C) in FormData doesn't create a JSON.stringify call on A.

# Details
Neither `A` nor `B` carries an explicit `type: object` of its own - that only appears on the leaf schemas being composed. The formData codegen currently decides whether to JSON.stringify a property via a direct `property.type === 'object'` check  res-req-types.ts, around the `else if (property.type === 'object' ...)` branch) after a single `resolveSchemaRef` dereference. That check does not merge through `allOf`, so it only happens to pass today because the dereferenced schema underneath a *single* level of $ref usually still exposes an explicit `type: object` (e.g. Rusts utoipa emits that on a flattened struct's own wrapper schema). Once a struct is nested one level deeper behind another pure `allOf: [$ref]` wrapper with no `type` field of its own, this check silently fails and the field is appended raw instead of JSON.stringified, breaking the request at runtime with no build-time signal.

## Bug Fixes

* Improved multipart form-data handling for schemas composed through nested definitions.
* Preserved raw serialization for scalar values represented through schema composition, including array items.
* Correctly serializes inherited object properties and applies required/optional rules across nested compositions.
* Improved handling of nested objects without an explicitly declared object type.
* Prevented recursive schema definitions from causing processing failures.


melloware (changes requested):
See AI feedback

melloware (approved):

PR #4110, opened 2026-09-14
refactor(core): drop unused debug dependency and createDebugger

wadakatu:
## Summary

`createDebugger` in `packages/core/src/utils/debug.ts` has no callers anywhere in the repo, so this removes:

- `packages/core/src/utils/debug.ts` and its barrel re-export in `utils/index.ts`
- the `debug` runtime dependency and `@types/debug` dev dependency of `@orval/core`
- the matching `bun.lock` entries (`debug` itself stays in the lockfile as a transitive dependency of tooling)

No behavior change. Follow-up to #4044; same precedent for removing unused `@orval/core` exports (#3840).

## Verification

- `bun install` regenerates the lockfile cleanly
- `bun run build`, `bun run lint` (type-aware) and `bun run test` pass across all packages

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed the previously available debug logging utility from the public utilities surface.
  * Debug-related runtime and development packages are no longer included.

## Context for reviewers

`createDebugger` was copied from Vite in 2021 with a single caller: timing logs in `file.ts` config loading (`DEBUG=orval:file-load`). That caller was removed in the ESM conversion (#2460, 2025-10), so the helper has been dead since. `ORVAL_DEBUG_FILTER` / `DEBUG=orval:*` are not documented anywhere (docs, README, CONTRIBUTING). If you'd rather keep a developer debug path, the better option is probably to wire it back into `file.ts` instead of keeping an uncalled helper — happy to go either way.


wadakatu:
Re Copilot's note on removing a public export: same situation as #4044 — `@orval/core`'s `utils` re-export has been treated as internal surface before (#3840 removed `getEnumNames` / `getEnumDescriptions` / `getEnumUnionFromSchema` in a minor release, v8.25.0). No callers in-repo and the env vars were never documented. Leaving the call to maintainers; see the "Context for reviewers" section above.

melloware:
We should remove anything that reduces the surface. If it's not used internally we don't care if someone is using it externally and rip it out!

melloware (approved):

melloware:
@wadakatu you can submit a follow up PR to finish the clean up if you would like!

Issue #4120, opened 2026-09-15
Consider opting into @scalar/json-magic SSRF hardening for external $ref resolution


melloware:
Found while reviewing the `@scalar/openapi-parser` 0.29.1 bump (#4114). Not a vulnerability report — an opt-in hardening option we are currently not taking.

## What changed upstream

`@scalar/json-magic@0.13.4` ([scalar/scalar#10079](https://github.com/scalar/scalar/pull/10079)) hardened external `$ref` resolution against SSRF and local file disclosure, and exposed it as **opt-in** options on the two bundling plugins:

- `fetchUrls` gains `blockPrivateNetworks` — refuses private, loopback, link-local and cloud metadata addresses
- `readFiles` gains `basePath` — confines local file reads to the document's directory

From the changelog: *"so other callers keep their current behavior unless they opt in."* We bumped to 0.13.4 in #4114 and did **not** opt in, so nothing changed.

## Why it is worth a look

`packages/orval/src/import-specs.ts` imports both plugins from `@scalar/json-magic/bundle/plugins/node` and already has its own external-`$ref` defences: `scanRefs`, an `isAllowedRef` allow-list, and loader wrappers that enforce the list on transitive refs. The upstream options overlap with that layer, so the questions are:

1. Does `blockPrivateNetworks` add anything our allow-list does not already cover? The allow-list is opt-in per user config, and `externalRefs.allow: ['*']` disables it entirely apart from a warning — a wildcard user currently has no protection against a spec that `$ref`s `http://169.254.169.254/...`.
2. Is `basePath` compatible with how orval resolves relative refs? Specs that legitimately reference a sibling directory (`../shared/refs.yaml`) would break if it is enabled naively, so this needs a look at the real corpus before switching on.
3. If we adopt either, should it be unconditional or a `parserOptions` escape hatch?

## Scope

- [ ] Decide whether `blockPrivateNetworks` should be on by default, particularly for the `externalRefs.allow: ['*']` path
- [ ] Assess `basePath` against the existing relative-`$ref` tests before enabling
- [ ] Tests covering whichever behaviour is chosen

melloware:
Decided against this.

Issue #4115, opened 2026-09-15
fix(orval): make resolveSpec emit a fully OAS 3.1-shaped document


melloware:
Sub-issue of #4113. **Depends on the Scalar bump landing first.**

## Goal

After the Scalar upgrade closes the nullable-`$ref` and nullable-`allOf` cases upstream, `resolveSpec` still lets residual OAS 3.0 syntax through. Close the remainder so the document handed to `importOpenApi` is fully 3.1-shaped.

## Remaining gaps

Scalar's upgrader bails when `nullable: true` has no sibling `type`, `$ref`, or `allOf` — the comment in its source says so outright: *"Otherwise there is nothing for `nullable` to attach to, so leave the schema untouched."* That leaves:

| Residual 3.0 input | Required 3.1 output |
| --- | --- |
| `{ enum: [...], nullable: true }` (no `type`) | `{ enum: [..., null] }` — `null` must be in the enum |
| `{ anyOf: [...], nullable: true }` | same combinator plus a `{ type: 'null' }` branch |
| `{ oneOf: [...], nullable: true }` | same combinator plus a `{ type: 'null' }` branch |
| `{ type: T, enum: [...], nullable: true }` | `{ type: [T, 'null'], enum: [..., null] }` |
| binary type unions not covered by the upgrader | `contentMediaType: 'application/octet-stream'` |
| `$ref` / `$dynamicRef` with invalid siblings | siblings hoisted out of the reference |

A live example of the untyped-enum case is already in the corpus: `tests/specifications/null-type.yaml`, schema `NullEnum` (`nullable: true` + `enum: [foo, null]`, no `type`).

Note `{ allOf: [...], nullable: true }` must become `{ oneOf: [{ allOf: [...] }, { type: 'null' }] }` — never append `{ type: 'null' }` to the `allOf` array, which would be unsatisfiable under JSON Schema AND semantics.

## Scope

- A normalization pass in `packages/orval/src/import-specs.ts` (post-`upgrade()`) covering the table above
- Unit tests asserting the boundary output contains **no** `nullable`, no boolean `exclusiveMinimum`/`exclusiveMaximum`, and no `format: 'binary'`
- Ideally a guard test that walks the post-`resolveSpec` document and fails on any residual 3.0 keyword

## Explicitly not `nullable` for these purposes

Internal support fields that happen to share the name — `combineSchemas({ nullable: '' })` and `GetObjectOptions.nullable`. Leave them alone.

## Out of scope

[...]

Issue #4113, opened 2026-09-15
Stop using OAS 3.0 in tests


snebjorn:
**⚠️ THIS IS A TRACKING ISSUE - DO NOT OPEN PRS AGAINST THIS ⚠️**

Choose sub-issues instead.

## Summary

`resolveSpec` is the last place OAS 3.0 (OpenAPI Specification - Version 3.0) is allowed. After it upgrades the document, every generator and every unit test that injects a schema should speak OAS 3.1.

Today that invariant is leaky in two places:

1. **Production.** Scalar upgrades typed `nullable`, boolean exclusivity, and most `format: binary` cases, but it leaves residual 3.0 forms untouched (`nullable` on `allOf`/`anyOf`/`oneOf`, invalid `$ref`/`$dynamicRef` siblings, untyped nullable schemas, and some binary type unions). Downstream code then has to handle both dialects.
2. **Tests.** Many suites under `packages/` construct schemas _after_ `resolveSpec` and still use OAS 3.0 keywords (`nullable: true`, `format: 'binary'`, boolean `exclusiveMinimum`/`exclusiveMaximum`, `openapi: '3.0.x'`). Those fixtures do not match what generators actually receive.

**This work has two main goals:**

- Make `resolveSpec` emit a fully OAS 3.1-shaped document.
- Rewrite every post-boundary test fixture to that same shape, without confusing OAS `nullable` with internal helpers such as `GetObjectOptions.nullable`.

OAS 3.0 and Swagger stay valid **only** as input to the import/upgrade pipeline (`import-specs` tests and full-pipeline `tests/specifications`).

**Sub goals:**

- Fix types used in tests. Remove `as unknown ` from `as unknown as <OpenApiType>`. As this is just a masked `any`.

## Problem

Program flow:

```text
orval.ts → generateSpec → importSpecs → resolveSpec (upgrade + normalize)
         → importOpenApi → core getters/generators/resolvers
         → client / Zod / Effect / mock → writeSpecs
```

Anything past `resolveSpec` should never see 3.0 schema syntax.

| 3.0 syntax                                       | Post-upgrade 3.1 equivalent                                                                      |
| ------------------------------------------------ | ------------------------------------------------------------------------------------------------ |
| `{ type: T, nullable: true }`                    | `{ type: [T, 'null'] }`                                                                          |
| `{ $ref, nullable: true }` (invalid even in 3.0) | `{ oneOf: [$ref, { type: 'null' }] }`                                                            |
[...]

PR #4121, opened 2026-09-15
chore(deps): upgrade @scalar/openapi-parser to 0.29.1

### LINKED ISSUE #4114: chore(deps): upgrade @scalar/openapi-parser to 0.29.x
Sub-issue of #4113.

## Why this goes first

`@scalar/openapi-upgrader@0.2.13` ([scalar#9827](https://github.com/scalar/scalar/pull/9827)) added the nullable-`$ref` / nullable-`allOf` upgrade that #4113 lists as a gap. We do not have it: `bun.lock` resolves `@scalar/openapi-parser@0.28.11` → `@scalar/openapi-upgrader@0.2.12`.

`0.2.12` only handles the typed case:

```js
if (schema.type !== undefined && schema.nullable === true) {
    schema.type = [schema.type, 'null'];
    delete schema.nullable;
}
```

`0.2.15` (via parser `0.29.1`) adds:

```js
else if (schema.nullable === true && schema.type === undefined) {
    if (typeof schema.$ref === 'string')  return { ...rest, anyOf: [{ $ref }, { type: 'null' }] };
    if (Array.isArray(schema.allOf))      return { ...rest, anyOf: [base, { type: 'null' }] };
}
```

That closes two rows of the conversion table upstream, so it has to land before we write any new normalization or rewrite any fixtures — it changes the shape everything else is written against.

## Scope

Bump `@scalar/openapi-parser` `^0.28.11` → `^0.29.1` in `packages/orval/package.json` and deal with the fallout:

1. **`normalizeNullableRefs` becomes mostly dead code.** It lives in `packages/orval/src/import-specs.ts` and runs *after* `upgrade()`. Scalar now strips `nullable` first, so the function no longer fires — and neither does its "your spec is non-conformant" warning from #3714. Decide: retire it, or keep a pre-`upgrade()` pass purely to preserve the warning. Either way its unit tests need updating.
2. **Snapshot churn in `tests/specifications`.** `{ allOf: [$ref], nullable: true }` now upgrades to `anyOf: [$ref, { type: 'null' }]`. See `tests/specifications/null-type.yaml` (`NullableRefContainer`) — review every diff rather than blanket-accepting.
3. **New validator.** `0.29.0` moves validation into `@scalar/openapi-validator`. Confirm no change in which user specs are accepted/rejected; this is the main user-visible risk in the bump.

## Type-level changes — already checked, no action needed

- `ErrorObject.path` is now `string | string[]`. `import-specs.ts` only `JSON.stringify`s the errors, so nothing narrows it.
- `AnyObject` / `UnknownObject` / `ValidationOutcome` / the internal `OpenApiDocument` alias are no longer exported. Nothing under `packages/` imports them.

## Acceptance criteria

- [ ] `@scalar/openapi-parser` at `^0.29.1`, lockfile updated
- [ ] `normalizeNullableRefs` resolved (retired or repositioned) with tests reflecting the decision
- [ ] Every `tests/specifications` snapshot diff reviewed and explained in the PR description
- [ ] Validation behaviour confirmed unchanged for the existing spec corpus

## Out of scope

Residual 3.0 normalization (#4113 gap 1 leftovers) and any test-fixture rewrites.

melloware:
Closes #4114. First sub-issue of #4113.

## What

Bumps `@scalar/openapi-parser` `0.28.11` → `0.29.1`, which brings `@scalar/openapi-upgrader` `0.2.12` → `0.2.15`. Also moves `@scalar/json-magic` to `^0.13.4` and the `@scalar/openapi-types` catalog entry to `0.9.5`, so the workspace and the parser resolve the same copies instead of duplicating them.

Upgrader `0.2.13` ([scalar/scalar#9827](https://github.com/scalar/scalar/pull/9827)) rewrites nullable `$ref` and nullable `allOf` into the 3.1 null-union form, so two rows of the #4113 conversion table are now handled upstream. That has three consequences, all handled here.

## `normalizeNullableRefs` moves ahead of `upgrade()`

It ran *after* the upgrade. Now that the upgrader strips `nullable` itself, the function was never reached in the pipeline and its "your spec is non-conformant" warning (#3714) would have silently stopped firing — while its unit tests kept passing, because they call the exported function directly.

Moving it ahead of `upgrade()` keeps the diagnostic and leaves the upgrader nothing to do for that case. It also makes the function's own doc comment true again: it already claimed to run before `upgrade()`.

## The `allOf` guard the upgrader does not have

The upgrader applies its rewrite with no `allOf`-context guard, which orval deliberately does:

```js
// @scalar/openapi-upgrader 0.2.15
else if (schema.nullable === true && schema.type === undefined) {
    if (typeof schema.$ref === 'string')  return { ...rest, anyOf: [{ $ref }, { type: 'null' }] };
```

So a `{ $ref, nullable: true }` member *inside* an `allOf` became `{ anyOf: [$ref, { type: 'null' }] }`. Orval reads through `allOf` members to collect the keys a schema guarantees, and a union member hides them:

```diff
- Required<Pick<NullableAllOfMemberWrapper, 'id'>>
+ Required<Pick<NullableAllOfMemberWrapper, Extract<keyof NullableAllOfMemberWrapper, 'id'>>>
```

which is exactly the regression `api-generation.spec.ts` pins with `.not.toContain('Extract<')`.

Per the Reference Object rules a sibling on a `$ref` is ignored, so that `nullable` was always a no-op. The pre-upgrade pass now drops it and hands the upgrader a plain `$ref`, restoring the `Pick` form.

## Snapshots

21 re-recorded, in three reviewed groups. No other snapshot moved.

**18 zod files** — `.object({...}).nullish()` → `.union([zod.object({...}), zod.null()]).optional()`. Both describe `T | null | undefined`.

```diff
  position: zod
-   .object({ id: zod.string(), quantity: zod.number() })
-   .nullish(),
+   .union([zod.object({ id: zod.string(), quantity: zod.number() }), zod.null()])
+   .optional(),
```

**2 mock/faker files** — optional `petId` now also generates `undefined`, which matches a property absent from `required`.

**2 regression models** — `(Base | null) & { marker?: string }` → `Base & { marker?: string }`. These are the same type: `null & { ... }` reduces to `never`, so the null branch was already inert. Confirmed with `tsc --strict` on a mutual-assignability check.

## Verification

- `typecheck` clean
- 4,650 unit tests pass (one added, for the new `allOf` branch)
- 7,011 snapshot tests pass
- `test:cli`: all 20 generated-API typecheck runs pass, generated-mock verification passes
- 493 regenerated sample files are byte-identical — the bump changes nothing outside the three groups above

## Type-level changes in 0.29.0 — no action needed

- `ErrorObject.path` is now `string | string[]`. It is only handed to `JSON.stringify` in `import-specs.ts`, so nothing narrows it.
- The removed `AnyObject` / `UnknownObject` / `ValidationOutcome` exports are not imported anywhere in the workspace.

## Notes for reviewers

`0.29.0` moves validation into a new `@scalar/openapi-validator` package. The full spec corpus still validates and generates identically, but that is the part of this bump with user-visible risk if a spec in the wild validates differently.

New transitive dependencies: `@scalar/openapi-validator` and `@scalar/types@0.19.0`, the latter pulling in `nanoid`, `type-fest`, and its own `zod@4.6.2`.

Two findings that are out of scope here are written up separately: the upgrader's lossy handling of nullable typed enums (#4115) and the new opt-in SSRF hardening in `@scalar/json-magic` (#4120).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved OpenAPI imports involving nullable references within `allOf` definitions.
  - Correctly handles ineffective `nullable` annotations on direct `allOf` references, avoiding incorrect nullable unions.
  - Preserves nullable conversions for nested references and supported contexts, with appropriate diagnostic warnings.
  - Corrected generated type nullability based on where `nullable` appears around `allOf` compositions.

- **Tests**
  - Added regression coverage for nullable reference placement and import behavior.


packages/orval/src/import-specs.ts
```diff
@@ -194,10 +203,29 @@ export function normalizeNullableRefs(
...
+  // `Base & { marker?: string }` are the same type, since `null & { ... }`
+  // reduces to `never`.
+  if (
+    inAllOf &&
+    '$ref' in obj &&
+    isString(obj.$ref) &&
+    (obj.nullable as boolean | undefined) === true
+  ) {
+    delete obj.nullable;
+  }
```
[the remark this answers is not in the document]
melloware: Valid — confirmed with a failing test before changing anything. `{ allOf: [{ type: 'object', properties: { pet: { $ref, nullable: true } } }] }` took the exemption branch and came out as a plain `$ref`, dropping the `| null`.

Root cause is the flag, not the branch: the recursion passed `inAllOf || key === 'allOf'`, so once inside an `allOf` every descendant inherited it. Rather than adding a depth check at the branch, `inAllOf` is now non-sticky — set when descending into `allOf`, carried to each member by the array branch, and cleared for every other key:

```ts
-      inAllOf || key === 'allOf',
+    obj[key] = normalizeNullableRefs(value, [...path, key], key === 'allOf');
```

That makes the flag mean what its name says, and it also fixes the mirror-image case that predates this PR: a nullable `$ref` nested below an `allOf` member used to skip the rewrite entirely and leak `nullable: true` past the boundary, which is exactly what #4113 is trying to eliminate.

Regression test added for the nested case. All 7011 snapshot tests still pass with no re-recording, so nothing in the corpus depended on the sticky behaviour.

Fixed in d8d467f51.

packages/orval/src/import-specs.test.ts
```diff
@@ -3354,6 +3354,27 @@ describe('normalizeNullableRefs', () => {
...
+    // Left as a plain `$ref` rather than rewritten into a union: orval reads
+    // through allOf members to collect the keys a schema guarantees, and a union
+    // member hides them (#3714).
+    expect(result).toEqual({
+      allOf: [
+        { $ref: '#/components/schemas/Pet' },
+        { type: 'object', properties: { marker: { type: 'string' } } },
+      ],
+    });
+  });
```
snebjorn: This is not right. Even though `{ $ref: '#/components/schemas/Pet', nullable: true }` is totally out of spec and straight up has no meaning according to OAS 3.0.
What the user tried to express was `Pet | null`.

So the result should probably be:
```
      allOf: [
        { oneOf: [{ $ref: '#/components/schemas/Pet' }, { type: 'null' }] },
        { type: 'object', properties: { marker: { type: 'string' } } },
      ],
```
But what does that even translate into in typescript? Something can't be `null & { marker: string }`
What do we do today with that input?

snebjorn (changes requested):

packages/orval/src/import-specs.test.ts
melloware: Fair challenge, and the answer turns out to be the point your last line is already circling.

## What we do today

I generated all three shapes to check rather than guess. On `master` (before this PR), your case emits:

```ts
export type NullableAllOfMemberWrapper = (NullableAllOfMemberBase | null) & {
  marker?: string;
};
```

So today we *do* try to preserve the intent. But that type does not mean what it looks like: `&` distributes over `|`, and `null & { marker?: string }` is `never`, so TypeScript reduces the whole thing to `NullableAllOfMemberBase & { marker?: string }`. I confirmed it with `tsc --strict` on a mutual-assignability check — the two are the same type, in both directions.

The `| null` was already being discarded. This PR just stops emitting it.

## Which is why `oneOf` there does not help either

Your suggested shape has the same problem:

```yaml
allOf:
  - oneOf: [{ $ref: Pet }, { type: 'null' }]
  - type: object
    properties: { marker: { type: string } }
```

That is an intersection whose first member may be `null`, so the null branch dies the same way. It also costs us something real: orval reads through `allOf` members to collect the keys a schema guarantees, and a union member hides them, degrading `Pick<Wrapper, 'id'>` to `Pick<Wrapper, Extract<keyof Wrapper, 'id'>>`. There is a regression test pinning that (`api-generation.spec.ts`, `.not.toContain('Extract<')`).

You cannot express "nullable" as a *member* of an `allOf`. The position is semantically void.

## Where it does work

Nullability has to sit on the composed schema, not inside it. Generated from the current branch:

```yaml
b_nullableOnWrapper:
  allOf:
    - $ref: '#/components/schemas/Base'
    - type: object
      properties: { marker: { type: string } }
  nullable: true          # on the wrapper, not on the member
```

```ts
export type HolderBNullableOnWrapper = Base & { marker?: string } | null;
```

That one survives, and the upgrader converts it to the canonical `anyOf: [{allOf: [...]}, {type: 'null'}]` for us.

## So: two options

1. **What this PR does.** Treat the sibling as the no-op the Reference Object rules say it is. Output is provably identical to today, and key collection keeps working. The user's `nullable` is dropped — but it was already being dropped by TypeScript.
2. **Honour the intent by hoisting.** Read `allOf: [{$ref: Pet, nullable: true}, {marker}]` as `(Pet & {marker}) | null` — lift the null out to the whole intersection, which is exactly what the wrapper form above already produces. That is a real behaviour change, and it needs a rule for what happens when two members are nullable, so I would not fold it into a dependency bump.

I lean towards 1 here and 2 as its own issue under #4113 if we want it. If you would rather see the emitted types pinned end-to-end instead of implied by a unit test, I am happy to add all three shapes to `regressions.yaml` so the difference is visible in the snapshots — say the word.

melloware: Added in f12255d58 — the placements are now pinned end-to-end in `regressions.yaml` instead of implied by a unit test, so the difference shows up in the snapshots.

The three that were missing, with the types they generate:

```yaml
NullablePlacementOnWrapper:        # nullable on the composed schema
  allOf: [$ref Base, {marker}]
  nullable: true
```
```ts
export type NullablePlacementOnWrapper =
  | (NullablePlacementBase & { marker?: string })
  | null;                                          // survives
```

```yaml
NullablePlacementHolder:           # bare $ref sibling, no composition
  properties:
    bare_ref: { $ref: Base, nullable: true }
```
```ts
bare_ref?: NullablePlacementBase | null;           // union + warning
```

```yaml
NullablePlacementBelowMember:      # below an allOf member, not on one
  allOf:
    - properties:
        nested_ref: { $ref: Base, nullable: true }
```
```ts
nested_ref?: NullablePlacementBase | null;         // union
```

The fourth placement — a sibling *on* an `allOf` member — is the existing `NullableAllOfMemberWrapper`. The new test asserts all four together, including that the member case emits no `| null`, with the reasoning in a comment so the next reader does not have to rediscover why that one differs.

That makes your original question answerable straight from the snapshots: nullability survives on the composed schema and at a plain property, and dies as an `allOf` member, because the member position is an intersection.

No existing snapshot moved. 7016 snapshot tests, all 20 generated-API typecheck runs, typecheck and lint all pass.

snebjorn: option 1, as this is what is outputted today

melloware: Option 1 it is — that is what the branch already does, so nothing changes.

One nuance so the diff does not surprise anyone at merge: the emitted *type* is what stays the same, not the emitted *text*. Two snapshots do move, because the inert branch stops being printed:

```diff
- export type NullableAllOfMemberWrapper = (NullableAllOfMemberBase | null) & {
+ export type NullableAllOfMemberWrapper = NullableAllOfMemberBase & {
    marker?: string;
  };
```

Same type to the compiler, in both directions — that is the `tsc --strict` check above. But the text is different, so those two files show up in the review.

The four placements are pinned in `regressions.yaml` now, with the reasoning in the test comment, so the next person who reads `NullableAllOfMemberWrapper` and wonders where the `| null` went has the answer next to it rather than in this thread.

packages/orval/src/import-specs.test.ts
```diff
@@ -3354,6 +3354,57 @@ describe('normalizeNullableRefs', () => {
     expect(result).toEqual({ type: 'string', nullable: true });
   });
 
+  it('should still rewrite a nullable $ref nested below an allOf member', () => {
```
snebjorn, line 3357: this seems to contradict the 'should drop a no-op nullable sibling from a $ref inside allOf' test.
I think "no-op" needs to be clarified so it doesn't read as a contradiction.

packages/orval/src/import-specs.test.ts
melloware: Yep that is what this PR does Option 1

tests/__snapshots__/default/all-of-primitive/endpoints.faker.ts
```diff
@@ -15,10 +15,13 @@ export const getGetPetsResponseMock = (): Pets[] =>
...
+        faker.helpers.arrayElement([
+          faker.string.alpha({ length: { min: 10, max: 20 } }),
+          null,
+        ]),
         faker.string.alpha({ length: { min: 10, max: 20 } }),
         null,
       ]),
-      faker.string.alpha({ length: { min: 10, max: 20 } }),
-      null,
+      undefined,
```
snebjorn, line 25: is undefined correct here?

tests/__snapshots__/fetch/zod-schema-response-suffix-single-no-runtime-validation/model/nullablePositionWrapperSchema.zod.ts
```diff
@@ -8,11 +8,14 @@ import * as zod from 'zod';
...
-    })
-    .nullish(),
+    .union([
+      zod.object({
+        id: zod.string(),
+        quantity: zod.number(),
+      }),
+      zod.null(),
+    ])
+    .optional(),
```
snebjorn, line 18: This change seems significant, though it looks like it expresses the same.
Are we sure this is correct?

snebjorn (changes requested):

packages/orval/src/import-specs.test.ts
melloware: Fair — "no-op" was doing two jobs and only one of them was true.

The sibling is out of spec in **both** positions; the Reference Object rules ignore it wherever it appears. So "no-op" cannot be what separates the two tests, and read together they look like we decided it means something in one place and nothing in the other.

What actually separates them is whether the union can *survive* where it sits:

| position | rewritten? | why |
|---|---|---|
| direct `allOf` member | no — dropped | members are intersected, and `null & { marker?: string }` is `never`, so the branch is inert |
| anywhere else, including *below* a member | yes — `anyOf` + warning | no intersection, so the `| null` reaches the emitted type |

The member case is not dropped because the sibling means less there. It is dropped because emitting the union changes no type and *costs* something: a union member hides the keys orval collects by reading through `allOf`, degrading `Pick<W, 'id'>` to `Pick<W, Extract<keyof W, 'id'>>`.

Reworded in e8b0550c7. The test is now `should drop a nullable sibling from a $ref that is a direct allOf member`, and the same framing replaces "equally meaningless" / "no-op" in the two branch comments and the function doc, which had the identical problem. Both comments now also point at where a nullable composition *should* go — `nullable` on the composed schema, pinned as `NullablePlacementOnWrapper` in `regressions.yaml`.

Comments and a test name only. 387 orval tests, lint and typecheck pass.

tests/__snapshots__/default/all-of-primitive/endpoints.faker.ts
melloware: Yes — and it is the more accurate of the two. `petId` is not in `required`, so absent is a legal response and the mock never generated it before.

The type did not change (`model/pets.ts` is byte-identical on both sides):

```ts
export interface Pets {
  petId?: ((DogId | null) & CatId) | null;
}
```

`petId?:` admits `undefined`, `null`, and a string. The old mock only ever produced the last two.

**What moved.** The property is `{ allOf: [DogId, CatId], nullable: true }`. I ran both upgraders over it directly to be sure rather than infer it from the snapshot:

```
0.2.12: {"allOf":[{"$ref":"DogId"},{"$ref":"CatId"}],"nullable":true}
0.2.15: {"anyOf":[{"allOf":[{"$ref":"DogId"},{"$ref":"CatId"}]},{"type":"null"}]}
```

So `nullable: true` used to survive the upgrade untouched — a 3.0 keyword left sitting in a document we had just declared to be 3.1. `0.2.15` rewrites it into the canonical 3.1 null-union.

That flips one branch in the mock generator, `packages/mock/src/faker/getters/object.ts`:

```ts
const hasNullable = 'nullable' in prop && prop.nullable === true;
...
if (!isRequired && !resolvedValue.overrided && !hasDefault) {
  const omitValue = mockOptions?.nonNullable || !hasNullable ? 'undefined' : 'null';
  return `${keyDefinition}: faker.helpers.arrayElement([${resolvedValue.value}, ${omitValue}])`;
}
```

Before, `hasNullable` was true, so the *omission* marker for the optional property was `null` — and `null` was then standing for two different things at once: "property absent" and "value is null". `undefined` was unreachable.

Now `nullable` is gone, `hasNullable` is false, the omission marker is `undefined`, and the `null` comes from the `{ type: 'null' }` union member instead. That is the extra nesting level in the diff — the outer `arrayElement` is the optional wrapper, the inner one is the union:

```js
petId: faker.helpers.arrayElement([   // optional: value or absent
  faker.helpers.arrayElement([        // anyOf: the allOf, or null
    faker.helpers.arrayElement([alpha, null]),  // DogId = string | null
    alpha,                                       // CatId
    null,                                        // type: 'null'
  ]),
  undefined,
]),
```

All three states the type permits are now reachable; before, one of them was not. Generated-mock verification and all 20 generated-API typecheck runs pass.

tests/__snapshots__/fetch/zod-schema-response-suffix-single-no-runtime-validation/model/nullablePositionWrapperSchema.zod.ts
melloware: It is the same schema — `.nullish()` is defined as `.nullable().optional()`. But "looks like it expresses the same" is not evidence, so I checked it two ways rather than argue from the definition.

**Runtime.** Built both snapshots against the repo's zod (4.4.2) and ran the same inputs through each:

```
SAME  present valid      old=ok {"position":{"id":"a","quantity":1}} | new=ok {"position":{"id":"a","quantity":1}}
SAME  null               old=ok {"position":null}                    | new=ok {"position":null}
SAME  undefined          old=ok {}                                   | new=ok {}
SAME  key absent         old=ok {}                                   | new=ok {}
SAME  wrong inner type   old=fail                                    | new=fail
SAME  missing inner      old=fail                                    | new=fail
SAME  scalar             old=fail                                    | new=fail
SAME  array              old=fail                                    | new=fail
SAME  string             old=fail                                    | new=fail
```

Identical accept/reject **and** identical parsed output on every case, including the stripped-key behaviour for `undefined`.

**Types.** The snapshot exports `zod.input<typeof Schema>`, so I checked both directions of assignability under `--strict`:

```ts
type Exact<A, B> = [A] extends [B] ? ([B] extends [A] ? true : false) : false;
const inputsMatch: Exact<OldIn, NewIn> = true;    // compiles
const outputsMatch: Exact<OldOut, NewOut> = true; // compiles
```

Clean. I also ran a negative control — swapping one side for `{ position: string }` fails with `TS2322` — so the check is not passing vacuously.

**Why it moved at all.** The source is the canonical OAS 3.0 nullable-ref idiom:

```yaml
position:
  nullable: true
  allOf:
    - $ref: '#/components/schemas/Position'
```

Confirmed against both upgrader versions directly:

```
0.2.12: {"nullable":true,"allOf":[{"$ref":"#/components/schemas/Position"}]}
0.2.15: {"anyOf":[{"$ref":"#/components/schemas/Position"},{"type":"null"}]}
```

`0.2.12` left `nullable: true` in place — a 3.0 keyword surviving into a document we had just declared to be 3.1 — and orval's zod generator picked it up through its own `nullable` fallback, collapsing it with the optional into `.nullish()`. `0.2.15` resolves it upstream into the 3.1 null-union, so the generator now takes the ordinary union path and emits `.optional()` for the non-required property.

Larger diff, less special-casing: the `| null` now arrives as structure rather than as a leftover 3.0 keyword. That is the same upstream change as the faker `undefined` above, and it is the point of [#4113](https://github.com/orval-labs/orval/issues/4113).

packages/orval/src/import-specs.test.ts
melloware: Correction: I was wrong above, and you were closer to right than my answer gave you credit for.

I told you the drop was safe because `null & { ... }` reduces to `never`. That is true — but only when the intersection *has* another part that cannot itself be null. I generalised it to every direct `allOf` member, and two shapes have no such part. Both silently lost the `| null`. Generated with the branch build, and again with the exemption disabled to isolate it:

| schema | exemption off | exemption on (as pushed before) |
|---|---|---|
| `allOf: [{$ref: Base, nullable: true}]` | `(Base \| null)` | `Base` ❌ |
| `allOf: [{$ref: Base, nullable}, {$ref: Other, nullable}]` | `(Base \| null) & (Other \| null)` | `Base & Other` ❌ |
| `allOf: [{$ref: Base, nullable}, {marker}]` | `(Base \| null) & {marker?}` | `Base & {marker?}` ✅ |

`(Base | null) & (Other | null)` is `(Base & Other) | null`, so neither member absorbs the other's branch. Confirmed under `tsc --strict`: the third row's two forms are mutually assignable, the first two rows' are not.

That was a real unsoundness — a consumer reads the generated type, drops a null check, and hits a `TypeError` on a response the API is entitled to send. Worse than the `| null` we were discussing being cosmetically absent, because this path emits no warning either.

Our own fixtures did not catch it: `NullableAllOfMemberWrapper` and `NullableParentWrapper` both happen to have a non-nullable shape in the intersection, which is exactly the case where my argument does hold.

**Fixed in 760a0439d.** The drop is now gated on `absorbsNullBranch` — some other member, or the enclosing schema's own constraints, has to be a non-nullable object. The enclosing half is not optional: `NullableParentWrapper` puts `properties` beside a *single*-member `allOf`, and that is what absorbs its null, so a sibling-only check would have broken it.

The gate is syntactic and deliberately does not follow a bare `$ref`, so `allOf: [{$ref: A, nullable: true}, {$ref: B}]` keeps the union even though `B` is usually a non-nullable object. That errs toward emitting a union that was not strictly needed, which is at worst noisier. Erring the other way is what produced this bug.

The same commit fixes a second one found alongside it — `inAllOf` was set from `key === 'allOf'` with no `Array.isArray` check, so a property literally *named* `allOf` took the drop branch:

```ts
// before
export interface PropNamedAllOf {
  allOf?: Base;        // nullable dropped, no warning
  other?: Base | null; // identical shape, handled correctly
}
```

Both shapes are now pinned end-to-end in `regressions.yaml` (`NullableAllOfLoneMember`, `NullableAllOfEveryMemberNullable`) rather than only in unit tests, so the placements table you asked for earlier is complete: nullability survives on the composed schema, at a plain property, below an `allOf` member, and now also inside an `allOf` that has nothing to absorb it. It dies only where the intersection really does kill it.

No existing snapshot moved, so the cases the exemption was built for are untouched. Full run: 4,655 unit tests, 7,020 snapshot tests, all 20 generated-API typecheck runs, generated-mock verification, typecheck and lint.

snebjorn (approved):
I think we're good now

PR #4122, opened 2026-09-16
feat(fetch): preserve error response types for custom mutators

### LINKED ISSUE #4111: Preserve error response status/body relationships for custom fetch mutators
### Description

I would like to narrow HTTP error bodies by their documented status in React Query queries and mutations.
To support this, I propose allowing custom fetch mutators to receive the complete error response type and the declared error statuses for each operation.

### Current behavior

Consider an operation that returns `{ message: string }` for 404 and `{ fields: string[] }` for 422.
With `client: 'react-query'`, `httpClient: 'fetch'`, `forceSuccessResponse: true`, and a custom mutator exporting `ErrorType`, the generated response types preserve the relationship between status and body.
However, query and mutation error types pass only the body union to `ErrorType`.

```ts
// Simplified example of the generated types

type GetItemResponseError =
  | { status: 404; data: { message: string }; headers: Headers }
  | { status: 422; data: { fields: string[] }; headers: Headers };

type GetItemQueryError = ErrorType<
  { message: string } | { fields: string[] }
>;
```

Adding `status: number` to the mutator's error type does not recover that relationship.
With the generated error type as it stands, consumers cannot narrow the body to access `fields` specifically for 422 or use a `never` check to detect unhandled documented statuses.

I reproduced this by building master at `5097831948f5dd436ed69260b4dfca32b089795d` and generating code for GET and POST operations with 404 and 422 responses.
After branching on status, the body remains a union: accessing the corresponding property produces TS2339, and assigning the remaining error to `never` also fails.
The generator passes `response.definition.errors` to the custom mutator's `ErrorType`.

https://github.com/orval-labs/orval/blob/5097831948f5dd436ed69260b4dfca32b089795d/packages/query/src/client.ts#L681-L693

### Proposed behavior

Allow `ErrorType` to receive the complete response union, preserving the relationship between status and body, for both queries and mutations.

```ts
type GetItemQueryError = ErrorType<GetItemResponseError>;
```

Also generate and pass each operation's declared error statuses and media types so that a shared mutator can distinguish documented HTTP errors from unexpected responses at runtime.
For example, a 422 response could be the JSON response declared for that operation or a response with an unexpected media type.

```ts
customFetch(url, options, {
  errorResponses: [
    { status: 404, contentType: 'application/json' },
    { status: 422, contentType: 'application/json' },
  ],
});
```

With this information, the mutator can use the documented error types while representing unexpected statuses, transport failures, and body parsing failures separately.
Matching a status and media type does not itself validate the body against its schema.
The mutator remains responsible for its error representation.

### Proposed solution

Add an opt-in setting for custom fetch mutators that passes the complete response union and the metadata shown above.
Existing `ErrorType<T>` implementations treat T as the body, so generation would remain unchanged unless the setting is enabled.

Pass the metadata as the third argument to keep it separate from caller-controlled `RequestInit`.
The current fetch generator calls the mutator with two arguments and passes the schema for `includeZodSchemaInArguments` in the second argument's `schema` property.
The proposal would preserve that format.

https://github.com/orval-labs/orval/blob/5097831948f5dd436ed69260b4dfca32b089795d/packages/fetch/src/index.ts#L690-L709

https://github.com/orval-labs/orval/blob/5097831948f5dd436ed69260b4dfca32b089795d/packages/fetch/src/index.ts#L821

I tested a prototype using a custom client extension against the master revision above.
For the same GET and POST operations, the body-narrowing and exhaustiveness checks that fail with the standard generator pass with the extension.
I also verified that it distinguishes a documented 422 from an unexpected 502 at runtime.

<details>
<summary>Custom client extension used for verification</summary>

Set `output.client` to the following `responseErrorsClient`.
It targets react-query/fetch with response envelopes and a named three-argument mutator exporting `ErrorType`.
This prototype handles explicit status codes and rejects `default` and status ranges during generation.

```ts
import type { OutputClientFunc } from "orval";

export const responseErrorsClient: OutputClientFunc = (clients) => ({
    ...clients["react-query"],
    client: async (operation, options, _client, output) => {
        const { mutator, response, override } = operation;
        if (
            options.context.output.httpClient !== "fetch" ||
            !override.fetch.includeHttpResponseReturnType ||
            !override.fetch.forceSuccessResponse ||
            !mutator ||
            mutator.default ||
            mutator.isHook ||
            !mutator.hasThirdArg ||
            !mutator.hasErrorType
        ) {
            throw new Error(
                "Response errors require react-query/fetch, response envelopes, forceSuccessResponse and a named three-argument mutator exporting ErrorType",
            );
        }
        // Current APIs use explicit statuses. A default/range response requires
        // a matching precedence rule in runtime metadata; reject it until that
        // contract is designed, rather than silently generating a false union.
        const errorResponses = response.types.errors.map(({ key, contentType }) => {
            if (!/^[1-5]\d\d$/.test(key))
                throw new Error(`Unsupported error status ${key} in ${operation.operationId}`);
            return { status: Number(key), contentType };
        });
        const name = `${operation.operationName}Request`;
        const result = await clients["react-query"].client(
            {
                ...operation,
                mutator: { ...mutator, name },
                response: {
                    ...response,
                    definition: {
                        ...response.definition,
                        errors: errorResponses.length
                            ? `${operation.typeName}ResponseError`
                            : "never",
                    },
                },
            },
            options,
            "react-query",
            output,
        );
        return {
            ...result,
            implementation: `const ${name} = <T>(url: Parameters<typeof ${mutator.name}>[0], options: Parameters<typeof ${mutator.name}>[1]): Promise<T> =>
    ${mutator.name}<T>(url, options, ${JSON.stringify({ errorResponses })});

${result.implementation}`,
        };
    },
});
```

Replacing `response.definition.errors` with the generated response union passes the status/body relationship through to `ErrorType` for queries and mutations.
The generated wrapper for each operation passes the derived `errorResponses` to the mutator's third argument.

</details>

Would you consider adding support for passing these response types and operation-specific definitions to custom mutators?

If this proposal is accepted, I can put together a PR with the implementation and tests.

eai04191:
- Closes #4111.

Adds `override.fetch.includeErrorResponseInMutator` so custom fetch mutators can receive:

- The full error response union through `ErrorType<T>`, allowing queries and mutations to narrow response bodies by status.
- Declared error statuses and media types through a third argument, separate from request options.

Error handling remains the mutator's responsibility. Requires response envelopes and `forceSuccessResponse`. Supports named and default exports; rejects hook mutators and any `default` or ranged error responses retained after `override.contentType` filtering.

Includes generation, TypeScript narrowing/exhaustiveness, and runtime tests covering documented and unexpected responses, plus coverage for Zod schema arguments and media type filtering.

byCodex

## Summary by CodeRabbit

* **New Features**
  * Added the opt-in `includeErrorResponseInMutator` setting for fetch clients.
  * Custom mutators can receive typed error responses and filtered status/content-type metadata.
  * Generated query and mutation error types now reflect included response details.
  * Added request-body date serialization when date transformation is enabled.
  * Added support for body-less error responses and content-type filtering.
  * Custom MCP handlers now emit structured content only for non-null, non-array objects.

* **Documentation**
  * Documented configuration requirements, supported mutator formats, error handling, unsupported status patterns, MCP responses, and date transformations.


packages/fetch/src/index.ts
```diff
@@ -81,6 +83,63 @@ const getRequestOptionsType = (mutator?: GeneratorMutator) => {
...
+  ) {
+    throw new Error(
+      'includeErrorResponseInMutator requires includeHttpResponseReturnType, forceSuccessResponse, and a non-hook mutator accepting three arguments.',
+    );
+  }
+
+  // Type resolution may supply a default JSON media type or collapse
+  // referenced content. Runtime metadata must preserve the source declaration.
+  const errorResponses = Object.entries(response.originalSchema ?? {})
+    .filter(([key]) => response.types.errors.some((entry) => entry.key === key))
```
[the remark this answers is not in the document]
eai04191: Kept validation after content-type filtering: excluded responses contribute neither error types nor runtime metadata, so they do not require range/default matching support. Retained `default` and ranged error responses still fail generation.

Clarified this boundary in the documentation and added integration coverage in 67d8ba34f. For both include and exclude filters, adding an excluded `default` or `4XX` response produces identical generated code; the generated consumer types also pass TypeScript checks.

byCodex

docs/content/docs/reference/configuration/output.mdx
```diff
@@ -2685,6 +2685,59 @@ produce `status: 200` and `status: Exclude<HTTPStatusCode2xx, 200>`.
...
+    output: {
+      target: './api.ts',
+      client: 'react-query',
+      httpClient: 'fetch',
+      override: {
+        mutator: { path: './mutator.ts', name: 'customFetch' },
+        fetch: {
+          includeHttpResponseReturnType: true,
+          forceSuccessResponse: true,
+          includeErrorResponseInMutator: true,
```
melloware: Shouldn't this be simplified to `includeHttpErrorResponse`

packages/core/src/types.ts
```diff
@@ -1490,6 +1490,7 @@ export interface AxiosOptions {
 export interface NormalizedFetchOptions {
   includeHttpResponseReturnType: boolean;
   forceSuccessResponse: boolean;
+  includeErrorResponseInMutator?: boolean;
```
melloware: includeHttpErrorResponse

melloware (changes requested):

docs/content/docs/reference/configuration/output.mdx
eai04191: renamed in a0d1b1780ab6655c7f32c45dbe1036fcdf731f2e

melloware (approved):

PR #4123, opened 2026-09-16
fix(core): escape the body type when wrapping props in BodyType<T>

Dev-next-gen:
I was looking at how the `BodyType<T>` envelope gets applied to the body parameter and noticed that four clients build the rewrite regex by interpolating `body.definition` straight into a `RegExp`:

```ts
toObjectString(props, 'implementation').replace(
  new RegExp(String.raw`(\w*):\s?${body.definition}`),
  `$1: ${mutator.bodyTypeName}<${body.definition}>`,
)
```

`body.definition` is a TypeScript type, so it regularly contains regex metacharacters. `@orval/query` already escapes this exact value in `wrapPropsBodyWithMutatorBodyType`; `packages/axios`, `packages/swr`, `packages/solid-start`, the Angular HttpClient generator and the request function in `packages/query/src/client.ts` (which only escaped the first `[]`) do not.

What that produces, with a mutator that exports `BodyType<T>` (I used `tests/mutators/custom-client.ts`) and a spec with an array request body and a `oneOf` request body:

```
# client: 'axios', before
const createPets = (
    pet: Pet[],                                     <- envelope silently dropped
const createAnimal = (
    petCat: CreateAnimalBodyType<Pet | Cat>| Cat,   <- widened back to the bare type
```

`Pet[]` ends with `[]`, an empty character class that can never match, so the whole pattern fails and the parameter keeps the raw type. `Pet | Cat` turns the pattern into an alternation, the `| Cat` branch survives into the output, and the parameter accepts a plain `Cat` again — which defeats the point of the envelope when `BodyType<T>` is not just `T & {...}`.

The Angular copy is a separate flavour of the same mistake. It was moved to `String.raw` without dropping the doubled escapes, so the pattern is `(\\w*):\\s?` and expects a literal backslash in the props string. It never wraps anything at all, and on a union body only the stray ` | Cat` alternative matches, with `$1` undefined:

```
# client: 'angular', before
 createPets<TData = void>(
    pet: Pet[],
 createAnimal<TData = void>(
    petCat: Pet |: CreateAnimalBodyType<Pet | Cat>,   <- generated file does not parse
```

Following review, the fix no longer builds a regex from the definition at all. Each of the five generators picks the prop whose type is `GetterPropType.BODY` before `toObjectString` and replaces its trailing `: ${body.definition}` with the envelope, so metacharacters, `$1`/`$&` in an inline enum body, a path param typed `PetStatus` next to a `Pet` body, and a path param of the same type as the body are all handled by construction. `Pet[]` and `Pet | Cat` now come out as `pet: BodyType<Pet[]>` and `pet: BodyType<Pet | Cat>` in the unit tests of every generator.

I added unit tests next to each generator (`Pet[]`, `Pet | Cat`, `(Pet | Cat)[]`, plus a plain `Pet` for Angular since that one never matched either). They fail on `master` — 3 in axios, 3 in angular, 2 in swr, 2 in solid-start — and pass with the change.

I did not add a snapshot fixture: none of the committed ones pairs a metacharacter body type with a `BodyType<T>` mutator, and I checked that regenerating `tests/configs/{axios,angular,swr,solid-start}.config.ts` leaves `tests/__snapshots__` byte-identical, so this is not a change to any currently generated output. `vp fmt --check`, `vp lint --type-aware --type-check packages`, typecheck and the unit suites of the four packages (plus core, query and fetch) are green.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed generated Angular, Axios, Solid Start, SWR, and Query clients so request bodies with object, array, union, and literal types are correctly wrapped in the configured body type.
  - Preserved other request parameters, including path parameters sharing the body’s type, when applying body type wrappers.
  - Improved handling of complex body definitions containing special characters, preventing incorrect request-generation substitutions.

- **Tests**
  - Added coverage for varied request body definitions and parameter combinations across supported client generators.


packages/angular/src/http-client.ts
```diff
@@ -522,7 +523,7 @@ export const generateHttpClientImplementation = (
     const propsImplementation =
       mutator.bodyTypeName && body.definition
         ? toObjectString(props, 'implementation').replace(
-            new RegExp(String.raw`(\\w*):\\s?${body.definition}`),
+            new RegExp(String.raw`(\w*):\s?${escapeRegExp(body.definition)}`),
```
melloware: **Newly-reachable mis-wrap: the pattern still has no right-hand boundary.**

This is the one site where the replacement was previously dead code (`\\w`/`\\s` required literal backslashes in the props string, so it never matched). Fixing the escapes turns it on for the first time — but the pattern is still applied to the *whole* props string and matches the first `name: <definition>` occurrence with no boundary after `<definition>`, so it can wrap a parameter that merely *starts with* the body type name.

Path/named-path/query props are emitted before the body prop by `getProps`, so for an operation with e.g. a path param typed `PetStatus` and `body.definition === 'Pet'`:

```
in : "petId: PetStatus,\n    pet: Pet,\n    "
out: "petId: BodyType<Pet>Status,\n    pet: Pet,\n    "
```

(verified by running this exact expression standalone). The wrong parameter gets the envelope, the body keeps the raw type, and `BodyType<Pet>Status` is not parseable TypeScript — so the generated Angular service fails to compile. Before this PR Angular just emitted no wrap at all, so this is a new failure mode introduced here, not a pre-existing one.

Since the body prop is identifiable (`type === GetterPropType.BODY`), rewriting that one prop's `implementation` instead of regexing the joined string would remove the whole class of problem. A minimal mitigation is a trailing boundary:

```suggestion
            new RegExp(String.raw`(\w*):\s?${escapeRegExp(body.definition)}(?![\w$])`),
```

The same missing boundary exists in the axios/swr/solid-start copies, but there the code path was already live.

packages/axios/src/index.ts
```diff
@@ -298,7 +299,7 @@ export type ${pascal(typeName)}Result = ${axiosResponse.name}`
     const propsImplementation =
       mutator.bodyTypeName && body.definition
         ? toObjectString(props, 'implementation').replace(
-            new RegExp(String.raw`(\w*):\s?${body.definition}`),
+            new RegExp(String.raw`(\w*):\s?${escapeRegExp(body.definition)}`),
```
melloware: **A fifth copy of this call site was missed: `packages/query/src/client.ts:382`.**

The PR description says "`@orval/query` already escapes this exact value in `wrapPropsBodyWithMutatorBodyType`". That is true of `packages/query/src/query-generator.ts`, but `packages/query/src/client.ts` has its own copy of the same block that does *not* use `escapeRegExp`:

```ts
const bodyDefinition = body.definition.replace('[]', String.raw`\[\]`);
const propsImplementation =
  mutator.bodyTypeName && body.definition
    ? toObjectString(props, 'implementation').replace(
        new RegExp(String.raw`(\w*):\s?${bodyDefinition}`),
        `$1: ${mutator.bodyTypeName}<${body.definition}>`,
      )
    : toObjectString(props, 'implementation');
```

This is the request-function generator for the react-query / vue-query / svelte-query clients, so it is the most used of the five. `String.prototype.replace` with a string pattern replaces only the **first** `[]`, and nothing at all is escaped for `|`, `(`, `)`, `{`, `}`, `.`, `?`, `+`, `*`. So the exact two symptoms this PR fixes elsewhere still reproduce there:

- `Pet | Cat` → pattern becomes the alternation `(\w*):\s?Pet ` / ` Cat`, and the output keeps `| Cat` outside the envelope.
- `(Pet | Cat)[]` → the parens become a capture group and `|` an alternation, so the wrap lands in the wrong place or not at all.
- `Pet[][]` → only the first `[]` is escaped, so the pattern can never match.

Worth folding into this PR so the fix covers the class rather than four of five sites (and it would let all five share one helper).

packages/swr/src/client.ts
```diff
@@ -96,7 +97,7 @@ const generateAxiosRequestFunction = (
     const propsImplementation =
       mutator.bodyTypeName && body.definition
         ? toObjectString(props, 'implementation').replace(
-            new RegExp(String.raw`(\w*):\s?${body.definition}`),
+            new RegExp(String.raw`(\w*):\s?${escapeRegExp(body.definition)}`),
```
melloware: Minor / same bug class, other half: the **replacement** string is still unescaped.

`escapeRegExp` fixes the pattern argument, but `body.definition` is also interpolated into the replacement on the next line, where `String.prototype.replace` expands `$&`, `` $` ``, `$'`, `$$` and `$1`–`$9`. `body.definition` is not always an identifier — `buildBody` joins the raw `resolveValue` output, so an inline request-body enum comes through as a literal union, and enum values containing `$` are not sanitized away.

```
def  = "'$1' | 'b'"
props= "pet: '$1' | 'b',\n"
out  = "pet: BodyType<'pet' | 'b'>,\n"   // $1 expanded to the captured prop name
```

```
def  = "'$&'"
out  = "pet: BodyType<'pet: '$&''>,\n"  // $& expanded to the whole match
```

(both verified by running the expression standalone). Passing a replacer function instead of a string sidesteps it entirely and keeps the four call sites identical:

```ts
.replace(
  new RegExp(String.raw`(\w*):\s?${escapeRegExp(body.definition)}`),
  (_match, name: string) => `${name}: ${mutator.bodyTypeName}<${body.definition}>`,
)
```

Applies to all four touched files (and to `packages/query/src/client.ts`).

melloware (changes requested):
See my comments

packages/angular/src/http-client.ts
Dev-next-gen: You're right: fixing the escapes turned this on for Angular for the first time, and your `petId: PetStatus` case reproduces. In 03308dade the pattern now ends in `(?=,)` rather than `(?![\w$])`, at all five sites. `toObjectString` joins props with `',\n    '` and appends a final `,` (packages/core/src/utils/string.ts:250), so the body type has to end exactly where its prop ends. That also rejects `Pet[]` as a prefix of `Pet[][]`, which the word-boundary version would still have let through.

The new tests put a `petId: PetStatus` prop (`petId: Pet[]Status` in the query one) ahead of the body. All of them fail against 3595a9c and pass now, and regenerating the axios, angular, swr, solid-start, react-query, vue-query, svelte-query, solid-query and angular-query configs leaves `tests/__snapshots__` unchanged.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

packages/axios/src/index.ts
Dev-next-gen: Thanks, I missed that copy, and my description was wrong to imply that query was already covered. In 03308dade `packages/query/src/client.ts` drops the `.replace('[]', …)` and uses the same `escapeRegExp` pattern, boundary and replacer as the other four, so all five blocks are now character-for-character identical. I added `generateAxiosRequestFunction` cases for `Pet`, `Pet[]`, `Pet[][]`, `Pet | Cat`, `(Pet | Cat)[]`, `'$1' | 'b'` and `'$&'`. All seven fail against 3595a9c and pass now (`Tests 7 failed | 234 passed` before).

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

packages/swr/src/client.ts
Dev-next-gen: Applied as you suggested, with the replacer function in all four files and in `packages/query/src/client.ts` (03308dade). `'$1' | 'b'` and `'$&'` are now in each test table. Both cases fail against 3595a9c in every suite, and they now pass with the output asserted verbatim as `pet: BodyType<'$1' | 'b'>` and `pet: BodyType<'$&'>`.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

packages/angular/src/http-client.ts
```diff
@@ -522,8 +523,11 @@ export const generateHttpClientImplementation = (
     const propsImplementation =
       mutator.bodyTypeName && body.definition
         ? toObjectString(props, 'implementation').replace(
-            new RegExp(String.raw`(\\w*):\\s?${body.definition}`),
-            `$1: ${mutator.bodyTypeName}<${body.definition}>`,
+            new RegExp(
+              String.raw`(\w*):\s?${escapeRegExp(body.definition)}(?=,)`,
+            ),
+            (_match, name: string) =>
+              `${name}: ${mutator.bodyTypeName}<${body.definition}>`,
```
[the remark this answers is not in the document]
Dev-next-gen: Confirmed: with a `petId: PetStatus` path param ahead of a `PetStatus` body, all five generators emitted `petId: BodyType<PetStatus>` and left `pet: PetStatus` bare. In 59ce31b each of them now rewrites the `GetterPropType.BODY` prop before `toObjectString`, so no regex is built from the definition any more; the new same-type test fails once per suite against 03308da (angular 1/319, axios 1/35, query 1/242, solid-start 1/37, swr 1/11), all five suites pass now, and regenerating the 14 snapshot configs leaves `tests/__snapshots__` unchanged.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

Dev-next-gen:
On 59ce31b both `pr-checks` jobs (ubuntu and windows) pass. The only red check is `publish-preview`, and it failed on the pkg.pr.new side: the upload got a 500, retried, and then got `Publishing failed (404): {"url":"/publish","statusCode":404,"statusMessage":"Not Found","message":"There is no workflow defined for gPXukJyPkt"}` ([job log](https://github.com/orval-labs/orval/actions/runs/35099879918/job/104809286118)). The same job passed on this branch at 03308da half an hour earlier ([run](https://github.com/orval-labs/orval/actions/runs/35097134102/job/104800851792)), and between those two heads the only changes are the five generator sources and their tests, with nothing in packaging or the workflow.

I tried `gh run rerun 35099879918 --failed` and GitHub refused it ("Must have admin rights to Repository"), so if you want the preview packages for 59ce31b, re-running that one job should be all it takes.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

melloware (approved):

PR #4125, opened 2026-09-16
fix(core): stringify allOf-wrapped array items in form data

Dev-next-gen:
After #4105 I went through the other places in `res-req-types.ts` that still decide "is this an object?" by reading `type` directly. #4105 switched the top-level array branch of `getSchemaFormDataAndUrlEncoded` to `isEffectivelyObjectSchema`, but the array branch of `resolveSchemaPropertiesToFormData`, the one used for each property of a multipart or url-encoded body, still tests `itemSchema.type === 'object'`.

So a property like this one, where the item is the same kind of pure `allOf` wrapper #4105 was about:

```yaml
Tag:
  type: object
  properties: { name: { type: string } }
TagRef:
  allOf: [{ $ref: '#/components/schemas/Tag' }]
# multipart body
tags:
  type: array
  items: { $ref: '#/components/schemas/TagRef' }
```

generates, on current master:

```ts
// arrayHandling: 'serialize'
createTagsRequestBody.tags.forEach(value => formData.append(`tags`, value));
// arrayHandling: 'explode'
createTagsRequestBody.tags.forEach((value, index) => formData.append(`tags[${index}]`, value));
```

`FormData.append` stringifies a plain object, so the server receives `[object Object]` for every tag. The fix is the same one-line swap #4105 made in the other branch: use `isEffectivelyObjectSchema(itemSchema, context)`. With it, serialize mode emits `JSON.stringify(value)` and explode mode recurses into the item, which already works for allOf wrappers since `resolveSchemaPropertiesToFormData` collects properties through `allOf`. Items that are an `allOf` around a scalar are left alone, as before.

I added two tests next to the #4105 ones in `res-req-types.test.ts`, one per array handling mode. Both fail on master (2 failed, 53 passed in that file) and pass with the change (55 passed); the whole `@orval/core` suite passes (2707 tests), and regenerating the default, axios, fetch, angular, react-query, swr and vue-query configs gives output identical to `tests/__snapshots__`. `vp fmt --check` and `vp lint` are clean on the two files.

Review follow-up in cb949a7: `allOf` wrappers around an array are now detected at both array sites, the explode recursion stops on recursive models (plain or behind an `allOf` wrapper) and appends the value as JSON instead of overflowing the stack, items with nothing to explode (maps, a bare `allOf: [{ type: object }]`, nested arrays) are appended as JSON instead of producing an empty `forEach`, and nullable object items are skipped when null. Five tests cover these; they fail on c15f154 and pass on cb949a7 (60 in that file), the `@orval/core` suite passes (2712 tests), and the default, axios, fetch, angular, react-query, swr, vue-query and zod configs still regenerate identical to the snapshots.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multipart form data handling for arrays of object-like items, including schemas defined through composition.
  * Ensured default serialization preserves each item as JSON.
  * Ensured exploded arrays submit nested item properties using bracketed field names.
  * Prevented recursive or cyclic data structures from causing endless expansion.
  * Correctly JSON-serializes items that cannot be safely expanded and skips null or undefined items during expansion.


packages/core/src/getters/res-req-types.ts
```diff
@@ -985,7 +985,10 @@ function resolveSchemaPropertiesToFormData({
       const propertyItems = getSchemaItems(property);
       if (propertyItems) {
         const { schema: itemSchema } = resolveSchemaRef(propertyItems, context);
-        if (itemSchema.type === 'object' || itemSchema.type === 'array') {
+        if (
```
melloware, line 1023: **Unguarded recursion: this widening makes generation crash on `allOf`-wrapped recursive models.**

Letting a pure `allOf` wrapper through here means it now reaches the EXPLODE branch below, which calls `resolveSchemaPropertiesToFormData` recursively with no cycle guard.

With `arrayHandling: 'explode'` and a multipart body containing `tags: { type: array, items: { $ref: TagRef } }`:

```yaml
Tag:
  type: object
  properties:
    name: { type: string }
    children: { type: array, items: { $ref: '#/components/schemas/TagRef' } }
TagRef:
  allOf: [ { $ref: '#/components/schemas/Tag' } ]
```

Before this commit that generates fine. After it, orval aborts with `RangeError: Maximum call stack size exceeded`.

The recursion hazard itself is pre-existing — a self-referential plain `type: object` item already crashes — but `allOf`-wrapped recursive models are exactly the shape this commit newly classifies as object-like, and a self-referential tree behind an `allOf` wrapper is a common way to write one. So this turns a latent hazard into a reachable one for a realistic spec.

Worth a `seen` set or a depth cap on the explode recursion, falling back to `JSON.stringify(value)` when a cycle is detected.

packages/core/src/getters/res-req-types.ts
```diff
@@ -985,7 +985,10 @@ function resolveSchemaPropertiesToFormData({
...
+          isEffectivelyObjectSchema(itemSchema, context) ||
+          itemSchema.type === 'array'
```
melloware: **EXPLODE can now emit an empty `forEach`, silently dropping the field from the request.**

When the item is object-like but `collectPropertiesThroughAllOf` yields no properties, `hasNonPrimitiveChild` is set to `true` here, which suppresses the primitive fallback at the bottom of the branch — and `resolvedValue` is `''`. The result is a loop that appends nothing:

`items: { additionalProperties: { type: 'string' } }` (a map, no explicit `type: object`)

```js
// before
tags.forEach((value, index) => formData.append(`tags[${index}]`, value));
// after
tags.forEach((value, index) => {
    });
```

`items: { allOf: [{ type: 'object' }] }` produces the same empty loop.

No error is raised — the field just disappears from the multipart body at runtime, which is a nastier failure than a wrong value.

`type: object` + `additionalProperties` was already broken this way before the commit, so this is a widening of an existing bug rather than a new one. But the newly-captured shapes are precisely the ones this PR is targeting, so it is worth closing here: if `resolvedValue.trim() === ''`, fall back to `JSON.stringify(value)` instead of emitting the empty loop.

packages/core/src/getters/res-req-types.ts
```diff
@@ -985,7 +985,10 @@ function resolveSchemaPropertiesToFormData({
       const propertyItems = getSchemaItems(property);
       if (propertyItems) {
         const { schema: itemSchema } = resolveSchemaRef(propertyItems, context);
-        if (itemSchema.type === 'object' || itemSchema.type === 'array') {
+        if (
+          isEffectivelyObjectSchema(itemSchema, context) ||
```
melloware, line 1024: **The `array` half of the condition was left as a raw type read, so the fix is only half-applied.**

`isEffectivelyObjectSchema(itemSchema, context)` resolves through `allOf`; `itemSchema.type === 'array'` sitting right next to it does not. So an `allOf`-wrapped *array* item still slips through:

```yaml
A:    { type: array, items: { type: string } }
ARef: { allOf: [ { $ref: '#/components/schemas/A' } ] }
# ... items: { $ref: '#/components/schemas/ARef' }
```

still generates

```js
tags.forEach(value => formData.append(`tags`, value));
```

— the array appended raw rather than stringified, which is the `[object Object]`-on-the-wire case this commit is meant to prevent. Given the title ("stringify allOf-wrapped array items in form data"), that looks in scope.

Resolving the array branch through `allOf` too would cover it. The same raw `type === 'array'` read exists at line 703, so whichever way you go it is probably worth keeping the two sites in step — #4105 already had to bring them into parity once.

Dev-next-gen: You're right, the array half was still a raw `type` read. cb949a7 adds `isEffectivelyArraySchema`, the same allOf walk as the object helper but for `type: array`, and uses it both at line 703 and in the property branch, so the two sites stay in step.

With your `ARef` spec, serialize mode now emits `` tags.forEach(value => formData.append(`tags`, JSON.stringify(value))) ``, where c15f154 still appended `value`. The new test for it fails on c15f154 and passes on cb949a7; the whole `@orval/core` suite passes (2712 tests), and regenerating the default, axios, fetch, angular, react-query, swr, vue-query and zod configs gives output identical to `tests/__snapshots__`.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

packages/core/src/getters/res-req-types.ts
Dev-next-gen: I reproduced it: with your `Tag`/`TagRef` spec in explode mode, c15f154 throws `RangeError: Maximum call stack size exceeded`, while the base commit generated `` formData.append(`tags[${index}]`, value) ``. A plain self-reference (`children: { items: { $ref: Tag } }`) throws on both.

cb949a7 passes the schemas currently being exploded down the recursion (`ancestors`, copied at each level so it only holds the current path). When an array item or an object property is already in it, the value is appended with `JSON.stringify` instead of recursing. Your spec now gives `tags[${index}].name` plus `` value.children.forEach((value, index1) => formData.append(`tags[${index}].children[${index1}]`, JSON.stringify(value))) ``. I put the same check on the object-property recursion, since `parent: { $ref: Tag }` overflowed the same way; there is a test for each case.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

packages/core/src/getters/res-req-types.ts
Dev-next-gen: Confirmed on c15f154: both `items: { additionalProperties: { type: 'string' } }` and `items: { allOf: [{ type: 'object' }] }` produced the empty `forEach` in explode mode.

In cb949a7 the branch only uses the nested form when the recursion actually returned something. Otherwise it falls through to the per-item line with `JSON.stringify(value)`, so both shapes now give `` tags.forEach((value, index) => formData.append(`tags[${index}]`, JSON.stringify(value))) ``. A nested array item (`items: { type: array }`) takes the same path, and the test loops over all three shapes.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

packages/core/src/getters/res-req-types.ts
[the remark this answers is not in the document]
Dev-next-gen: This one was real. With `items: { type: ['object', 'null'], properties: { name: ... } }` and the data `[null, { name: 'x' }]`, the explode code generated on c15f154 throws `TypeError: Cannot read properties of null (reading 'name')`. The base commit didn't treat that item as an object, so it appended `tags[0]=null` and `tags[1]=[object Object]`.

In cb949a7, when the item schema is nullable (`nullable: true`, or a type list that contains `null`), the loop body is wrapped in `if (value !== null && value !== undefined)`, and the same data now only appends `tags[1].name=x`. Non-nullable items keep the unguarded body, so the existing generated output doesn't change.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

melloware (approved):

PR #4127, opened 2026-09-16
feat(mcp): wrap non-object responses in structuredContent

soartec-lab:
## Summary

Follow-up to #4126. Non-object success responses (arrays, scalars, unions, records) now get `structuredContent` and a matching `outputSchema` by wrapping the body as `{ result: body }`, the same way FastMCP does. Plain-object responses are passed as is. Tools without a usable response schema (no body, `zod.void()`, or `allOf` intersections) still declare neither.

- `@orval/zod`: `isPlainObjectResponseSchema` is replaced by two predicates, `isObjectResponseSchema` and `hasResponseSchema`, both built on the same `parseBodyAndResponse` call as the generator.
- `@orval/mcp`: `outputSchema: { result: XResponse }` and `structuredContent: { result: res.data }` for non-object responses; the SDK accepts the raw shape, so `server.ts` needs no `zod` import.
- Custom handlers receive a third argument, `toStructuredContent(data)`, generated per operation (`data as Record<string, unknown>` / `{ result: data }` / `undefined`). `ctx` and `options` are now required in the generated handlers. Docs and fixtures updated.

### Before

```ts
// handler returns text only

tools.listPets = server.registerTool(
  'listPets',
  {
    inputSchema: { ... }
   // no outputSchema
  },
 ...
);
```

### After

```ts
// handler returns structuredContent: { result: res.data }

tools.listPets = server.registerTool(
  'listPets',
  {
    inputSchema: { ... },
    outputSchema: { result: ListPetsResponse }
  },
  ...
);
```

Custom handler signature:

```ts
export const customHandler = async (
  fetcher: (overrides?: RequestInit) => Promise<{ status: number; data: unknown; headers: Headers }>,
  ctx: RequestHandlerExtra<ServerRequest, ServerNotification>,
  toStructuredContent: (data: unknown) => Record<string, unknown> | undefined,
): Promise<CallToolResult> => {
  const res = await fetcher();
  // ...
  return {
    content: [{ type: 'text', text: JSON.stringify(res.data ?? null) }],
    structuredContent: toStructuredContent(res.data),
  };
};
```

### Why `allOf` responses get no `outputSchema`

`allOf` (and `oneOf` with sibling properties) renders as a zod intersection. The SDK converts it with `io: 'output'`, which adds `additionalProperties: false` to every member, so the resulting JSON Schema `allOf` rejects every value and the client-side validation would fail on each call. Nested intersections (e.g. `PetWithTag.pet`) are still affected; this is a zod `toJSONSchema` limitation, not addressed here.

### Breaking change notice

Custom handlers must accept the third `toStructuredContent` argument and use it for `structuredContent`. Generated MCP output changes for every non-object response (see the updated snapshots under `tests/__snapshots__/mcp` and `samples/mcp`).

## Summary by CodeRabbit

* **New Features**
  * MCP tools now expose response schemas where applicable.
  * Structured responses are validated against their schemas, with unsupported fields removed.
  * Validation failures return the original response with an error message and `isError: true`.
  * Non-object responses are wrapped as `{ result: ... }`; empty responses omit structured content.
  * Custom MCP handlers support consistent structured-response formatting and error handling.
  * Generated servers preserve configured and per-request cancellation signals.

* **Documentation**
  * Updated MCP handler guidance and examples in English and Chinese to reflect response validation, error handling, and handler parameters.


melloware (dismissed):

soartec-lab:
@melloware 
I apologize for contacting you repeatedly. This will be my final fix, so I would appreciate it if you could post a review.

melloware (approved):

PR #4129, opened 2026-09-16
fix: use queryOptions helper for suspense query functions (#1788)

### LINKED ISSUE #1788: Tanstack Query: Use `queryOptions` type helper instead of UseQueryResult
Tanstack Query recommends using the `queryOptions` type helper instead of manually typing query results. This is now their [recommended standard](https://github.com/TanStack/query/issues/8390).

Not using this helper has already led to some problems because `QueryOptions` and `SuspenseQueryOptions` are now incompatible. This means that generated  `getUse*SuspenseQueryOptions()` are not usable in `useSuspenseQuery` / `useSuspenseQueries` in the current version.

```typescript
Type 'UseSuspenseQueryOptions<ValueType, ErrorType, ValueType, QueryKey>' is not assignable to type 'UseSuspenseQueryOptions<ValueType, unknown, ValueType, QueryKey>'.
  Types of property 'refetchInterval' are incompatible.
    Type 'number | false | ((query: Query<ValueType, ErrorType, ValueType, QueryKey>) => number | false | undefined) | undefined' is not assignable to type 'number | false | ((query: Query<ValueType, unknown, ValueType, QueryKey>) => number | false | undefined) | undefined'.
      Type '(query: Query<ValueType, ErrorType, ValueType, QueryKey>) => number | false | undefined' is not assignable to type 'number | false | ((query: Query<ValueType, unknown, ValueType, QueryKey>) => number | false | undefined) | undefined'.
        Type '(query: Query<ValueType, ErrorType, ValueType, QueryKey>) => number | false | undefined' is not assignable to type '(query: Query<ValueType, unknown, ValueType, QueryKey>) => number | false | undefined'.
          Types of parameters 'query' and 'query' are incompatible.
            Type 'Query<ValueType, unknown, ValueType, QueryKey>' is not assignable to type 'Query<ValueType, ErrorType, ValueType, QueryKey>'.
              Types of property 'state' are incompatible.
                Type 'QueryState<ValueType, unknown>' is not assignable to type 'QueryState<ValueType, ErrorType>'.
                  Type 'unknown' is not assignable to type 'ErrorType'.typescript(2322)
```

### Minimal, reproducible example

https://codesandbox.io/p/devbox/rough-brook-drv2hx

### Steps to reproduce

```typescript
const [{ error, data }] = useSuspenseQueries({
  queries: [getSuspenseQueryOptions()],
});

function getSuspenseQueryOptions(): UseSuspenseQueryOptions<ValueType,  ErrorType> {
  return {};
}

type ValueType = { some: string; value: string };
type ErrorType = { error: unknown };
```

ErfanBagheri404:
Fixes #1788

Replaces the `as UseSuspenseQueryOptions` cast on suspense query options functions with TanStack Query's `queryOptions()` builder. This fixes `useSuspenseQueries` TypeScript errors when consuming orval-generated suspense query options.

**What changed:**
- Added `getQueryOptionsHelperTypes()` to the adapter interface to opt in suspense types per framework
- React adapter opts in for `SUSPENSE_QUERY` only (suspense-infinite stays on the cast because the user `Partial<>` override spread makes `getNextPageParam` optional, which `infiniteQueryOptions()` rejects)
- Generated suspense queries now emit `return queryOptionsBuilder({ queryKey, queryFn, ...queryOptions })` instead of `return { ... } as UseSuspenseQueryOptions<...>`
- The `queryOptions` import is aliased to `queryOptionsBuilder` to avoid name-shadowing with the destructured local `const { query: queryOptions } = options ?? {}`
- Mutator path left unwrapped (custom mutators bypass the builder)

**Breaking note:** The `queryOptions()` builder uses its own `DataTag` branding semantics, so the internal query key literal no longer carries the explicit `& { queryKey: DataTag<...> }` annotation. This is the correct approach per [TanStack Query docs](https://tanstack.com/query/latest/docs/framework/react/guides/query-options#queryoptions), but the type semantics are different enough to warrant a major version bump. All existing tests and generated snapshots pass (7017/7017).

## Summary by CodeRabbit

- **Improvements**
  - Generated React Query suspense configurations now use TanStack Query’s recommended option builders where supported.
  - Improved compatibility with TanStack Query v5 through more reliable suspense-query typing and inference.
  - Existing customization remains supported, including custom option transformations and suspense infinite-query configurations.

- **Tests**
  - Added regression coverage for suspense option typing, query-key inference, generated hooks, and supported query-cache operations.
  - Added dedicated compile-time validation for suspense query configurations.


melloware (changes requested):
Build failed

samples/react-query/basic/src/api/endpoints/petstoreFromFileSpecWithTransformer.ts
```diff
@@ -481,11 +482,7 @@ export const getListPetsSuspenseQueryOptions = <
     signal,
   }) => listPets(params, version, signal);
 
-  return { queryKey, queryFn, ...queryOptions } as UseSuspenseQueryOptions<
-    Awaited<ReturnType<typeof listPets>>,
-    TError,
-    TData
-  > & { queryKey: DataTag<QueryKey, TData, TError> };
+  return queryOptionsBuilder({ queryKey, queryFn, ...queryOptions });
```
melloware: The committed snapshot for this sample was not regenerated, so `test:snapshots` will fail in CI.

`samples/react-query/basic/api-generation.spec.ts` diffs `src/api/endpoints` against `samples/react-query/basic/__snapshots__/endpoints`. On `master` those two copies of `petstoreFromFileSpecWithTransformer.ts` are byte-identical; on this branch only the `src/api/endpoints` copy carries the `queryOptions as queryOptionsBuilder` import and this `return queryOptionsBuilder(...)`, while `__snapshots__/endpoints/petstoreFromFileSpecWithTransformer.ts` still has the old `as UseSuspenseQueryOptions<...> & { queryKey: DataTag<...> }` cast.

`.github/workflows/pr-checks.yaml` runs `vp run -w test:snapshots`, which regenerates and compares — it will report the mismatch. Running `bun run test:snapshots:update` (or updating that one file) fixes it.

packages/query/src/query-generator.ts
```diff
@@ -908,11 +920,15 @@ ${hookOptions}
       }
 
    return  ${
-     queryOptionsMutator
-       ? 'customOptions'
-       : `{ queryKey, ${queryFnProperty}, ${queryOptionsImp}}`
+     useQueryOptionsHelper
+       ? // The builder brands the query key and structurally checks the
+         // literal, so no cast is emitted for this branch.
+         `${queryOptionsHelperName}({ queryKey, ${queryFnProperty}, ${queryOptionsImp}})`
```
melloware: Dropping the annotation here silently re-brands the exported `queryKey` and changes the options type — worth calling out in the changeset, and it is inconsistent with the hook in the same generated file.

TanStack's builder returns `queryKey: DataTag<TQueryKey, **TQueryFnData**, TError>`, whereas the cast this replaces used `DataTag<QueryKey, **TData**, TError>`. Emitting `.d.ts` for the new `tests/__snapshots__/react-query/mutator/endpoints.ts` gives:

```ts
export declare const getListPetsSuspenseQueryOptions: <TData = ..., TError = Error>(...) =>
  OmitKeyof<UseQueryOptions<Pets, TError, TData, QueryKey>, "queryFn">
  & { queryFn?: QueryFunction<Pets, QueryKey, never> }
  & { queryKey: QueryKey & { [dataTagSymbol]: Pets; [dataTagErrorSymbol]: TError } };
```

So for a caller that uses `select`:

```ts
const opts = getListPetsSuspenseQueryOptions<string[]>(params, 1, { query: { select: (p) => p.map((x) => x.name) } });
queryClient.setQueryData(opts.queryKey, names); // used to accept string[]; now requires Pets[]
```

I confirmed with `tsc` that `DataTag<QueryKey, string[], MyError>` is no longer assignable from `opts.queryKey` (`TError` still infers correctly; only the data brand moved). Meanwhile `useListPetsSuspense(...)` in the same file still declares `queryKey: DataTag<QueryKey, TData, TError>`, and the non-suspense `getListPetsQueryOptions` keeps the `TData` brand — so the same key now has two different brands depending on which generated symbol you go through.

Secondary effect of the same line: the return type is no longer `UseSuspenseQueryOptions`, it is a plain query-options type, so `enabled` / `throwOnError` / `placeholderData` (deliberately omitted from `UseSuspenseQueryOptions` because suspense ignores them) are now type-visible on the result, e.g. `useSuspenseQuery({ ...getListPetsSuspenseQueryOptions(), enabled: false })` compiles and silently does nothing.

packages/query/src/dependencies.ts
```diff
@@ -152,6 +152,15 @@ const REACT_QUERY_DEPENDENCIES: GeneratorDependency[] = [
...
       { name: 'useInfiniteQuery', values: true },
       { name: 'useSuspenseInfiniteQuery', values: true },
+      // Aliased so the generated options functions can keep their local
+      // `const { query: queryOptions } = options ?? {}` destructure without
+      // shadowing the builder. See #1788.
+      { name: 'queryOptions', alias: 'queryOptionsBuilder', values: true },
+      {
+        name: 'infiniteQueryOptions',
+        alias: 'infiniteQueryOptionsBuilder',
+        values: true,
```
melloware: `infiniteQueryOptionsBuilder` is unreachable today and carries a latent version gap.

The react adapter returns only `[QueryType.SUSPENSE_QUERY]`, so `queryOptionsHelperName`'s `isInfiniteQuery(type) ? 'infiniteQueryOptionsBuilder' : ...` branch in `query-generator.ts` never fires and this export entry is never emitted (imports are filtered by `alias` in `addDependencyFromIdentifiers`, so it is harmless today — just dead).

If a future adapter adds an infinite type to `getQueryOptionsHelperTypes()`, the `hasQueryV5` gate is too coarse: `isQueryV5` is `>= 5.0.0`, and `queryOptions` has been there since 5.0.0 but `infiniteQueryOptions` was added later in the 5.x line. A project on an early 5.x would then get `import { infiniteQueryOptions as infiniteQueryOptionsBuilder } from '@tanstack/react-query'`, which does not exist. Either drop this entry until it is used, or gate it on its own version check like the other `hasQueryV5With*` flags.

melloware (approved):

PR #4130, opened 2026-09-16
fix(core,query): serialize format: date request bodies as calendar days

### LINKED ISSUE #4124: useDatesTransform sends format: date fields as datetimes, so strict date parsers reject them
### Description

`useDatesTransform` deserializes `format: date` response fields to `Date`, but generates no request-side serializer. A `Date` in a `format: date` slot therefore goes out through `Date.prototype.toJSON()` as a full RFC 3339 datetime, which strict `date` parsers (Pydantic v2, Java `LocalDate` bindings) reject.

This is the outbound follow-up noted when `useDatesTransform` landed — https://github.com/orval-labs/orval/issues/1939#issuecomment-4982653944.

### Output client

react-query (axios), and the other axios-based clients wired for `useDatesTransform`.

```ts
output: { override: { useDatesTransform: true } }
```

### Environment

orval 8.27.0 through 8.33.0 (current master).

### Expected behavior

```
PUT /items  {"day": "2026-07-01"}
```

### Actual behavior

```
PUT /items  {"day": "2026-07-01T00:00:00.000Z"}
```

```yaml
openapi: 3.1.0
info: { title: example, version: 1.0.0 }
paths:
  /items:
    put:
      operationId: putItem
      requestBody:
        required: true
        content:
          application/json:
            schema: { $ref: '#/components/schemas/Item' }
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema: { $ref: '#/components/schemas/Item' }
components:
  schemas:
    Item:
      type: object
      required: [day]
      properties:
        day: { type: string, format: date }
```

GET returns `{"day": "2026-07-01"}`; the generated client parses it to a `Date`; PUT sends it back as a datetime.

### Additional context

`format: date-time` is unaffected — `toJSON()` is already correct for an instant.

On the calendar-day convention: orval already treats a `Date` in a `format: date` slot as its **UTC** calendar day. `packages/mock/src/faker/constants.ts:11` renders that format as `faker.date.past().toISOString().slice(0, 10)`, and with `useDates` on the mock generator already round-trips exactly that way (`new Date(faker.date.past().toISOString().slice(0, 10))`). No local-parts date formatting exists anywhere in the repo. So the fix should follow the existing convention rather than introduce a second one, which also means inbound parsing does not need to change.

Happy to open the PR — the implementation is ready.

sami286:
Closes #4124.

`useDatesTransform` converts `format: date` response fields to `Date` but generated no request serializer, so those fields went out as full RFC 3339 datetimes and strict `date` parsers rejected them. This adds the outbound half that #1939 noted was missing.

An operation whose JSON request body declares a `format: date` field now gets a generated `serialize<Op>Request`, chained at the call site:

```ts
export const updateAppointment = (appointment: Appointment) => {
  return customInstance<Appointment>({
    url: `/appointments`,
    method: 'PUT',
    headers: { 'Content-Type': 'application/json' },
    data: serializeUpdateAppointmentRequest(appointment),
  }).then(deserializeUpdateAppointmentResponse);
};
```

**Convention.** A `Date` in a `format: date` slot is read as its UTC calendar day (`toISOString().slice(0, 10)`) — the convention orval's mock generator already uses for that format, so clients and mocks agree. Inbound parsing is unchanged.

**Scope.** `format: date-time` emits nothing (`toJSON()` is already correct). `readOnly` properties are skipped, since OpenAPI `readOnly` means response-only. Only single-JSON-media-type bodies are covered; multi-media-type, multi-JSON and blob bodies emit nothing rather than wrong code, as does a schema that is simultaneously array- and object-shaped. Date-only values in query and path parameters are still ISO datetimes — documented as a known gap.

**On the shared walk.** The first commit extracts the existing response walk behind a `DateTransformMode` so both directions share ref resolution, cycle guards, `allOf` merging, property iteration and discriminated-union dispatch; only the leaf, container traversal and property filtering differ. It is a pure refactor: no emitted output changes, which the existing tests and zero fixture drift both confirm. The alternative was a second parallel walk, which would have duplicated ~300 lines that then rot apart.

One consequence worth flagging: the request leaf guards on `instanceof Date` so it is idempotent. The response leaf is idempotent by accident (`new Date(aDate)` clones), but `toISOString()` is not, and a schema that re-declares an inherited date property (`allOf: [$ref Base, {properties: {createdOn: ...}}]`) legitimately reaches the same accessor twice.

**Testing.** Unit coverage for the emitter and the generator function; wiring tests for both the mutator and plain-axios branches including ordering and flag-off; a runtime spec that stubs the axios adapter and asserts the actual JSON body (`{"day":"2026-07-01"}`, with `date-time` keeping its instant) and that the caller's object is not mutated; plus fixture coverage in `tests/specifications/dates-transform.yaml`, which had no `format: date` fields before.

With the flag off, generated output is byte-identical.

A follow-up PR adds `additionalProperties` map traversal on top of this one.

## Summary by CodeRabbit

* **New Features**
  * Request bodies can serialize OpenAPI `format: date` fields as UTC calendar dates when date transformation is enabled.
  * Supports nested objects, arrays, nullable fields, optional bodies, and preserves original request data.
  * Date-time fields and query/path date parameters retain existing behavior.

* **Bug Fixes**
  * Read-only properties are excluded from request serialization.
  * Unsupported body configurations are skipped, including multiple media types, blob bodies, and ambiguous schemas.

* **Documentation**
  * Expanded guidance on request-body serialization behavior and limitations.


melloware:
Looks like merge conflicts

packages/core/src/generators/date-transform.ts
```diff
@@ -193,13 +199,20 @@ const buildResolvedStatements = ({
...
+  if (mode.isLeaf(schema)) {
     result = {
-      statements: [`${writeAccessor ?? accessor} = new Date(${accessor});`],
+      statements: [mode.leafStatement(accessor, writeAccessor ?? accessor)],
       cyclicRefs: new Set(),
     };
+  } else if (
+    mode.dropArrayObjectConflict &&
+    schema.items &&
+    needsObjectCopy(schema, context)
```
melloware, line 213: The array/object conflict guard only looks at `schema.items` on the resolved schema, so it misses the case where the array shape arrives through an `allOf` branch instead of a sibling `items`. `needsObjectCopy` *does* recurse into `allOf`, so the two halves disagree.

Repro (verified against this branch by calling `buildRequestDateSerializeStatements` directly):

```yaml
x:
  allOf:
    - { type: array, items: { type: string, format: date } }
    - { type: object, required: [d], properties: { d: { type: string, format: date } } }
```

emits

```ts
copy.x = { ...copy.x };
copy.x = copy.x.map((item0) => { /* ... */ });   // TypeError: copy.x.map is not a function
copy.x.d = /* ... */;
```

and the same schema at the body root produces `const copy = { ...data };` followed by `copy = copy.map(...)`, i.e. assignment to a `const` — the exact two failure modes the `dropArrayObjectConflict` doc comment says it prevents. Checking for an array-shaped `allOf` branch (a mirror of `needsObjectCopy`) would close it.

packages/core/src/generators/date-transform.ts
```diff
@@ -381,27 +486,33 @@ const buildPropertiesStatements = ({
...
+        writeAccessor: mode.propertyWriteAccessor(
+          accessor,
+          key,
+          propertySchema,
+        ),
+        mode,
       });
       if (inner.statements.length === 0) return inner;
 
+      const statements = [
```
melloware, line 529: The copy prelude runs unguarded for a **required, non-nullable** object property, so a caller that omits it silently gains a key it never sent: `copy.a = { ...copy.a }` with `copy.a === undefined` evaluates to `{}`, and the request body goes out as `{"a":{}}` instead of `{}`.

That is a new behaviour, not just a stricter one — before this PR the field was simply absent and the server's own validation reported the missing field. It bites untyped (JS) consumers and anyone building a body incrementally. For comparison, the required-array case fails loudly instead (`copy.slots = copy.slots.map(...)` throws `Cannot read properties of undefined (reading 'map')`), so the two container kinds also disagree with each other.

Emitting the prelude only under `if (target != null)` regardless of `required` would make both cases a no-op and keep the payload byte-identical to what the caller passed.

packages/core/src/generators/date-transform.ts
```diff
@@ -495,7 +609,260 @@ export const buildDateTransformStatements = ({
...
+ * Guarded on `instanceof Date`: the same accessor can be reached twice (an
+ * `allOf` branch re-declaring a property its base already declares, or a
+ * discriminated-union variant re-declaring one its parent schema's own
+ * `properties` also converts), and unlike the response leaf's `new Date(...)`
+ * — a clone, harmless to repeat — this conversion produces a string. Running
+ * it twice without the guard would call `.toISOString()` on that string and
+ * throw at runtime. The guard makes a second pass a no-op instead.
+ */
+const serializeLeafStatement = (read: string, write: string): string =>
+  `${write} = ${read} instanceof Date ? (${read}.toISOString().slice(0, 10) as unknown as Date) : ${read};`;
```
melloware, line 666: Flagging the UTC convention as the most likely real-world failure, since it is silent and off by a whole day.

The round trip through orval's own deserializer is consistent (`new Date("2026-07-01")` -> UTC midnight -> `.toISOString().slice(0, 10)` -> `"2026-07-01"`), but a `Date` that did not come from a response almost never is. Every common browser source of a date is local midnight — `new Date(2026, 6, 1)`, `<input type="date">` parsed with `new Date(y, m, d)`, most date-picker components — and at any positive UTC offset that is the previous day in UTC: a user in Berlin (UTC+2) picking 1 July sends `"2026-06-30"`. No error, no type complaint, just the wrong day on the server.

I see this is documented and that it matches the mock generator, so it may well be the intended trade-off — but `getFullYear()`/`getMonth()`/`getDate()` would be correct for both sources (a response-derived UTC-midnight `Date` still reads back as its own calendar day locally, except at negative offsets... which is the symmetric problem). Worth a deliberate decision rather than inheriting the mock generator's convention, given `format: date` bodies are usually user-entered.

melloware (changes requested):
See feedback

packages/core/src/generators/date-transform.ts
sami286: This is a design/product decision we should discuss, I preferred using UTC over local time since its the pattern used by mock generators.
Both approaches have pros and cons, I have no preference but I understand this is important to be discussed openly

packages/core/src/generators/date-transform.ts
sami286: Good catch — fixed in c190412d6. The conflict guard now uses an `isArrayShaped` check that recurses through `allOf`/`$ref` the same way `needsObjectCopy` does, so the two halves agree: your schema emits nothing in the request direction, both as a property and at the body root. An `allOf` with only an array branch still gets its `.map` conversion, and the response direction is unchanged. Both shapes are covered by new tests.

packages/core/src/generators/date-transform.ts
sami286: Agreed — fixed in c190412d6. In the request direction, any property whose statements write through a container (object copy, array `.map`, union dispatch) is now wrapped in `if (x != null)` regardless of `required`, so an omitted container stays out of the payload instead of becoming `{}` or throwing, and objects and arrays now behave the same. Required date leaves stay unguarded, since `x instanceof Date ? … : x` already leaves `undefined` alone. There's a unit test that executes the emitted serializer on a body missing a required object and a required array, plus a runtime test through the generated `updateAppointment`. The response direction is unchanged.

packages/core/src/generators/date-transform.ts
melloware: Agreed

melloware (approved):

PR #4132, opened 2026-09-16
fix(query): add exactMatch parameter to setQueriesData for better cache control

### LINKED ISSUE #3260: Bulk updates when setting TanStack query data
### Description

Generated helper functions for updating cached query data when enabling `override.query.useSetQueryData` are something like this:

```ts
export const useSetListPetsQueryData = () => {
  return (
    params: ListPetsParams, // Note, required
    updater:
      | Awaited<ReturnType<typeof listPets>>
      | undefined
      | ((
          old: Awaited<ReturnType<typeof listPets>> | undefined,
        ) => Awaited<ReturnType<typeof listPets>> | undefined),
  ) => {
    queryClient.setQueryData(getListPetsQueryKey(params), updater);
  };
};
```

Note, that in `getListPetsQueryKey`, params are optional

```ts
export const getListPetsQueryKey = (params?: ListPetsParams) => {
  return [`/pets`, ...(params ? [params] : [])] as const;
};
```

### Current behavior

What if we update the name of a pet? We want to update the pet's name in every cached list we have regardless of params like `sortOrder` or `search` or whatever is in `ListPetsParams`.

Since `params` is required, there's no way to easily update all queries for pets regardless of params (even though the query key getter doesn't require params).

### Proposed behavior

Support updating based on the prefix of the query key (i.e. without params). This would allow something like

```ts
const updatePetList = useSetListPetsQueryData();
useUpdatePet({
  mutation: {
    onSuccess: (pet) => updatePetList(undefined /* regardless of params */, (old) => {
      // update pet details based on matching id
    })
  }
})
```

### Proposed solution

TanStack supports this if you use `setQueriesData` instead of `setQueryData`:

From [`setQueryData`](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientsetquerydata):

> To update multiple queries at once and match query keys partially, you need to use [queryClient.setQueriesData](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientsetqueriesdata) instead.

From [`setQueriesData`](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientsetqueriesdata):
> `setQueriesData` is a synchronous function that can be used to immediately update cached data of multiple queries by using filter function or **partially matching the query key**.

ephibian-adietrich:
## Summary

- Add `exactMatch` parameter to generated `useSetXxxQueryData` / `setXxxQueryData` helpers to control the forwarded call to `queryClient.setQueriesData<TData>(...)` so that calls can specify exact match queries rather than prefix-path sharing queries.
- Set this parameter to `true` by default to match existing (pre 8.11.0) behavior.

## Summary by CodeRabbit

* **Behavior Changes**
  * Query data helpers now use exact query-key matching by default when updating cached data.
  * Set `exactMatch` to `false` to update entries sharing a query-key prefix.
  * With unspecified parameters or request bodies, prefix matching can update all matching cached entries.
  * Updates affect existing cached entries only; new entries are not created.

* **Documentation**
  * Updated framework-specific guides and configuration references with matching examples and upgrade guidance.


ephibian-adietrich:
Linking to #3260

melloware (changes requested):
See AI feedback

melloware:
## Review

Thanks for this! The generator change itself is sound — I verified all 6 affected snapshot fixtures were genuinely regenerated (the multi-line vs. single-line breaks match prettier at 80 cols for each key expression), no `samples/` output uses `setQueriesData`, and no unit test asserts these signatures.

The findings are all documentation/comment drift: flipping the default to `exact: true` invalidates several statements that 8.11.0 introduced and that this PR left in place.

### Medium

1. **`docs/content/docs/reference/configuration/output.mdx:1389`** — not updated, and now factually wrong: *"Query keys are matched by prefix … Passing `undefined` updates every cached entry that shares the same path."* With `exact` defaulting to `true`, `setListPetsQueryData(undefined, updater)` matches nothing and silently updates zero entries.

2. **`docs/content/docs/zh/reference/configuration/output.mdx:1244`** — same stale claim in the Chinese reference.

3. **`docs/content/docs/zh/guides/{react,angular,vue,solid,svelte}-query.mdx`** — all five Chinese guides still show the pre-change helper signature (no `$exactMatch`, no `exact:`) and state prefix matching is the default. A zh reader copies the documented `updatePetList(undefined, updater)` call and gets a silent no-op.

4. **`packages/query/src/query-generator.ts:1051`** — the comment justifying `widenNonPath: allowUndefinedParam` still says *"passing `undefined` updates every cached entry sharing the path prefix"*. That's exactly the behavior the new default disables, so the comment now argues for a feature the code turns off — a maintainer reads it as a reason to keep widening without learning that `$exactMatch: false` is required.

### Low

5. **`packages/query/src/query-generator.ts:319`, `:324`** — `$exactMatch` was moved out of `propsSig` into the renderer's two template literals, but the `body` that consumes it is still supplied by the caller (~line 1146). Nothing ties the two together, and the renderer's JSDoc gave no hint: a future caller passing a body that doesn't reference `$exactMatch` emits an unused parameter into generated code.

6. **`docs/content/docs/guides/react-query.mdx:159`** — the "those args are widened to accept `undefined`" sentence was deleted here but kept in the other four guides, so the React guide never explains why `params` accepts `undefined`.

7. **`docs/content/docs/guides/vue-query.mdx:172`** — trailing whitespace inside the `ts` code fence.

8. **English guides Callout (all five)** — `$exactMatch` written unbackticked in MDX prose; *"Prior to 8.11.0 … After this change"* mixes two different release anchors and reads as a single change; missing terminal period.

### Behavior note (not a defect — intended, but worth calling out)

9. Defaulting `exact: true` silently turns the prefix-update pattern that 8.11.0 introduced *and documented* (`setXxxQueryData(undefined, updater)`) into a no-op, with no runtime warning. Anyone who adopted the 8.11.0 helper gets a second silent semantics change in consecutive releases — this deserves explicit release-note framing rather than only a docs footnote.

   Also worth correcting in the PR description: `setQueriesData` with `exact: true` still never *creates* a missing entry, so this does not fully restore pre-8.11.0 `setQueryData` behavior.

### Pre-existing gap (not from this diff)

No svelte/solid/angular snapshot fixture exercises `useSetQueryData`, so the non-React `(queryClient, …)` shape of this helper has zero generated coverage.

---

I have fixes for 1–8 ready locally (docs rewrites in both locales, the two generator comments — comments only, so the committed snapshots stay correct; `typecheck` passes). Happy to push them to this branch or open a follow-up, whichever you prefer.

melloware:
Pushed the fixes for 1–8 to this branch as 7b08a23 — docs rewrites in both locales plus the two generator comments. Comments and docs only, so the committed snapshots stay correct; `typecheck` passes.

Left #9 alone since that's the intent of the PR — just flagging it for the release notes.

melloware (approved):

ephibian-adietrich:
Yep, that all sounds good to me. 

I didn't touch the Chinese documentation because I sadly do not speak the language, so I appreciate the assist.

Agreed on the flag for release notes. There is an argument for switching to a split between setQueriesData and setQueryData based on the $exactMatch flag to _create_ a query key entry where none previously exists (or even adding an options structure to modify defaults, but that may just overcomplicate the helper). I leaned toward the single method call to keep the generated relatively small/simple, but it should be an easy change if needed in the future.

PR #4139, opened 2026-09-17
fix(core): guard required containers in the response date deserializer

sami286:
With `useDatesTransform`, the generated response deserializer walks required properties without a null guard. For a container that means code like:

```ts
for (let i0 = 0; i0 < data.slots.length; i0++) { … }
```

so a response that omits a required array, object or discriminated-union property throws `TypeError: Cannot read properties of undefined` inside the generated `.then(deserialize…)`, before the caller can handle anything. It has been there since `useDatesTransform` shipped (#3731).

#4130 fixed the same problem on the request side behind a `guardRequiredContainers` mode flag that was only enabled for requests. This enables the guard for responses too; since the flag is now `true` in both directions it is removed rather than left as a constant:

```ts
const deserializeUpdateAppointmentResponse = (data: Appointment): Appointment => {
  if (data == null) return data;
  data.day = new Date(data.day);
  if (data.slots != null) {
    for (let i0 = 0; i0 < data.slots.length; i0++) { … }
  }
  return data;
};
```

**Scope.** Only containers are guarded. Required date leaves are unchanged in both directions, so a response missing a required scalar date still yields `Invalid Date` exactly as before — guarding every leaf would touch every generated deserializer for a case that doesn't throw. Request-side output is byte-identical.

**Testing.** Unit tests for required objects, arrays and discriminated-union properties, plus one that executes the emitted deserializer on a payload missing a required object and array and checks it neither throws nor adds keys. The runtime spec's mock response no longer needs the `{ slots: [] }` workaround, and a new runtime test asserts a response omitting `slots` resolves. The fixture snapshot diff is only the new guards inside `deserialize…Response` functions.

Since #4131 landed, a required `additionalProperties` map is covered too: the map loop in the deserializer is guarded the same way, with unit tests, an executed test and a runtime test for a response omitting the required map.

## Summary by CodeRabbit

- **Bug Fixes**
  - Response processing now safely handles omitted required arrays, objects, and maps without throwing errors.
  - Date transformations now correctly process values nested within map-like response and request data.
  - Improved handling for nested objects, arrays, maps, and discriminated data structures.
  - Empty or mixed array-and-map structures are handled more reliably.
  - Omitted response containers no longer create fabricated properties or keys.


melloware:
OK you can rebase this one now!

melloware (approved):

PR #4155, opened 2026-09-19
refactor: remove downstream OAS 3.0 handling from generators

### LINKED ISSUE #4113: Stop using OAS 3.0 in tests

### LINKED ISSUE #4119: refactor: remove downstream OAS 3.0 handling from generators
Sub-issue of #4113. **Last. Optional — worth an explicit decision before anyone starts.**

## Context

Once `resolveSpec` provably emits 3.1-only documents and every post-boundary fixture matches, the 3.0-handling branches scattered through the generators are unreachable. This issue is about whether to delete them.

## Surface

~175 non-test `nullable` references across the generator packages:

| File | Count |
| --- | --- |
| `core/src/getters/combine.ts` | 21 |
| `zod/src/index.ts` | 20 |
| `core/src/getters/object.ts` | 19 |
| `core/src/getters/scalar.ts` | 17 |
| `core/src/generators/date-transform.ts` | 15 |
| `mock/src/faker/resolvers/value.ts` | 13 |
| `effect/src/index.ts` | 11 |
| remainder (ref, enum, options, assertion, …) | ~59 |

Some of these are the internal `nullable` support fields (`combineSchemas({ nullable: '' })`, `GetObjectOptions.nullable`), **not** OAS `nullable` — they stay. Separating the two is the first task here, and it is not mechanical.

## The decision to make first

Deleting this code is the only genuinely risky step in #4113. Arguments both ways:

- **For:** it is the actual payoff. Two dialects in the generators is the cost the whole tracking issue is trying to remove.
- **Against:** it is defence-in-depth against a spec path that skips or partially fails the upgrade, and against future Scalar regressions. It costs nothing at runtime.

A middle option: keep the branches but make them assert/warn instead of silently handling, so a residual 3.0 keyword becomes a reported bug rather than a quiet fallback.

## If it proceeds

- [ ] Classify every `nullable` reference as OAS keyword vs. internal support field
- [ ] Remove only the unreachable OAS-keyword branches, one package per PR
- [ ] Full snapshot + sample suites green, with diffs reviewed rather than accepted wholesale

melloware:
Closes #4119. Last sub-issue of #4113.

## The decision

#4119 asked for an explicit decision before anyone started: delete the unreachable OAS 3.0 branches, make them warn, or leave them as defence-in-depth. This PR **deletes them**. Two dialects in the generators is the cost the tracking issue set out to remove, and the invariant is now guarded rather than assumed.

## Why this is safe now

`resolveSpec` runs `upgrade()` and then `normalizeToOpenApi31()` (`packages/orval/src/import-specs.ts`), and the boundary guard added in #4140 (`import-specs.test.ts`, `resolveSpec boundary`) walks the document handed to `importOpenApi` and fails on any residual 3.0 keyword — `nullable`, a boolean or undefined `exclusiveMinimum`/`exclusiveMaximum`, or a `format` of `binary`/`base64`/`byte`.

The user `input.override.transformer` hook runs *before* both, so nothing downstream of the boundary can reintroduce 3.0 syntax.

## Classification

The ~175 `nullable` references split two ways, and separating them was the first task in the issue:

**OAS keyword reads — removed (~18 sites).** Most carried an `as boolean | undefined` cast or an `'nullable' in schema` probe, because `OpenAPIV3_1.SchemaObject` has no such keyword. The types already said this code was unreachable.

| Package | Sites |
| --- | --- |
| `core` | `getters/combine.ts` (3) · `getters/scalar.ts` · `getters/res-req-types.ts` (2) · `generators/date-transform.ts` · `generators/schema-definition.ts` · `resolvers/ref.ts` · `resolvers/value.ts` · `utils/assertion.ts` |
| `zod` | the `nullable` const, `CHAINABLE_SIBLINGS`, `applyChainableSiblings`, three intersection casts |
| `effect` | the `nullable` const |
| `mock` | `isSchemaNullableAtRoot` · `isNullableArrayItem` · the `nullable` spread in `resolveMockValue` |

**Internal support fields — untouched.** `combineSchemas({ nullable: '' })`, `GetObjectOptions.nullable`, the local `const nullable = ' | null'` suffixes, `getEnum(..., nullable: boolean)`, `getNullable(value, nullable)`, `nullableAtRoot` and `nonNullable` are orval's own vocabulary, not the OAS keyword.

Also removed: the boolean spelling of `exclusiveMinimum`/`exclusiveMaximum` in `zod` and `effect`, which ran an `isBoolean(raw) ? (raw ? min : undefined) : raw` conversion under a comment headed *"OpenAPI 3.0 vs 3.1 compatibility"*. The same guard covers it. The `mock` faker scalar only had a stale comment — its code already read the numeric form.

## Deliberately out of scope: `format: binary`

`format: 'binary'` is on the guard's list, so the incoming reads look dead — but orval **synthesizes it internally**. `mock/src/msw/mocks.ts` builds `{ type: 'string', format: 'binary' }`, and `content-type.ts` uses `'binary' | 'text'` as its own fileType token that `core/getters/scalar.ts`, `body.ts`, `response.ts`, `res-req-types.ts`, `zod`, `effect` and `mock-types.ts` all read.

Deleting those would not be removing dead 3.0 handling — it would be migrating orval's file-type vocabulary to `contentMediaType`, which decides `Blob | File` typing. That is a different and riskier change and belongs in its own issue.

## Tests

The ~24 post-boundary cases that asserted the deleted behaviour are gone or rewritten:

- **Rewritten to the 3.1 spelling** where the coverage still means something — `{ $ref, type: ['object', 'null'] }` for the two zod split-mode regressions (that sibling *is* still propagated by `resolvers/ref.ts`), `{ type: ['string', 'null'] }` for the `#3780` member row, which renders `zod.string().nullable()` exactly as the old row rendered `zod.unknown().nullable()`, and a `description` sibling for the `$dynamicRef` reusable-schema case so that path keeps coverage.
- **Deleted** where the case only ever exercised the dead branch, or had an exact 3.1 twin asserting the same output. Two mock cases had become vacuous — with the keyword ignored the schema is simply not nullable, so "does not null-randomize" passed for the wrong reason.
- Three of the removed zod/core cases used `as never` / `as unknown as` on `{ $ref, nullable: true }`, the masked-`any` casts #4113 asks to remove.

Comments that *describe* the boundary (what `resolveSpec` makes of a 3.0 input) are left in place — they explain why the surviving code reads only the 3.1 spellings.

## Verification

- `bun run test` — 2829 core + 417 zod + 419 mock + 41 effect + all remaining packages, **0 failures**
- `bun run typecheck`, `bun run lint`, `bun run format:check` — clean
- `bun run test:snapshots` — full rebuild, regeneration of every sample and `tests/specifications`, then the snapshot suites (7256 assertions in `tests` alone), **0 failures**
- **`git status` after regeneration: zero changed files.** The corpus was genuinely rewritten (verified by mtime) and came back byte-identical, which is the bar #4119 set: no generated output moves.

## Summary by CodeRabbit

- **Behavior Changes**
  - Nullability is now recognized from OpenAPI 3.1 `type` unions containing `null`, rather than the OpenAPI 3.0 `nullable: true` flag.
  - Generated types, validation schemas, request/form serialization, references, mocks, and Zod output now follow this standardized nullability format.
  - Exclusive minimum and maximum constraints now use OpenAPI 3.1 numeric values directly; legacy boolean representations are no longer handled at this stage.
- **Tests**
  - Updated coverage to reflect OpenAPI 3.1 nullability and constraint formats.


snebjorn (approved):
LGTM

No snapshot changes means that this is most likely safe :)

PR #4179, opened 2026-09-21
fix(core): only url-encode path parameters under urlEncodeParameters

### LINKED ISSUE #4178: `urlEncodeParameters` also encodes a runtime `baseUrl`
### Description

With `urlEncodeParameters: true` and a runtime `baseUrl`, the base URL also gets wrapped in `encodeURIComponent`. So `http://localhost:3000` ends up as `http%3A%2F%2Flocalhost%3A3000` and every request goes to a broken URL. Same thing happens with `fetch`, `angular` and `react-query`.

### Output client

fetch

```ts

export default defineConfig({
  api: {
    input: './spec.yaml',
    output: {
      target: './src/api.ts',
      client: 'fetch',
      urlEncodeParameters: true,
      baseUrl: {
        runtime: 'getBaseUrl()',
        imports: [{ name: 'getBaseUrl', importPath: './base-url' }],
      },
    },
  },
});
```

### Environment

```
  System:
    OS: macOS 15.6.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 690.09 MB / 32.00 GB
    Shell: Unknown
  npmPackages:
    orval: ^8.36.0 => 8.36.0
```

### Expected behavior

Only the path param is encoded:

```ts
return `${getBaseUrl()}/items/${encodeURIComponent(String(id))}`
```

### Actual behavior

The base URL is encoded too:

```ts
return `${encodeURIComponent(String(getBaseUrl()))}/items/${encodeURIComponent(String(id))}`
```

```yaml
openapi: 3.0.3
info:
  title: Demo
  version: 0.0.0
paths:
  /items/{id}:
    get:
      operationId: getItem
      parameters:
        - name: id
          in: path
          required: true
          schema:
            type: string
      responses:
        '200':
          description: OK
```

### Additional context

`makeRouteSafe` wraps every `${...}` in the route, and by then the runtime base URL has already been joined in, so it gets wrapped along with the params. Similar to #3910, except here it's the base URL instead of an `allowReserved` param. I have a fix that only wraps the actual path params and can open a PR.

mridang:
`makeRouteSafe` wrapped every `${...}` in the route in `encodeURIComponent`, including a runtime `baseUrl` or runtime server variable, so with `urlEncodeParameters: true` the base URL got encoded and every request broke.

`wrapRouteParameters` and `makeRouteSafe` now take an optional `only` set, and the fetch, angular and query clients pass their path param names, so only real path params get wrapped. Callers that don't pass it behave as before. Added two tests in `route.test.ts` for a runtime base URL and a runtime server variable. Existing snapshots are unchanged.

Fixes #4178

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected URL encoding for generated requests so runtime base URLs and server-provided values remain intact.
  - Path parameters continue to be safely encoded, including when requests use URL encoding or reserved-parameter settings.
  - Applied consistent route handling across supported request clients and resource integrations.
  - Added coverage for routes combining runtime substitutions with standard path parameters.


melloware (approved):

mridang:
@melloware thank you for the promptness. 🙇🏽‍♂️

Issue #2410, opened 2025-09-29
C# IFormFile resolves to Blob in generated Typescript files


Zylvian:
`TS2345: Argument of type Blob is not assignable to parameter of type string`

Generated `observation.ts`:
```ts
const formUrlEncoded = new URLSearchParams();

if(postObservationCreateBody.Image !== undefined) {
 formUrlEncoded.append(`Image`, postObservationCreateBody.Image)
 }

...

export type PostObservationCreateBody = {
  Count?: number;
  UserId?: string;
  Density?: number;
  TowVolume?: number;
  TowDistance?: number;
  FrameDiameter?: number;
  SampleType?: string;
  TaxaID?: number;
  Latitude?: number;
  Longitude?: number;
  Image?: IFormFile;
  UploadedImageId?: string;
};

...

export type IFormFile = Blob;

```

From this C# endpoint and DTO:
```cs
    public async Task<ActionResult<ObservationDto>> CreateObservation([FromForm] CreateObservationDto dto)
    {
        
        var created = await repository.CreateAsync(dto);
        return CreatedAtAction(nameof(GetObservationById), new { id = created.Id }, created);
    }
```
```cs
public record CreateObservationDto
{
    public IFormFile? Image { get; set; }
}
```

Let me know if more data is required! Orval and Typescript have been updated to their latest versions.

Zylvian:
THinking the issue is related to the URLSearchParams, as I'm not sure why it does "FromForm" to URL Search params.

melloware:
you might want to post the OpenAPI JSON that is creating this as Orval only looks at the OpenAPI spec.

Zylvian:
[openApi_blob_issue.json](https://github.com/user-attachments/files/22636437/openApi_blob_issue.json)

@melloware Here you go! :-)

melloware:
Yep you are right nothing stands out as odd to me.

Issue #4180, opened 2026-09-22
Query Invalidation - onSuccess Type Error


fabiankranewitter-aqt:
### Description

Query invalidation is currently not working due to a TypeScript type error in the generated onSuccess callback.

The issue occurs when the query invalidation feature is enabled, in both cases:

When using a custom mutator/modifier.
When using the default generated client without a custom mutator/modifier.

The error appears to be caused by a mismatch between Orval's generated TContext type and the TContext | undefined expected by the installed TanStack Query types.

Without **mutationOptions**:
```ts
src/api/client/new/example/client.ts:343:46 - error TS2322: Type '(data: Awaited<ReturnType<typeof postNotes>>, variables: PostNotesMutationVariables, onMutateResult: TContext, context: MutationFunctionContext) => void' is not assignable to type '(data: postNotesResponse200 & { headers: Headers; }, variables: PostNotesMutationVariables, onMutateResult: TContext | undefined, context: MutationFunctionContext) => unknown'.
  Types of parameters 'onMutateResult' and 'onMutateResult' are incompatible.
    Type 'TContext | undefined' is not assignable to type 'TContext'.
      'TContext' could be instantiated with an arbitrary type which could be unrelated to 'TContext | undefined'.

343     return { ...mutationOptions, mutationFn, onSuccess };
                                                 ~~~~~~~~~

  node_modules/.pnpm/@tanstack+query-core@5.89.0/node_modules/@tanstack/query-core/build/modern/hydration-iULCH7y8.d.ts:1196:5
    1196     onSuccess?: (data: TData, variables: TVariables, onMutateResult: TOnMutateResult | undefined, context: MutationFunctionContext) => Promise<unknown> | unknown;
             ~~~~~~~~~
    The expected type comes from property 'onSuccess' which is declared here on type 'UseMutationOptions<postNotesResponse200 & { headers: Headers; }, TError, PostNotesMutationVariables, TContext>'
```

With **mutationOptions**:
```ts
src/api/client/new/example/client.ts:345:43 - error TS2322: Type '(data: Awaited<ReturnType<typeof postNotes>>, variables: PostNotesMutationVariables, onMutateResult: TContext, context: MutationFunctionContext) => void' is not assignable to type '(data: postNotesResponse200 & { headers: Headers; }, variables: PostNotesMutationVariables, onMutateResult: TContext | undefined, context: MutationFunctionContext) => unknown'.
[...]

PR #4183, opened 2026-09-22
fix(core): type multipart file parts in shared request body schemas

### LINKED ISSUE #4177: Multipart file uploads with shared request body schema do not get `File` type
### Description

Edge case followup from #3662: the implementation currently supports generating `File | Blob` types for multipart request bodies with file uploads. However, this does not work properly when the request body is aliased as a shared schema in the OpenAPI spec. Presumably this is because, when generating the shared schema type, `formDataContext` is `false` ([here](https://github.com/orval-labs/orval/blob/a146e4e5b5dd2c8f460d19d842425177e2212d1b/packages/core/src/getters/scalar.ts#L228)).

It would be great to support the `File | Blob` pattern in shared schemas as well.

### Output client

fetch

### Environment

Orval 8.33.0

```yaml
openapi: 3.1.0
info:
  title: Demo
  version: 0.0.0
paths:
  /foo:
    post:
      description: This is an upload endpoint
      requestBody:
        required: true
        content:
          multipart/form-data:
            schema:
              $ref: "#/components/schemas/upload"
      responses:
        '204':
          description: Success
components:
  schemas:
    upload:
      type: object
      properties:
        bar:
          type: string
        files:
          type: array
          items:
            type: string
            contentMediaType: application/octet-stream
```

melloware:
Fixes #4177

### Problem

A `multipart/form-data` body written **inline** gets a `FormDataContext` from `getResReqContentTypes`, which is what types its binary parts as `Blob | File` (#3662 / #3915) and applies the media type `encoding`.

A body **aliased behind a shared schema** (`$ref: '#/components/schemas/upload'`) is emitted once by `generateSchemasDefinition` instead, detached from the operation that declares the media type — so those parts fell back to the context-free `Blob`.

Same spec, two bodies, before this PR:

```ts
export interface Upload {            // $ref'd shared schema
  files?: Blob[];
}

export type UploadInlineBody = {     // inline schema
  files?: (Blob | File)[];
};
```

A part carrying `encoding.contentType: text/csv` was worse than cosmetic: the model said `string` while the generated FormData builder already wrapped the value in a `Blob`.

### Fix

Index the `components.schemas` entries used directly as a multipart body, together with the `encoding` of the media type referencing them, and generate those schemas with that context.

- `application/x-www-form-urlencoded` deliberately stays out of the index: its context narrows binary fields to `string`, and the shared model may well be referenced by a multipart endpoint too, where `Blob` is correct (#2410).
- The index is memoized per spec document (a `WeakMap` keyed on `context.spec`) rather than on the context object, since contexts are shallow-copied throughout the generators.
- Request bodies declared under `components.requestBodies` are swept too, so `$ref`'d bodies are covered.

### Impact

Every resulting change **widens** a request-body property (`Blob` → `Blob | File`, `string` → `Blob | File | string`), so callers that build these bodies keep compiling. Schemas that are never a multipart body are untouched — the new regression spec pins that with a `Metadata` component used only as a JSON response.

Four existing snapshots move from `Blob` to `Blob | File`; all four are shared multipart bodies.

### Tests

- `packages/core/src/getters/form-data-components.test.ts` — the index itself (paths, `components.requestBodies`, encoding merge, url-encoded and inline exclusions).
- `packages/core/src/generators/schema-definition.test.ts` — the emitted models.
- `tests/specifications/issue-4177-shared-form-data-body.yaml` + `fetch.config.ts` — end-to-end snapshot covering the shared body, the inline body next to it, and an untouched non-multipart component.

`vp check`, `vp test`, the generated-output typecheck and the snapshot suites all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Shared schemas used in `multipart/form-data` request bodies now generate binary fields with the correct `Blob | File` typing.
  * Multipart field encoding is preserved when schemas are reused across endpoints, including schemas extended through aliases.
  * JSON, response-only, and URL-encoded usages remain unaffected.

* **Tests**
  * Added coverage for shared multipart schemas, request-body references, encoding behavior, alias extensions, and mixed usage scenarios.


melloware:
Both points addressed.

**Out-of-scope `packages/query` changes** — real, and my mistake: I branched off `fix/4180-optional-on-mutate-result` instead of `master`, so the PR carried #4180's commit. Since #4180 landed as 67d3ea7da, I rebased onto `master` and dropped the duplicate. The PR is now the #4177 commit plus the fix below.

**`indexBySpec` keyed on document identity** — valid, and the cache did not earn its keep. `generateSchemasDefinition` is the only caller, so the index is now built once at the top of it and each schema's context is passed down. Same single scan per generation, no state surviving between calls, so a document that changes between runs can no longer be described by a stale index. `schema-definition.test.ts` gained a regression test that runs the generator twice against the same document object with `paths` emptied in between and asserts the second run drops back to `Blob`.

`vp check`, `vp test`, the snapshot suites (7441) and the generated-output typecheck (21 runs) all pass; samples regenerate with no diff.

melloware:
Both fixed.

**Nested ref passing the prefix check** — valid. `#/components/schemas/Wrapper/properties/upload` starts with the prefix, and `getRefInfo` names a ref after its last token, so an unrelated `components.schemas.upload` could have been handed the form-data context. Rather than a `refPaths.length` check I gated on the existing `isComponentRef`, which already encodes "a single named component slot" (`^#/components/(schemas|…)/[^/]+$`) and is what `resolveValue` uses for the same distinction.

**`formDataContext` missing on the alias extra schemas** — also valid, and it is reachable: the bound-alias branch returns before the context is applied, so the extra `allOf` members intersected into an aliased multipart body emitted `Blob`. Passing the context there is right — those members' properties are parts of the same body.

Both have regression tests, and I confirmed each fails with its fix reverted:
- `form-data-components.test.ts` — a nested property ref leaves the index empty.
- `schema-definition.test.ts` — a bound alias whose extra `allOf` member has a `format: binary` property now emits `avatar?: Blob | File`.

`vp check`, `vp test` (473 + 62), the snapshot suites (7441) and the generated-output typecheck all pass; tests and samples regenerate with no diff.

